Issue #7, opened 2017-04-26
Add 3rd party dependencies wget file 


smlambert:
We depend on some 3rd party jars, which we pick up from external sites.  We can add the file that contains the links to the TestConfig/lib folder, so that a user can run wget -f dependencies.txt to pick them up and have them in TestConfig/lib (where we expect them to be found).

We will also need to update README to indicate these dependencies (will be done as part of issue 1).

karianna:
@smlambert We've tended to use Ansible or a set of shell scripts to pull those down.  Might be worth a chat to @gdams and @neomatrix369 to see if we can merge some of those efforts?

smlambert:
Actually, I guess we do have a shell script that does this https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/get.sh, which we should edit, so that the fetch of the adoptopenjdk binary is optional if you want to run tests with your own SDK.

sophia-guo:
With #12 people could do following to wget third party jars:
cd openjdk-test
get.sh

PR #12, opened 2017-04-26
update get.sh #7

sophia-guo:
update get.sh support wget third party jar for local run.  

This branch is branched from UpdateJenkins. So probably need to first merge pull #8?

or could just close pull #8 and update this comment?

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -12,19 +12,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-echo 'Get binary openjdk...'
-cd $1
-mkdir openjdkbinary
-cd openjdkbinary
-download_url=`curl https://api.github.com/repos/AdoptOpenJDK/openjdk-releases/releases/latest | jq -r '.assets[] | select(.browser_download_url | contains("x64_Linux_jdk")) | {url: .browser_download_url} | select (.url | contains("tar.gz")) | .[]'`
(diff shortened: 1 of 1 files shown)


sophia-guo:
@ShelleyLambert did a rebase. Please take a review.

get.sh
```diff
@@ -12,19 +12,26 @@
...
+if [ "$#" -eq 2 ]
+then
+	openjdktest=$2
+	echo 'Get binary openjdk...'
+	cd $1
+	mkdir openjdkbinary
+	cd openjdkbinary
+	download_url=`curl https://api.github.com/repos/AdoptOpenJDK/openjdk-releases/releases/latest | jq -r '.assets[] | select(.browser_download_url | contains("x64_Linux_jdk")) | {url: .browser_download_url} | select (.url | contains("tar.gz")) | .[]'`
+	wget "${download_url}"
+	#wget https://github.com/AdoptOpenJDK/openjdk-releases/releases/download/jdk8u152-b01/OpenJDK8_x64_Linux_jdk8u152-b01.tar.gz
```
smlambert: I am presuming you are leaving this comment in as a convenience, in case folks want to run this script on a machine that doesn't have jq installed.

sophia-guo: Correct. Also this will be updated when we do have more specific git api available.

PR #13, opened 2017-05-01
Running openjdk regression tests

sophia-guo:


sophia-guo:
Enable regression tests under java/lang as intial try. see https://ci.adoptopenjdk.net/view/OpenJDK%20tests/job/AdoptOpenJDK_Test/25/console

PR #22, opened 2017-05-04
Jenkinsfile update.

sophia-guo:
1. Rename jenkinsfile per platform
2. Delete when success to help debug when build fails,
   use Jenkins Git plugin delete workspace


karianna:
@sophia-guo The code change doesn't match the description in the PR?

gdams:
@karianna does it not? I can see the file was renamed and also a function that deletes the directory on success

PR #30, opened 2017-05-09
Enable more regression test

sophia-guo:


karianna:
Is there a way to template the common XML config?

sophia-guo:
Yes and no. We are sharing the same general xml template between this openjdk tests(jtreg) and all other FV tests(testNG). Probably later also for SV tests, PERF tests. That means we don't have a specific template for the jtreg tests. If this becomes a general requirement we can enhance the test framework to support this.

PR #31, opened 2017-05-10
Using groups instead of pacakges to specify tests components

sophia-guo:
Using TEST.groups defined groups specify test components


sophia-guo:
Openjdk doesn't parameterize -Xmx we will probably keep same.

PR #74, opened 2017-06-15
XServer on :1 and DISPLAY address configuration.

allandequeiroz:
After few tries using different approaches, I think this one was less intrusive.

As you can see, I managed to execute tests pointing to the XServer address that I used to start it.

screen shot 2017-06-15 at 21 38 30

I had to use a package called ```xserver-xorg-legacy``` that provides a way to enable XServer startup by any user through ```/etc/X11/Xwrapper.config``` file configuration to avoid the necessity of giving sudo powers to the user that will start the tests.

I'm assuming that tests will be kicked by **jenkins** user, let me know if I'm wrong.

Hope it help.

Thanks.

### DIFF
--- a/Build/ansible/playbooks/ubuntu_14_16_x86.yml
+++ b/Build/ansible/playbooks/ubuntu_14_16_x86.yml
@@ -27,16 +27,16 @@
         - make
         - openjdk-7-jdk
         - zip
+        - wget		
+        - git		
 
     - name: Install Test tools
       apt: name={{item}} state=installed
       with_items:    
         - ant
         - ant-contrib
-        - git
         - jq
         - perl
-        - wget
 
     - name: Create jenkins user
       action: user name=jenkins state=present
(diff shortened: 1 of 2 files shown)


allandequeiroz:
Thanks @ShelleyLambert, after merging how can we try that?

Build/ansible/playbooks/ubuntu_14_16_x86.yml
```diff
@@ -135,6 +135,39 @@
             job="/usr/bin/apt-get update && /usr/bin/apt-get -y upgrade"
             state=present
 
+    - name: Install Test tools
+      apt: name={{item}} state=installed
+      with_items:
+        - ant
+        - ant-contrib
+        - git
```
gdams: hmm so wget and git should be added to the build tools area as they are also required there

allandequeiroz: Hi @gdams, I didn't change this bit, still just like before, but if you guys think that would be good, I can change that.

gdams: yes please

allandequeiroz: Done

karianna:
@gdams You happy for this to land now?

PR #75, opened 2017-06-22
Enable junit result

sophia-guo:

### DIFF
--- a/Build/jenkins/Jenkinsfile_x86-64_linux
+++ b/Build/jenkins/Jenkinsfile_x86-64_linux
@@ -47,20 +47,13 @@ pipeline {
 					}
             }
         }
-        stage('Deploy') {
-            steps {
-				timestamps {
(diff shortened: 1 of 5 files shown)


Build/jenkins/Jenkinsfile_x86-64_linux
```diff
@@ -51,16 +51,16 @@ pipeline {
             steps {
 				timestamps {
                 	echo 'Deploying....'
-                	step([$class: "TapPublisher", testResults: "**/*.tap"])
```
smlambert: The deploy step is now empty.  In the future, I imagine us uploading test results to a cloud-based DB.  If you think that action would live here, then we can keep this step, otherwise, perhaps we remove it.

sophia-guo:
Removed empty stage 'deploy'.

karianna:
@smlambert Good to land this now?

Issue #73, opened 2017-06-15
jdk_sound: javax/sound/midi/Devices/InitializationHang.java fails (MidiUnavailableException)


smlambert:
avax/sound/midi/Devices/InitializationHang.java fails with javax.sound.midi.MidiUnavailableException: Can not open line ... Caused by: java.lang.IllegalArgumentException: No line matching interface SourceDataLine supporting format PCM_SIGNED 44100.0 Hz, 16 bit, stereo, 4 bytes/frame, little-endian is supported.

20:53:10 HOME=/home/jenkins \
20:53:10 LANG=en_US.UTF-8 \
20:53:10 PATH=/bin:/usr/bin \
20:53:10     /home/jenkins/workspace/openjdk_test_x86-64_linux/openjdkbinary/j2sdk-image/jre/bin/../../bin/javac \
20:53:10         -J-ea \
20:53:10         -J-esa \
20:53:10         -J-Xmx512m \
20:53:10         -J-Dtest.src=/home/jenkins/workspace/openjdk_test_x86-64_linux/openjdk-test/OpenJDK_Playlist/openjdk-jdk8u/jdk/test/javax/sound/midi/Devices \
20:53:10         -J-Dtest.src.path=/home/jenkins/workspace/openjdk_test_x86-64_linux/openjdk-test/OpenJDK_Playlist/openjdk-jdk8u/jdk/test/javax/sound/midi/Devices \
20:53:10         -J-Dtest.classes=/home/jenkins/workspace/openjdk_test_x86-64_linux/jvmtest/OpenJDK_Playlist/SE80/work/classes/2/javax/sound/midi/Devices \
20:53:10         -J-Dtest.class.path=/home/jenkins/workspace/openjdk_test_x86-64_linux/jvmtest/OpenJDK_Playlist/SE80/work/classes/2/javax/sound/midi/Devices \
20:53:10         -J-Dtest.vm.opts='-ea -esa -Xmx512m' \
20:53:10         -J-Dtest.tool.vm.opts='-J-ea -J-esa -J-Xmx512m' \
20:53:10         -J-Dtest.compiler.opts= \
20:53:10         -J-Dtest.java.opts= \
20:53:10         -J-Dtest.jdk=/home/jenkins/workspace/openjdk_test_x86-64_linux/openjdkbinary/j2sdk-image/jre/bin/../.. \
[...]

PR #76, opened 2017-06-24
Pulseaudio for sound tests

allandequeiroz:
I've put some comments here #73 
Thanks!

### DIFF
--- a/Build/ansible/playbooks/ubuntu_14_16_x86.yml
+++ b/Build/ansible/playbooks/ubuntu_14_16_x86.yml
@@ -173,11 +173,9 @@
       static: yes
 
     - name: Include Jenkins to audio group
-      replace:
-        backup: yes		
-        dest: /etc/group
-        regexp: '^(audio(?!.*\bjenkins\b).*)$'
-        replace: '\1jenkins'
+      user: name=jenkins
+        groups=audio
+        append=yes
 
     - name: Install Pulseaudio
       apt: pkg=pulseaudio state=latest update_cache=yes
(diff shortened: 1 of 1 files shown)


gdams:
rather than doing a regex replace you should be able to do something like this I think:
``` yaml
- name: Adding existing user jenkins to group audio
  user: name=jenkins
        groups=audio
        append=yes
```

allandequeiroz:
Hi @gdams thanks a lot for let me know about this way. Just changed.

Build/ansible/playbooks/ubuntu_14_16_x86.yml
```diff
@@ -172,6 +172,14 @@
     - include: slack_notification/successful_slack.yml
       static: yes
 
+    - name: Include Jenkins to audio group
```
gdams: nit: Include Jenkins to audio group ==> Add jenkins to the audio group

allandequeiroz: 👍  Sorry about that, I'll change that tonight when I get back to home.

allandequeiroz:
Hi there. I've just updated.

Let me know if we should group "apts" in a different way,
we sort of did it for Test and Build tools in another issue.

Maybe we could put everything together under the same task and
Include comments with the former tasks names just to explain
what those packages are related to.

karianna:
@bblondin OK to land this now?

bblondin:
@karianna I'm ok with it...

PR #82, opened 2017-08-01
Temporarily disable group jdk_awt and jdk_sound

sophia-guo:

### DIFF
--- a/OpenJDK_Playlist/playlist.xml
+++ b/OpenJDK_Playlist/playlist.xml
@@ -284,23 +284,6 @@
 		</tags>
 	</test>
 	<test>
-		<testCaseName>jdk_awt</testCaseName>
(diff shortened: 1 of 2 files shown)


OpenJDK_Playlist/playlist.xml
```diff
@@ -352,23 +335,6 @@
...
-	$(TEST_STATUS)</command>
-		<subsets>
-			<subset>SE80</subset>
-		</subsets>
-		<tags>
-			<tag>extended</tag>
-		</tags>
-	</test>
-	<test>
 		<testCaseName>jdk_swing</testCaseName>
```
smlambert: There are also many x11 server failures in the jdk_swing group, if you are temporarily excluding tests have known failures of great numbers you may want to include it.

sophia-guo: Thanks for pointing out that. Updated.

Issue #100, opened 2017-09-20
Wget get 403 error when get SDK binary 


sophia-guo:
Using wget to download sdk binary from https://api.adoptopenjdk.net/$jvm_version/releases/$platform/latest/binary get 403 error, which emerged recently. 

Directly downladable with browser.

sophia-guo:
Will try with Cookie.

PR #101, opened 2017-09-20
Issue #100 Add wget Header option to avoid ERROR 403.

sophia-guo:
Also add status check, turn off the verbose output


sophia-guo:
It's intermittent failure. I tried a few times seems good.

PR #102, opened 2017-09-22
dockerize openjdk test

TianyuZuo:
* new Dockerfile added to Build/docker 
* test executable dependencies are installed in Dockerfile
* docker image can be used to execute test automatically
* documented shell script to run openjdk tests within a docker container

### DIFF
--- a/Build/docker/Dockerfile
+++ b/Build/docker/Dockerfile
@@ -1,0 +1,62 @@
+# (C) Copyright IBM Corporation 2017.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This Dockerfile in Build/docker dir is used to create an image with
+# AdoptOpenJDK jdk binary installed. Basic test dependent executions
+# are installed during the building process.
+#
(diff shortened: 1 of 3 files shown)


TianyuZuo:
Can you help to review PR
@ShelleyLambert

Build/docker/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+# `--build-arg list` to specify your base image
+#
+# Build example: `docker build --build-arg IMAGE_NAME=<image_name> --build-arg IMAGE_VERSION=<image_version >-t adoptopenjdk-test .`
+
+ARG IMAGE_NAME=adoptopenjdk/openjdk8
+ARG IMAGE_VERSION=latest
+
+FROM ${IMAGE_NAME}:${IMAGE_VERSION}
+
+MAINTAINER Tianyu Zuo <tianyu@ca.ibm.com>
```
smlambert: You will likely not be the only maintainer of this file, so may choose to leave this out (as we have done in the Dockerfiles at the OpenJ9 project).

Build/docker/mktest.sh
```diff
@@ -0,0 +1,50 @@
...
+	$JAVA_BIN/java -version
+fi
+
+export SPEC=linux_x86-64
+export JAVA_VERSION=SE80
+
+
+cd /test/TestConfig
+make -f run_configure.mk
+make getdependency
```
smlambert: This reminds me that we should fold this step into the compile (target depends on target) as we have done elsewhere so it is not needed as an explicit step.

Build/docker/Dockerfile
TianyuZuo: yes, PR updated

Build/docker/mktest.sh
TianyuZuo: yes, thanks for reminding me this. OpenJDK is downloading all dependencies from the get.sh script, and we are going to have the `getdependency` integrated into compile. Change set updated.

TianyuZuo:
Change set is ready for another review.

PR #103, opened 2017-09-27
integrate dockerized scala test with testKitGen

TianyuZuo:
* Add containerTest folder for all dockerized test
* Add scala-test under containerTest project
* build.xml is used to build scala-test docker image
* playlist.xml is used to execute the dockerized scala-test image
* dockerized scala test now run "jvm, pos and net" test suite
* more test suites can be added if needed
* the dockerized scala test can now be built and triggered by testkitGen
* To build and run scala test, you need to export environment variable:
  - export BUILD_LIST="containerTest/scala-test"

### DIFF
--- a/thirdparty_containers/scala/build.xml
+++ b/thirdparty_containers/scala/build.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0"?>
+<project name="Scala-Test" default="build" basedir=".">
+	<taskdef resource="net/sf/antcontrib/antlib.xml" />
+	<description>
+		Build Scala test Docker image
+	</description>
+
+	<!-- set properties for this build -->
+	<property name="DEST" value="${BUILD_ROOT}/containerTest/scala-test" />
+	<property name="dist" location="${DEST}/${JAVA_VERSION}" />
+	<property name="src" location="." />
+
+	<target name="init">
+		<mkdir dir="${dist}"/>
+	</target>
+
+	<target name="clean_image" depends="init" description="build scala test docker image">
+		<exec executable="docker">
+			<arg line="rmi -f adoptopenjdk-scala-test" />
+		</exec>
+	</target>
+
+	<target name="build_image" depends="clean_image" description="build scala test docker image">
+		<exec executable="docker">
+			<arg line="build -t adoptopenjdk-scala-test -f dockerfile/Dockerfile ." />
+		</exec>
+	</target>
+
+	<target name="dist" depends="build_image" description="generate the distribution">
+		<copy todir="${DEST}">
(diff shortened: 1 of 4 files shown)


TianyuZuo:
Reviewer: @ShelleyLambert

smlambert:
This is looking good, but for aesthetic purposes, I'd like to introduce a standardized naming of the directories.  We currently have OpenJDK_Playlist and containerTest (mixing uppercase/lowercase/nospaces/underscores.  In the openj9 repo, we have yet more styles of names for different test groups/directories.  Let's standardize to attempt to best communicate what the groups of tests are. We can talk about some options, but consider changing the top-level directory names to openjdk_regression and thirdparty_app ?  also, perhaps the sub-directories thirdparty_app/scala/docker ?

TianyuZuo:
Agree to have a test project naming standard, which I think will increases the readability and ease the difficulty of understanding the repo structure.

The name of `openjdk_regression` is obvious and easy understanding. For  `thirdparty_app`, are we going to put other SV tests into this folder (other than dockerized tests) ? If yes, I think we would better to have a stand alone dir under root folder or under `thirdparty_app` (e.g. `thirdparty_app/dockerized_test/scala`), as the building and running process of dockerized thirdparty tests are different from regular SV tests  (e.g. STF or plain SV tests). Does this suit our purpose?

For other test projects existing in root folder and future tests, I would suggest use lower case and underscore as the naming standard. Then the openjdk-test repo is going to be like:
```
build/
openjdk_regression/
test_example/
utils/
```

For the testKitGen folder `TestConfig`, I think we need a more meaningful name which can be easily recognized as the start entry of running all the tests. This will need suggestions from other.

smlambert:
When I want to test a single make target for a directory, I type "make test_<name_of_dir>"  and it should read like a sentence...  what am I about to do?  make test_openjdk_regression, make test_thirdparty_app, etc...  so perhaps, thirdparty_containers is better, as I want to test_thirdparty_containers, in which case it is nice to go into that directory and see the list, scala, derby, solr_lucene, etc...

Leave TestConfig as is for now, as I'd like a bigger discussion and as there are references to it in READMEs and I do not want to hold up this PR.  Also, it is a special directory, not containing test code, so is okay if it doesn't follow standard for now... We want to set a precendence for new test code directories first.

TianyuZuo:
Thanks for the clarification Shelley, changeset updated to meet our test project naming standard. For `openjdk_regression`, I will create a new PR for that. A few files is linking to the name of `OpenJDK_Playlist`, and the build is running, I think the modification with OpenJDK_Playlist needs testing build verification

sophia-guo:
ok to test

PR #109, opened 2017-10-02
Add returncode to stop jenkins pipeline step when fails

sophia-guo:


sophia-guo:
Included in https://github.com/AdoptOpenJDK/openjdk-tests/pull/129

PR #129, opened 2017-10-03
Exclude failure tests on linux

sophia-guo:
1. Open issues for failure tests
2. Exclude failure tests on local
3. Delete jdk_core group
4. Add return code when setting up failed.

### DIFF
--- a/OpenJDK_Playlist/ProblemList.txt
+++ b/OpenJDK_Playlist/ProblemList.txt
@@ -39,6 +39,63 @@
 #
 #############################################################################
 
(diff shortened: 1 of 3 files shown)


sophia-guo:
FYI @ShelleyLambert

sophia-guo:
Some tests failed on MAC. Will do MAc clean next step.

sophia-guo:
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_x86-64_linux_experimental/68/ job green.

PR #140, opened 2017-10-04
clone scala src into docker image

TianyuZuo:
* clone scala src into docker image
  - ease triage effort
  - easy to reproduce same defect


TianyuZuo:
Reviewer: @ShelleyLambert  

Per talk with Lan, it may be better to integrate the Scala src into docker image. Once there is a defect found, we can share the image directly and the same scala src is in the docker image. In this way, developers can reproduce the same error easier once it is found.

Issue #149, opened 2017-10-10
README doesn't work for building and running tests locally


sophia-guo:
Readme is obsolete. Will also update get.sh to be more flexible and readable

sophia-guo:
Sandeep Konchady <sandeep.konchady@Oracle.com> tried with building and running tests locally and noticed this.

skonchady:
Thanks Sophia for your feedback on steps to build and run tests using this project. As discussed, some  areas that could use some clarification in the README, especially for novice developers are:

1. Pre-requisite for building and running this project is not clear. Perhaps it may help novice developers to get a list of pre-requisite that's specific to this project. One example is Apache ANT and ant-contrib. I am here that git, make etc are obvious pre-requisite as they are listed in the usage.
2. The existence of local ProblamList.txt, in addition to OpenJDK ProblemList.txt, that captures known issues in this project
3. "get.sh" does not have a help option and it is hard to know what the command-line options to provide are. One can deduce what the command-line options are by reading the script, but it may not be accurate

Thanks,
Sandeep

chandrams:
Also point to the prerequisites required to run the openjdk-systemtest.

smlambert:
I am closing this issue, as the readme was updated.  The prereqs for running openjdk-systemtest are included in its readme, and when we clone and use its materials, the build.xml files pick up the dependencies required.

For further improvements on the documentation, we can open new issues.

PR #152, opened 2017-10-12
Issue #149 update get.sh

sophia-guo:
Rewrite with functions, add usage function.


karianna:
@sophia-guo conflict?

sophia-guo:
Agreed. In case neither is installed we still need a requirement and suppose it is installed by default.

PR #169, opened 2017-11-06
Support releases, nightly and upstream different build version

sophia-guo:
Support grab sdk from adoptOpenjdk API releases or nightly build.
Also support grab sdk from specific sdk build, which could also be used by upstream build to automatically trigger build.
 

sophia-guo:
When merges all upstream sdk build configure should be updated.

sophia-guo:
If SDK_RESOURCE are set either "releases" or "nightly" those two parameters will be ignored. 
If SDK_RESOURCE are set upstream, and manually set the wrong JOB name and number jenkin will fail with "cannot find the build job".

The main scenario is when upstream build(sdk build) triggers test builds with predefined parameters:
SDK_RESOURCE=upstream
UPSTREAM_JOB_NAME=$JOB_NAME
UPSTREAM_JOB_NUMBER=$BUILD_NUMBER
which cannot fails to set.

If we want to manually reproduce those upstream triggered test builds we need to manually set those three parameters and have to set it correctly.

PR #182, opened 2017-12-13
Issue 180 : Add system directory and playlist

Mesbah-Alam:

### DIFF
--- a/buildenv/jenkins/Jenkinsfile
+++ b/buildenv/jenkins/Jenkinsfile
@@ -5,8 +5,8 @@ pipeline {
 		choice (choices: 'x64_linux\nx64_mac\ns390x_linux\nppc64le_linux\naarch64_linux', description: 'PLATFORM?', name: 'PLATFORM')
 		choice (choices: 'SE90\nSE80', description: 'What is JAVA_VERSION?', name: 'JAVA_VERSION')
 		choice (choices: 'openjdk9-openj9\nopenjdk8-openj9\nopenjdk9\nopenjdk8', description: 'What is JVM Version?', name: 'JVM_VERSION')
-		choice (choices: 'openjdk\nsanity\nextended\nruntest', description: 'What is test level?', name: 'TARGET')
-		string (defaultValue: 'linux&&x64&&test', description: 'What is machine label?', name: 'MACHINE_LABEL')
+		choice (choices: 'openjdk\nsanity\nextended\ntest_systemtest\nruntest', description: 'What is test level?', name: 'TARGET')
+		choice (choices: 'linux&&x64&&test\nopentest&&linux', description: 'What is machine label?', name: 'MACHINE_LABEL')
 		choice (choices: 'releases\nnightly\nupstream\ncustomized', description: 'Where is sdk?', name: 'SDK_RESOURCE')
 		string (defaultValue: "N/A", description: 'Upstream build name, only set when SDK_RESOURCE is upstream', name: 'UPSTREAM_JOB_NAME')
 		string (defaultValue: "N/A", description: 'Build number of upstream build, working with UPSTREAM_JOB_NAME, only set when SDK_RESOURCE is upstream', name: 'UPSTREAM_JOB_NUMBER')
@@ -84,6 +84,11 @@ pipeline {
             archiveArtifacts artifacts: '**/*.tap', fingerprint: true
             junit allowEmptyResults: true, keepLongStdio: true, testResults: '**/work/**/*.xml, **/junitreports/**/*.xml'
             archiveArtifacts artifacts: '**/work/**/*.xml, **/junitreports/**/*.xml', fingerprint: true
+             script {
+            	if (params.TARGET == 'test_systemtest') {  
+	       	 		sh 'tar -zcvf openjdk-systemtest-results.tar.gz $WORKSPACE/openjdk-test/TestConfig/test_output_*'
+		    	}    
+	    	}  
 		}
  	}
     
--- a/systemtest/build.xml
+++ b/systemtest/build.xml
@@ -1,0 +1,117 @@
+<?xml version="1.0"?>
+<project name="systemtest" default="build" basedir=".">
+	<taskdef resource="net/sf/antcontrib/antlib.xml" />
+	<description>Build STF based System Tests </description>
+
+	<!-- set global properties for this build -->
+	<property name="SYSTEMTEST_ROOT" value="${basedir}" />
+	<property name="DEST" value="${BUILD_ROOT}/systemtest/${JAVA_VERSION}" />
+	<property environment="env" />
+
+	<target name="init">
+		<if>
+			<not>
+				<available file="${DEST}" type="dir" />
+			</not>
+			<then>
+				<mkdir dir="${DEST}" />
+			</then>
+		</if>
+	</target>
+
+	<target name="clone_stf">
+		<exec executable="git" failonerror="false">
+			<arg value="clone" />
+			<arg value="--depth" />
+			<arg value="1" />
+			<arg value="--single-branch" />
+			<arg value="https://github.com/AdoptOpenJDK/stf.git" />
+			<arg value="${SYSTEMTEST_ROOT}/stf" />
+		</exec>
+	</target>
+
+	<target name="clone_systemtest">
+		<exec executable="git" failonerror="false">
+			<arg value="clone" />
+			<arg value="--depth" />
+			<arg value="1" />
+			<arg value="--single-branch" />
+			<arg value="https://github.com/AdoptOpenJDK/openjdk-systemtest.git" />
+			<arg value="${SYSTEMTEST_ROOT}/openjdk-systemtest" />
+		</exec>
+	</target>
+
+	<target name="check_systemtest" depends="init">
+		<if>
+			<not>
+				<available file="${SYSTEMTEST_ROOT}/stf" type="dir" />
+			</not>
+			<then>
+				<echo message="${SYSTEMTEST_ROOT}/stf does not exist, clone from GitHub" />
+				<antcall target="clone_stf" inheritall="true" />
+			</then>
+			<else>
+				<echo message="${SYSTEMTEST_ROOT}/stf exists, skip cloning" />
+			</else>
+		</if>
+		<if>
+			<not>
+				<available file="${SYSTEMTEST_ROOT}/openjdk-systemtest" type="dir" />
+			</not>
+			<then>
+				<echo message="${SYSTEMTEST_ROOT}/openjdk-systemtest does not exist, clone from GitHub" />
+				<antcall target="clone_systemtest" inheritall="true" />
+			</then>
+			<else>
+				<echo message="${SYSTEMTEST_ROOT}/openjdk-systemtest exists, skip cloning" />
+			</else>
+		</if>
+	</target>
+
+	<target name="configure_stf" depends="check_systemtest">
+		<ant antfile="${SYSTEMTEST_ROOT}/stf/stf.build/build.xml" dir="${SYSTEMTEST_ROOT}/stf/stf.build/" target="configure" inheritAll="false" />
+		<ant antfile="${SYSTEMTEST_ROOT}/openjdk-systemtest/openjdk.build/build.xml" dir="${SYSTEMTEST_ROOT}/openjdk-systemtest/openjdk.build/" target="configure" inheritAll="false" />
+	</target>
+
+	<target name="compile_stf" depends="configure_stf">
+		<ant antfile="${SYSTEMTEST_ROOT}/openjdk-systemtest/openjdk.build/build.xml" dir="${SYSTEMTEST_ROOT}/openjdk-systemtest/openjdk.build/" inheritAll="false" />
+	</target>
+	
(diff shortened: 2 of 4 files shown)


buildenv/jenkins/Jenkinsfile
```diff
@@ -5,7 +5,7 @@ pipeline {
 		choice (choices: 'x64_linux\nx64_mac\ns390x_linux\nppc64le_linux\naarch64_linux', description: 'PLATFORM?', name: 'PLATFORM')
 		choice (choices: 'SE90\nSE80', description: 'What is JAVA_VERSION?', name: 'JAVA_VERSION')
 		choice (choices: 'openjdk9-openj9\nopenjdk8-openj9\nopenjdk9\nopenjdk8', description: 'What is JVM Version?', name: 'JVM_VERSION')
-		choice (choices: 'openjdk\nsanity\nextended\nruntest', description: 'What is test level?', name: 'TARGET')
+		choice (choices: 'openjdk\nsanity\nextended\ntest_systemtest\nruntest', description: 'What is test level?', name: 'TARGET')
```
smlambert: for now this is fine, to leverage the ability to run directories of tests via a make target structured as test_dirName, shortly we will want to update this as we will add a make target called 'system'.

buildenv/jenkins/Jenkinsfile
```diff
@@ -73,19 +73,16 @@ pipeline {
         }
     }
     post {
-    	failure {
```
smlambert: I think you removed this while you were trying things out, but don't accidentally remove the slack notifications piece from the jenkinsfile

buildenv/jenkins/Jenkinsfile
```diff
@@ -73,19 +73,16 @@ pipeline {
...
     	always {
  			step([$class: "TapPublisher", testResults: "**/*.tap"])
             archiveArtifacts artifacts: '**/*.tap', fingerprint: true
             junit allowEmptyResults: true, keepLongStdio: true, testResults: '**/work/**/*.xml, **/junitreports/**/*.xml'
             archiveArtifacts artifacts: '**/work/**/*.xml, **/junitreports/**/*.xml', fingerprint: true
+            script {
+            	if (params.TARGET == 'test_systemtest') {  
+	       	 		sh 'tar -zcvf openjdk-systemtest-results.tar.gz $WORKSPACE/openjdk-test/TestConfig/test_output_*'
+		    	}    
+	    	}
```
smlambert: I think it is true that TARGET is only ever a single target, but please confirm, it is never a comma-separated list of targets passed in?

systemtest/build.xml
```diff
@@ -0,0 +1,97 @@
...
+	</target>
+	
+	<target name="build" depends="build-dependencies">
+	</target>
+		
+	<target name="clean">
+		<ant antfile="${SYSTEMTEST_ROOT}/openjdk-systemtest/openjdk.build/build.xml" dir="${SYSTEMTEST_ROOT}/openjdk-systemtest/openjdk.build" inheritAll="false" target="clean" />
+		<ant antfile="${SYSTEMTEST_ROOT}/stf/stf.build/build.xml" dir="${SYSTEMTEST_ROOT}/stf/stf.build" inheritAll="false" target="clean" />
+	</target>
+</project>
```
smlambert: Please confirm that this will work alongside the work that @TianyuZuo is doing for the jck folder (to minimize the duplication of stf and systemtest copies).  I will assume so, based on the check of pre-existence of test materials.

systemtest/playlist.xml
```diff
@@ -0,0 +1,340 @@
...
+		<tags>
+			<tag>extended</tag>
+		</tags>
+		<subsets>
+			<subset>SE80</subset>
+			<subset>SE90</subset>
+		</subsets>
+	</test>
+	
+</playlist>
```
smlambert: As per our earlier discussion, thank you for stripping out the extra make layer provided by systemtest, as I think keeping the layers of indirection to a minimum will help us going forward to evolve our tools to the leanest, most necessary set possible (and easier to debug and maintain).

smlambert: Also, as an initial playlist this is good.  We can add some additional tags, to be even more flexible on how we categorize these tests, once we move to using the upgraded testkitgen tooling.

lumpfish: Be aware that in "stripping out the extra make layer" you lose the provided capabilities to generate correctly formatted Windows vs unix command lines (slashes, quotes etc.).

buildenv/jenkins/Jenkinsfile
Mesbah-Alam: Yes, I'll add this back before delivering. I had removed it just to turn off the notification while I am running this job off of my branch.

buildenv/jenkins/Jenkinsfile
Mesbah-Alam: I believe it will always be a single target, at least at the level of this Jenkins job. We may invoke this Jenkins script with various different targets, but it should not be a coma separated list passed into this. @sophia-guo - can you please confirm?

systemtest/playlist.xml
smlambert: We have $Q and $D already, so its a good point.  @Mesbah-Alam please update the playlist.xml file to use $Q for quotes and $D for delimiter (slash) - look at the openjdk_regression/playlist for examples.  

To answer the other question, if I can strip tooling to its simplest form, there is a hope of it not being a massive mess and maintenance nightmare in the future.  That is what we choose to do when we fetch and run openjdk regression tests, and that is what we choose to do when we grab the system tests.  

We are already pulling in an extra (STF) framework for running the JCKs that is not actually necessary to actually run them (as we have done in the past), but possibly offers some convenience.  I won't go so far as suggesting we also strip that layer out in the upcoming JCK PR, but when I look ahead a year or two from now, my motivation is to make sure we make choices that simplify and reduce, rather than add and make more complex.  

The detritus of tooling accumulates quickly and becomes unmanageable over time, my mission is to avoid the accumulation when possible.

systemtest/build.xml
Mesbah-Alam: Thanks @TianyuZuo . I have merged in the new changes to the build.xml file to this PR.

systemtest/playlist.xml
Mesbah-Alam: @smlambert , Hi Shelley, I am using $Q for quotes and $D for delimiter (slash) in the playlist.xml.

buildenv/jenkins/Jenkinsfile
```diff
@@ -1,12 +1,12 @@
...
-    agent {label "${params.MACHINE_LABEL}&&x64&&test"}
+    agent {label "${params.MACHINE_LABEL}"}
     parameters {
 		choice (choices: 'x64_linux\nx64_mac\ns390x_linux\nppc64le_linux\naarch64_linux', description: 'PLATFORM?', name: 'PLATFORM')
 		choice (choices: 'SE90\nSE80', description: 'What is JAVA_VERSION?', name: 'JAVA_VERSION')
 		choice (choices: 'openjdk9-openj9\nopenjdk8-openj9\nopenjdk9\nopenjdk8', description: 'What is JVM Version?', name: 'JVM_VERSION')
-		choice (choices: 'openjdk\nsanity\nextended\nruntest', description: 'What is test level?', name: 'TARGET')
-		choice (choices: 'linux\nmac', description: 'What is machine label?', name: 'MACHINE_LABEL')
+		choice (choices: 'openjdk\nsanity\nextended\ntest_systemtest\nruntest', description: 'What is test level?', name: 'TARGET')
+		choice (choices: 'linux&&x64&&test\nopentest&&linux\nmac', description: 'What is machine label?', name: 'MACHINE_LABEL')
```
smlambert: @sophia-guo may have already done this tidy up for the hard-coded machine label, please be aware as you may have to update to grab that recent change

buildenv/jenkins/Jenkinsfile
Mesbah-Alam: This has been put back.

buildenv/jenkins/Jenkinsfile
Mesbah-Alam: Thanks for pointing this out. I have merged in Sophia's change and refreshed the PR.

sophia-guo:
I have updated the MACHINE_LABEL to be a string as it could be any string. Any reason to change it back to choice one? Any changes to Jenkinsfile parameter may need to update all related jenkins jobs.

sophia-guo:
As getSystemTests() is no longer needed in get.sh could you also all systemtests related variables  or functions in get.sh?

Also I'd like to confirm that in this way do you still need the file jenkinsfile_openjdk-systemtest? I'm a little confused. Thanks.

systemtest/playlist.xml
lumpfish: Has the playlist been tested on Windows?

Mesbah-Alam:
@sophia-guo , The reason to update MACHINE_LABEL to a choice was that we are passing in two different labels for this job from two places - one for ci.adopt systems, and one for the internal Jenkins and it helps if we have a list of labels to pick from. If all other Jenkins jobs fail due to this change, we should update it and make it String again, and pass the appropriate string for each build run.

Mesbah-Alam:
@sophia-guo , I have created a new PR in which I have put back choices with Strings for MACHINE_LABEL, removed Jenkinsfile_systemtest script, and also updated get.sh and removed variables related to system test. Could you please review it : https://github.com/AdoptOpenJDK/openjdk-tests/pull/188

sophia-guo:
Has it been tested on Mac? Test has been enabled on Mac.

sophia-guo:
https://ci.adoptopenjdk.net/view/OpenJDK%20tests/job/openjdk9_openj9_test_x86-64_linux/88/consoleFull

12:20:18 BUILD FAILED
12:20:18 /home/jenkins/workspace/openjdk9_openj9_test_x86-64_linux/openjdk-test/TestConfig/scripts/build_test.xml:75: The following error occurred while executing this line:
12:20:18 /home/jenkins/workspace/openjdk9_openj9_test_x86-64_linux/openjdk-test/systemtest/build.xml:73: The following error occurred while executing this line:
12:20:18 /home/jenkins/workspace/openjdk9_openj9_test_x86-64_linux/openjdk-test/systemtest/openjdk-systemtest/openjdk.build/include/top.xml:64: The following error occurred while executing this line:
12:20:18 /home/jenkins/workspace/openjdk9_openj9_test_x86-64_linux/openjdk-test/systemtest/openjdk-systemtest/openjdk.build/include/top.xml:443: Execute failed: java.io.IOException: Cannot run program "${java8_bindir}/java" (in directory "/home/jenkins/workspace/openjdk9_openj9_test_x86-64_linux/openjdk-test/systemtest/openjdk-systemtest/openjdk.build"): error=2, No such file or directory

Could you take a look at this?

sophia-guo:
same failure in openjdk8_openj9_test_x86-64_linux
https://ci.adoptopenjdk.net/view/OpenJDK%20tests/job/openjdk8_openj9_test_x86-64_linux/21/console

Mesbah-Alam:
I reran the openjdk9_openj9_test_x86-64_linux job with 'nightly' build and it worked just fine : https://ci.adoptopenjdk.net/view/OpenJDK%20tests/job/openjdk9_openj9_test_x86-64_linux/90/console

It was likely a machine issue, @sophia-guo

sophia-guo:
passing build:
setup-java8-properties:
22:44:01      [echo] java8_bindir set to /usr/lib/jvm/java-9-openjdk-amd64/bin

failed build: tes
setup-java8-properties:
12:19:49      [echo] java8_bindir set to ${java8_bindir}
java8_bindir can not be set.

In Systemtest, how this properties is set up or pass in?

smlambert:
According to https://github.com/AdoptOpenJDK/stf/issues/18 that java8_bindir is set via this chain:

${java8_home} (in stf.build/include/top.xml), and ${java8_home} is set based ${java.home} (in openjdk.build/include/top.xml, target `setup-java8-properties).

sophia-guo:
java8_bindir is set in  openjdk.build/include/top.xml
 <set-platform-properties platform.property="java8_platform"
								 platformclass.property="java8_platformclass"
								 **java.bindir.property="java8_bindir"**
								 java.compiler.property="java8_compiler"
								 osname.property="java8_osname"
								 arch.property="java8_arch"
								 bits.property="java8_bits"
								 java.home="${java8_home}"
								 java.id="java8"/>
Should check how it works.

sophia-guo:
set-platform-properties are defined in same file
 <macrodef name="set-platform-properties" description="Sets various properties for the java at the provided java_home location">

Mesbah-Alam:
We can use the following issue to track progress on the java8_bindir issue : https://github.com/AdoptOpenJDK/stf/issues/18

smlambert:
Whatever the fix ends up to be, there should be an effort made to simplify the properties and variables setting, rather than the current story that seems fairly convoluted and indirect.

lumpfish:
This was to to the java being used for the build not having javac.  So a 'fix' for this issue would be for the build to terminate with a more user friendly error message in the event that someone is trying to run the build using a jre which does not contains javac, rather than to then use a property with an unusable value.

PR #184, opened 2017-12-14
Separate problem list according jvm_version 

sophia-guo:
- ProblemList_openjdk8 is the original problem list, which is for openjdk8

- Other jvm version problem lists are initialized as empty, will set up gradually by running test builds

- Remove hard code part of machine label


sophia-guo:
Once commit, jenkins job need to update Machine Label parameter

openjdk_regression/ProblemList_openjdk8-openj9.txt
```diff
@@ -0,0 +1,128 @@
...
+#      - the javac compiler
+#      - the OS (depends on what the testcase does)
+#
+# If you managed to really fix one of these tests, here is how you can
+#    remove tests from this list:
+#  1. Make sure test passes on all platforms with samevm, or mark it othervm
+#  2. Make sure test passes on all platforms when run with it's entire group
+#  3. Make sure both VMs are tested, -server and -client, if possible
+#  4. Make sure you try the -d64 option on Solaris
+#  5. Use a tool like JPRT or something to verify these results
```
smlambert: Quite sure that some of these instructions are not currently possible or correct to do at AdoptOpenJDK, but I realize they are a carry-over of the set of instructions from OpenJDK.  Some instructions are vague and offer no real guidance as to what actual actions should happen next.  At some point we should revisit the instructions in this file, but that is a whole other issue for another day.  I will not block this PR for it.

sophia-guo: Yes, need to revisit it later

PR #185, opened 2017-12-14
Refactor adoptopenjdk dependency

sophia-guo:
Move category based tests dependency to each own folder
Keep repo level get.sh focous on repo dependency


smlambert:
Looking good, it will be great that each directory will be hermetic and have all the info to be able to run independently.  

Based on the comments in the Slack channel, Muneer mentioned that the openjdk regression tests require the asmtools to be in the jtreg/lib dir, can you please make sure that is present, as it likely explains a few of the test failures we are seeing.

PR #190, opened 2017-12-20
enabling testKitGen to run STF based JCK test

TianyuZuo:
* jck and systemtest are using the same src of stf & openjdk-systemtest
* require JCK_ROOT to be exported before compiling and running JCK tests
* add api-java_math as an example to run STF JCK tests
* add jck subset
* works with target jck_${JAVA_VERSION}


TianyuZuo:
change set tested on my local machine, it passed on both SE80 and SE90

PR #191, opened 2017-12-20
Issue 18 : Re-enable systemtest + Add JAVA_HOME export to Jenkins file

Mesbah-Alam:

### DIFF
--- a/buildenv/jenkins/Jenkinsfile
+++ b/buildenv/jenkins/Jenkinsfile
@@ -53,7 +53,8 @@ pipeline {
             steps {
 				timestamps{
                 	echo 'Building tests...'
-                	withEnv(["JAVA_BIN=$WORKSPACE/openjdkbinary/j2sdk-image/${(params.JAVA_VERSION == 'SE80') ? 'jre/' : ''}bin",
+                	withEnv(["JAVA_HOME=$WORKSPACE/openjdkbinary/j2sdk-image/${(params.JAVA_VERSION == 'SE80') ? 'jre/' : ''}",
+                	"JAVA_BIN=$WORKSPACE/openjdkbinary/j2sdk-image/${(params.JAVA_VERSION == 'SE80') ? 'jre/' : ''}bin",
                 	"JAVA_VERSION=${params.JAVA_VERSION}"]) {
                 		sh '$OPENJDK_TEST/maketest.sh $OPENJDK_TEST'
 					}
@@ -64,7 +65,8 @@ pipeline {
             steps {
 				timestamps{
                 	echo 'Running tests...'
-                	withEnv(["JAVA_BIN=$WORKSPACE/openjdkbinary/j2sdk-image/${(params.JAVA_VERSION == 'SE80') ? 'jre/' : ''}bin",
+                	withEnv(["JAVA_HOME=$WORKSPACE/openjdkbinary/j2sdk-image/${(params.JAVA_VERSION == 'SE80') ? 'jre/' : ''}",
+                	"JAVA_BIN=$WORKSPACE/openjdkbinary/j2sdk-image/${(params.JAVA_VERSION == 'SE80') ? 'jre/' : ''}bin",
                 	"JAVA_VERSION=${params.JAVA_VERSION}"]) {
                 		sh "$OPENJDK_TEST/maketest.sh $OPENJDK_TEST ${params.TARGET}"
 					}
--- a/systemtest/build.xml
+++ b/systemtest/build.xml
@@ -68,12 +68,11 @@
 		</if>
 	</target>
 
-	<target name="configure_stf" depends="check_systemtest">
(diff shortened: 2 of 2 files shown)


Mesbah-Alam:
As discussed with Simon and Shelley, setting JAVA_HOME to the SDK under test in the Jenkinsfile solves the problem, as STF requires JAVA_HOME to be set in order for it to figure out which JDK to use for compiling STF and systemtest artifacts.

Mesbah-Alam:
I made the update to Jenkinsfile in my branch and ran ran a few builds to test:

    Here's a openjdk9_openj9_systemtest_x64_linux build I launched manually with "nightly" SDK : https://ci.adoptopenjdk.net/view/work%20in%20progress/job/openjdk9_openj9_systemtest_x64_linux/33/consoleFull

    Here's a openjdk9_openj9_test_x86-64_linux build launched with upstream SDK: https://ci.adoptopenjdk.net/view/OpenJDK%20tests/job/Mesbah_openjdk9_openj9_test_x86-64_linux/1/console

In both the cases, java.home , java8_bindir, and java8_compiler are now pointing to the SDK under test.

Mesbah-Alam:
@sophia-guo - could you please review the changes and confirm that adding JAVA_HOME export in Jenkinsfile will not have any negative side effects in any other jobs?

sophia-guo:
I have some comments on the changes:
1. The change doesn't explain why ${java8_bindir} was set up by JAVA_HOME occasionally or on some machine without this explicitly export.
2. The change actually set java8_bindir to java9_bindir if it's SE90. Is this expected?
3. @smlambert mentioned before this is unnecessary for other tests. Instead of exporting in Jenkinsfile at both build and test stage could it be set up by JAVA_BIN directly in build.xml itself? JAVA_HOME=$JAVA_BIN/..   and JAVA_BIN is defined in makefile and passed in to all build.xml already. For running tests I see all tests have been set up by `export JAVA_HOME=$(JDK_HOME)$(D)` : 
`command>export JAVA_HOME=$(JDK_HOME)$(D)  && \
	perl $(TEST_RESROOT)$(D)stf$(D)stf.core$(D)scripts$(D)stf.pl  \
	-test-root=$(Q)$(TEST_RESROOT)$(D)stf;$(TEST_RESROOT)$(D)openjdk-systemtest$(Q) \
	-systemtest-prereqs=$(Q)$(TEST_RESROOT)$(D)systemtest_prereqs$(Q)  \
	-results-root=$(REPORTDIR)  \
	-test=JdiTest; \
	$(TEST_STATUS)</command>`
So should be fine. 

In this way there will be no concern for local running. Otherwise "export JAVA_HOME=....." will be a must for local systemtest running.

Mesbah-Alam:
@sophia-guo , replying to your previous comment above: 
1) The explanation was actually given in the related issue: https://github.com/AdoptOpenJDK/stf/issues/18. In short, the reason why it was occasionally working without this explicit export is because, when the explicit export of JAVA_HOME is not done, it uses whatever Java ANT_HOME is pointing to, and in the failing cases, the test was running on machines where ANT_HOME was pointing to a JRE, which did not have javac in it, instead of a full SDK. 
2. The change actually set java8_bindir to java9_bindir if it's SE90. Is this expected? -- Yes. It is expected.
3. The excerpt you used in your example is actually from a playlist.xml file, but this failure happens during compilation stage of tests, which happens earlier, so the export in playlist.xml didn't actually work.

smlambert:
@Mesbah-Alam, for point 3 above, please re-read Sophia's comment, she is talking about the build.xml file and saying that you could do a similar thing in the build.xml file for compilation, **like** what we do for execution in playlist.xml file.

Mesbah-Alam:
@sophia-guo @ShelleyLambert, 
I did misunderstand Sophia's comment earlier. I've tried the idea out of setting JAVA_HOME in systemtest/build.xml instead of Jenkinsfile, and it worked with the systemtest build (e.g.: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/openjdk9_openj9_systemtest_x64_linux/42/).
The only thing that needed to be added was, when systemtest/build.xml (in which JAVA_HOME is defined) is calling the openjdk.build/build.xml to compile system test projects, we'd have to pass in java.home = JAVA_HOME, as a property and set inheritAll="false"; this ensures that whatever default value Ant set its java.home property to was not inherited but the passed in value was used by the child build.xml files. 

@sophia-guo Thanks for the tip! 

I have also tested and ensured that the change does not break the non-systemtest builds (e.g. https://ci.adoptopenjdk.net/view/OpenJDK%20tests/job/openjdk9_openj9_test_x86-64_linux/108/console).

Mesbah-Alam:
@sophia-guo : I have asked you to review the PR containing the change above: https://github.com/AdoptOpenJDK/openjdk-tests/pull/196

PR #192, opened 2017-12-27
WIP: Adding IdleMicroBenchmark Tests

sabkrish:
IdleMicroBenchmark are simple performance test intended to test the
Idle detection and management feature of J9 vm controlled by jvm
options which J9 understands but the test is compatible with Hotspot vm also.

Signed-off-by: sabkrish <sabkrish@in.ibm.com>


smlambert:
@sabkrish - we will merge this PR, note however that there is an upcoming change to how we will tag the different implementations (hotspot vs openj9), so at present the perf target will not be enabled in automated builds.

sabkrish:
@smlambert - Only performance/idle_micro/build.xml had the copyright snippet little different from others. Can you confirm if the current one is right or needs to be like in other files?

smlambert:
@sabkrish build.xml can use same copyright snippet as playlist.xml (and I see that the other build.xml files in the repo will need to be updated to match)

sabkrish:
I've changed the copyright snippet. I've tested with both hotspot and j9 vm builds.

PR #198, opened 2018-01-06
Add Jenkins_personal for personal build

sophia-guo:

### DIFF
--- a/buildenv/jenkins/Jenkinsfile_personal
+++ b/buildenv/jenkins/Jenkinsfile_personal
@@ -1,0 +1,95 @@
+#!groovy
+pipeline {
+    agent {label "${params.MACHINE_LABEL}"}
+    parameters {
+		choice (choices: 'x64_linux\nx64_mac\ns390x_linux\nppc64le_linux\naarch64_linux', description: 'PLATFORM?', name: 'PLATFORM')
+		choice (choices: 'SE90\nSE80', description: 'What is JAVA_VERSION?', name: 'JAVA_VERSION')
+		choice (choices: 'openjdk9-openj9\nopenjdk8-openj9\nopenjdk9\nopenjdk8', description: 'What is JVM Version?', name: 'JVM_VERSION')
+		string (defaultValue: 'runtest', description: 'Test Target, e.g.openjdk sanity extended test_systemtest test_DirectoryName', name: 'TARGET')
+		choice (choices: 'linux&&x64&&test\nmac&&x64&&test\nwindows&&x64&&test', description: 'What is machine label?', name: 'MACHINE_LABEL')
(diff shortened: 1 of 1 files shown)


buildenv/jenkins/Jenkinsfile_personal
```diff
@@ -0,0 +1,95 @@
+#!groovy
+pipeline {
+    agent {label "${params.MACHINE_LABEL}"}
+    parameters {
+		choice (choices: 'x64_linux\nx64_mac\ns390x_linux\nppc64le_linux\naarch64_linux', description: 'PLATFORM?', name: 'PLATFORM')
+		choice (choices: 'SE90\nSE80', description: 'What is JAVA_VERSION?', name: 'JAVA_VERSION')
+		choice (choices: 'openjdk9-openj9\nopenjdk8-openj9\nopenjdk9\nopenjdk8', description: 'What is JVM Version?', name: 'JVM_VERSION')
+		choice (choices: 'openjdk\nsanity\nextended\ntest_systemtest\nruntest', description: 'What is test level?', name: 'TARGET')
+		string (defaultValue: 'linux&&x64&&test', description: 'What is machine label?', name: 'MACHINE_LABEL')
```
smlambert: This could almost be auto-filled based on PLATFORM selection... (except that the labels may be different at different Jenkins servers..., so we can leave it as an explicitly filled out choice.

buildenv/jenkins/Jenkinsfile_personal
```diff
@@ -0,0 +1,95 @@
+#!groovy
+pipeline {
+    agent {label "${params.MACHINE_LABEL}"}
+    parameters {
+		choice (choices: 'x64_linux\nx64_mac\ns390x_linux\nppc64le_linux\naarch64_linux', description: 'PLATFORM?', name: 'PLATFORM')
+		choice (choices: 'SE90\nSE80', description: 'What is JAVA_VERSION?', name: 'JAVA_VERSION')
+		choice (choices: 'openjdk9-openj9\nopenjdk8-openj9\nopenjdk9\nopenjdk8', description: 'What is JVM Version?', name: 'JVM_VERSION')
+		choice (choices: 'openjdk\nsanity\nextended\ntest_systemtest\nruntest', description: 'What is test level?', name: 'TARGET')
```
smlambert: I kind of think it would be useful to allow for more than these choices.  Perhaps we should consider a string parameter, so that we can run individual test cases.  If we keep it as a choice parameter, I'd like to see perf, external, jck added as choices.

sophia-guo: Yes, updated and also put examples in description.

buildenv/jenkins/Jenkinsfile_personal
sophia-guo: updated.

PR #205, opened 2018-01-09
SCM test job's property information

sophia-guo:

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -1,0 +1,73 @@
+#!groovy
+/* template jenkinsfile for adoptopenjdk test builds*/
+JCL_VERSION='current'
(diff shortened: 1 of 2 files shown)


sophia-guo:
Initial version.  Different directory or level test job need to pass in its own TARGET. Suppose test build job is triggered by upstream build job for now. Later we may trigger the test jobs the same way as jdk builds are triggered.

smlambert:
Would you mind to name the templates with all lowercase characters, much like the builds are named?  so openjdk8_linux  ...  FYI, I have put a page up describing how we will name the builds here: https://github.com/AdoptOpenJDK/openjdk-tests/wiki/Naming-Convention-for-Test-Builds-in-Jenkins

sophia-guo:
@smlambert , rename to openjdk8_x86-64_linux. Omit the "impl" and "jobtype" part as it is a template.

smlambert:
Updates look good.

PR #207, opened 2018-01-10
Exclude GenerateJLIClassesPluginTest for openJ9 implementation

### LINKED ISSUE #206: Exclude GenerateJLIClassesPluginTest for openJ9 impl
As per https://github.com/eclipse/openj9/issues/893, the GenerateJLIClassesPluginTest in the openjdk regression suite is not applicable to openj9 implementation.  At present, we are not running it, but when we turn that test group on for openj9, we will need to exclude it.

This can/will be done through use of a separate problemsList.txt file for the openj9 implementation.  Currently we are using the same problemsList for both implementations, but it was known/expected to have a need for separate ones (similar as we do for the functional tests in openj9, with its separate exclude files per Java version).

sophia-guo:
The test is not applicable to openj9 implementation

close #206


smlambert:
Thanks Sophia!  Presumably, there will be one additional change, where in the playlist we reference this exclude file in the jdk_tools test case for openj9 (but agree that can come in a later PR, when we actually enable that group of tests for that implementation.

sophia-guo:
@ShelleyLambert I may not get your point. Currently each test case jdk_*** are referenced to specific exclude file by JVM_VERSION( for openj9 is openjdk8/9-openj9:
`-exclude:$(Q)$(TEST_RESROOT)$(D)ProblemList_$(JVM_VERSION).txt$(Q) \`

smlambert:
ah great, my mistake, I did not see we were already doing that here (thought the change was only in the openj9 repo), thanks @sophia-guo !

PR #209, opened 2018-01-10
Add step to upload systemtest result zip + fix math load test command

Mesbah-Alam:


smlambert:
As per desk-side review, revise change to Jenkinsfile to only need allowEmptyArchives: true (not needing to reorganize code, and updating to include this change in all lines of all Jenkinsfiles.

smlambert:
also, updating to only archive work jtr.xml files (not all xml files)

Mesbah-Alam:
The changes for this PR work fine, verified with the following personal build : https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/82/

PR #210, opened 2018-01-10
Add all jenkins property files

sophia-guo:


buildenv/jenkins/openjdk10_aarch64_linux
```diff
@@ -0,0 +1,16 @@
+#!groovy
+/* Template for test jobs on openjdk8 linux. Configure test job as parameterized.
+	Set parameter TARGET(openjdk, system, external, perf, jck etc.).
+	Set UPSTREAM_JOB_NAME(version_impl_jobtype_arch_os, for example: openjdk8_hs_openjdktest_x86-64_linux)
```
smlambert: If the UPSTREAM_JOB_NAME will mainly be a build/compile job, so perhaps use a build job name as an example rather than a test job name (same goes for all templates).

sophia-guo: Correct, my bad.

sophia-guo:
Add all other jenkins property files. Using those template to set up specific implement-jobtype test job. Update when have any specific problem.

PR #215, opened 2018-01-12
Add BUILD_LIST to compile based on TARGET

sophia-guo:
Only TARGET related folder will be compiled. No other folders will be compiled. In this way openjdk8_hs_openjdktest_x86-64_linux wont' compile systemtest and won't be affected by systemtest compile failures.


sophia-guo:
Tested with openjdk/extended/system. No existing test job configurations would be affected.

PR #216, opened 2018-01-15
Re-enable tests under issues 62 and 155

akolarkunnu:
The bugs 62 and 155 are not reproducible on latest build.


sophia-guo:
add to whitelist

Issue #144, opened 2017-10-06
jdk_lang: Failed. Unexpected exit from test [exit code: 1] - cannot allocate memory


sophia-guo:
<pre>
19:52:10 ACTION: junit -- Failed. Unexpected exit from test [exit code: 1]
19:52:10 REASON: User specified action: run junit/othervm/timeout=2500 -XX:+IgnoreUnrecognizedVMOptions -XX:-VerifyDependencies -esa test.java.lang.invoke.MethodHandlesTest 
19:52:10 TIME:   24.416 seconds
19:52:10 messages:
19:52:10 command: junit -XX:+IgnoreUnrecognizedVMOptions -XX:-VerifyDependencies -esa test.java.lang.invoke.MethodHandlesTest
19:52:10 reason: User specified action: run junit/othervm/timeout=2500 -XX:+IgnoreUnrecognizedVMOptions -XX:-VerifyDependencies -esa test.java.lang.invoke.MethodHandlesTest 
19:52:10 Mode: othervm [/othervm specified]
19:52:10 elapsed time (seconds): 24.416
19:52:10 configuration:
19:52:10 STDOUT:
19:52:10 #
19:52:10 # There is insufficient memory for the Java Runtime Environment to continue.
19:52:10 # Native memory allocation (mmap) failed to map 65536 bytes for committing reserved memory.
19:52:10 # An error report file with more information is saved as:
19:52:10 # /home/jenkins/workspace/openjdk8_test_x86-64_linux/jvmtest/OpenJDK_Playlist/SE80/work/scratch/3/hs_err_pid30798.log
19:52:10 #
19:52:10 # Compiler replay data is saved as:
19:52:10 # /home/jenkins/workspace/openjdk8_test_x86-64_linux/jvmtest/OpenJDK_Playlist/SE80/work/scratch/3/replay_pid30798.log
19:52:10 STDERR:
19:52:10 OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x00007ff8e54a0000, 65536, 1) failed; error='Cannot allocate memory' (errno=12)
19:52:10 rerun:
19:52:10 HOME=/home/jenkins \
19:52:10 LANG=en_US.UTF-8 \
19:52:10 PATH=/bin:/usr/bin \
[...]

Issue #111, opened 2017-10-02
jdk_net :ACTION: main -- Failed. Execution failed: `main' threw exception: java.lang.RuntimeException: Failed


sophia-guo:
REASON: Assumed action based on file name: run main GetMacAddress 

java/net/NetworkInterface/GetMacAddress.java

akolarkunnu:
I rerun and not able to reproduce : https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/114/testReport/

akolarkunnu:
Re-enabled these tests by PR #230

Issue #107, opened 2017-09-29
jdk_io: java/io/Serializable/unresolvableObjectStreamClass/UnresolvableObjectStreamClass.java fail


sophia-guo:
Execution failed: `main' threw exception: java.lang.NoClassDefFoundError: MySerializable

akolarkunnu:
I rerun and not able to reproduce:
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/116/testReport/
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/117/testReport/

akolarkunnu:
Re-enabled these tests by PR #230

PR #230, opened 2018-01-25
Re-enable tests under 107, 111 and 144

akolarkunnu:
Bugs #107, #111 and #144 are not reproducible.


smlambert:
Change looks good, but may I ask that you leave the comment lines for the different groups in the file, as an easy guide for others who may exclude tests.

akolarkunnu:
Ok, I will update the changes.

akolarkunnu:
Kept back the different group names in the file

Issue #228, opened 2018-01-24
Enable Java10 testing in Jenkins 


smlambert:
As there is a Java10 build available to test, we should enable the openjdk_regression and systemtest suites at minimum.

karianna:
Please note that the Java 10 is an old build, see https://github.com/AdoptOpenJDK/openjdk-build/pull/209 which needs to land first before we can sanely test.

sophia-guo:
Will "openjdk-jdk10" in "https://github.com/AdoptOpenJDK/openjdk-jdk10" be changed to openjdk-jdk?

smlambert:
We will need to update the openjdk_regression/playlist file for a few tests that require additional commandline option to point to location of native tests for Java10 testing (see below, and https://download.java.net/openjdk/testresults/10/docs/howtoruntests.html for details).

-nativepath:{location of the jdk test native library}: This has to be generated at build time of the test jdk by invoking "make test-image". These libraries are located in the build directory under "images/test/jdk/jtreg/native/"

sophia-guo:
Will the OpenjdkBuild build test-image or test build do it?

smlambert:
I presume it makes sense to have the native test libs be built alongside the JVM (as they have been done for Java8 and Java9), so if they are not currently being built as part of the Java10 builds, we should open an issue on the openjdk-build repo.

sophia-guo:
I have double check no test image is built in Java10 builds. Will open an issue in build repo.

sophia-guo:
Open an issue https://github.com/AdoptOpenJDK/openjdk-build/issues/248

smlambert:
Closing as now listed in summary issue #346

PR #232, opened 2018-01-25
Enable SE100 in testkitgen and openjdk_regression tests

sophia-guo:


sophia-guo:
One successful build and disabled for now:

https://ci.adoptopenjdk.net/view/work%20in%20progress/job/openjdk10_hs_openjdktest_x86-64_linux/

sophia-guo:
Related with #228

sophia-guo:
Good catch! Thanks, updated. Not knowing why I initially put this duplicate one.

smlambert:
One additional place that should likely get updated is https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/Jenkinsfile_personal#L8 and L9, though its fine to put it in a separate PR if preferred.

sophia-guo:
PR Build will also need to support it. I probably should try if it works for SCM with current Jenkins version. Otherwise maintain it separately is kind of pain.

Issue #186, opened 2017-12-15
Create the set of criteria (a test plan) for marking builds 'good'


smlambert:
This issue is to discuss and decide what criteria we should use to mark an AdoptOpenJDK binary "good"?   To kick off the discussion, I propose the following goals:

For **release builds**, all tests at our disposal should pass, where "all" includes:
- openjdk regression tests
- functional tests
- system tests
- jck tests (including the manual/interactive tests)
- optionally external (third party application tests) 
- optionally performance benchmarks

For **nightly builds**, a subset of all tests should be run and pass, where we explicitly state what tests are in the subset, and as more machines are available, we keep adding to the subset (to be as close to the entire list of tests that get run against release builds as we can, given the set of resources we have), starting off with:
- openjdk regression tests
- functional tests
- system tests
- optionally a subset of the jck tests (not including the manual/interactive tests)

For **pull request builds**, a small subset of tests should be run and pass.  Ideally, this set is dynamic and selected to best test the change in the PR, but as a starting point, this set would be a short list that represents a sample from the broad spectrum of full tests we have.  

For background, here is brief presentation on testing at AdoptOpenJDK: https://youtu.be/R3rdLIC089k

tellison:
A good discussion Shelley.  Let me try to describe how I might define those criteria you have suggested.

**Constraints**
In an ideal world, we would be running all the tests all the time, however, there are a number of factors that impact on the amount of testing we can perform, and these will guide the criteria we apply around the different quality assurances we assign to binaries.

The factors that impact our test plans include: the length of time available to run the tests, the number and types of machines available to run tests, and the number of tests we have at our disposal.  _Any more constraints?_

**Quality Levels**
As you note, externally we have identified three groups of binaries that are consumed by users of AdoptOpenJDK. _Releases_ which are our highest quality binaries designed for production use, _nightlies_ which have a reasonable level of confidence that they are working as expected and in particular that no serious regressions have been introduced since the last good nightly build, and the _pull request_ build that assures an individual developer that their change will not cause a nightly or release build to be marked broken.

[...]

PR #235, opened 2018-01-30
adding more jck test targets into jck playlist

TianyuZuo:
* update jck build.xml to ensure it uses the same javac in JAVA_BIN 
* add JCK_VERSION to use same test target on all JCK versions
* and JCK custom test targets and JCK_TEST_TARGET variable
* user can run JCK test subset with new added JCK custom test targets
* export JCK_TEST_TARGET to define which JCK test subset to run 
* add daily and weekly JCK test tags and test targets

### DIFF
--- a/jck/README.md
+++ b/jck/README.md
@@ -23,9 +23,9 @@
 1. Put unarchived jck test materials (jck8b or jck9) into an empty folder, for example:
 * `/jck/jck8b/` and `/jck/jck9`
 
-2. Export `JCK_ROOT=/jck` as an environment variable or pass it in makefile when run make commands
+2. Export `JCK_ROOT=/jck` as an environment variable or pass it in when run as a make command
 
-3. Export `JCK_VERSION=<your_jck_version>` as an environment variable or pass it in makefile when run make commands. For example `export JCK_VERSION=jck8b` 
+3. Export `JCK_VERSION=<your_jck_version>` as an environment variable or pass it in when run as a make command. For example `export JCK_VERSION=jck8b` 
 
 4. Export `JAVA_HOME=<your_JDK_root>` as an environment variable
 
@@ -43,8 +43,10 @@ There are three custom JCK test targets `jck-runtime-custom`, `jck-compiler-cust
 
 1. Follow the Steps 1 - 4 mentioned above. 
 
-2. Export `JCK_TEST_TARGET=<jck_test_subset>` as an environment variable or pass it in makefile when run make commands. For example `export JCK_TEST_TARGET=api/java_math`
+2. Export `JCK_TEST_TARGET=<jck_test_subset>` as an environment variable or pass it in when run as a make command. For example `export JCK_TEST_TARGET=api/java_math`
 
 3. Make sure the JCK test subset is available in JCK test material folder, a.k.a. `$(JCK_ROOT)/$(JCK_VERSION)/`.
 
-4. Follow the steps remaining in `openjdk-tests/README.md`
+4. If you need to add extra Java options to JCK tests, you could export `JCK_JAVA_ARGS="<java_options>"`. Then extra added Java options would be added to JCK test during execution.
+
+5. Follow the steps remaining in `openjdk-tests/README.md`
--- a/jck/README.md
+++ b/jck/README.md
@@ -1,3 +1,16 @@
+<!--
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
 
 # How-to Run JCK Tests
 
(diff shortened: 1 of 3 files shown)


jck/README.md
```diff
@@ -10,13 +23,28 @@
...
 
 
 This test directory contains:
   * build.xml file - that clones AdoptOpenJDK/stf repo to pick up a test framework
   * playlist.xml - to allow easy inclusion of JCK tests into automated builds
+
+
+# How-to Run customized JCK test targets
+
+There are three custom JCK test targets `jck-runtime-custom`, `jck-runtime-custom` and `jck-runtime-custom`. With these three test targets, you can run custom JCK subsets.
```
smlambert: Is this a cut'n'paste error, the 3 targets have identical names?  do you mean to include jck-devtools-custom and jck-compiler-custom in that list.

TianyuZuo: ah, thanks Shelley, this is a cut'n error. updated with newer changeset.

TianyuZuo:
For the weekly builds (a.k.a. whole set of JCK tests), I have some thoughts but it may be wrong. Please correct me if any part is inappropriate. 

I think the definition of daily and weekly would be different between OpenJ9 and AdoptOpenJDK.
AdoptOpenJDK is providing a whole JDK so it needs to run the 'whole' set of JCK tests, which means it need to run runtime, compiler and devtools. 
However for OpenJ9, it provides the VM part of JDK, I assume the `whole` set of JCK tests for OpenJ9 is the JCK runtime suites. With the same reason, the requests of 'daily' tests may be different as well. Do we want to use the same playlist for both OpenJ9 and OpenJDK?

Another issue is about how the whole set of JCK tests will be executed: serial or parallel? Serial execution would be easy to implement as just one target can do the task. If it is parallel, do we want to parallel weekly targets which have been defined in the jck playlist, or run jck.xxx.custom target several times with different passed in JCK test subset?

TianyuZuo:
I have deleted those weekly targets since the weekly run are not clear yet.

smlambert:
@TianyuZuo - it is true that different projects will have different preferences for what JCKs they want to run against different builds, which is a reason to be careful not to over-label.

At AdoptOpenJDK there is no weekly build.  There are **nightly** builds and **release** builds.  

There is a JCK test build that is run against the release build that should include ALL of the JCK tests, therefore can use "make jck" to accomplish it.

For nightly builds at AdoptOpenJDK, we may choose to run a subset of the JCKs (topic under discussion in issue #186 ).  That JCK test build is limited by time, and I would imagine would be a subset from the vm and api tests.

For other projects like that wish to use this test material from AdoptOpenJDK (assuming they have their own JCK test materials), they can choose the full test suite and run "make jck" or once we decide on the nightly subset, use the tag name that we agree upon for the make target, or run a set of individual make targets of their choosing.  I am saying hold off on this tag for now, as you know we are about to update testkitgen, and we'd either add a level called 'nightly' or leverage the existing 'sanity' or 'extended' levels for that purpose.  At present, I want to try and use the same playlist for both AdoptOpenJDK and OpenJ9 testing.  

For now, I do not care to optimize the running of these tests in parallel, as we do not really have enough machines to support this.  If we eventually do support parallel testing, we would be doing that in a Jenkinsfile and likely would do that using the jck.xxx.custom targets as you suggest.   Note we will always still need to support a serial/manual run outside of Jenkins.

smlambert:
If you have tried running this from your branch, and are satisfied that it works as expected, then I will merge this for now.  We will have to update the playlist when we bring in the updates to testkitgen, at which point we can finalize what level we use for which builds.

TianyuZuo:
Hi @smlambert  
Thanks for the info, which helps a lot. So I will keep the plyalist.xml as is (without weekly test targets, only contains daily targets). And once the tkg is synced from OpenJ9, they will also be updated.

Tests have been tested locally, worked and passed as expected.

jck/README.md
```diff
@@ -10,13 +23,30 @@
 1. Put unarchived jck test materials (jck8b or jck9) into an empty folder, for example:
 * `/jck/jck8b/` and `/jck/jck9`
 
-2. Export `JCK_ROOT=/jck` as an environment variable or pass it in makefile when run make command
+2. Export `JCK_ROOT=/jck` as an environment variable or pass it in makefile when run make commands
 
-3. Export `JAVA_HOME=<your_JDK_root>` as an environment variable
+3. Export `JCK_VERSION=<your_jck_version>` as an environment variable or pass it in makefile when run make commands. For example `export JCK_VERSION=jck8b`
```
Mesbah-Alam: Typo?.. Could be phrased as "when run as a make command".

jck/README.md
```diff
@@ -10,13 +23,30 @@
...
   * playlist.xml - to allow easy inclusion of JCK tests into automated builds
+
+
+# How-to Run customized JCK test targets
+
+There are three custom JCK test targets `jck-runtime-custom`, `jck-compiler-custom` and `jck-devtools-custom`. With these three test targets, you can run custom JCK subsets.
+
+1. Follow the Steps 1 - 4 mentioned above. 
+
+2. Export `JCK_TEST_TARGET=<jck_test_subset>` as an environment variable or pass it in makefile when run make commands. For example `export JCK_TEST_TARGET=api/java_math`
```
Mesbah-Alam: Typo?.. Could be phrased as "when run as a make command".

jck/README.md
```diff
@@ -10,13 +23,30 @@
...
+
+3. Make sure the JCK test subset is available in JCK test material folder, a.k.a. `$(JCK_ROOT)/$(JCK_VERSION)/`.
+
+4. If you need to add extra Java options to JCK tests, you could export `JCK_JAVA_ARGS="<java_options>"`. Then extra added Java options would bee added to JCk test during execution.
```
Mesbah-Alam: Typo: be, JCK

jck/README.md
TianyuZuo: Thanks for pointing it out Mesbah, updated with newer chageset

jck/README.md
TianyuZuo: updated

jck/README.md
TianyuZuo: updated

PR #241, opened 2018-02-06
change daily tagged jck tests to sanity

TianyuZuo:
* we are not going to have daily or weekly level
* we are going to use sanity and extended as daily and weekly
* change all daily tagged jck tests to sanity level 


TianyuZuo:
reivewer: @smlambert 
reviewer: @llxia

PR #244, opened 2018-02-08
Add sanity tags for subset of regression tests

smlambert:
Add sanity tags to jdk_io, jdk_lang, jdk_net, jdk_math, jdk_nio, jdk_util, jdk_management
Add exclusions for hotspot tests in sanity group

### DIFF
--- a/buildenv/jenkins/Jenkinsfile_personal
+++ b/buildenv/jenkins/Jenkinsfile_personal
@@ -27,6 +27,7 @@ pipeline {
                 BUILD_LIST="$TESTPROJECT"
                 REPO_URL="https://github.com/$personalRepo/openjdk-tests.git"
                 JVM_VERSION="${JVM_VERSION}"
+				DIAGNOSTICLEVEL ='failure'
     }
     
     stages {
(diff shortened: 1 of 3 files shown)


sophia-guo:
Could you please also update line 11:
string (defaultValue: 'runtest', description: 'Test Target to execute, default is runtest which will run all tests. You can also select sub-targets such as:\ntop level: openjdk, sanity, extended, system, jck, functional, perf, etc\ndirectory level: prefix a directory with test (eg test_systemtest)\nindividual test: jdk_beans_SE80_0, etc', name: 'TARGET')  

test: jdk_beans_SE80_0  to test: jdk_beans_0?

In the new framework target will not need JAVA_VERSION any more. I forgot that in my last commit. Thanks!

buildenv/jenkins/Jenkinsfile_personal
```diff
@@ -27,6 +27,7 @@ pipeline {
                 BUILD_LIST="$TESTPROJECT"
                 REPO_URL="https://github.com/$personalRepo/openjdk-tests.git"
                 JVM_VERSION="${JVM_VERSION}"
+				DIAGNOSTICLEVEL ='failure'
```
sophia-guo: Thanks!

smlambert:
Updated description for individual test as you suggested.

PR #247, opened 2018-02-08
update JCK test build process and test targets

TianyuZuo:
* update JCK test build process to meet new testKitGen mechanism
* update JCK test playlist to meet new testKitGen
* add more JCK test target to run whole JCK test runtime suite

### DIFF
--- a/jck/build.xml
+++ b/jck/build.xml
@@ -21,12 +21,11 @@
 	<!-- set global properties for this build -->
 	<property name="SYSTEMTEST_ROOT" value="${basedir}/../systemtest" />
 	<property name="DEST" value="${BUILD_ROOT}/jck" />
-	<property name="dist" location="${DEST}/${JAVA_VERSION}" />
-	<property name="SYSTEMTEST_BUILD_ROOT" value="${BUILD_ROOT}/systemtest/${JAVA_VERSION}" />
+	<property name="SYSTEMTEST_BUILD_ROOT" value="${BUILD_ROOT}/systemtest" />
 	<property environment="env" />
 
 	<target name="init">
-		<mkdir dir="${dist}" />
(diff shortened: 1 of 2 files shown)


TianyuZuo:
verified locally and on Jenkins server

TianyuZuo:
the git diff didn't correctly show the changes in `jck/playlist.xml`. the real changes are:
* custom targets are deleted since we add all JCK test subset into playlist.
* set up api, lang, vm and xml_schema to include whole JCK runtime suite subset. 


reply to Shelley's question:
all JCK runtime test suites are now in playlist, I didn't add devtools and compiler test suites, do we also want to add them?

llxia:
In current JCK playlist.xml, sanity test is a subset of extended.
Talked with Terry, I understand this is due to structure of JCK tests ( too many subfolders and it is hard to separate). Just a note here, there is tests overlap between JCK sanity and extended.

smlambert:
Ok thanks for the answers, for now we will leave out devtools and compiler tests.  

And yes, we are leveraging the sanity tag to tag JCKs that we may want to run nightly and/or very often.  And using the extended tag for the full set (which for now does not include devtools & compilers, but eventually will).

sophia-guo:
Just a note here. If we are going to run all sanity tests later( tests from all projects with sanity tag), there will be a problem for JCK tests. JCK tests are not supposed to run externally that is we need to exclude/disable sanity tests from JCK.

smlambert:
To your comment, in our builds at AdoptOpenJDK, we can be explicit _sanity.system, _sanity.openjdk, etc.  

In builds not at AdoptOpenJDK, where we wish to run sanity and there is a JCK_ROOT defined, then we can just proceed to include them.

PR #249, opened 2018-02-09
Add logic to calculate JVM_VERSION in third party test build.xml

Mesbah-Alam:

### DIFF
--- a/thirdparty_containers/build.xml
+++ b/thirdparty_containers/build.xml
@@ -26,9 +26,57 @@
 	<target name="init">
 		<mkdir dir="${DEST}" />
 	</target>
-
+	
+	<if>
+		<equals arg1="${JAVA_VERSION}" arg2="SE80" />
+		<then>
+			<property name="JDKVERSION" value="openjdk8" />
+		</then>
+		<elseif>
+			<equals arg1="${JAVA_VERSION}" arg2="SE90" />
+			<then>
+				<property name="JDKVERSION" value="openjdk9" />
+			</then>
+		</elseif>
+		<elseif>	
+			<equals arg1="${JAVA_VERSION}" arg2="SE100" />
+			<then>
+				<property name="JDKVERSION" value="openjdk10" />
+			</then>
+		</elseif>
+		<else>
+			<echo message="Unsupported JAVA_VERSION supplied"/>
+		</else>
+	</if>
+	
+	<echo message="JAVA_VERSION = ${JAVA_VERSION}" />
+	<echo message="JAVA_IMPL = ${JAVA_IMPL}" />
+	
+	<if>
+	    <contains string="${JAVA_IMPL}" substring="openj9" />
+	    <then>
+	    	<property name="JVM_VERSION" value="${JDKVERSION}-openj9" />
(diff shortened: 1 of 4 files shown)


Mesbah-Alam:
This PR is verified with the following personal build:https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/213/console

smlambert:
If more than thirdparty_containers need this JVM_VERSION variable, should it be defined at a more common place?

Mesbah-Alam:
From what I understand, the more common place would be TestKitGen/makefile compile target, we could pass JVM_VERSION into the invocation of build_test.xml. This way it'd be available to all test projects, and we wouldn't have to do anything extra to figure it out based on JAVA_IMPL and JAVA_VERSION (something we are doing in openjdk_regression, and also in thidparty_containers with this PR). 

@llxia, @sophia-guo : Is there a reason why we can't do the above?

sophia-guo:
From adoptopenjdk side JVM_VERSION is a must input variable. The common place to define it in test framework is TestConfig/setttings.mk, which can be consumed by test compile and run target. Though JVM_VERSION could be parsing and combined by JAVA_IMPL and JAVA_VERSION I would perfer to define it directly in settings.mk
@llxia @renfeiw

Mesbah-Alam:
@sophia-guo : Agreed. This PR can be discarded once we have the change in for JVM_VERSION.

llxia:
JVM_VERSION is basically another string representation of JAVA_IMPL and JAVA_VERSION value.
We are passing JAVA_IMPL and JAVA_VERSION in compile target. https://github.com/eclipse/openj9/pull/1146
It is redundant to also passing JVM_VERSION. As far as I know, most of test components do not need it for compilation, except container test.
The original thought it that specific variable should be defined in its own setting.mk. But if multiple test component needs it, I am ok to put it in a common place. Maybe a common place for adoptopenjdk as openj9 does not need it.

Mesbah-Alam:
@llxia : Thanks for your input. As mentioned above, there are already more than one tests that are requiring JVM_VERSION (regression and third party), so we are duplicating the work of composing JVM_VERSION from JAVA_IMPL and JAVA_VERSION in each. We need to put it in a common place. And the best place to put it, as Sophia mentioned, is settings.mk. 

> Maybe a common place for adoptopenjdk as openj9 does not need it
Are you proposing we add a new .mk file under TestConfig ?

llxia:
Talked with Sophia about this. Basically,
- jck, systemtest and openj9 do not use JVM_VERSION variable
- regression test uses JVM_VERSION in execution (openjdk_regression.mk)
- container test uses JVM_VERSION in compilation (build.xml)

If we were putting JVM_VERSION in a common place, it needs to be defined in TestConfig\settings.mk and passed in ant call in TestConfig\makefile

smlambert:
I can imagine benchmarks added to performance dir may also use it, so common location makes sense.

Mesbah-Alam:
Created the following issue to export JVM_VERSION  https://github.com/AdoptOpenJDK/openjdk-tests/issues/251

PR #252, opened 2018-02-12
Remove JVM_VERSION calculation as JVM_VERSION is now exported from

Mesbah-Alam:
TestConfig


Mesbah-Alam:
@sophia-guo - could you please review the PR?

get.sh
```diff
@@ -95,7 +95,7 @@ getBinaryOpenjdk()
 getTestKitGen()
 {
 	cd $TESTDIR
-	git clone https://github.com/eclipse/openj9.git
+	git clone https://github.com/Mesbah-Alam/openj9.git -b addExportForJvmVersion
```
sophia-guo: This should be updated.

sophia-guo:
Please also update JenkinsfileBase to export JVM_VERSION explicitly.

get.sh
Mesbah-Alam: Yep, done. I had only added it to test it off of my branch in a personal build.

PR #259, opened 2018-02-14
Updates on JenkinsfileBase and Jenkinsfile_personal

sophia-guo:
A couple of minor updates on JenkinsfileBase and Jenkinsfile_personal:

1. Archive *.jtr instead of *.jtr.xml, which contains full error logs
2. Enable SPEC parameter instead of hard coded in personal build
3. Set JAVA_IMPL as a  global environment variable, which is consistent with Testkitgen MK story
4. Set DIAGNOSTICLEVEL ='noDetails' for all jdk test

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -67,8 +67,8 @@ def test() {
 			timestamps{
 				step([$class: "TapPublisher", testResults: "**/*.tap"])
 				archiveArtifacts artifacts: '**/*.tap', fingerprint: true, allowEmptyArchive: true
-				junit allowEmptyResults: true, keepLongStdio: true, testResults: '**/work/**/*.xml, **/junitreports/**/*.xml'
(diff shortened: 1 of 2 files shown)


sophia-guo:
Current jenkins archive doesn't support archiving files only in the failure case. https://issues.jenkins-ci.org/browse/JENKINS-33048
The only concerns is those *.jtr files are much bigger than *.jtr.xml files. As we already have the limited disk space issue before, not sure if this could cause that space issue again.

smlambert:
I will go through and make sure we are only keeping the last x number of Jenkins builds, otherwise as you mention, the archived artifacts may overload the Jenkins master node.

We could consider offloading artifacts to some other storage (as we do with internal builds, using Artifactory, if it becomes a problem, but for now lets see how this approach goes).

smlambert:
Did you consider using this approach?

 post {
        always {
            junit '**/target/*.xml'
        }
        failure {
        }
    }

sophia-guo:
Thanks Shelley, updated. As we have tap file which set build status to be unstable if there is test failure. Using UNSTABLE to archive test materials.

Issue #262, opened 2018-02-21
getBinaryOpenjdk in get.sh need to be updated to avoid folder name collision


TianyuZuo:
in the get.sh file, the last three cmds in the function
```
getBinaryOpenjdk()
{
...
        jarDir=`ls -d */`
	dirName=${jarDir%?}
	mv $dirName j2sdk-image
}
```
will cause rename collision when the unzipped folder is also named as j2sdk-image. The error message would be:
`$ mv j2sdk-image/ j2sdk-image/
mv: rename j2sdk-image/ to j2sdk-image/j2sdk-image/: Invalid argument`

Since the `j2sdk-image` is a common name for some JDK's unzipped folder, My first thought is to change the target folder name ` j2sdk-image` to `testSDK` or other more meaningful names to avoid folder name collision. 

Did a search and found that `j2sdk-image` is used in JenkinsBase, windows and personal Jenkinsfiles, so changing `j2sdk-image` to other names also need to change these Jenkinsfiles.

TianyuZuo:
Hi @sophia-guo  @Mesbah-Alam  Do you have any comments or suggestions?

sophia-guo:
j2sdk-image is a common image name in openjdk. There is a decision about the sdk folder name in adoptopenjdk. Like jdk8u172-b03, which should not get collision. Could you first confirm which sdk you are using? Even if there is collision we can do like:

```
if dirName!='j2sdk-image'
  mv $dirName j2sdk-image
fi
```

TianyuZuo:
The JDK is openj9-java8 compiled internally, which is different from the one compiled i adopt ci. I will add the check to the get.sh script.

Mesbah-Alam:
@TianyuZuo - please update `get.sh` with the check.

PR #275, opened 2018-02-26
Add JCK signaturetest to playlist and fix dir rename issue

TianyuZuo:
* add JCK signaturetest to jck playlist and limited it to linux_x86
  machines since only linux_x86 machines have Svfb installed
* add dir name check before rename it to j2sdk-image
* clone openj9 with only main branch and history 2 to shrink clone time
  which also shrink filter-branch time.

fixed:#262

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -87,13 +87,17 @@ getBinaryOpenjdk()
 	fi
 	jarDir=`ls -d */`
 	dirName=${jarDir%?}
-	mv $dirName j2sdk-image
+	if [ "$dirName" != "j2sdk-image" ]; then
+		mv $dirName j2sdk-image
+	else
+		echo "dirName is equal to j2sdk-image, skip moving"
+	fi
 }
 
 getTestKitGen()
 {
 	cd $TESTDIR
-	git clone https://github.com/eclipse/openj9.git
+	git clone --depth 1 https://github.com/eclipse/openj9.git
 	cd openj9
 	git filter-branch --subdirectory-filter test/TestConfig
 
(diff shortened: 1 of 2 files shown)


get.sh
```diff
@@ -87,13 +87,17 @@ getBinaryOpenjdk()
...
-	mv $dirName j2sdk-image
+	if [ "$dirName" != "j2sdk-image" ]; then
+		mv $dirName j2sdk-image
+	else
+		echo "dirName is equal to j2sdk-image, skip moving"
+	fi
 }
 
 getTestKitGen()
 {
```
smlambert: If some other test project has retrieved testkitgen already, then we could skip this step.  It would be good to check first if the clone is required.

get.sh
```diff
@@ -87,13 +87,17 @@ getBinaryOpenjdk()
...
 	cd $TESTDIR
-	git clone https://github.com/eclipse/openj9.git
+	git clone --single-branch --depth 2 https://github.com/eclipse/openj9.git
```
sophia-guo: To my understand if using --depth option --single-branch would be default option?

```
 git clone --depth 1 https://github.com/eclipse/openj9.git 

```
Would do same function?

get.sh
sophia-guo: @ShelleyLambert For jenkins story sdk is needed for all projects. When build is triggered by upstream sdk download is skipped.

For local running parameter SDKDIR is used to check if sdk is needed to download.
```
if [[ "$SDKDIR" != "" ]]; then
	getBinaryOpenjdk
fi

```

sophia-guo: @TianyuZuo  this should close issue #262

TianyuZuo: Updated to check if TestConfig dir exists.

smlambert:
Ok, thanks for the explanation

PR #279, opened 2018-02-26
Add ability to Jenkinsfiles to compile and run JCK test

TianyuZuo:
* add JCK_VERSION to openjdk8 and 9 Jenkinsfiles
* add getJCK funciton to prepare JCK test materials
* variables like JCK_MATERIAL_SERVER_ID and BUILD_LIST are passed in 
  from job configure, if they are not passed in, the exception will be
  caught and handle by Jenkinsfile

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -11,6 +11,51 @@ def getBuildList() {
 	def TESTPROJECT = TESTPROJECTS[simpleTarget]
 	return TESTPROJECT
 }
+
+def getJCK(){
+	dir("$WORKSPACE/jck/"){
+		def jck_suite = "runtime"
+		def jck_version = "0"
+		if (env.JCK_VERSION == "jck8b") {
+			jck_version = "8b"
+		} else if (env.JCK_VERSION == "jck9") {
+			jck_version = "9"
+		} else if (env.JCK_VERSION == "jck10") {
+			jck_version = "10"
+		} else {
+			error_message =  "FAILED: JCK version ${env.JCK_VERSION} are not correctly defined \n"
+			echo error_message
+			currentBuild.result = 'FAILURE'
+			error(error_message)
+		}
+		def server = Artifactory.server 'na.artifactory.swg-devops'
+		def downloadSpec_JCK = """{
+		"files": [
+			{
+			"pattern": "${JCK_MATERIAL_SERVER_ID}/jck-test/jck-src/${env.JCK_VERSION}/JCK-${jck_suite}-${jck_version}.zip",
+			"target": "./"
+			},
+			{
+			"pattern": "${JCK_MATERIAL_SERVER_ID}/jck-test/jck-src/${env.JCK_VERSION}/${env.JCK_VERSION}_configs.zip",
+			"target": "./"
+			}
+		]
+		}"""
+		server.download(downloadSpec_JCK)
+		sh """
+			df -h
+			mv ./jck-test/jck-src/${env.JCK_VERSION}/* ./
+			ls ./
+			echo "unzipping jck materials"
+			unzip -o -q ./${env.JCK_VERSION}_configs.zip  -d ./${env.JCK_VERSION}/
+			unzip -o -q ./JCK-${jck_suite}-${jck_version}.zip -d ./${env.JCK_VERSION}/
+			echo "unzipping jck materials successfully"
+			ls ./${env.JCK_VERSION}/JCK-${jck_suite}-${jck_version}/
+			ls ./${env.JCK_VERSION}/
+		"""
+	}
+}
+
 def test() {
 	timeout(time: 6, unit: 'HOURS') {
 		stage('Setup') {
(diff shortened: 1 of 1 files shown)


TianyuZuo:
Reviewer: @smlambert 
Reviewer: @sophia-guo 

Need to mention that the getJCK function is a temporary version, once all infrastructure is setup (github jck material), it will be updated to use the new infrastructure

buildenv/jenkins/JenkinsfileBase
```diff
@@ -26,7 +72,33 @@ def test() {
 					CUSTOMIZED_SDK_URL = ''
 				}
 				env.SPEC = "${SPEC}"
-				env.BUILD_LIST= "${TESTPROJECT}"
+				//check if customized BUILD_LIST is defiend
+				try{
```
smlambert: Nitpick, but please use consistent spacing ( a single space ) between keywords and braces { and brackets (

sophia-guo:
I may miss some information. I have some questions.

1. why we need to pass in BUILD_LIST? To reduce passed in parameters for jenkins template job we have getBuildlist function to get BUILD_LIST by TARGET. What are the expected jck build targets?

2. Seems that the main job of getJCK is to download jck test material. Wondered if it is possible to do getJCK by build.xml?  My understand is if target is jck(could also be sanity/extend.jck or jck**)the BUILD_LIST is jck. And JCK folder will be compiled, which means all needed materials(including jck test material ) should be setup. Otherwise local running will be a different story. 

3. JCK_VERSION could be jck8b, jck9 which are related with JAVA_VERSION. Could be set as env.JCK_VERSION= "${(JAVA_VERSION == 'SE80') ? 'jck8b' : 'jck9'". Using if else could also set for jck10. In this way no template jenkinsfile need to add redundant extra parameter.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -26,7 +72,33 @@ def test() {
 					CUSTOMIZED_SDK_URL = ''
 				}
 				env.SPEC = "${SPEC}"
-				env.BUILD_LIST= "${TESTPROJECT}"
+				//check if customized BUILD_LIST is defiend
```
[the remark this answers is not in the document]
TianyuZuo: this code snippet was removed since Sophia's change can grab build list based on test target

buildenv/jenkins/JenkinsfileBase
TianyuZuo: updated

TianyuZuo:
1. BUILST_LIST was used for single test targets, prepared for parallel testing. Since the BUILD_LIST can be set based on test target now, removed this code snippet.

2. Using build.xml needs pass in many parameters and put the Jenkins credentials to public, so use pipeline function and Artifactory plugin to download JCK test materials for now. Once the story of getting JCK from GitHub has been completed, it will be updated eventually.

3. This is a good suggestion, updated in the PR.

TianyuZuo:
Ready for another reviewe @smlambert  @sophia-guo

buildenv/jenkins/JenkinsfileBase
```diff
@@ -26,6 +75,30 @@ def test() {
 					CUSTOMIZED_SDK_URL = ''
 				}
 				env.SPEC = "${SPEC}"
+				env.BUILD_LIST = "${getBuildList()}"
+				echo "env.BUILD_LIST is ${env.BUILD_LIST}"
```
sophia-guo: This is redundant.  sh 'printenv' print all environmental variables.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -26,6 +75,30 @@ def test() {
...
+				
+				//download JCK test materials if BUILD_LIST is jck
+				if ( env.BUILD_LIST == "jck" ){
+					if (fileExists("$WORKSPACE/jck/")) {
```
sophia-guo: I would prefer  relative path, which would be consistent with other code. Though absolute path may be a better idea?

buildenv/jenkins/JenkinsfileBase
```diff
@@ -11,6 +11,55 @@ def getBuildList() {
 	def TESTPROJECT = TESTPROJECTS[simpleTarget]
 	return TESTPROJECT
 }
+
+def getJCK(){
+	if(!fileExists("$WORKSPACE/jck/${env.JCK_VERSION}/")){
```
sophia-guo: This check is redundant. I personal think checking should be caller's functionality.

sophia-guo:
Talked with @TianyuZuo getJCK() in Jenkinsfile is a temporary solution, which will eventually move to build.xml when github jck material is available. 

getJCK() use a lot of environmental variables. I would prefer to pass in those parameters. Otherwise function will heavily depend on the environment settings. I'm ok for now as this is  a temporary solution.

buildenv/jenkins/JenkinsfileBase
TianyuZuo: updated

buildenv/jenkins/JenkinsfileBase
TianyuZuo: updated

buildenv/jenkins/JenkinsfileBase
TianyuZuo: The absolute path will be pass to JCK_ROOT, which is used within Ant build.xml. relative path will cause build failure.

sophia-guo: I mean if (fileExists("$WORKSPACE/jck/) and dir("$WORKSPACE/jck/

smlambert:
Yes, I agree with @sophia-guo comments.  Will approve, and expect another round of changes to require less of this in the Jenkinsfile.  For now the implication for manually running the JCKs is that one needs to set up some of the test materials themselves (or run on pre-configured machines).

Somewhat unrelated, looking at the README in the jck folder, are all of the instructions still valid?  Custom targets still a thing?

TianyuZuo:
Custom targets was removed to run JCK tests without `test target failures`. Now we can include extra makefile to define default `JCK_TEST_TARGET`, will add those custom test targets back soon.

PR #285, opened 2018-02-28
Exclude SharedClasse softmx increase JitAot test from linux ppc64le

Mesbah-Alam:

### DIFF
--- a/systemtest/playlist.xml
+++ b/systemtest/playlist.xml
@@ -695,6 +695,7 @@
 		</impls>
 	</test>
 	
+	<!-- Exclude the following test on Linux ppc64le. Reason: AdoptOpenJDK/openjdk-systemtest/issues/79 -->
 	<test>
 		<testCaseName>SharedClassesWorkloadTest_Softmx_Increase_JitAot</testCaseName>
 		<variations>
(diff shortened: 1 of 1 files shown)


systemtest/playlist.xml
```diff
@@ -695,6 +695,7 @@
 		</impls>
 	</test>
 	
+	<!-- Exclude the following test on Linux ppc64le. Reason: AdoptOpenJDK/openjdk-systemtest/issues/79 -->
 	<test>
```
smlambert: I don't see the < disabled > tag

smlambert: I see, so the original test has been repurposed as the non-linux one, and a new test added for the linux platforms with ^ppcle for now.  

Eventually when the defect is fixed, we drop back to only one entry for that test in the playlist (for all platforms).  

This change could have been done with only one test entry, by just adding ^ppcle to the list?  or you would then have to explicitly mention all platforms it could run on?

Mesbah-Alam: Yes, if we just added ^ppc in the single test entry, then that'd exclude the test from ppc64 aix too, so in that case we would have had to explicitly mention each platform it runs on. Yes, once the test i fixed, we will remove the second entry for the test and also remove the `platformRequirements` tag from the first entry.

PR #287, opened 2018-02-28
Update jdk and hotspot test dir to work with restructuring of openjdk10 test

### LINKED ISSUE #284: openjdk10_hs_openjdktest_x86-64_linux got "Error: Group spec does not specify a valid test suite:"
Build openjdk10_hs_openjdktest_x86-64_linux didn't run any tests due to error"Error: Group spec does not specify a valid test suite: "

Looking at code structure openjdk10 has changed code structure,  which separate source code and test code from each component. All test code has been moved to test folder. Where can we find those changes document or how can we be notified when those changes happens?

sophia-guo:
In openjdk jdk/hotspot test has been moved from {openjdk source top directory}/jdk(hotspot)/test in jdk8/9 to {openjdk source top directory}/test/jdk(hotspot).  

This change update jdk/hotspot test dir to work for both jdk8/9 and jdk10.

Also disable hotspot test, which need extra command option -nativepath:{location of the hotspot test native library}.

Close #284 

Signed-off-by: Sophia Guo sophiag@ca.ibm.com

### DIFF
--- a/openjdk_regression/playlist.xml
+++ b/openjdk_regression/playlist.xml
@@ -58,6 +58,7 @@
 		<impls>
 			<impl>hotspot</impl>
 		</impls>
+		<disabled>https://github.com/AdoptOpenJDK/openjdk-tests/issues/228</disabled>
 	</test>
 	<test>
 		<testCaseName>hotspot_jre</testCaseName>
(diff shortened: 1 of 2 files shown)


openjdk_regression/playlist.xml
```diff
@@ -58,6 +58,7 @@
 		<impls>
 			<impl>hotspot</impl>
 		</impls>
+		<diabled>https://github.com/AdoptOpenJDK/openjdk-tests/issues/228</diabled>
```
smlambert: typo in tag name: diabled should be disabled

PR #289, opened 2018-03-01
add jck custom test target back

TianyuZuo:
* add jck runtime custom test target
* add default JCK_TEST_TARGET value into jck.mk, in this case if
  JCK_TEST_TARGET is not define by user, it will have a default value
  to run the jck test
* add default JCK_VERSION and JCK_ROOT in jck.mk
* add jck-runtime test target as an example since we only need runtime
  for now, if extra test suites are needed, will update the playlist 
  upon request.

### DIFF
--- a/jck/README.md
+++ b/jck/README.md
@@ -29,17 +29,18 @@
 
 4. Export `JAVA_HOME=<your_JDK_root>` as an environment variable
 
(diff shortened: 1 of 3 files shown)


TianyuZuo:
Reviewer: @smlambert 
Reviewer: @llxia

PR #290, opened 2018-03-02
Set common env variables globally

sophia-guo:
Set common env variables globally in Jenkinsfile_personal


smlambert:
actually, before you commit this @sophia-guo, is it possible for you to add a label to the choices called sw.tool.docker ?  as we have been using linux machines for it, but now I am not certain that all linux machines have it (possibly just 4 of 6).  

And once its installed on other platforms, selecting docker will give a bigger set of machines to send jobs to

sophia-guo:
@smlambert updated with label sw.tool.docker and compressedref spec.

smlambert:
haha, ok.  You and @Mesbah-Alam are moving fast!  It is great... now that I see how many different choices there are, I wonder that we might just switch this param to be a string (with all the current choices showing up below the text field in the Jenkins UI, so we avoid having to continue to change this file for this field).  see note on #298

sophia-guo:
Agreed. Actually this field was a choice, then changed to a string, then changed back to a choice :) Will change back to a string.

smlambert:
Yes, apologies for the churn on this one...  :)  and thanks for taking it on.

Issue #292, opened 2018-03-05
All openj9 test build fails : Error: failed to get JDK properties for /home/jenkins/workspace/openjdk8_j9_openjdktest_x86-64_linux/openjdkbinary/j2sdk-image/jre/bin/../../bin/java -ea -esa -Xmx512m -Xnocompressedrefs; exit code 255


sophia-guo:
All openj9 test build fails : Error: failed to get JDK properties for /home/jenkins/workspace/openjdk8_j9_openjdktest_x86-64_linux/openjdkbinary/j2sdk-image/jre/bin/../../bin/java -ea -esa -Xmx512m -Xnocompressedrefs; exit code 255

https://ci.adoptopenjdk.net/view/OpenJDK%20tests/job/openjdk8_j9_openjdktest_x86-64_linux/35/console

sophia-guo:
Recent testkitgen changes for j9. If it's non compressed sdk -Xnocompressedrefs is added as default jvm options. If it's compressed sdk -Xcompressedrefs is added as default jvm options. 

In adoptopenjdk comrpessedrefs is default. There is no nocompressedrefs or compressedrefs sdk.

smlambert:
is this not covered by env variable SPEC, and are we using the correct one at Adopt?

sophia-guo:
Refer to #294

PR #294, opened 2018-03-06
Update SPEC with cmprssptrs substring

sophia-guo:
Update SPEC with comprssptrs substring as default value as we need to set tests against openj9 sdk with jvm option -Xcompressedrefs.


smlambert:
And to clarify, for hotspot implementations, testkitgen strips off the -Xcompressedrefs command line option for that JAVA_IMPL=hotspot or sap, so tests will run well against all 3 supported implementations.

sophia-guo:
close #292

PR #296, opened 2018-03-06
Enable parallelism for test execution and remove jck dependent functions in Jenkinsfile

TianyuZuo:
* update base file to enable parallelism
* use single LABEL to define Jenkins nodes
* reuse LABEL to parallel test jobs
* remove jck dependent functions in Jenkinsfile
  since JCK materials are staging in ant build.xml
* archive test output dir if test was unstable

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -11,101 +11,82 @@ def getBuildList() {
 	return TESTPROJECT
 }
 
-def getJCK(){
-	dir("$WORKSPACE/jck/"){
-		def jck_suite = "runtime"
-		def jck_version = "0"
-		if (env.JCK_VERSION == "jck8b") {
-			jck_version = "8b"
-		} else if (env.JCK_VERSION == "jck9") {
-			jck_version = "9"
-		} else if (env.JCK_VERSION == "jck10") {
-			jck_version = "10"
-		} else {
-			error_message =  "FAILED: JCK version ${env.JCK_VERSION} are not correctly defined \n"
-			echo error_message
-			currentBuild.result = 'FAILURE'
-			error(error_message)
+def archiveTestBinaries() {
+	dir("$WORKSPACE") {
+		sh "tar -zcf test-binaries.tar.gz ./openjdk-tests ./jvmtest ./openjdkbinary"
+	}
+	archiveArtifacts artifacts: '**/test-binaries.tar.gz', fingerprint: true, allowEmptyArchive: false
+}
+
+def stageTestBinaries() {
+	sh "wget -q --no-check-certificate --header 'Cookie: allow-download=1' ${env.BUILD_URL}artifact/test-binaries.tar.gz " 
+	sh "tar -xzf test-binaries.tar.gz"
+	echo pwd()
+	sh "ls"
+}
+
+def setupEnv() {
+	env.JCL_VERSION = "current"
+	env.JAVA_BIN = "$WORKSPACE/openjdkbinary/j2sdk-image/${(JAVA_VERSION == 'SE80') ? 'jre/' : ''}bin"
+	env.JAVA_HOME = "${JAVA_BIN}/.."
+	env.JAVA_VERSION = "${JAVA_VERSION}"
+	env.JVM_VERSION = "${JVM_VERSION}"
+	env.SPEC = "${SPEC}"
+	env.BUILD_LIST = "${getBuildList()}"
+	
+	if (env.BUILD_LIST == "jck") {
+		if ( JAVA_VERSION == "SE80" ) {
+			env.JCK_VERSION = "jck8b"
+		} else if ( JAVA_VERSION == "SE90" ) {
+			env.JCK_VERSION = "jck9"
+		} else if ( JAVA_VERSION == "SE100" ) {
+			env.JCK_VERSION = "jck10"
 		}
-		def server = Artifactory.server 'na.artifactory.swg-devops'
-		def downloadSpec_JCK = """{
-		"files": [
-			{
-			"pattern": "${JCK_MATERIAL_SERVER_ID}/jck-test/jck-src/${env.JCK_VERSION}/JCK-${jck_suite}-${jck_version}.zip",
-			"target": "./"
-			},
-			{
-			"pattern": "${JCK_MATERIAL_SERVER_ID}/jck-test/jck-src/${env.JCK_VERSION}/${env.JCK_VERSION}_configs.zip",
-			"target": "./"
-			}
-		]
-		}"""
-		server.download(downloadSpec_JCK)
-		sh """
-			df -h
-			mv ./jck-test/jck-src/${env.JCK_VERSION}/* ./
-			ls ./
-			echo "unzipping jck materials"
-			unzip -o -q ./${env.JCK_VERSION}_configs.zip  -d ./${env.JCK_VERSION}/
-			unzip -o -q ./JCK-${jck_suite}-${jck_version}.zip -d ./${env.JCK_VERSION}/
-			echo "unzipping jck materials successfully"
-			ls ./${env.JCK_VERSION}/JCK-${jck_suite}-${jck_version}/
-			ls ./${env.JCK_VERSION}/
-		"""
+		env.JCK_ROOT = "$WORKSPACE/../jck_root"
+		echo "env.JCK_ROOT is: ${env.JCK_ROOT}, env.JCK_VERSION is: ${env.JCK_VERSION}"
+		if( params.JCK_GIT_REPO ) {
+			env.JCK_GIT_REPO = params.JCK_GIT_REPO
+			echo "env.JCK_GIT_REPO is ${env.JCK_GIT_REPO}"
+		} else {
+			echo "params.JCK_GIT_REPO was not defined"
+		}
+	}
+	
+	if (JVM_VERSION.contains('openj9')) {
+		JAVA_IMPL = 'openj9'
+	} else if (JVM_VERSION.contains('sap')) {
+		JAVA_IMPL = 'sap'
+	} else {
+		JAVA_IMPL = 'hotspot'
+	}
+	env.JAVA_IMPL= "${JAVA_IMPL}"
+	if (env.BUILD_LIST == 'openjdk_regression' ||  env.BUILD_LIST == 'thirdparty_containers') {
+		env.DIAGNOSTICLEVEL ='noDetails'
+	}
+	sh 'printenv'
+}
+
+def setupParallelEnv() {
+	stage('setupParallelEnv') {
+		timestamps{
+			cleanWs()
+			setupEnv()
+		}
 	}
 }
 
 def setup() {
 	stage('Setup') {
 		timestamps{
-			env.JCL_VERSION = "current"
-			env.JAVA_BIN = "$WORKSPACE/openjdkbinary/j2sdk-image/${(JAVA_VERSION == 'SE80') ? 'jre/' : ''}bin"
-			env.JAVA_HOME = "${JAVA_BIN}/.."
-			env.JAVA_VERSION = "${JAVA_VERSION}"
-			env.JVM_VERSION = "${JVM_VERSION}"
-			env.SPEC = "${SPEC}"
-			env.BUILD_LIST = "${getBuildList()}"
+			setupEnv()
+			// get TestKitGen and test JDK for test
 			try{
 				CUSTOMIZED_SDK_URL = "${CUSTOMIZED_SDK_URL}"
 			} catch (MissingPropertyException e) {
 				CUSTOMIZED_SDK_URL = ''
 			}
-			//download JCK test materials if BUILD_LIST is jck
-			if (env.BUILD_LIST == "jck") {
-				if ( fileExists ("$WORKSPACE/jck/") ) {
-					dir("$WORKSPACE/jck/") {
-						deleteDir()
-					}
-				}
-				
-				if ( JAVA_VERSION == "SE80" ) {
-					env.JCK_VERSION = "jck8b"
-				} else if ( JAVA_VERSION == "SE90" ) {
-					env.JCK_VERSION = "jck9"
-				} else if ( JAVA_VERSION == "SE100" ) {
-					env.JCK_VERSION = "jck10"
-				}
-				echo "env.BUILD_LIST is jck, starting preparing JCK test materials."
-				getJCK()
-				env.JCK_ROOT = "$WORKSPACE/jck/"
-				
-			}
-			
-			if (JVM_VERSION.contains('openj9')) {
-				JAVA_IMPL = 'openj9'
-			} else if (JVM_VERSION.contains('sap')) {
-				JAVA_IMPL = 'sap'
-			} else {
-				JAVA_IMPL = 'hotspot'
-			}
-			env.JAVA_IMPL= "${JAVA_IMPL}"
-			if (env.BUILD_LIST == 'openjdk_regression' ||  env.BUILD_LIST == 'thirdparty_containers') {
-				env.DIAGNOSTICLEVEL ='noDetails'
-			}
-			sh 'printenv'
-			
-			sh "chmod 755 ${WORKSPACE}/openjdk-tests/maketest.sh"
-			sh "chmod 755 ${WORKSPACE}/openjdk-tests/get.sh"
+			sh "chmod 755 $WORKSPACE/openjdk-tests/maketest.sh"
+			sh "chmod 755 $WORKSPACE/openjdk-tests/get.sh"
 			if (fileExists('openjdkbinary')) {
 				dir('openjdkbinary') {
 					deleteDir()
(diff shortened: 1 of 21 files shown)


TianyuZuo:
reviewer: @smlambert 
reviewer: @sophia-guo 

This PR enables parallelism for https://github.com/AdoptOpenJDK/openjdk-tests/pull/291

sophia-guo:
Ideally we only need to parallel test part. That is saying that setup part and build part we only need to do once. All parallel test part should share setup and build part. Otherwise We don't take the benefit of parallelism. In this current version all parallel tests are repeat setup and build.

TianyuZuo:
Agree to that the compile-archive-shared process are more intuitive to some cases (for example openjdk and system). The archive-share process may take several minutes less than the current process, but I don't think the saved time slot would make a big difference for tests which need parallelization.

The compile-archive-shared process doesn't fit all test processes. Thirdparty_containers tests need the test compilation and execution on the same machine at the same time. And for jck tests, the archive-share process would double the test material preparation time, since JCK materials are really large and zipping them would cost nearly 1 hour. 

So I would suggest we keep the current process since it works for all test scenarios. And we would keep refining it with the test updates.

smlambert:
This approach would be super costly for FV tests (which is likely where Sophia is coming from), but agree that unzipping JCKs is time-consuming.  

I would be interested to know actual numbers, rather than a guess at it being similar execution time.  

I also want to know that this change does not break the other tests (so please run test_personal build from your branch for openjdk and system target (or some individual test targets from those groups).

TianyuZuo:
Hi @smlambert  and @sophia-guo  

This PR was tested and worked as expected for adopt test jobs. It won't break current job.
The time consumption of compile-archive-share approach for openjdk, system and openj9 functional was in the table below
```
openj9 functional https://urv-jenkins.canlab.ibm.com:9095/view/JCK_Tests/job/temp-timecount/30/console
	- compile from 11:42:13 to 11:45:53


openjdk_regression https://urv-jenkins.canlab.ibm.com:9095/view/JCK_Tests/job/temp-timecount/24/consoleFull
	- compile from  10:03:09  to  10:09:29
	- archive  from  10:09:29 to  10:12:45
	- download and unzip  from  10:12:46 to  10:13:25
	
systemtest https://urv-jenkins.canlab.ibm.com:9095/view/JCK_Tests/job/temp-timecount/25/consoleFull
	- compile from   10:03:01  to  10:05:47
	- archive  from  10:05:47 to  10:06:10
	- download and unzip  from  10:06:10 to  10:06:20
```

My initial thought was to make the Jenkinsfile more generic to fit most jobs, so the parallel-compile-test approach was created. This would fit most scenarios like "compile and test need to be in one job", "artifacts are too large to be archived and unzipping -- time consuming", "normal compile-run tests".

I agree that the compile-archive-share approach is a bit more time efficient for "compile-runtest" scenarios, and this will save time for more jobs. If we want to use this compile-archive-share approach, we need to separate some special cases in Jenkinsfile like "compile and test need to be in one job" and "artifacts are too large to be archived and unzipping -- time consuming". Once there is another special case doesn't fit the archive-share approach, we need to manually add that special case in our Jenkinsfile. 

Another thought is about the disk space of Jenkins master node.  Job artifacts are archived in Jenkins master node, if the number of our running jobs numbers increased or test materials size increased, master disk space will be consumed. Do we need to take it into account or infra team will take care of it?

smlambert:
fyi @llxia 

Thanks for the measurements @TianyuZuo - though it is not entirely clear what you are sharing.
1) Are the 1st times reported the ones not in parallel mode, then second in parallel mode?

For openj9 functional example, compile from 11:42:13 to 11:45:53.  11:42:13 is the amount of time to compile all the test material in non-parallel mode, and 11:45:53 is in parallel mode?  (and so a small increase, presumably worth it based on improvement in overall execution time, which isn't listed here but would depend on how we broke up the workload and how many machines available to share across).

2) Remember also, that we intend to try the storage of JCK materials unzipped in a repo, and doing a pure clone (rather than current way, which requires unzipping each time), that we should keep in mind that the unzip step may not be a time hog going forward.  Should be trying that out to see what is possible.

3) For now, I don't think we need to worry about the master running out of space.  Internally, we are moving to one with more capacity.  Externally the ci-jck master has no other jobs on it.

llxia:
@TianyuZuo , do you have measurements for JCK tests?

TianyuZuo:
Thanks for pointing it out Shelley, I updated the previous comment to make it easier to follow.

1. The work flow is download the whole src, start compilation, archive the test src and compile materials, then trigger parallel jobs. In parallel jobs, they download and unzip the archived test materials.

`compile` means the total compilation time, begins at xxx to ends at  xxx, the difference between these two is the total compilation time.
`archive` means the total time consumption in archiving test src and compiled materials
`download and unzip` means the total time consumption of each parallel job downloading and unzipping the archived test materials

I didn't archive and parallel the OpenJ9 functional tests, since the compilation time (3 minutes) is smaller than openjdk_regression (6 minutes), I assume the archive and parallel step will be faster then openjdk_regression.

2. Yes, the way to get JCK test materials will be updated to use GitHub and steps will be moved to build.xml eventually. Then I will compare the time consumption between these two: a) get zip version from GitHub then unzip them and b) get unarchied version from GitHub directly. 

Both of these two method will be affected by how the parallelism designed. 

**If we compile-archive-share**: download JCK test materials from GitHub during compilation process ==> archive JCK test materials and stage them to master ==> each parallel job will download the JCK test materials and unpack them.  JCK test materials will be archived 1 time, downloaded 2 times.

**If we parallel-compile-runtest**: download JCK test materials from GitHub during compilation ==> run the tests.

These two approached need to be compare to find which one is faster. The first approach has two steps of transmitting large archived files, which would be vulnerable on network issues.

3. that's good news, thanks for the info Shelley.

TianyuZuo:
JCK test materials are currently downloaded to `$WORKSPACE/jck/`, which is not in side the archived folder: jvmtest or openjdk-test. The JCK test compilation (TKG part) will be the same as system test. 

I will try to put the jck test materials inside the openjdk-tests folder and check the time consumption of JCK parallelism.

TianyuZuo:
JCK  test time consumption is below:
```
jck8 
	-download and unzip jck test material 16:53:42 to 16:55:47 (total 2')
	- compile begin   16:56:23  end 16:59:00 (total 2'30'')
	- archive  begin  16:59:00  end 17:01:39 (total 1'30'')
	- download and unzip  begin  17:01:39  end  17:04:58 (total 3'10'')
	
jck9
	-download and unzip jck test material 17:32:32 to 17:39:32 (total 7')
	- compile begin   17:41:14  end 17:44:45 (total 3'30'')
	- archive  begin  17:44:45  end 17:55:40 (total 10')
	- download and unzip  begin  18:12:48  end  18:35:15 (total 23')
```

karianna:
There's merge conflicts in this PR..

TianyuZuo:
JCK materials will be staged onto test machines and now we can use `archive-stage-runtest` for most scenarios. This PR is ready for review.

This PR is depending on https://github.com/AdoptOpenJDK/openjdk-tests/pull/357

buildenv/jenkins/JenkinsfileBase
```diff
@@ -186,14 +170,48 @@ def post() {
...
+						post()
+					}
+				}
+			}
+			parallel parallel_tests
+		} else {
+			echo "running test in default mode"
+			runTest("$build_list")
+		}
 		post()
```
llxia: We should move `post()` to `else`.

llxia:
In post(), we should tar test result using different names (i.e., name contains testSubDir) so that they will not override each other when move master in parallel case.
https://github.com/TianyuZuo/openjdk-tests/blob/9ec818e639263fbdffc1a7dc7e64f9442abe5c96/buildenv/jenkins/JenkinsfileBase#L169

TianyuZuo:
@llxia  Thanks for the review, updated the JenkinsfileBase and the test results archive file will be like
```
  | jck-runtime.api-test-results.tar.gz | 9.04 KB | view
  | jck-runtime.vm-test-results.tar.gz | 5.84 KB | view
  | test_output_15226958497592/TestTargetResult.tap | 16.57 KB | view
  | test_output_15226958866875/TestTargetResult.tap | 37.91 KB | view
  | test-binaries.tar.gz | 336.11 MB | view
```

TianyuZuo:
adopt Jenkins has installed sshagent plugin.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -181,20 +165,56 @@ def post() {
 					archiveArtifacts artifacts: '**/openjdk-systemtest-results.tar.gz', fingerprint: true, allowEmptyArchive: true
 				}
```
llxia: - In parallel testing, systemtest result tar file will have same override issue.
Why we need to check test type? Can we just tar the test_output_* dir when the build is unstable?
- Can we prepend $test_target to *.tap file?

smlambert:
Agree with Lan's review comments, I approve pending you amend PR to archive for all test types and better identify result files with prefixed names.

buildenv/jenkins/JenkinsfileBase
TianyuZuo: Updated per request.

* remove check and archive test_output dir if test is unstable.
* since we already archive test_output dir and tap file was in the dir, remove arching tap result step. This will not affect Jenkins tap plugin since tap results were parsed before archiving.

TianyuZuo:
updated changeset was tested in openjdk_regresssion and system in 
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox2/173/console
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox2/174/console

buildenv/jenkins/JenkinsfileBase
```diff
@@ -12,101 +11,82 @@ def getBuildList() {
...
-		def jck_version = "0"
-		if (env.JCK_VERSION == "jck8b") {
-			jck_version = "8b"
-		} else if (env.JCK_VERSION == "jck9") {
-			jck_version = "9"
-		} else if (env.JCK_VERSION == "jck10") {
-			jck_version = "10"
+def archiveTestBinaries() {
+	dir("$WORKSPACE") {
+		sh "tar -zcf test-binaries.tar.gz ./openjdk-tests ./jvmtest ./openjdkbinary"
```
sophia-guo: This basically get the whole workspace tar-ed, how big it will be? If tests don't need the test source code /jvmtest should be enough. Otherwise with build.xml updated /openjdk-tests should be right directory to tar-ed.
There is also no need to tar /openjdkbinary again, which has sdk tar-ed.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -12,101 +11,82 @@ def getBuildList() {
...
+	}
+	archiveArtifacts artifacts: '**/test-binaries.tar.gz', fingerprint: true, allowEmptyArchive: false
+}
+
+def stageTestBinaries() {
+	sh "wget -q --no-check-certificate --header 'Cookie: allow-download=1' ${env.BUILD_URL}artifact/test-binaries.tar.gz "
```
sophia-guo: Any reason not using Copy Artifact Plugin to get artifacts?

buildenv/jenkins/JenkinsfileBase
```diff
@@ -158,43 +142,72 @@ def runTest() {
...
 				echo 'Running tests...'
-				sh "$OPENJDK_TEST/maketest.sh $OPENJDK_TEST _$TARGET"
+				sh "$WORKSPACE/openjdk-tests/maketest.sh $WORKSPACE/openjdk-tests/$subDir _$TARGET"
 			}
 		}
 	}
 }
 
-def post() {
+def post(test_target) {
```
sophia-guo: Naming dir or subdir as parameter may be better as later passing in dir when calling this function.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -158,43 +142,72 @@ def runTest() {
...
 				archiveArtifacts artifacts: '**/work/**/*.jtr, **/junitreports/**/*.xml', fingerprint: true, allowEmptyArchive: true
-				if (params.TARGET == 'system') {
-					sh 'tar -zcf openjdk-systemtest-results.tar.gz $WORKSPACE/openjdk-tests/TestConfig/test_output_*'
-					archiveArtifacts artifacts: '**/openjdk-systemtest-results.tar.gz', fingerprint: true, allowEmptyArchive: true
-				}
-				if (env.BUILD_LIST == 'jck') {
-					sh 'tar -zcf jck-test-results.tar.gz $WORKSPACE/openjdk-tests/TestConfig/test_output_*'
-					archiveArtifacts artifacts: '**/jck-test-results.tar.gz', fingerprint: true, allowEmptyArchive: true
-				}
+				sh "tar -zcf $test_target-test-results.tar.gz $WORKSPACE/openjdk-tests/TestConfig/test_output_*"
```
sophia-guo: Suggest using ${test_target}-test-results.tar.gz, which is more readable.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -158,43 +142,72 @@ def runTest() {
...
 }
+
 def testBuild() {
 	timeout(time: 6, unit: 'HOURS') {
+		// prepare environment and compile test projects
 		setup()
 		buildTest()
-		runTest()
-		post()
+		def build_list = getBuildList()
```
sophia-guo: This is redundant. We have env.BUILD_LIST = "${getBuildList()}". Can using env directly.

buildenv/jenkins/openjdk10_aarch64_linux
```diff
@@ -3,7 +3,8 @@
 	Set parameter TARGET(openjdk, system, external, perf, jck etc.).
 	Set UPSTREAM_JOB_NAME(version_build_arch_os, for example: openjdk8_build_x86-64_linux)
 	Set JVM_VERSION(openjdk8, openjdk8-openj9, openjdk9, openjdk8-openj9, etc.*/
-node('linux&&arm64&&test') {
+LABEL='linux&&arm64&&test'
```
sophia-guo: Using "def LABEL='linux&&arm64&&test" to define variable as a local one. Same for all others.

buildenv/jenkins/JenkinsfileBase
llxia: If we are going to change the name, I think it should be something like `${test_target}_test_output.tar.gz`.
`test-results` is confusing.

buildenv/jenkins/JenkinsfileBase
TianyuZuo: sorry that I am not sure what is the question, could you give more details?

buildenv/jenkins/JenkinsfileBase
TianyuZuo: updated in the new change set

buildenv/jenkins/JenkinsfileBase
TianyuZuo: updated in the new change set

buildenv/jenkins/openjdk10_aarch64_linux
TianyuZuo: `LABEL` will be used for parallelism in JenkinsfileBase, will this change has a side effect on that? Meanwhile, `LABEL` is the variable which defines which machine to run tests on, a global variable is making more sense since each Jenkinsfile is to used to support one specific platform. 

If this concern is about using a yml file to store labels, then labels and other variables will be return from a function and stored in a global variable. Did I miss any concerns?

buildenv/jenkins/JenkinsfileBase
TianyuZuo: Didn't think about it initially, thanks for the suggestion. I will try that plugin on both private and public Jenkins server and verify if there is any issue with that. I will create a separate PR to update the JenkinsfileBase to use Copy Artifact Plugin, which will depend on this change set.

buildenv/jenkins/JenkinsfileBase
sophia-guo: archiveTestBinaries() is called after compile stage. That is all materials related with specific test type have been staged. Say regression test, there will be the whole openjdk repo. I don't think it is pretty small. Though it may not be a concern for now if jck tests material is small and only jck tests will be enabled parallelism. It stays as a concern for later.

jdk is a tar under ./jdkbinary I really don't think it's necessary to re-tar(which will tar all the unzipped jdk and zipped jdk). Can update archiveArtifacts artifacts: '**/test-binaries.tar.gz, yoursdkfilename', fingerprint: true, allowEmptyArchive: false 

and get test.tar and sdktar by copyArtifacts filter: 'test.tar, sdktar', fingerprintArtifacts: true, projectName: 'gsdgsg'.

buildenv/jenkins/openjdk10_aarch64_linux
sophia-guo: I personally  would not try to use global variable for it's difficult to know the side effect.

PR #299, opened 2018-03-06
Remove impls tag for testcases working with all exsiting impls

sophia-guo:
1. Remove impls tag for testcases working with all exsiting impls
(openj9,
hotspot, sap).  When there is a new impls tests can run against it
automatically. If test doesn't work with new impls we can explicitly add
impls tag.
2. Add ProblemList_openjdk10-sap.txt to enable tests on openjdk10-sap


smlambert:
Yes, good.  As the default is to assume if its not explicitly specified, the test case is valid for all implementations, this will allow for even more implementations to be supported without having to update the playlist. 👍

PR #318, opened 2018-03-12
clean JCK test job workspaces to avoid consuming too much disk space

TianyuZuo:
* each JCK test job may consume 5gb disk space
* some slaves only have 50gb disk space in total
* JCK test jobs will archive its test failure results and 
  clean its workspace after each run

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -182,6 +182,7 @@ def post() {
 				if (env.BUILD_LIST == 'jck') {
 					sh 'tar -zcf jck-test-results.tar.gz $WORKSPACE/openjdk-tests/TestConfig/test_output_*'
 					archiveArtifacts artifacts: '**/jck-test-results.tar.gz', fingerprint: true, allowEmptyArchive: true
+					cleanWs()
 				}
 				cleanWs cleanWhenFailure: false
 			}


buildenv/jenkins/JenkinsfileBase
```diff
@@ -182,6 +182,7 @@ def post() {
 				if (env.BUILD_LIST == 'jck') {
 					sh 'tar -zcf jck-test-results.tar.gz $WORKSPACE/openjdk-tests/TestConfig/test_output_*'
 					archiveArtifacts artifacts: '**/jck-test-results.tar.gz', fingerprint: true, allowEmptyArchive: true
+					cleanWs()
```
sophia-guo: We have cleanWS() at the end of build jobs no matter what tests is running. Why need extra cleanWs()?

TianyuZuo:
The JCK test sometimes failed (with failed status), and consumed many disk spaces with its test materials. Assuming we have daily-java8, daily-java9, weekly-java8 and weekly-java9, it would consume 3 or 4 * 8 gb. At the same time, the same machines are also being used for OpenJ9 compilation and test execution. This could lead to using up all disk spaces (and happened recently based on releng team comments). So I add this line to delete the workspace every time this job finished.

TianyuZuo:
Just talked with Sophia, it seems this post function won't be executed when the job failed during execution. close this PR

Issue #267, opened 2018-02-22
runtime/RedefineTests/RedefineInterfaceCall.java got Error. Can't find library: /test/lib


sophia-guo:
18:56:41 TEST: runtime/RedefineTests/RedefineInterfaceCall.java
18:56:41 TEST JDK: /home/jenkins/workspace/openjdk8_hs_openjdktest_x86-64_linux/openjdkbinary/j2sdk-image/jre/bin/../..
18:56:41 
18:56:41 TEST RESULT: Error. Can't find library: /test/lib

https://ci.adoptopenjdk.net/view/OpenJDK%20tests/job/openjdk8_hs_openjdktest_x86-64_linux/35/consoleFull

This is a new test, which is running at Feb 19. Checked https://github.com/AdoptOpenJDK/openjdk-jdk8u there is a merge 5 days ago. 

This is a jre test and need extra lib. Believe extra settings are needed. @akolarkunnu Any idea?

sophia-guo:
I have run a personal build issue stays 
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/330/testReport/junit/runtime_RedefineTests_RedefineInterfaceCall/java/RedefineInterfaceCall/

Probably it is not in adopt jdk8u repo yet: https://github.com/AdoptOpenJDK/openjdk-jdk8u

Not sure how adopt jdk8u repo get sync with jdk8u hotspot.

smlambert:
This is actually a good point to remind for openjdk_regression failures to follow this guidance for raising defects: 
https://github.com/AdoptOpenJDK/openjdk-tests/wiki/Guidance-for-Creating-Test-Defects

akolarkunnu:
The changeset http://hg.openjdk.java.net/jdk8u/jdk8u-dev/hotspot/rev/de8045923ad2 is there in adopt repo also: https://github.com/AdoptOpenJDK/openjdk-jdk8u/blob/master/hotspot/test/runtime/RedefineTests/RedefineInterfaceCall.java.
@sophia-guo: Can you please confirm personal runs take test cases from repo https://github.com/AdoptOpenJDK/openjdk-jdk8u?

sophia-guo:
Yes, the link in former comment is using adopt repo  https://github.com/AdoptOpenJDK/openjdk-jdk8u.

[...]

PR #350, opened 2018-03-20
Use master branch instead of dev to git clone adoptOpenJDK openjdk repo

sophia-guo:
Use master branch instead of dev to git clone adoptOpenJDK openjdk repo
as master has been updated as the default branch.
Also use shallow clone to save clone time.
 
### DIFF
--- a/openjdk_regression/ProblemList_openjdk8.txt
+++ b/openjdk_regression/ProblemList_openjdk8.txt
@@ -90,7 +90,6 @@ runtime/ErrorHandling/TestExitOnOutOfMemoryError.java	121	generic-all
 runtime/ErrorHandling/TestOnOutOfMemoryError.java	122	generic-all
 runtime/NMT/NMTWithCDS.java	124	generic-all
 runtime/memory/ReserveMemory.java	124	generic-all
-runtime/RedefineTests/RedefineInterfaceCall.java	267 generic-all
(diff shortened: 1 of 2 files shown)


sophia-guo:
When it's done, we can check #267 in https://github.com/AdoptOpenJDK/openjdk-tests/issues/345

karianna:
@sophia-guo Just to be clear - this is so you can compare the AdoptOpenJDK binary (built on `dev`) vs the 'pure' OpenJDK binary cloned from mercurial?

`dev` contains extra patches that we maintain over and above master...

sophia-guo:
@karianna we git clone from adoptopenjdk openjdk8u repo to get latest openjdk test materials, which guarantee we can always get newest test source code and newest test fixes. 
Patches for build binary are not necessary for test build. Besides seems dev is not sync with mercurial, some tests are stale in dev branch.

karianna:
@sophia-guo We should fix `dev` being stale in that case as that's the branch the binaries are built from that we should be testing.  I've opened https://github.com/AdoptOpenJDK/openjdk-build/issues/269

sophia-guo:
@karianna agreed that  dev branch should be in sync with openjdk mercurial especially in openjdk10   binary build will also build test image.

For regression tests we grab tests and run tests against any supported sdk binary.   A pure and in sync branch with openjdk mercurial should be the place to clone.

PR #364, opened 2018-03-28
Add functional test structure

sophia-guo:

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -2,7 +2,7 @@
 /* template jenkinsfile for adoptopenjdk test builds*/
 OPENJDK_TEST="$WORKSPACE/openjdk-tests"
 def getBuildList() {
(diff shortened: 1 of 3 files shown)


sophia-guo:
FYI @llxia

sophia-guo:
@llxia I just realized that we need test playlists before compile stage. I'd like to confirm that for openj9 tests build will you use openj9 testkitgen directly or use testkitgen from adopt indirectly?

llxia:
Yes, there are couple of things that we are missing. I updated functional/build.xml to include the subdirs build.xml in my local branch. The current failure seems related to jar files.
https://ci.eclipse.org/openj9/view/Test/job/Test-Sanity-Functional-JDK8-linux_390-64_cmprssptrs/4/console

Ideally, we should not directly use TKG that is nested under openj9/test. We should use TKG that is cloned at same level as functional folder, so we have unified story for all test types.

PR #372, opened 2018-04-04
use copy artifact plugin to copy compiled test binary

TianyuZuo:
* use copy artifact plugin to copy compiled test binary
* remove `/` from testSubDir


TianyuZuo:
reviewer: @llxia  @sophia-guo 

change set tested and pass on Jenkins

PR #378, opened 2018-04-06
set outputTapToConsole to false in TapPublisher

llxia:
- Avoid to echo TAP file content to console output. User can
check the result in the TAP UI.


llxia:
Personal build: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox2/176/console

PR #383, opened 2018-04-11
Add openjdk10-openj9 sdk specific TEST.ROOT

sophia-guo:
openjdk-jdk10/test/jdk/TEST.ROOT doesn't work with openjdk10-openj9. Use openjdk10-openj9/TEST.ROOT instead of openjdk-jdk10/test/jdk/TEST.ROOT to run test/jdk against openjdk10-openj9 sdk.

Manually update this file may be needed if
openjdk-jdk10/test/jdk/TEST.ROOT is updated or openjdk10-openj9
implemented with corresponding package related with system properties.
 
### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -163,7 +163,6 @@ def post(test_target) {
 				archiveArtifacts artifacts: "**/${test_target}_test_output.tar.gz", fingerprint: true, allowEmptyArchive: true
 			}
 			cleanWs cleanWhenFailure: false
-			
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


sophia-guo:
Note for myself: just for track the issue: do get some jenkins issue https://ci.adoptopenjdk.net/view/work%20in%20progress/job/openjdk10_j9_openjdktest_x86-64_linux/7/consoleFull

Which is related with [Durable Task Plugin](https://github.com/jenkinsci/durable-task-plugin/blob/master/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java

and it is the first time with this kind of issue. 
A new build is good https://ci.adoptopenjdk.net/view/work%20in%20progress/job/openjdk10_j9_openjdktest_x86-64_linux/8/console. If happens again will open an issue as a separate one.

Issue #380, opened 2018-04-09
Get latest sdk per platfom doesn't work any more


sophia-guo:
With updated sdk deployment story( deploy per platform) openjdk-api query about latest sdk per platform will fail.

https://api.adoptopenjdk.net/$JVMVERSION/$SDK_RESOURCE/$PLATFORM/latest/binary

The reason is some platform may not have latest binary.
For example: 
openjdk8 only have 4 platform available for latest ( April 9 for now: linux, ppcle, aix, windows)
https://api.adoptopenjdk.net/openjdk8/nightly

In that case if we want get latest mac sdk, which latest version is Mar 8, wget will fail as there is no sdk available. 
wget https://api.adoptopenjdk.net/openjdk8/nightly/x64_mac/latest/binary

sophia-guo:
Refer to https://github.com/AdoptOpenJDK/openjdk-api/issues/52

smlambert:
This has now been fixed by a set of PRs in the website/website-backend/build repos.

PR #384, opened 2018-04-13
Update download latest sdk from adoptopenjdk API

sophia-guo:
Use curl to download latest sdk from adoptopenjdk API
Add return check by download sdk name

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -99,23 +99,32 @@ parseCommandLineArgs()
 getBinaryOpenjdk()
 {
 	cd $SDKDIR
+	mkdir openjdkbinary
+	cd openjdkbinary
+	jar_file_name=""
 	if [[ "$CUSTOMIZED_SDK_URL" == "" ]]; then
 		if [[ "$SDK_RESOURCE" == "nightly" || "$SDK_RESOURCE" == "releases" ]]; then
 			echo 'Get binary openjdk...'
-			mkdir openjdkbinary
 			download_url="https://api.adoptopenjdk.net/$JVMVERSION/$SDK_RESOURCE/$PLATFORM/latest/binary"
-			wgetSDK
+			curl -OLJ $download_url
+			jar_file_name=`ls`
+			if [ "$jar_file_name" = "binary" ]; then
+				echo "No available jdk binary, exiting"
+				exit 1
+			fi
 		fi
 	else 
 		download_url=$CUSTOMIZED_SDK_URL
-		wgetSDK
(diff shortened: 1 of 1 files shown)


smlambert:
This assumes curl is available on all machines, which it could be, I am not sure.  wget was added to the prereqs list and is in ansible scripts, etc.  you could keep a function called getSDK, and use curl if it is present, if not fallback to wget.  I recognize that we need to use curl for z/OS, and it is likely sufficient for all other platforms, but should we consider a fallback plan?  

This PR then would imply we change our Prerequisites.md doc, if we no longer require wget (though I think we still do require it inside of build.xml files)

sophia-guo:
Yeah, didn't realize our prerequisites issue as [adoptOopenJDK](https://github.com/AdoptOpenJDK/openjdk-api) suggests curl. 
The hidden issue is if there is no latest sdk available  "wget https://api.adoptopenjdk.net/$JVMVERSION/$SDK_RESOURCE/$PLATFORM/latest/binary" won't return error code(non-zero), which fail the test build at later stage. A little bit annoying for debugging failures.

llxia:
> which fail the test build at later stage.

What is the current behavior? Will it fail at unzip?

sophia-guo:
Yes, un-tar should have problem.

sophia-guo:
The other choice is to check the size of sdk, which is just  not straightforward. Though should work.

smlambert:
What I was suggesting is that we use curl for all cases (not just adopt nightly & release URLs), unless we find that curl is not present on the machine, in which case attempt to fall back to use  wget.  

We can talk about it, I will not block this PR if we would want to investigate the proposal to use curl for all cases as another PR.  What you have in this PR is a dependency on curl (which we have not documented), our prereqs list only mentions wget (so we may need to update that).

sophia-guo:
Temporarily change it to WIP as this is not a blocking issue, which only give accurate error information if there is no nightly sdk. It's not a fix of #380  .

sophia-guo:
Rebase and retested  on arm64 and xlinux [ personal build 640-645](https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/644/).  Checked infrastructure curl is used as a base tool to configure other prerequisites.

llxia:
Just double check, curl is available on all machines? This script is used across sites (adopt, openj9 and internal).

smlambert:
If we are to believe what we are told, all projects use the ansible scripts at Adopt, so in theory if its present in the ansible scripts, it should be present on all machines.

PR #388, opened 2018-04-18
WIP: Add grinder and options to test_personal

smlambert:
- ITERATIONS is the number of times to execute a test target
- EXTRA_OPTIONS appends commandline options to the target being run
- JVM_OPTIONS is supposed to wholesale replace the commandline options being run

### DIFF
--- a/buildenv/jenkins/Jenkinsfile_personal
+++ b/buildenv/jenkins/Jenkinsfile_personal
@@ -78,13 +78,12 @@ pipeline {
 						if (env.BUILD_LIST == 'openjdk_regression' ||  env.BUILD_LIST == 'thirdparty_containers') {
 							DIAGNOSTICLEVEL ='noDetails'
 						}
-					
-						withEnv(["DIAGNOSTICLEVEL=${DIAGNOSTICLEVEL}"]) {
-							sh 'printenv'
-							int iterations = "${ITERATIONS}".toInteger()
-							for (i = 0; i < iterations; i++) {
+					}
+					withEnv(["DIAGNOSTICLEVEL=${DIAGNOSTICLEVEL}"]) {
+						sh 'printenv'
+						int iterations = "${ITERATIONS}".toInteger()
+						for (i = 0; i < iterations; i++) {
    								sh "$OPENJDK_TEST/maketest.sh $OPENJDK_TEST _${params.TARGET}"
-							}
 						}
 					}
 				}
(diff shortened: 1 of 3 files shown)


smlambert:
I will be changing the playlists shortly to have a < test > like that in jck/playlist.xml (jck-runtime-custom) to target individual test classes for personal test builds.

buildenv/jenkins/Jenkinsfile_personal
```diff
@@ -70,10 +76,14 @@ pipeline {
...
 							DIAGNOSTICLEVEL ='noDetails'
 						}
-					}
-					withEnv(["DIAGNOSTICLEVEL=${DIAGNOSTICLEVEL}"]) {
-						sh 'printenv'
-                		sh "$OPENJDK_TEST/maketest.sh $OPENJDK_TEST _${params.TARGET}"
+					
+						withEnv(["DIAGNOSTICLEVEL=${DIAGNOSTICLEVEL}"]) {
+							sh 'printenv'
+							int iterations = "${ITERATIONS}".toInteger()
```
sophia-guo: Not sure if the following script  needs  to be wrapped in script {}.

smlambert: A previous thing I tried did need it, I will try now without it to see if it works without.  Thanks!

smlambert: Need to keep the .toInteger() inside the block

smlambert:
I would additionally like to re-add the NODE_LABEL field (so that someone could pass in a particular node name, if they wanted to grind on a machine of their choice), but may do that in a separate pull request.

openjdk_regression/playlist.xml
```diff
@@ -14,6 +14,27 @@
 -->
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
 	<include>openjdk_regression.mk</include>
+	<test>
+		<testCaseName>openjdk-custom</testCaseName>
```
sophia-guo: Better use same name converntions as other testcases: jdk_custom. As we map testcasename or target to build_list.

smlambert: Yes, I was debating on that, or on making a standard naming that would be applied at each of the groups...  to match the jck-custom, openjdk-custom, perf-custom, external-custom, functional-custom... but those other types are difficult to customize, so I will use jdk_custom for now.

smlambert:
Replaced by #412

PR #395, opened 2018-04-24
update jck playlist to add JCK10 specific test buckets

TianyuZuo:
* update jck playlist to add JCK10 specific test buckets
* update Dockerfile to to use default keyboard layout
* update docker readme to make it more readable


llxia:
@ShelleyLambert Could you do a post review? I will merge this PR, so it does not block other tasks. If update is need, we can address it in another PR. Thanks.

PR #406, opened 2018-05-14
Enable tests on Windows

sophia-guo:


sophia-guo:
Depends on [PR in eclipse openj9](https://github.com/eclipse/openj9/pull/1910).
Related jobs are temporarily using this branch winNewSh, will update when dependent PR is merged.

openjdk_regression/build.xml
```diff
@@ -51,6 +51,17 @@
 	</target>
 	
 	<target name="getOpenjdk" depends="openjdk-jdk.check" unless="jdkdir.exists">
+		<!-- Windows API limitation of file paths having 260 characters or fewer,
+		     run the config command to allow file paths of 4096 characters 
+		-->
+		<if>
+			<contains string="${PLATFOMR}" substring="win" />
```
smlambert: is it a typo PLATFOMR vs PLATFORM ?

sophia-guo: Yes, it's typo. Thanks! Though confused why it works. Double checking ...

sophia-guo: Got it. When I tested on machine itself the configuration has been written to gitconfig file. So works even with this typo. Will fail when new machine enabled. The better way should be directly configure git itself on windows machines? Will keep this temporarily if reconfigure won't cause extra troubles.

sophia-guo:
Updated typo.

PR #424, opened 2018-05-23
Specify default jdk_custom test class

### LINKED ISSUE #423: Tests in openjdk8 regression tests builds are doubled
In nightly or releases tests builds CUSTOM_TARGET is empty, which causes all enabled tests are running under target jdk_custom. That is all tests are running twice. Builds time are doubled.

sophia-guo:
Specify default jdk_custom test class be a simple test to avoid running whole test set
Remove jtreg exclude options for jdk_custom target. In this way re-running or grinder jdk_custom target do not need any changes. 

close #423 


smlambert:
Thanks for catching this @sophia-guo !

PR #425, opened 2018-05-24
Enable playlists for Java 9 modularity tests in systemtest builds at Adopt

Mesbah-Alam:
This change adds the playlists for Java 9 modularity tests.

### DIFF
--- a/systemtest/playlist.xml
+++ b/systemtest/playlist.xml
@@ -2,7 +2,7 @@
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
 	
 	<test>
-		<testCaseName>ClassLoadingTest</testCaseName>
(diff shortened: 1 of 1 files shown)


sophia-guo:
You may need to do rebase as I can see the PR include other's commit.
The only difference between all testcases is testcase name and passing in options `-test` `-test-args`. I'm good as long as it passes personal build and works as expected.

The only suggestion is if we could add prefix - say `system_` to all testCaseName it would help mapping test TARGET to test category folder. That improves test flexibility. Though it could be done in a separate PR.

Mesbah-Alam:
Thanks for pointing it out Sophia. 

There was a problem with my branch , it contained someone else's commit. I recreated the PR in a new branch here: https://github.com/AdoptOpenJDK/openjdk-tests/pull/426

Mesbah-Alam:
@sophia-guo , about adding the prefix `system_`- I will add them in a separate PR if you need them. Will talk to you about it off line

PR #429, opened 2018-05-31
Add win32 and zos specs in ci script

smlambert:
Add ability to run personal testing against win32 & zos builds
Ability to incorporate Win32 and x64_linux_largeHeap to nightly build pipeline for Java8

### DIFF
--- a/buildenv/jenkins/Jenkinsfile_personal
+++ b/buildenv/jenkins/Jenkinsfile_personal
@@ -4,7 +4,7 @@ pipeline {
 	parameters {
 		string (defaultValue: "https://github.com/AdoptOpenJDK/openjdk-tests.git", description: 'personal repo from AdoptOpenJDK/openjdk-tests', name: 'ADOPTOPENJDK_REPO')
 		string (defaultValue: "master", description: 'personal AdoptOpenJDK branch', name: 'ADOPTOPENJDK_BRANCH')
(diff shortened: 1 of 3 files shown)


smlambert:
Related to https://github.com/AdoptOpenJDK/openjdk-build/pull/355

buildenv/jenkins/Jenkinsfile_personal
```diff
@@ -169,6 +169,14 @@ def getPlatformAndLabel(SPEC) {
 			platformAndLabel[0] = 'x64_win'
 			platformAndLabel[1] = 'hw.arch.x86&&sw.os.win'
 			break
+		case ~/win_x86/:
+			platformAndLabel[0] = 'x64_win'
+			platformAndLabel[1] = 'hw.arch.x86&&sw.os.win'
+			break
+		case ~/zos_390-64.*/:
+			platformAndLabel[0] = 's390x_zos'
```
sophia-guo: `PLATFORM` is used to retrieve nightly/releases sdk. Do we have zOS sdk published? Note if not using personal build launch tests on Zos should be careful. That is user can only use custmized URL to download existing zOS sdk.

sophia-guo:
We will also need to add separate PLATFROM for both Win32 and x64_linux_largeHeap in https://github.com/AdoptOpenJDK/openjdk-api. If eventually those sdk need to be published.

buildenv/jenkins/Jenkinsfile_personal
smlambert: You are right, we currently do not have an official nightly build of z/OS, so this was intended to only be used with customized, not nightly or released (in those latter 2 cases it will fail to find SDK, in much the same way as an openjdk9 will fail to find a release build, as there are none published, since they were cleared out).

smlambert:
Ok, I will ask you about what needs to change in the API repo tomorrow, thanks.

sophia-guo:
Regarding the PLATFORM probably need a discussion with infra or build people, which could do later and should not block this change.

llxia:
Approve the PR.
I agree, we need to discuss with infra/build team. In OpenJ9, build is called `XXX_cmprssptrs`. It is better to have an unified story.

Issue #446, opened 2018-06-13
Add getDependency job in Jenkins


llxia:
[getDependency](https://raw.githubusercontent.com/eclipse/openj9/master/test/TestConfig/scripts/tools/getDependencies.pl) is called every time we run functional test build. The script downloads test dependent jars from various urls. There is no guarantee these sites are available 24/7. Since we run functional test frequently (multiple builds per day), we need a stage job to pre-download the jars in case of these sites are down.

mesbaha:
We may want to think about scripting the "staging job" to take care of staging a super-set of prerequisite jars-- that include jars that `getDependency` script downloads + systemtest prereqs.

llxia:
- what are systemtest prereqs?
- Ideally, getDependency job should only be called when functional test is invoked. If system test needs similar job, it should be created separately.

Mesbah-Alam:
systemtest prereqs are listed here: https://github.com/AdoptOpenJDK/openjdk-systemtest/blob/master/openjdk.build/docs/build.md

Sure, we can have a separate job to stage the systemtest prereqs too.

llxia:
Jenkins job `test.getDependency` runs daily at midnight: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test.getDependency/

llxia:
PR #450 is merged.

PR #450, opened 2018-06-15
get pre-staged jars from test.getDependency build

llxia:
Issue: #446

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -145,7 +145,14 @@ def buildTest() {
 						env.PERF_PASSWORD = PASSWORD_VAR
 				}
 			}
-		
+
+			try {
+				//get pre-staged jars from test.getDependency build before test compilation
+				copyArtifacts fingerprintArtifacts: true, projectName: "test.getDependency", selector: lastSuccessful(), target: 'openjdk-tests/TestConfig/lib'
+			} catch (Exception e) {
+				echo 'Cannot run copyArtifacts from test.getDependency. Skipping copyArtifacts...'
+			}
+
 			if( params.JCK_GIT_REPO ) {
 				sshagent (credentials: ["$params.SSH_AGENT_CREDENTIAL"]) {
 					sh "$WORKSPACE/openjdk-tests/maketest.sh $WORKSPACE/openjdk-tests"


llxia:
- Jenkins job `test.getDependency` runs daily at midnight: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test.getDependency/

- Tested with jdk_math: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox2/236/

- Tested with invalid build name. Test should continue if the build does not exist. https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox2/235/

sophia-guo:
The point of pre-stage is in case that third-party web is unstable( I believe here especially asmtools from https://ci.adoptopenjdk.net). However if ci.adoptopenjdk.net is down like last weekend with per-staged  [test.getDependency](https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test.getDependency/)  all the Artifacts can not be accessed.

smlambert:
I assumed the point was that the other sites (in the getDependency.pl script) were unstable.  If you also want to protect against ci.adoptopenjdk.net being down, then the url for the test.dependency job should be a parameter that defaults to ci.adoptopenjdk.net, but we can override with the ci.eclipse.org/openj9 job.

sophia-guo:
Unrelated issue, seems your tests are running on AIX build machine with labeled `test` https://ci.adoptopenjdk.net/computer/build-osuosl-ppc64-aix-71-1/ @ShelleyLambert is that machine for test so we can enable tests on AIX?

smlambert:
Answer to the AIX question, I have enabled "weekly" tests on AIX build machine.  Though I need to change the job config to use Jenkinsfilebase instead of the personal file (as that seems to reset to use wrong params), if possible (though there is no upstream job and job number, they are triggered on a schedule).

sophia-guo:
I believe by now it only happens to asmtools from Adopt twice. I haven't heard any issue with other libs.  However, it may happen to others later.

@llxia mentioned "Ideally, getDependency job should only be called when functional test is invoked. If system test needs similar job, it should be created separately" in that case could we limit this step to function tests, i.e. only run this step when target is functional?
I

sophia-guo:
Thanks @ShelleyLambert I recalled the story of AIX.

llxia:
Yes, this work is mainly for OpenJ9 Jenkins test jobs. For AdoptOpenJDK Jenkins test jobs, we are at least protected from other unstable sites like Shelley mentioned.

llxia:
>"Ideally, getDependency job should only be called when functional test is invoked. If system test needs similar job, it should be created separately" in that case could we limit this step to function tests, i.e. only run this step when target is functional?

That is our long term goal, which needs a bit more work to be done in build.xml files https://github.com/eclipse/openj9/issues/1708
Currently all tests runs getDependency perl before compilation in make. I can limited the script to run copyArtifacts for just functional target, but current solution protects other component tests as well.

PR #459, opened 2018-06-22
Enable systemtest on z/OS

Mesbah-Alam:
Resolves: https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/139

### DIFF
--- a/buildenv/jenkins/Jenkinsfile_personal
+++ b/buildenv/jenkins/Jenkinsfile_personal
@@ -71,11 +71,6 @@ pipeline {
 				timestamps{
                 	echo 'Building tests...'
                 	script {
-	                	if (JAVA_VERSION == 'SE80') {
-							sh 'chmod 755 ${JAVA_BIN}/java'
-							sh 'chmod 755 ${JAVA_BIN}/../../bin/javac'
-							sh 'chmod 755 ${JAVA_BIN}/../../bin/java'
-						}
 						if ( SPEC.contains('win') ) {
 							bat "$OPENJDK_TEST/maketest.bat $OPENJDK_TEST"
 						} else {
--- a/buildenv/jenkins/openjdk_s390x_zos.xml
+++ b/buildenv/jenkins/openjdk_s390x_zos.xml
@@ -1,0 +1,22 @@
+#!groovy
+/* Template for test jobs on z/OS. Configure test job as parameterized.
+	Set parameter TARGET(openjdk, system, external, perf, jck etc.).
+	Set UPSTREAM_JOB_NAME(version_build_arch_os, for example: openjdk8_build_s390x)
+	Set JVM_VERSION(openjdk8, openjdk8-openj9, openjdk9, openjdk8-openj9, etc.*/
+LABEL='hw.arch.s390x&&sw.os.zos'
+
+node ("master") {
+	checkout scm
+	def jenkinsfile = load "${WORKSPACE}/openjdk-tests/buildenv/jenkins/JenkinsfileBase"
+	jenkinsfile.setLabelParam()
+	cleanWs()
+}
+
+node("$LABEL") {
+    PLATFORM = 's390x_zos'
+    SDK_RESOURCE = 'upstream'
+    SPEC='zos_390-64_cmprssptrs'
+    checkout scm
+    def jenkinsfile = load "${WORKSPACE}/openjdk-tests/buildenv/jenkins/JenkinsfileBase"
+    jenkinsfile.testBuild()
+}
--- a/get.sh
+++ b/get.sh
@@ -115,7 +115,12 @@ getBinaryOpenjdk()
 		fi
 	else 
 		download_url=$CUSTOMIZED_SDK_URL
-		wget -q --no-check-certificate --header 'Cookie: allow-download=1' ${download_url}
+		if [[ "$PLATFORM" == "s390x_zos" ]];
+		then
+			curl -O ${download_url} 
+		else
+			wget -q --no-check-certificate --header 'Cookie: allow-download=1' ${download_url}
+		fi
 		if [ $? -ne 0 ]; then
 			echo "Failed to retrieve the jdk binary, exiting"
 			exit 1
@@ -152,6 +157,8 @@ getBinaryOpenjdk()
 				mv $jar_dir_name j2sdk-image
 			fi
 		done
+		
+	chmod -R 755 j2sdk-image
 }
 
 getTestKitGenAndFunctionalTestMaterial()
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -153,12 +153,7 @@ def buildTest() {
 	stage('Build') {
 		timestamps{
 			echo 'Building tests...'
-			if (JAVA_VERSION == 'SE80') {
-				sh 'chmod 755 ${JAVA_BIN}/java'
-				sh 'chmod 755 ${JAVA_BIN}/../../bin/javac'
-				sh 'chmod 755 ${JAVA_BIN}/../../bin/java'
-			}
-		
+					
 			if ( params.PERF_CREDENTIALS_ID ) {
 				withCredentials([usernamePassword(credentialsId: "$params.PERF_CREDENTIALS_ID",
 					passwordVariable: "PASSWORD_VAR", usernameVariable: "USERNAME_VAR")]) {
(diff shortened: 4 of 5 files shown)


sophia-guo:
We will combine different java version jenkins files to a general one. That is each build will pass in JAVA_VERSION.  See https://github.com/AdoptOpenJDK/openjdk-tests/pull/483.

get.sh
```diff
@@ -115,7 +115,13 @@ getBinaryOpenjdk()
 		fi
 	else 
 		download_url=$CUSTOMIZED_SDK_URL
-		wget -q --no-check-certificate --header 'Cookie: allow-download=1' ${download_url}
+		if [[ "$PLATFORM" == "s390x_zos" ]];
```
sophia-guo: We have added curl as a prerequisite for all platforms. So this condition could be removed and use curl for all cases.

get.sh
```diff
@@ -115,7 +115,13 @@ getBinaryOpenjdk()
...
+		then
+			curl -OLJ ${download_url}
```
sophia-guo: OLJ options could be removed as it is only required when redirection is needed. Besides some lower curl version doesn't support it.

get.sh
```diff
@@ -115,7 +115,13 @@ getBinaryOpenjdk()
 		fi
 	else 
 		download_url=$CUSTOMIZED_SDK_URL
-		wget -q --no-check-certificate --header 'Cookie: allow-download=1' ${download_url}
+		if [[ "$PLATFORM" == "s390x_zos" ]];
+		then
+			curl -OLJ ${download_url} 
+			jar_file_name=`ls`
```
sophia-guo: Same this jar_file_name is  used to check if download is successful when redirection happens. Not used for this case please remove this line.

Mesbah-Alam: When we are using a customized sdk, e.g., when running in Grinder, jar_file_name needs to be defined here, I tested without it, and it fails as it can't find the sdk.

get.sh
Mesbah-Alam: We should use a different PR to switch to using Curl in all cases.

get.sh
Mesbah-Alam: Done.

smlambert:
I'd prefer you just put a generic openjdk_s390x_zos file (and not individual ones for 8,9,10), I will be merging #483 today, so you will be able to create and run builds with little delay.

Mesbah-Alam:
I've updated the PR with a generic `openjdk_s390x_zos` data file and removed the 3 Java version specific ones.

get.sh
sophia-guo: ah, that's because any issue, ```if [[ $jar_name == *zip || $jar_file_name == *jar ]]; then
				unzip -q $jar_file_name -d .
			```else
jar_file_name should be updated to jar_name, which is in @llxia PR.

llxia: https://github.com/AdoptOpenJDK/openjdk-tests/pull/486

buildenv/jenkins/openjdk_s390x_zos.xml
```diff
@@ -0,0 +1,22 @@
+#!groovy
+/* Template for test jobs on z/OS. Configure test job as parameterized.
+	Set parameter TARGET(openjdk, system, external, perf, jck etc.).
+	Set UPSTREAM_JOB_NAME(version_build_arch_os, for example: openjdk8_build_s390x)
+	Set JVM_VERSION(openjdk8, openjdk8-openj9, openjdk9, openjdk8-openj9, etc.*/
+LABEL='ci.project.openj9&&hw.arch.s390&&hw.arch.z14&&sw.os.zos'
```
llxia: - label is updated. `hw.arch.s390` will not have match. Please change it to `hw.arch.s390x`
- Please remove `ci.project.openj9`
- do we need to have `hw.arch.z14`?

buildenv/jenkins/Jenkinsfile_personal
```diff
@@ -72,9 +72,7 @@ pipeline {
                 	echo 'Building tests...'
                 	script {
 	                	if (JAVA_VERSION == 'SE80') {
-							sh 'chmod 755 ${JAVA_BIN}/java'
-							sh 'chmod 755 ${JAVA_BIN}/../../bin/javac'
-							sh 'chmod 755 ${JAVA_BIN}/../../bin/java'
+							sh 'chmod -R 755 $WORKSPACE/openjdkbinary/j2sdk-image/'
```
llxia: Is it possible to have this chmod in get.sh, so we do not need to do it in two separate files. By putting it in get.sh, it will solve the problem during local run as well.

get.sh
```diff
@@ -156,6 +162,11 @@ getBinaryOpenjdk()
 
 getTestKitGenAndFunctionalTestMaterial()
 {
+	if [[ "$PLATFORM" == "s390x_zos" ]];
+	then
+		OPENJ9_REPO="git://github.com/eclipse/openj9.git"
```
llxia: OPENJ9_REPO is a parameter, we do not need to hardcode here.

llxia: We can pass in as parameter from Jenkins file. That's how we do personal build.

buildenv/jenkins/openjdk_s390x_zos.xml
llxia: does the matched machines look right to you?
https://hyc-runtimes-jenkins.swg-devops.com/label/hw.arch.s390x&&sw.os.zos/

get.sh
Mesbah-Alam: OK that makes sense for personal builds. From where will the value of OPENJ9_REPO be passed for regular non-personal builds?

buildenv/jenkins/openjdk_s390x_zos.xml
Mesbah-Alam: Since I am not working on labels much, and they are being changed, I am not an expert on how to validate which combination of labels will find a match. For example I did not know you could use this URL to validate a combination. Thanks for pointing it out. Since you know how to confirm a combination - do you have an answer to your question - "do we need to have hw.arch.z14?" - ?

get.sh
llxia: You can create parameter in Jenkins file configuration and pass in value for OPENJ9_REPO
https://github.com/AdoptOpenJDK/openjdk-tests/blob/5416139f7a6c673e5ceea35b5153aa3e06feed93/buildenv/jenkins/JenkinsfileBase#L122

buildenv/jenkins/openjdk_s390x_zos.xml
Mesbah-Alam: Thanks for letting me know there is this easy way to validate combination of labels. The matched machines are right.

buildenv/jenkins/Jenkinsfile_personal
Mesbah-Alam: It should be possible. I'd have to test it with a build.

get.sh
Mesbah-Alam: Okay. Makes sense. But why then are we having a default value for this parameter in get.sh for regular case? Why aren't we passing that in via Jenkins file configuration too?

llxia: We need a default value in case this value is not provided.

Mesbah-Alam: So can we consider my change to be the default value for the case of z/OS?

llxia: If we make zos value default, then we have to change all job configuration to pass in https value for OPENJ9_REPO for all other platforms. It does not make sense to me.

Mesbah-Alam: Not asking to make the z/OS value default. I meant, the way the value is in this PR, can we leave it as is, instead of moving it to a data file (with the check for zos around it), as the "default" value when we are on z/OS.

buildenv/jenkins/Jenkinsfile_personal
Mesbah-Alam: I've moved the chmod to get.sh. Tested with z/OS and it works fine. Thanks for the suggestion.

get.sh
Mesbah-Alam: OK. Removed.

get.sh
Mesbah-Alam: Merged https://github.com/AdoptOpenJDK/openjdk-tests/pull/486 with this PR and removed `jar_file_name` from here.

Mesbah-Alam:
@lanx @sophia-guo  - is there any other change you would recommend? I'd like to deliver this PR along with the other 3 if not.

LanX:
Hi

I'm the wrong person : LanX not Lan Xia!

cheers
  LanX

2018-07-23 15:45 GMT+02:00 Mesbah Alam <notifications@github.com>:

Mesbah-Alam:
@LanX - corrected.

PR #460, opened 2018-06-26
WIP: Enable win

llxia:
- remove if wins, then .bat. We will use one .sh script
- remove WORKSPACE key word as it does not work in sh on wins
- change maketest.sh file permission so there is no need to do
chmod in Jenkins file


sophia-guo:
@llxia Could this be closed as a duplicate of https://github.com/AdoptOpenJDK/openjdk-tests/pull/506?

llxia:
yes, close this PR as it is delivered in #506

PR #471, opened 2018-07-09
[WIP] Update result zip logic for systemtest to only include junit xml reports

Mesbah-Alam:
Resolves https://github.com/AdoptOpenJDK/openjdk-tests/issues/470

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -215,12 +215,21 @@ def post(test_target) {
 		timestamps{
 			step([$class: "TapPublisher", testResults: "**/*.tap", outputTapToConsole: false])
 			junit allowEmptyResults: true, keepLongStdio: true, testResults: '**/work/**/*.jtr.xml, **/junitreports/**/*.xml'
+			
 			if (currentBuild.result == 'UNSTABLE') {
 				archiveArtifacts artifacts: '**/work/**/*.jtr, **/junitreports/**/*.xml', fingerprint: true, allowEmptyArchive: true
 				if ( SPEC.contains('win') ) {
-					bat "tar -zcf ${test_target}_test_output.tar.gz $WORKSPACE/openjdk-tests/TestConfig/test_output_*"
+					if (env.BUILD_LIST == 'external') {  
+						bat "tar -zcf ${test_target}_test_output.tar.gz $WORKSPACE/openjdk-tests/TestConfig/test_output_*/*/junitreports/*.xml"
+					} else {
+						bat "tar -zcf ${test_target}_test_output.tar.gz $WORKSPACE/openjdk-tests/TestConfig/test_output_*"
+					}
 				} else {
-					sh "tar -zcf ${test_target}_test_output.tar.gz $WORKSPACE/openjdk-tests/TestConfig/test_output_*"
+					if (env.BUILD_LIST == 'external') {
+						sh "tar -zcf ${test_target}_test_output.tar.gz $WORKSPACE/openjdk-tests/TestConfig/test_output_*/*/junitreports/*.xml"
+					} else {
+						sh "tar -zcf ${test_target}_test_output.tar.gz $WORKSPACE/openjdk-tests/TestConfig/test_output_*"
+					}
 				}
 				archiveArtifacts artifacts: "**/${test_target}_test_output.tar.gz", fingerprint: true, allowEmptyArchive: true
 			}
--- a/thirdparty_containers/lucene-solr/playlist.xml
+++ b/thirdparty_containers/lucene-solr/playlist.xml
@@ -15,7 +15,7 @@
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
 	<test>
 		<testCaseName>lucene_solr_nightly_smoketest</testCaseName>
-	<command>docker run --rm -v $(JDK_HOME):/java -v $(REPORTDIR)/junitreports/lucene-solr:/lucene-solr/lucene/build/core/test adoptopenjdk-lucene-solr-test:latest ; \
+	<command>docker run --rm -v $(JDK_HOME):/java -v $(REPORTDIR)/junitreports/:/home/jenkins/lucene-solr/lucene/build/core/test adoptopenjdk-lucene-solr-test:latest ; \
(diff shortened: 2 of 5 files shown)


Mesbah-Alam:
@karianna - could you please explain what you mean by "applying some DRY"?

Mesbah-Alam:
@karianna  - thanks for the suggestion. I can make that update. 

@sophia-guo - yes, the plan was to remove those changes before delivery of course. But it took longer than expected to test this change with some infra issues. And besides, I am dealing with a different related issue at the moment mentioned in my last comment here: https://github.com/AdoptOpenJDK/openjdk-tests/issues/329. So, this PR should be moved to WIP.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -215,12 +215,21 @@ def post(test_target) {
 		timestamps{
 			step([$class: "TapPublisher", testResults: "**/*.tap", outputTapToConsole: false])
 			junit allowEmptyResults: true, keepLongStdio: true, testResults: '**/work/**/*.jtr.xml, **/junitreports/**/*.xml'
+			
 			if (currentBuild.result == 'UNSTABLE') {
 				archiveArtifacts artifacts: '**/work/**/*.jtr, **/junitreports/**/*.xml', fingerprint: true, allowEmptyArchive: true
 				if ( SPEC.contains('win') ) {
-					bat "tar -zcf ${test_target}_test_output.tar.gz $WORKSPACE/openjdk-tests/TestConfig/test_output_*"
+					if (env.BUILD_LIST == 'external') {  
+						bat "tar -zcf ${test_target}_test_output.tar.gz $WORKSPACE/openjdk-tests/TestConfig/test_output_*/*/junitreports/*.xml"
```
smlambert: Not sure if you need to have the separation here.  Have you considered/tried just putting the output files directly into the test_output folder? and then changing the regex to search for all *.xml there?

In the case of openjdk regression, a.k.a., we generate .xml files and put them in the test_output folder, so you perhaps could change the regex for all cases, not just 'external', avoiding the need for the if/else handling.

thirdparty_containers/lucene-solr/playlist.xml
```diff
@@ -15,7 +15,7 @@
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
 	<test>
 		<testCaseName>lucene_solr_nightly_smoketest</testCaseName>
-	<command>docker run --rm -v $(JDK_HOME):/java -v $(REPORTDIR)/junitreports/lucene-solr:/lucene-solr/lucene/build/core/test adoptopenjdk-lucene-solr-test:latest ; \
+	<command>docker run --rm -v $(JDK_HOME):/java -v $(REPORTDIR)/junitreports/:/lucene-solr/lucene/build/core/test adoptopenjdk-lucene-solr-test:latest ; \
```
smlambert: You could try remove the junitreports dir from this command, to dump reports directly into test_output_xxxx dir (REPORTDIR)

buildenv/jenkins/JenkinsfileBase
Mesbah-Alam: @smlambert , Hi Shelley, We have different kinds of tests in thirdparty_containers, would it be ideal to dump output of all of them in the same folder?

(Right now, I am using $(REPORTDIR) while creating the mapping as the host folder which points to, e.g., <workspace>/openjdk-tests/TestConfig/test_output_15313396945308/lucene_solr_nightly_smoketest_0/. 

The reason to use "junitreports" is cos I didn't want to touch what's already being used in the rege x in  `archiveArtifacts artifacts` command in the Jenkinsfile.)

Mesbah-Alam: But either way, we need to fix this problem first: https://github.com/AdoptOpenJDK/openjdk-tests/issues/329#issuecomment-404172867. 

I've been trying to fix it by having a user called `jenkins` (with the same UID as the 'jenkins' user of the host machine) inside the docker container with proper rights, and have this user needs to own everything that happens inside the container, so that the junit report xmls that we are downloading off of the container onto the host machine are owned by `jenkins`-- otherwise we run into the issue of these files being owned by `root`, and `jenkins` can't touch them.

Mesbah-Alam: The easiest thing to do would be if we could somehow change the ownership of the xml report files from 'root' to 'jenkins' on the host machine itself, or change the file permission of them to readable to all. But since they are owned by 'root', 'jenkins' user can't do those. Is there a way jenkins could sudo to be root and do this @sxa555 ?

smlambert: Ok, I am slow today.  I am just realizing that you are tarring the contents in these lines, not sure why you need to do this, as I thought our discussion led us to the conclusion you do not need to archive these artifacts.  
Simply want to map the junit files to the host machines so that the Jenkins junit plugin can find and display them, which sounds like you are working through.

smlambert: I thought that we discussed in the issue you have opened for this that there is little value in archiving the output, so let's not.  Feedback we have from dev team is that they do not want to download, unzip and then access the output.  We just need a way to make triage faster as scrolling through reams of console output takes a long time.  One way to improve triage time would be 'allow us to see test results via the junit plugin'.

Mesbah-Alam: Thanks for reminding me! That settles it then. We are not going to zip the xml output that we map on disc. 

There however remains a different problem. Even if we don't zip them up, they still remain on disc on the host machine, and they can't be deleted by the 'jenkins' user. So, for example, if we map them anywhere inside the workspace directory of the build, then the workspace clean up will keep on failing due to this.

Mesbah-Alam:
I am closing this PR at this point since we do not want to archive xml reports but only want to display it in Jenkins output via the Junit plugin. The task mapping host user to docker container user so that clean up of mapped files can be done properly in jeknins workspace will be handled by https://github.com/AdoptOpenJDK/openjdk-tests/issues/475. Further discussion on this topic should happen there.

buildenv/jenkins/JenkinsfileBase
Mesbah-Alam: Thanks for the link @smlambert. Further discussion to be continued on this topic here: https://github.com/AdoptOpenJDK/openjdk-tests/issues/475

PR #476, opened 2018-07-12
Remove volume map in lucene-solr tests until user id mapping resolved

Mesbah-Alam:
Remove volume map in lucene-solr tests until user id mapping is resolved


Mesbah-Alam:
We need to revert the change we made to add volume mapping for lucene-solr tests until https://github.com/AdoptOpenJDK/openjdk-tests/issues/475 is fixed. This will allow us to continue running the external test builds without causing WS clean up issues.

PR #479, opened 2018-07-13
Support builds also with JRE image and enable JRE tests

### LINKED ISSUE #417: Support the fact that builds will produce JRE builds as well as JDK
There is a PR open against the build repo which will cause the builds to produce both JRE and full JDK artefacts. Any jobs that copy from this will need to tolerate this (and probably allow for picking up either version).

The PR is https://github.com/AdoptOpenJDK/openjdk-build/pull/331 (naming convention will be discussed in that PR so I'm not going to specify it here in case it goes out of date)

This build has been modified to build from the code in my PR so can be seen as an example: https://ci.adoptopenjdk.net/job/openjdk8_openj9_build_x86-64_linux/

### LINKED ISSUE #437: Make JRE test build ready
We'd like to enable JRE package test build ready as there was a request at OpenJ9 to provide JRE packages in addition to JDK packages. 

JRE package basically is a subset of JDK package. Reasonable smoke tests should be enough for now. Regression tests do define a jre sub group. We can use that group as our initial smoke tests. We probably need adding jre/smoke either as a separate group or separate level.

sophia-guo:
Support builds with both JRE as well as JDK image.

Enable jdk_math tests for JRE image produced by openjdk8-openj9. When others builds make jre image available we can easily turn on this test against different JVM version.

Fix #417 , fix #437 

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -27,6 +27,7 @@ def stageTestBinaries() {
 
 def setupEnv() {
 	env.JAVA_BIN = "$WORKSPACE/openjdkbinary/j2sdk-image/${(JAVA_VERSION == 'SE80') ? 'jre/' : ''}bin"
+	env.JRE_IMAGE = "$WORKSPACE/openjdkbinary/j2jre-image"
 	env.JAVA_HOME = "${JAVA_BIN}/.."
 	env.JAVA_VERSION = "${JAVA_VERSION}"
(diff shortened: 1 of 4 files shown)


sophia-guo:
[test with jdk_math_jre against openjdj8-openj9](https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox/423/)
[test without jdk_math_jre openjdk8, as openjdk8 build with only SDK enabled](https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox/423/)

PR #483, opened 2018-07-19
Java11 updates

smlambert:
- updates to enable Java11 testing


buildenv/jenkins/JenkinsfileBase
```diff
@@ -26,10 +25,15 @@ def stageTestBinaries() {
 }
 
 def setupEnv() {
+	// env.JAVA_VERSION = "${getJavaVersionFromJvmVersion()}"
```
smlambert: Leaving this as comment, as I wish to return to it next week in a separate PR

buildenv/jenkins/openjdk_x86-64_windows
```diff
@@ -0,0 +1,18 @@
+#!groovy
+LABEL='hw.arch.x86&&sw.os.win'
```
llxia: Should we use `sw.os.windows` according to https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/93?
Eclipse OpenJ9 is already updated to use `sw.os.windows`

llxia:
Approve the change so we can get generic files in. wins file is not used in Eclipse Openj9, so it can be addressed later.

buildenv/jenkins/openjdk_x86-64_windows
smlambert: Yes we should use sw.os.windows, I will update and merge.  Thanks!

smlambert:
Since I have not removed any of the versioned data jenkinsfiles, all the old builds will work as is (using the old files).  We decided to push this in now, before teh getJavaVersionFromJvmVersion() method is completely working, but I left the commented line in, as I plan to return to it this week, because then we could remove all of the versioned data jenkinsfiles and not have to make any updates to job config.

PR #491, opened 2018-07-26
Enable jdk_math tests for JRE Images

sophia-guo:
Enable jdk_math tests for JRE Image with
	- hotspot/openj9 impls
	- Java9and up version 


sophia-guo:
We already enabled this test for openjdk8-openj9. Some other version of JRE has been published https://adoptopenjdk.net/nightly.html . We should enable the test for those versions. 
Test failure is expected as @johnoliver new script to output both jres and jdks have not delivered.

PR #500, opened 2018-08-01
suppress output for tests that passed

sophia-guo:
Reduce console output while result files *.jtr are kept. 


sophia-guo:
Passed test name and number keep same, which will keep junit or tap result same. As we don't need to triage the passing test there is no need to keep detailed information.

Issue #490, opened 2018-07-26
Update script to support API v2


llxia:
Current script does not work with [API v2](https://api.adoptopenjdk.net/README)

With API v2, url contains special characters (i.e., ?, &, etc) which cannot read correctly by get.sh 

Grinder build: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder/93/console

with CUSTOMIZED_SDK_URL: https://api.adoptopenjdk.net/v2/binary/nightly/openjdk10?openjdkImpl=openj9&os=linux&arch=x64&release=latest&type=jdk

```
00:00:25.673 [Grinder] Running shell script
00:00:26.091 + /home/jenkins/workspace/Grinder/openjdk-tests/get.sh+  -s /home/jenkins/workspace/Grinder -tos=linux /home/jenkins/workspace/Grinder/openjdk-tests
00:00:26.092  -p x64_linux -v openjdk10-openj9 -r customized -c https://api.adoptopenjdk.net/v2/binary/nightly/openjdk10?openjdkImpl=openj9+ 
00:00:26.092 arch=x64
00:00:26.092 + type=jdk --openj9_repo https://github.com/eclipse/openj9.git --openj9_branch master
00:00:26.092 + release=latest/home/jenkins/workspace/Grinder@tmp/durable-0de63df1/script.sh: 2: /home/jenkins/workspace/Grinder@tmp/durable-0de63df1/script.sh: 
00:00:26.092 --openj9_repo: not found
```

List of issues that should be fixed:
- [ ] customized url should take special characters
- [ ] update to use curl (we can remove wget)
- [ ] update url for release and nightly

smlambert:
Thanks for capturing this in an issue @llxia !

sophia-guo:
Those special characters are eaten when passed in Jenkins sh command. So jenkinsfile need to be updated.
Not sure what does "update url for release and nightly" mean? In V1 we generate url with releases and nightly. For CUSTOMIZED_URL user need to generate URL according release type( releases or nightly) /v2/<request type>/<release type>/<version>

llxia:
>Those special characters are eaten when passed in Jenkins sh command. So jenkinsfile need to be updated.

Not only Jenkins file needs to be updated when parsing in url, but get.sh has to be update to have "" around.

[...]

PR #503, opened 2018-08-07
Update script to support API v2 in CUSTOMIZED_SDK_URL

llxia:
- add double quotes
- remove wget (simplify the code to use only curl)
- merge multiple if statements into one
- add echo messages in get.sh
- change maketest.sh permission to 755
- remove unnecessary chmod
- fix node label for zos

Issue: #490


llxia:
Tested:
- set CUSTOMIZED_SDK_URL to v2 api: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox/456/
- set CUSTOMIZED_SDK_URL to archive SDK url from upstream build: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox/457/
- set SDK_RESOURCE to nightly: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox/458/
- tested on zos internally (passed download and got error with sha not match which is not related to this PR)

llxia:
A separate PR will be created for updating `download_url="https://api.adoptopenjdk.net/$JVMVERSION/$SDK_RESOURCE/$PLATFORM/latest/binary"` to v2 API.

llxia:
fyi @jdekonin

get.sh
```diff
@@ -98,49 +98,42 @@ parseCommandLineArgs()
...
+		download_url="https://api.adoptopenjdk.net/$JVMVERSION/$SDK_RESOURCE/$PLATFORM/latest/binary"
+	else
+		download_url=""
+		echo "--sdkdir is set to $SDK_RESOURCE. Therefore, skip download jdk binary"
+	fi
+	
+	if  [ "${download_url}" != "" ]; then
+		echo "curl -OLJks "${download_url}""
+		curl -OLJks "${download_url}"
 		if [ $? -ne 0 ]; then
```
sophia-guo: This check may not work for curl -O options. That is saying when curl -OLJ fails to download SDK the return CODE is still 0.

llxia:
- added check for binary file and tested with invalid url
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox/463
- regular case:
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox/462/

This binary file check only catches invalid url for current release and nightly url which end with binary. I think we should come up with a better solution once nightly and release url is updated to API v2.

PR #508, opened 2018-08-13
add **/*.jar into CopyArtifact filter

llxia:


llxia:
SDK may be archived into *.jar format. Add **/*.jar into CopyArtifact filter, so we also copy in this case.

PR #510, opened 2018-08-14
Support variable number of parameters

sophia-guo:
Enhance maketest.sh support variable number of parameters as make target or make variables.

### DIFF
--- a/maketest.sh
+++ b/maketest.sh
@@ -24,9 +24,9 @@ if [ "$#" -eq 1 ];then
 		exit 1
 	fi
 	$MAKE compile
-elif [ "$#" -eq 2 ];then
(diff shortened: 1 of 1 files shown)


sophia-guo:
Yes, this can still run without extra parameters.
Personal build for without/with extra parameters:
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/708/
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/test_personal/709/

The change is related with https://github.com/AdoptOpenJDK/openjdk-tests/pull/501 to flexibly variable number of parameters in case that later we could have requirements to set multiple variables for one/more custom target. 

This file is used by jenkins job stage `build` and `test` underneath. We don't have --help or .md yet.

PR #512, opened 2018-08-14
Add use of username/passwd credential for use with custom sdk urls

jdekonin:
Signed-off-by: Joe deKoning <joe_dekoning@ca.ibm.com>

fyi @llxia @AdamBrousseau

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -139,7 +139,13 @@ def setup() {
 
 			dir( WORKSPACE) {
 				sshagent(credentials:["${params.USER_CREDENTIALS_ID}"], ignoreMissing: true) {
-					sh "$GET_SH_CMD"
+					if (params.CUSTOMIZED_SDK_URL_CREDENTIAL_ID) {
+						withCredentials([usernamePassword(credentialsId: "${params.CUSTOMIZED_SDK_URL_CREDENTIAL_ID}", usernameVariable: 'USERNAME', passwordVariable: 'PASSWORD')]) {
+					           sh "$GET_SH_CMD"
+					      }
+				        } else {
+					      sh "$GET_SH_CMD"
+                                        }
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


llxia:
- Ideally, username/password credential should pass into get.sh and get handled in one place. 
Talked with Joe, it seems we have trouble use username/password credential in get.sh. Joe, what was the error that you encountered? 

- curl does not have option `-J`. It will not break in current case. But if we pass in an URL with credential and need redirect in the future, it will not work. I know some of machines do not support `-J` option yet (https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/404). If you want to skip `-J` for now, could you add a comment?

- https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/Jenkinsfile_personal needs to be updated, so that Grinder can work. This can be addressed in a separated PR.

llxia:
fyi @ShelleyLambert @sophia-guo

jdekonin:
I no longer have the changes that did that work but error was simple enough, the credentials weren't coming through to get.sh at all.  I couldn't determine why as a simplified test it worked just fine.  But my suspicion is that is had something to do with the sshagent credentials wrapper.

We could add the `-J` now to be consistent with the other use case.  If that is preferred then I would like this pr put on hold until I can update the build farm machines I use to work with a newer version of curl.  SLES12, CentOS and RH6 machines default curl package does not support the use of `-J`. I do not have an eta on when I could have that completed and I would rather not break my testing ability.

I can update the Jenkinsfile_personal as well.  Looks like its a fair bit different than the JenkinsfileBase, I might need some aid in testing.

jdekonin:
So my simple testcase to show how it was broken only showed me that I needed to look again.  I have it working with this last rework.  I believe this addresses 2 of the points you brought up @llxia.  I'll work on the personal version in a separate PR if there no objections.

jdekonin:
Also important to note, USERNAME and PASSWORD reference with a withCredentials block will not be visible within job output.  Nor will it show in the printenv listing that is issue in get.sh within the withCredentials block.  When used within a withCredentials block output will appear like this `--user ***:***`. Outside the block USERNAME and PASSWORD have no value.

Issue #517, opened 2018-08-17
update label in windows scripts


llxia:
According to https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/93, Windows label should be `sw.os.windows`

https://github.com/AdoptOpenJDK/openjdk-tests/blob/2424010ee6835e99b0e9066602929dd2740feb54/buildenv/jenkins/openjdk8_x86-32_windows#L6

https://github.com/AdoptOpenJDK/openjdk-tests/blob/2424010ee6835e99b0e9066602929dd2740feb54/buildenv/jenkins/openjdk8_x86-64_windows#L6

PR #518, opened 2018-08-17
Change label to sw.os.windows

llxia:
Fix: #517 


llxia:
@ShelleyLambert Could you also update the windows machine label in AdoptOpenJDK Jenkins? Thanks.

sophia-guo:
I guess Jenkinsfile_personal also need to do the same update?

llxia:
yep, updated. Thanks

Issue #521, opened 2018-08-20
Remove build parameter JVM_VERSION


sophia-guo:
Originally JVM_VERSION( openjdk8, openjdk8-openj9, etc.) is a main path option of openjdk-api binary query https://github.com/AdoptOpenJDK/openjdk-api/blob/master/README.v1. Corresponding setting JVM_VERSION as a build parameter is reasonable. Now V2 api updated to use use openjdk version and openjdk_impl instead. 

In the TestKitGen openjdk version and openjdk_impl (namely java_versio, impl) are required variables, while JVM_VERSION is a generated variable based on java_version and impl.

To make the story simple and consistent I'd like to remove this parameter and corresponding set openjdk version and impl as required build parameters. Besides we are in the process of using generic jenkinsfile (not individual ones for different openjdk version 8,9,10)(https://github.com/AdoptOpenJDK/openjdk-tests/pull/483), which also request openjdk version is a required build parameter.

sophia-guo:
Personal build is an example https://github.com/AdoptOpenJDK/openjdk-tests/pull/522  https://github.com/AdoptOpenJDK/openjdk-tests/pull/519

sophia-guo:
just notice https://github.com/AdoptOpenJDK/openjdk-tests/pull/483 is using JVM_VERSION to get java version.

llxia:
getJavaVersionFromJvmVersion() is not used.
Currently, user needs to provide JAVA_VERSION for running generic jenkinsfile.

sophia-guo:
JAVA_VERSION is set per openjdk version Jenkinsfile. Current job will need to provide JAVA_IMPL. However if we plan to use generic jenkinsfile job will also need to provide JAVA_VERSION. Both will need to update jenkinsjob parameters I'd like to do it at once. 

@ShelleyLambert #483 planed to use JVM_VERSION to get java version though it's not used now. What's your opinion about add JAVA_VERSION and JAVA_IMPL as build parameters?

smlambert:
Fixed by #522 and #523

PR #522, opened 2018-08-20
Remove temp variable JVM_VERSION in JenkinsPersonal

sophia-guo:
This generated temp variable is not needed by personal build anymore. We have changed to use java version and impl directly as get.sh passing in parameters. 


sophia-guo:
@llxia remove JVM_VERSION from JenkinsfileBase will do in a separate PR, which I have created an issue in https://github.com/AdoptOpenJDK/openjdk-tests/issues/521

sophia-guo:
The difference is JVM_VERSION is a temp generated variable in JenkinsPersonal, which is not referenced any more due to the API v2 update. However in JenkinsfileBase JVM_VERSION is a build Parameter, which is used to generate JAVA_IMPL. I have detailed discription in #521

PR #524, opened 2018-08-21
Remove destination directory from git clone command and use exec dir

Mesbah-Alam:
Resolves https://github.com/AdoptOpenJDK/openjdk-tests/issues/513 

### DIFF
--- a/systemtest/build.xml
+++ b/systemtest/build.xml
@@ -29,13 +29,12 @@
 	</target>
 
 	<target name="clone_stf">
-		<exec executable="git" failonerror="false">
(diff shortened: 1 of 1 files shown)


systemtest/build.xml
```diff
@@ -90,22 +88,15 @@
 				</exec>
 			</then>
 		</if>
-		<if>
-			<contains string="${JVM_VERSION}" substring="openj9"/>
```
smlambert: Understand that this is redundant as that target is already called on line 158.

Mesbah-Alam: Hi @ShelleyLambert , yes. The redundant call was removed.

PR #531, opened 2018-08-28
Correct TEST.ROOT copy if logic

andrew-m-leonard:

Original change used "openjdk1?-openj9" but it seems that is not supported.
Changed it to use \<or\> operation.


andrew-m-leonard:
@smlambert Shelly, please can you review this? it seems original fix didn;t work...

sophia-guo:
My original suggest is `matches `  not `not equals`  `  <matches string="${JVM_VERSION}" pattern="openjdk1?-openj9"> ` .   Though `or` is good too.

PR #533, opened 2018-08-30
Add Thorntail (WildFly-Swarm) MicroProfile TCK tests to build

Mesbah-Alam:
Resolves https://github.com/AdoptOpenJDK/openjdk-tests/issues/532


Mesbah-Alam:
This build does a test run for these tests : https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox2/354/testReport/ 

One concern is that it takes quite a while to build thorntail and run these tests (scope for optimization).

smlambert:
I am thinking we should look at the tests defined in thirdparty_containers, and reconsider which should be tagged as sanity versus extended.  Then run the sanity set nightly, and the extended set weekly.

PR #515, opened 2018-08-16
Pull lucene-solr junit reports from docker container to host machine for jenkins reporting

Mesbah-Alam:
Resolves https://github.com/AdoptOpenJDK/openjdk-tests/issues/514

Signed-off-by: Mesbah <Mesbah_Alam@ca.ibm.com>

PR #534, opened 2018-08-30
WIP: Add jitaas to third party containers

harryyu1994:
- this enables testing for the soon-to-be-open-sourced JITaaS project
- the JITaaS project is still in private repos therefore the changes can't be tested through Jenkins personal yet
- the current Dockerfile first builds the JITaaS binaries then runs tests. (Currently only running openj9's sanity tests)

### DIFF
--- a/thirdparty_containers/jitaas/build.xml
+++ b/thirdparty_containers/jitaas/build.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0"?>
+<project name="jitaas" default="build" basedir=".">
+	<taskdef resource="net/sf/antcontrib/antlib.xml" />
+	<description>
+		Build JITaaS OpenJ9 Sanity Test Docker image
+	</description>
+
+	<!-- set properties for this build -->
+	<property name="DEST" value="${BUILD_ROOT}/thirdparty_containers/jitaas" />
+	<property name="dist" location="${DEST}/${JAVA_VERSION}" />
+	<property name="src" location="." />
+	<property name="jvm_version" location="${JVM_VERSION}" />
+
+	<target name="init">
+		<mkdir dir="${dist}"/>
+	</target>
+
+	<target name="clean_image" depends="init" description="clean jitaas openj9 test docker image if there is one">
+		<exec executable="docker">
+			<arg line="rmi -f jitaas-openj9-test" />
+		</exec>
+	</target>
+
+	<target name="build_image" depends="clean_image" description="build jitaas openj9 test docker image">
+		<exec executable="docker"  failonerror="true">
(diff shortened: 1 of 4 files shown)


harryyu1994:
Question: will there be a nightly build for JITaaS in OpenJDK? We can eliminate the build process in the Dockerfile if we do.
Question: how can this be manually tested?
Question: how will the test results be displayed?

smlambert:
Thanks for pulling this together @harryyu1994 !  I will try to answer some of your questions.

1) If JITaaS will be built into / as part of an openj9 build (which I believe it will be), then we will have nightly builds of it at AdoptOpenJDK, once the code is committed to the eclipse/openj9 repo. 

2) We can follow the [test instructions for running 'external' tests](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/doc/userGuide.md#set-environment-variables-configure-build-and-run-tests) or run a [Grinder/(test_personal) job](https://github.com/AdoptOpenJDK/openjdk-tests/wiki/How-to-Run-a-Personal-Test-Build-on-Jenkins).  If you are intending on running the test_personal job on the AdoptOpenJDK Jenkins server, I may need to add you to the test group.  If you want to run it on the internal Jenkins server, we should make sure we have some internal machines labeled with sw.tool.docker label (Docker installed on them).  By your PR (line 28&31 of the playlist.xml file), you are adding this test to the 'extended.external' target.  But you can also just run the 'jitaas' test target directly.

3) The test results are available in the console, but if we apply the same approach as this PR #515, we can display them in Jenkins in the same way as we do for normal builds (by pulling them out of the Docker image).

I will try to launch a test_personal job shortly from your fork/branch and see how far we get.

llxia:
Just couple of general comments:
- You can use following as example to set up your dockerfile and script
 https://github.com/AdoptOpenJDK/openjdk-tests/tree/master/buildenv/docker

- for functional test, you need `BUILD_LIST=functional`

- we are replacing JAVA_VERSION with JDK_VERSION eclipse/openj9#2786. Currently, both works for now, but it is better to switch now.

harryyu1994:
The reason I didnt set up the dockerfile and script like  https://github.com/AdoptOpenJDK/openjdk-tests/tree/master/buildenv/docker is because I'm building then testing inside docker. Once we have a JITaaS nightly build available, I will change everything. It's probably better to wait till we have JITaaS nightly builds available before I work on this further.

smlambert:
btw @llxia - irregardless of where the JITaaS testing lands (on real machines being preferred), @chandrams will want to run some functional/system tests in Docker (as part of the verification requirements of the Cloud team)

smlambert:
@harryyu1994 - do you still want this PR, or do you have a sufficient approach to JITaaS testing now?

harryyu1994:
We are able to do JITaaS testing on real machines so the docker testing isn't needed anymore.I'm going to close this PR for now.

PR #537, opened 2018-09-04
Add docker cp step to copy jenkins test result xmls to host for reporting

Mesbah-Alam:

### DIFF
--- a/thirdparty_containers/jenkins/playlist.xml
+++ b/thirdparty_containers/jenkins/playlist.xml
@@ -15,7 +15,10 @@
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
 	<test>
 		<testCaseName>jenkins_test</testCaseName>
-	<command>docker run --rm -v $(JDK_HOME):/java  adoptopenjdk-jenkins-test:latest "jvm pos neg"; \
+	<command>docker run -v $(JDK_HOME):/java --name jenkins-test adoptopenjdk-jenkins-test:latest; \
+			 docker cp jenkins-test:/jenkins/cli/target/surefire-reports $(REPORTDIR)/external_test_reports; \
+			 docker cp jenkins-test:/jenkins/core/target/surefire-reports $(REPORTDIR)/external_test_reports; \
+			 docker rm jenkins-test; \
 		$(TEST_STATUS)</command>
 		<subsets>
 			<subset>SE80</subset>


buildenv/jenkins/openjdk8_x86-64_linux_docker
```diff
@@ -3,7 +3,7 @@
 	Set parameter TARGET(openjdk, system, external, perf, jck etc.).
 	Set UPSTREAM_JOB_NAME(version_build_arch_os, for example: openjdk8_build_x86-64_linux)
 	Set JVM_VERSION(openjdk8, openjdk8-openj9, openjdk9, openjdk8-openj9, etc.*/
-node('linux&&x64&&test&&sw.tool.docker') {
+node('test-packet-ubuntu1604-x64-1&&test-packet-ubuntu1604-x64-3&&test-scaleway-ubuntu1604-x64-1') {
```
llxia: Is this intended? We should not hardcode machine name.  And I do not think this will match any machine.

thirdparty_containers/jenkins/playlist.xml
```diff
@@ -15,7 +15,9 @@
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
 	<test>
 		<testCaseName>jenkins_test</testCaseName>
-	<command>docker run --rm -v $(JDK_HOME):/java  adoptopenjdk-jenkins-test:latest "jvm pos neg"; \
+	<command>docker run --rm -v $(JDK_HOME):/java --name jenkins-test adoptopenjdk-jenkins-test:latest "jvm pos neg"; \
+			 docker cp jenkins-test:/jenkins/cli/target/surefire-reports $(REPORTDIR)/external_test_reports; \
+			 docker cp jenkins-test:/jenkins/core/target/surefire-reports $(REPORTDIR)/external_test_reports; \
```
llxia: should we remove the image at the end?

smlambert: Given that we have space issues on test machines, I think it might be the right thing to do (remove image, at least in the case of successful runs).

Mesbah-Alam: Good catch @smlambert ! I missed the step to remove the image! Added it.

buildenv/jenkins/openjdk8_x86-64_linux_docker
Mesbah-Alam: Please ignore this change. I only did it for testing purposes to avoid a failing machine.

llxia: Could you remove it from PR? If we merge the PR, this change will be in.

Mesbah-Alam: @llxia - of course. The idea is to remove it before merging - which is what I have done.

Mesbah-Alam:
@ShelleyLambert thanks! I've removed `"jvm pos neg"` and re-ran the test and it works fine: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/TestBuild_Sandbox2/362/testReport/

Those options must have been carry overs from other applications. I've removed them from the Jenkins test.

PR #546, opened 2018-09-10
Enable SE100 and SE110 on external tests

Mesbah-Alam:
Enable SE100 and SE110 on external tests

### DIFF
--- a/thirdparty_containers/derby/playlist.xml
+++ b/thirdparty_containers/derby/playlist.xml
@@ -17,9 +17,6 @@
 		<testCaseName>derby_test_junit_all</testCaseName>
 	<command>docker run --rm -v $(JDK_HOME):/java  adoptopenjdk-derby-test:latest ; \
 		$(TEST_STATUS)</command>
-		<subsets>
-			<subset>SE90</subset>
-		</subsets>
 		<levels>
 			<level>sanity</level>
 		</levels>
--- a/thirdparty_containers/elasticsearch/playlist.xml
+++ b/thirdparty_containers/elasticsearch/playlist.xml
@@ -28,10 +28,6 @@
 			docker rm elasticsearch-test; \
 			$(TEST_STATUS)
 		</command>
-		<subsets>
-			<subset>SE80</subset>
-			<subset>SE90</subset>
-		</subsets>
 		<levels>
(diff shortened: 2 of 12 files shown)


sophia-guo:
Should be ` SE100`

smlambert:
If you have checked that all of those applications work against an openjdk10 and openjdk11, then you can actually remove the < subsets > block entirely.  When it is not present, testkitgen generates targets for all versions (so we won't have to update when openjdk12 comes along, unless the applications do not support it... in which case we are back to explicitly entering each version as a subset).  Also, currently it is SE100 (for openjdk10), not SE10.  

There is a PR coming soon to stop using the old SExx versions, but its not committed yet.  When it gets committed, we can use <subset>8</subset> etc.  without the weirdness of SE prefix and an extra zero suffix.

Mesbah-Alam:
Grinder to test extended.external tests (e.g. mp tcks) on JDK11: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder/197/parameters/

smlambert:
@Mesbah-Alam - I believe the jdk11 image is now available at docker hub.  Can you please confirm, and merge this if so?  Thanks!

Mesbah-Alam:
New Grinder launched to check which external test apps compile with jdk11 https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder/272/consoleFull

Mesbah-Alam:
external test extended build runs fine with JDK11 hotspot - the thorntail test fails - but openliberty and payara passes - So, I'll enable the extended external test on jdk11 : https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder/272/tapResults/

Mesbah-Alam:
No out-of-the-ordinary behavior noticed running `sanity.external` target on jdk11 either : https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder/282/console. 

We should be good merging this PR. This will mean that now the external test builds on jdk11 will be executing the tests instead of running empty.

Mesbah-Alam:
Note: I've updated the PR after it was approved and removed the `subsets` blocks completely from the playlist entries - as with this change, the tests are now effectively supporting all 4 subsets.

PR #560, opened 2018-09-18
Update problem list from recent test results

ben-walsh:
Job which produced results - https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/325

Signed-off-by: Ben Walsh ben_walsh@uk.ibm.com


smlambert:
This PR needs to follow the same format as https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/openjdk_regression/ProblemList_openjdk8-openj9.txt where you include not only the excluded file name, but also issue number and platform(s) on which to exclude.

ben-walsh:
Update all entries to include placeholder issue "000" and "generic-all" for platform.

PR #562, opened 2018-09-18
Add debug information to the JDK download curl

sxa555:
Will be useful to diagnose https://github.com/AdoptOpenJDK/openjdk-tests/issues/561

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -154,7 +154,9 @@ getBinaryOpenjdk()
 			echo "curl -OLJks ${curl_options} $file"
 			curl -OLJks ${curl_options} $file
 			if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


AdamBrousseau:
Does Curl exit non-zero if the file fails to download?
I've found one of my old scripts where I check the return code is 200. This curl is for the Github API but I think the http_code part is what we're interested in.
```
RESULT=$(curl --write-out %{http_code} -H "Authorization: token $TOKEN" -H 'Accept: application/vnd.github.v3.raw' -O -L $URL)
if [ "$RESULT" != "200" ]
then
    # Failed
fi
```

sophia-guo:
Correct, if file fails to download the curl return code is non-zero.

PR #567, opened 2018-09-19
Always Clean Workspace after build

AdamBrousseau:
- Wrap all steps in try/finally.
- Move CleanWs to finally block.
- This ensures CleanWs always runs even if
  the build has failed a step.
- This also reverts #320 which stopped the
  clean from running if the build failed.
- Proper archiving should ensure we don't need
  to leave workspaces around for test failure
  investigation.

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -220,7 +220,6 @@ def post(test_target) {
 				sh "tar -zcf benchmark_test_output.tar.gz ${benchmark_test_output_dir}"
 				archiveArtifacts artifacts: '**/benchmark_test_output.tar.gz', fingerprint: true, allowEmptyArchive: true
 			}
-			cleanWs cleanWhenFailure: false
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


AdamBrousseau:
Marking as WIP to give time for discussion.
@llxia @sophia-guo @ShelleyLambert

AdamBrousseau:
I agree with your first 2 sentences. I'm not sure I see the value the workspace will provide in the cases where a test has failed hard. The situations I can think of when a hard failure will occur:
- Timeout
- Build aborted
- Machine config issue
- Network issue
Unless I'm missing a use case, I don't see the value in keeping or archiving the workspace in any of these. Maybe there's a case where a failed test will actually cause the test harness to fall over?

The issue I'm trying to solve is machines running out of space because test workspaces getting left behind.

llxia:
I would suggest to have a flag added in finally block, so that we can turn on and off to fit different needs.

AdamBrousseau:
Updated with `KEEP_WORKSPACE`

PR #570, opened 2018-09-19
To update the jtreg exclude lists on jdk11 with the lists from jdk8

adamfarley:
The tests excluded from the jdk8 test runs should not be run against
jdk11 until the issues that they are raised against have been resolved.

### DIFF
--- a/openjdk_regression/ProblemList_openjdk11-openj9.txt
+++ b/openjdk_regression/ProblemList_openjdk11-openj9.txt
@@ -21,10 +21,100 @@
 ############################################################################
 
 # jdk_lang
+jdk/lambda/vm/InterfaceAccessFlagsTest.java	126	linux-all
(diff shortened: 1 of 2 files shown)


karianna:
As discussed on Slack, I'm pretty sure each version should maintain a mutually exclusive list.  That is sometimes issues can be fixed in a later version but not backported and so forth.

I could be wrong on this and so it's worth checking with Oracle on this assumption.

adamfarley:
I agree with the concept, but only for existing exclude lists. 

If a test is excluded from version 8, then version 9 should begin with the same exclude list until determined otherwise. These tests can then be removed from *just* 9, or even *just* 8, as needed.

My issue covers the concept that each new version shouldn't start with us re-discovering the same failures all over again. It adds work we don't need, and leads to us ignoring all of the jtreg failures until we have the time to triage the same failures all over again.

Plus, starting with a green screen makes it easy for us to spy a newly-failing test. Starting with 300 failures makes it difficult to notice the 301st failure.

adamfarley:
Closing.

PR #584, opened 2018-09-24
Support for jitaas testing

renfeiw:
- Invoke teardown before and after to ensure jitaas is killed

related PR: https://github.com/eclipse/openj9/pull/3059

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -24,6 +24,15 @@ def stageTestBinaries() {
 	sh "ls"
 }
 
+def makeTest(testParam) {
+	try {
+		sh "make -f ./openjdk-tests/TestConfig/testEnv.mk testEnvTeardown"
+		sh "./openjdk-tests/maketest.sh $testParam" 
+	} finally {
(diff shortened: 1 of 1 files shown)


sophia-guo:
Is this WIP? Confused that maketest_jitaas.sh is exactly same as maketest.sh?

renfeiw:
Yes, maketest_jitass.sh is same as maketest.sh for now. It's a template for the JIT team to add pretest and posttest targets. Once this is in, please feel free to add them. @harryyu1994
IMO the server watchdog should be added in the maketest_jitass.sh, as the Jenkins file does not have visibility of what subprocesses are running in the makefile.

AdamBrousseau:
While I agree that this approach makes sense from a test harness perspective, from an infrastructure perspective, my concern is that if the makefile has started the server, and then has a hard failure during the tests, the cleanup command will not get executed and the server will continue to run.

smlambert:
Is there a way of thinking about this, naming this more generically?  jitaas may not be the only feature in the future to need additional things set up in advance of the test execution.  Most modern test harnesses have a setup/execute/teardown piece.  Calling the script make_jitaas may be confusing if we have other features later that also require some sort of pre-test setup.

llxia:
This cannot be merged before https://github.com/eclipse/openj9/pull/3059

Issue #583, opened 2018-09-24
Include system test in thirdparty containers


chandrams:
Include system test in third party container testing
- Include dockerfile and script to clone stf, adopt/openjdk-systemtest and eclipse/openj9-systemtest
- Build.xml and playlist.xml
- Ensure results are copied to host
- Results use TAP format, this might have an implication on overall report for external tests as most of them use junit

chandrams:
Tried running the mauve tests but they fail, will create a separate issue to fix this. 

LT  Test failed:
LT  java.lang.ClassNotFoundException: gnu.testlet.java.util.AbstractCollection.AcuniaAbstractCollectionTest
LT      at java.lang.Class.forNameImpl(Native Method)
LT      at java.lang.Class.forName(Class.java:339)
LT      at net.adoptopenjdk.loadTest.adaptors.MauveAdaptor.executeTest(MauveAdaptor.java:50)
LT      at net.adoptopenjdk.loadTest.LoadTestRunner$2.run(LoadTestRunner.java:177)
LT      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
LT      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
LT      at java.lang.Thread.run(Thread.java:813)
LT  <<<
LT
LT  07:35:33.242 - Test failed. Details recorded in execution log.
LT  07:35:33.565 - Test failed. Details recorded in execution log.
LT  07:35:34.694 - Thread completed. Suite=0 thread=0
LT  07:35:34.781 - Load test completed
LT  07:35:34.782 -   Ran     : 3864
LT  07:35:34.782 -   Passed  : 3861
LT  07:35:34.782 -   Failed  : 3

Include NioLoadTest to test my changes submitted a personal build

Mesbah-Alam:
Copy-pasting from slack for records.. 

The issue that was causing failure in compilation in systemtestes inside your docker container has been fixed  (https://github.com/AdoptOpenJDK/openjdk-systemtest/pull/186)). See this Grinder for a successful run of docker based system test:  https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/351/console

chandrams:
Closing this issues as the initial set of system tests to run in docker containers has been added.

PR #606, opened 2018-10-01
Included system test mauve in thirdparty containers

chandrams:
Included system test mauve in thirdparty containers

issue #583 

Signed-off-by: Chandrakala <chandra-ms@in.ibm.com>

### DIFF
--- a/thirdparty_containers/system-test/dockerfile/Dockerfile
+++ b/thirdparty_containers/system-test/dockerfile/Dockerfile
@@ -1,0 +1,97 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This Dockerfile in thirdpart_containers/system-test/dockerfile dir is used to create an image with
+# AdoptOpenJDK jdk binary installed. Basic test dependent executions
+# are installed during the building process.
+#
+# Build system: `docker build -t adoptopenjdk-system-test .`
+#
+# This Dockerfile builds image based on adoptopenjdk/openjdk8:latest.
+# If you want to build image based on other images, please use
+# `--build-arg list` to specify your base image
+#
+# Build systemtest: `docker build --build-arg IMAGE_NAME=<image_name> --build-arg IMAGE_VERSION=<image_version > -t adoptopenjdk-system-test .`
+
+ 
+ARG SDK=openjdk8
+ARG IMAGE_NAME=adoptopenjdk/$SDK
+ARG IMAGE_VERSION=latest
+
+FROM $IMAGE_NAME:$IMAGE_VERSION
+
+# Install test dependent executable files
+RUN apt-get update \
+	&& apt-get -y install \
+	ant \
+	apt-transport-https \
+	ca-certificates \
+	dirmngr \
+	curl \
+	git \
+	make \
+	unzip \
+	vim \
+	wget \
+	cpio \
+	gcc
+
+# Install ant
+ENV ANT_VERSION 1.10.5
+RUN wget --no-check-certificate https://www.apache.org/dist/ant/binaries/apache-ant-${ANT_VERSION}-bin.tar.gz && \
+    gunzip apache-ant-${ANT_VERSION}-bin.tar.gz && \
+    tar -xvf apache-ant-${ANT_VERSION}-bin.tar
+
+RUN apt-get -y install \
+        libtext-csv-perl \
+        libjson-perl
+
+
+VOLUME ["/java"]
+#ENV JAVA_HOME=/java \
+#    PATH=/java/bin:$PATH \
+#    JAVA_TOOL_OPTIONS="-Dfile.encoding=UTF8"
+ENV  JAVA_TOOL_OPTIONS="-Dfile.encoding=UTF8"
+
+
+# This is the main script to run system tests.  
+COPY ./dockerfile/system-test.sh /system-test.sh
+
+# Clone the repo where the 3rd party application code and tests reside. 
+WORKDIR /
+RUN pwd
+
+ENV ANT_HOME ${WORKDIR}/apache-ant-${ANT_VERSION}
+
+RUN wget https://sourceforge.net/projects/ant-contrib/files/ant-contrib/1.0b3/ant-contrib-1.0b3-bin.tar.gz && \
+    tar -zxvf ant-contrib-1.0b3-bin.tar.gz && \
+    cp ant-contrib/ant-contrib-1.0b3.jar ${ANT_HOME}/lib
+
+
+# Clone AdoptOpenJDK openjdk-tests repo
+RUN git clone https://github.com/AdoptOpenJDK/openjdk-tests.git
+
+ENV JAVA_BIN=$JAVA_HOME/bin
+ENV SPEC=linux_x86-64_cmprssptrs
+ENV BUILD_LIST=systemtest
+
+RUN cd /openjdk-tests && \
+    ./get.sh -t /openjdk-tests  
+
+# Generates test targets applicable to the platform, implementation and version
+RUN echo "Generating test targets applicable to the platform, implementation and version..." && \
+    cd /openjdk-tests/TestConfig && \
+    make -f run_configure.mk && \
+    make compile
+
+ENTRYPOINT ["/bin/bash", "/system-test.sh"]
+CMD ["--version"]
--- a/thirdparty_containers/system-test/dockerfile/system-test.sh
+++ b/thirdparty_containers/system-test/dockerfile/system-test.sh
@@ -1,0 +1,66 @@
+#/bin/bash
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#Set up Java to be used by the system-test
+
+if [ -d /java/jre/bin ];then
+	echo "Using mounted Java8"
+	export JAVA_BIN=/java/jre/bin
+	export JAVA_HOME=/java
+	export PATH=$JAVA_BIN:$PATH
+	java -version
+elif [ -d /java/bin ]; then
+	echo "Using mounted Java9"
+	export JAVA_BIN=/java/bin
+	export JAVA_HOME=/java
+	export PATH=$JAVA_BIN:$PATH
+	java -version
+else
+	echo "Using docker image default Java"
+	java_path=$(type -p java)
+	suffix="/java"
+	java_root=${java_path%$suffix}
+	export JAVA_BIN="$java_root"
+	echo "JAVA_BIN is: $JAVA_BIN"
+
+	java_home=$(dirname ${JAVA_BIN})
+	export JAVA_HOME=$java_home
+
+	$JAVA_BIN/java -fullversion > version.log 2>&1
+        if grep  "1.8.0" version.log
+        then
+               JAVA_BIN="$java_home/jre/bin"
+	       JDK_VERSION="8"
+               echo "Set JAVA_BIN to $JAVA_BIN for Java 8"
+        fi
+        rm version.log
+fi
+
+TEST_SUITE=$1
+
+echo "PATH is : $PATH"
+echo "JAVA_HOME is : $JAVA_HOME"
+echo "type -p java is :"
+type -p java
+echo "java -version is: \n"
+java -version
+
+testparam=$@
+
+test=_$testparam
+
+echo "Running test $test..."
+#Run tests
+sh /openjdk-tests/maketest.sh /openjdk-tests $test
+
--- a/thirdparty_containers/system-test/playlist.xml
+++ b/thirdparty_containers/system-test/playlist.xml
@@ -1,0 +1,273 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+<playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
+	 <test>
+                <testCaseName>MauveSingleThreadLoadTest</testCaseName>
+                <command>docker run --name mauve-system-test adoptopenjdk-system-test:latest MauveSingleThreadLoadTest; \
+                 docker rm mauve-system-test; \
+                $(TEST_STATUS)</command>
+                <subsets>
+                        <subset>8</subset>
+                        <subset>9</subset>
+                </subsets>
+                <levels>
+                        <level>sanity</level>
+                </levels>
+                <groups>
+                        <group>system</group>
+                </groups>
+        </test>
+
+	 <test>
+                <testCaseName>MauveSingleInvocationLoadTest</testCaseName>
+                <command>docker run --name mauve-system-test adoptopenjdk-system-test:latest MauveSingleInvocationLoadTest; \
+                 docker rm mauve-system-test; \
+                $(TEST_STATUS)</command>
+                <subsets>
+                        <subset>8</subset>
+                        <subset>9</subset>
+                </subsets>
+                <levels>
+                        <level>sanity</level>
+                </levels>
+                <groups>
+                        <group>system</group>
+                </groups>
+        </test>
+
+	<test>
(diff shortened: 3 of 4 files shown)


chandrams:
@smlambert - Can you please review this. Thanks.

chandrams:
Grinder run - https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder/247/consoleFull

I 'm on vacation next week, will address any review comments once I 'm back.

thirdparty_containers/system-test/playlist.xml
```diff
@@ -0,0 +1,36 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+<playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
+	<test>
+		<testCaseName>system-test</testCaseName>
```
smlambert: Will assume the intent be to add other system-tests (beyond mauve).  If so, then testCaseName as system-test is fine, otherwise it may be good to give a more specific name for this target (system-test-mauve).

thirdparty_containers/system-test/playlist.xml
```diff
@@ -0,0 +1,36 @@
...
+		<variations>
+			<variation>-XX:+UseContainerSupport</variation>
+		</variations>
+		<command>docker run -v $(JDK_HOME):/java --name system-test adoptopenjdk-system-test:latest ; \
+		 docker cp system-test:/tmp/stf  $(REPORTDIR)/external_test_reports; \
```
smlambert: Is the content of external_test_reports all of the stf output plus the generated .tap file?

Mesbah-Alam: I am not sure about what the content of /tmp/stf here would be. System tests do not generate any junit xml report like the application tests. So the `docker cp` command here can be removed for the time being. When /if we update system tests to generate reports, we can have this step added back in.

Mesbah-Alam:
@chandrams - the reason why corba files are not deleted in the case of running systemtests in docker is that the check which decides whether to delete corba files (for jdk11 and up) or to keep them depends on JDK_VERSION variable - and that variable gets set by the build but inside the docker env, it doesn't get exported. 

I am actually in the process of updating that logic in mauve build.xml to not depend on JDK_VERSION but on an ant property -- this in theory should solve the issue. I'll update you when I have the fix ready.

chandrams:
I 'm making the changes to use testkitgen instead of system test make files as suggested by @smlambert.

chandrams:
@smlambert - I have made the changes as discussed and included shared classes tests too. Please review. Thanks.

https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder/413/console
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder/415/console

chandrams:
@smlambert - Hi, did you get a chance to review this?

thirdparty_containers/system-test/dockerfile/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+# Build system: `docker build -t adoptopenjdk-system-test .`
+#
+# This Dockerfile builds image based on adoptopenjdk/openjdk8:latest.
+# If you want to build image based on other images, please use
+# `--build-arg list` to specify your base image
+#
+# Build systemtest: `docker build --build-arg IMAGE_NAME=<image_name> --build-arg IMAGE_VERSION=<image_version > -t adoptopenjdk-system-test .`
+
+ 
+ARG SDK=openjdk8
```
smlambert: do you wish to hardcode this?

smlambert:
Sorry for my delay @chandrams !  On a quick look, I definitely prefer this approach, I will complete a review today.

thirdparty_containers/system-test/playlist.xml
```diff
@@ -0,0 +1,50 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+<playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
+	<test>
+		<testCaseName>mauve-system-test</testCaseName>
+		<variations>
+			<variation>-XX:+UseContainerSupport</variation>
```
smlambert: Given this is an openj9 only option, another reason to explicitly state the `<impls><impl>openj9</impl></impls>` in this `<test>` block, otherwise the underlying mauve tests may try to execute against hotspot and fail

thirdparty_containers/system-test/playlist.xml
```diff
@@ -0,0 +1,50 @@
...
+		</variations>
+		<command>docker run -v $(JDK_HOME):/java --name mauve-system-test adoptopenjdk-system-test:latest mauve; \
+		 docker rm mauve-system-test; \
+		$(TEST_STATUS)</command>
+		<subsets>
+			<subset>SE80</subset>
+			<subset>SE90</subset>
```
llxia: Please change it to 
```
<subset>8</subset>
<subset>9</subset>
```

thirdparty_containers/system-test/dockerfile/system-test.sh
```diff
@@ -0,0 +1,92 @@
...
+	make _SharedClassesWorkloadTest_Softmx_Increase_JitAot_Linux
+	make _SharedClasses.SCM01.SingleCL
+	make _SharedClasses.SCM01.MultiCL
+	make _SharedClasses.SCM01.MultiThread
+	make _SharedClasses.SCM01.MultiThreadMultiCL
+	make _SharedClasses.SCM23.SingleCL
+	make _SharedClasses.SCM23.MultiCL
+	make _SharedClasses.SCM23.MultiThread
+	make _SharedClasses.SCM23.MultiThreadMultiCL
+fi
```
llxia: Could we make this more generic? Instead of passing `TEST_SUITE` and redefine test groups in the file, can we just reuse https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/maketest.sh? In this way, we can pass in any existing test targets.

https://github.com/AdoptOpenJDK/openjdk-tests/blob/ac0012d53f1223456c4b8dfc9a5a99675aa279ba/buildenv/jenkins/JenkinsfileBase#L31

thirdparty_containers/system-test/dockerfile/system-test.sh
```diff
@@ -0,0 +1,92 @@
...
+
+TEST_SUITE=$1
+
+echo "PATH is : $PATH"
+echo "JAVA_HOME is : $JAVA_HOME"
+echo "type -p java is :"
+type -p java
+echo "java -version is: \n"
+java -version
```
llxia: If we check `java -version` in line 46, we can remove the check in line 22, 28, 36.
In fact, we added check in ./get.sh for `java -version`.

thirdparty_containers/system-test/playlist.xml
chandrams: -XX:+UseContainerSupport - this is turned on by default now, will remove this variation

thirdparty_containers/system-test/playlist.xml
```diff
@@ -0,0 +1,273 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+<playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
+	 <test>
+                <testCaseName>MauveSingleThreadLoadTest</testCaseName>
+                <command>docker run -v $(JDK_HOME):/java --name mauve-system-test adoptopenjdk-system-test:latest MauveSingleThreadLoadTest; \
```
llxia: Do we want to run tests selectively? Can we just run two targets `sanity.system` and `extended.system`?

thirdparty_containers/system-test/dockerfile/system-test.sh
```diff
@@ -0,0 +1,69 @@
...
+
+cd /openjdk-tests
+
+./get.sh -t /openjdk-tests -p x64_linux -i openj9
```
llxia: If you donot use `get.sh` to get SDK, then we do not need `-p x64_linux -i openj9`

thirdparty_containers/system-test/dockerfile/system-test.sh
```diff
@@ -0,0 +1,69 @@
...
+
+cd TestConfig
+
+export SPEC=linux_x86-64_cmprssptrs
+export BUILD_LIST=systemtest
```
llxia: Do we set JDK_IMPL?

chandrams: Not sure if  we need to set JDK_IMPL, wouldn't it pick from the playlist. For mauve tests, I have not specified any impl in the playlist as they are applicable to both hotspot and openj9. But for shareclasses I have specified openj9 as impl.

llxia: - `JDK_IMPL` value should be set based on the SDK that we are testing. The playlist checks `JDK_IMPL` value to determine if the test is applicable or not.

   If `JDK_IMPL` is not provided, the program will use the default value `openj9`
https://github.com/eclipse/openj9/blob/5839468bd3fe2217b7590eb27368c2c246157f04/test/TestConfig/run_configure.mk#L61-L63

- Actually, to make the story simple, can we just mount test material directly when running docker? Similar to how to mount java. In this way, there is no need to run `get.sh` in this script.

chandrams: Mounted material would need to be updated regularly, I would rather prefer moving get.sh and compilation part into the docker image and just have the running part in shell script. 

@smlambert @Mesbah-Alam - Any thoughts?

thirdparty_containers/system-test/dockerfile/system-test.sh
chandrams: Changes made

Mesbah-Alam:
The Java used by the container based tests are not actually "mounted" from host machine, but obtained, via the `FROM` clause used in the dockerfile, from the adoptopenjdk dockerhub, e.g., https://hub.docker.com/r/adoptopenjdk/openjdk8/ .

thirdparty_containers/system-test/dockerfile/system-test.sh
llxia: > Mounted material would need to be updated regularly

No. Test material is checked out during every build. We do not need to update it at all. This is the same idea as you mount jdk in playlist.xml, something like:

```
docker run -v $(JDK_HOME):/java -v <path to test root folder>:/test <docker_image_name> sanity.system
```

Since we are passing in repo, branch and sha for regular build and Grinder, it is all been taking care of by mounting the existing test material. If we choose to use run `get.sh` again in docker, not only we git clone twice, but also we need to get repo, branch and sha parameters from the build and pass them into `get.sh` in docker.

smlambert:
To reinforce @Mesbah-Alam 's comment, by FROM-ing/pulling the jdks from docker hub, we are testing what gets pushed to Docker hub.

smlambert:
@llxia - do you mind if we defer the decision to mount or not to mount test material for now?  @chandrams can shift the fetching of test material so that its only called once in the Docker container.  If we decide we want to mount test material, we should likely decide to be consistent and do this same approach for all the external test targets (third party apps).

smlambert:
Reading more carefully @llxia review comments, I am fine with either approach.  If we mount SVT material, this won't be quite like the other external tests, but that is fine, it is a slightly different beast anyway, compared to the functional suites from applications.

llxia:
I am ok with not to mount test material. Let's get this working first and we can refine this later.

sophia-guo:
I'm confused... Should all container based tests testing against jdks from docker hub? Or only this system test? Based on the currently file structure if docker run with sdk mounted ` docker run -v $(JDK_HOME):/java.....`definitely the mounted sdk from host machine is tested.

```
VOLUME ["/java"]
...
ENTRYPOINT ["/bin/bash", "/system-test.sh"]
CMD ["--version"]
``` 
in Dockerfile

And export in system-test.sh
```
if [ -d /java/jre/bin ];then
	echo "Using mounted Java8"
	export JAVA_BIN=/java/jre/bin
	export JAVA_HOME=/java
	export PATH=$JAVA_BIN:$PATH
	java -version
elif [ -d /java/bin ]; then
	echo "Using mounted Java9"
	export JAVA_BIN=/java/bin
	export JAVA_HOME=/java
	export PATH=$JAVA_BIN:$PATH
	java -version
else
	echo "Using docker image default Java"       <--- only if no mount, FROM-ing/pulling the jdks from        
	java_path=$(type -p java)                                        docker hub will be tested
	suffix="/java" 
	java_root=${java_path%$suffix}
	export JAVA_BIN="$java_root"
	echo "JAVA_BIN is: $JAVA_BIN"
	$JAVA_BIN/java -version
fi

```
The original design is to support mounted java or default java(FROM-ing/pulling one). If we do want to test jdks from docker hub then we should not mount java from host machines.

Mesbah-Alam:
@sophia-guo Good catch. Thanks.. You are right, I've removed the volume map for jdk : https://github.com/AdoptOpenJDK/openjdk-tests/pull/750

chandrams:
> 
> 
> I am ok with not to mount test material. Let's get this working first and we can refine this later.

Thanks for the review Lan @llxia 

I have moved the get.sh and compiling part to dockerfile.  But as discussed, I 'm seeing this error while building, even though I set the ENV JAVA_BIN=/java/bin, I think /java will not be available at building stage.  
7      [exec] Cannot find java executable in JAVA_BIN: /java/bin!
17:58:47      [exec] The command '/bin/sh -c cd /openjdk-tests &&     ./get.sh -t /openjdk-tests -p x64_linux -i openj9 &&     cd TestConfig' returned a non-zero code: 1

https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/635/console

Also just gave a try with ENV JAVA_BIN=/home/jenkins/workspace/Grinder/openjdkbinary/j2sdk-image/bin, as I saw this location in the console output, but looks like this directory too doesn't exist.
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/652/console

Is there any other java that we can point to? Any suggestions? Thanks.

Mesbah-Alam:
@chandrams - Please see the docker image's default java being used here: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/624/consoleFull

The JAVA_BIN is being set to `/opt/java/openjdk/bin` 

And it is being constructed in the .sh file associated with the external test, like this: 

```
echo "Using docker image default Java"
	java_path=$(type -p java)
	suffix="/java"
	java_root=${java_path%$suffix}
	export JAVA_BIN="$java_root"
	echo "JAVA_BIN is: $JAVA_BIN"
	$JAVA_BIN/java -version
```

chandrams:
Thanks @Mesbah-Alam I will need to include that logic in the Dockerfile as I have moved the compilation of the test material from the shell script to Dockerfile, for now I have just it to JAVA_HOME=/opt/java/openjdk to check if the test material is built and it runs. It seems to build fine, but fails with **FAILED** JVM for 'javahome-generation' is not pointing at a valid JVM build: /opt/java/openjdk/bin/../../

I have set JAVA_HOME explicitly in the system-test.sh for docker image case, but it seems to be overridden to /opt/java/openjdk/bin/../../ when the test is run. In the system test playlist there is export JAVA_HOME=${JDK_HOME}, trying to find where JDK_HOME is getting picked from and why we have ../.. instead of just ../

chandrams:
JDK_HOME is set in settings.mk in openj9/test/TestConfig as below:

JDK_HOME := $(JAVA_BIN_TMP)$(D)..
ifeq ($(JDK_VERSION),8)
JDK_HOME := $(JAVA_BIN_TMP)$(D)..$(D)..
endif

This doesn't seem to work with my changes as JAVA_BIN is obtained using type -p java if a java docker image is used. This is the logic used for all external tests and it works fine for other tests.  Only for system tests that use testkitgen it fails as JAVA_BIN doesn't contain jre in its path and for Java8 the logic is to go two dirs back.  Instead of checking for version is 8, can we check if JAVA_BIN contains jre then go two dirs back?  @llxia - Why is JDK_HOME set in TestConfig? Since you are more familiar with the openj9 functional tests, can you please take a look at this and comment. Thanks.

llxia:
> @llxia - Why is JDK_HOME set in TestConfig? 

JDK_HOME is needed by the system test framework and OpenJDK regression. It is set by TKG based on JAVA_BIN, so the user does not need explicitly set it.

Please try to remove `ENV JAVA_HOME=/opt/java/openjdk` as it is not required.

chandrams:
Thanks @llxia 
I have removed the JAVA_HOME env, and set JAVA_BIN and JDK_VERSION for Java8 for now. Please review.

thirdparty_containers/system-test/playlist.xml
```diff
@@ -0,0 +1,273 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+<playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
+	 <test>
+                <testCaseName>MauveSingleThreadLoadTest</testCaseName>
```
llxia: Please do not list all targets as we actively add/remove/modify system test targets.
Please do it the same as your functional test PR, just run `sanity.system`.

chandrams: @llxia 

Our plan was to run only some of the mauve and shared classes tests in containers. Not the entire sanity.system. Shared classes tests are listed as extended.system, so if we go by these groups we would need to run extended.system.

chandrams: Those were the initial set of tests that we wanted to run in containers.  We just wanted some stress test and check shared classes related test to check the JVM optimizations included for containers.

llxia:
Talked with Chandrakala, Shelley and Mesbah. We will merge this change and refine it later.

PR #642, opened 2018-10-15
Update

ben-walsh:


smlambert:
@ben-walsh - needs conflict resolution ...

PR #654, opened 2018-10-19
WIP: Updates for OSX testing for OpenJ9 

smlambert:
This is also to allow running of functional test suites, subsets of which we may eventually want to use at AdoptOpenJDK for all implementations.


smlambert:
Related https://github.com/eclipse/openj9/pull/3494

Issue #651, opened 2018-10-17
Expand test on OpenJ9's concurrent scavenge support


0dvictor:
OpenJ9 is expanding its support on Concurrent Scavenge support and hence the related tests should be enabled/introduced.

smlambert:
Fixed by #662

PR #662, opened 2018-10-26
Enable concurrent scavenge related tests on X86

0dvictor:
OpenJ9 has expanded its concurrent scavenge support on X86, and hence related tests should be enabled.

Issue #651 

Signed-off-by: Victor Ding <dvictor@ca.ibm.com>

### DIFF
--- a/systemtest/playlist.xml
+++ b/systemtest/playlist.xml
@@ -2645,7 +2645,7 @@
 	</test>
 	<!-- Tests above pertain to Java 9 Modularity -->
 	
-	<!-- The tests below are added to test concurrent scavenge on z/OS and z/Linux using OpenJ9 SDK -->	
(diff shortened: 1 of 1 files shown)


karianna:
Is this going to work with older versions?  My only concern is that an existing pipeline will fail if the OpenJ9 binaries we produce aren't new enough yet.

0dvictor:
> just to confirm, is the support for CS on x86 for all versions - jdk8, jdk11? (am presuming so, as this will generate test targets for all versions).

Yes, CS on is supported on all available JDK versions that targeting Linux X86-64 (both compressed refs and large heap).

smlambert:
Quick test here: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/327/
will merge once a few other quick tests run.

smlambert:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/328/

Issue #602, opened 2018-09-28
OpenJDK 11 jdk_nio, jdk_net failures


sophia-guo:
Linux:
java/nio/file/attribute/BasicFileAttributeView/UnixSocketFile.java 
java/nio/channels/AsyncCloseAndInterrupt.java	
java/nio/charset/spi/CharsetProviderBasicTest.java	
java/nio/channels/DatagramChannel/ConnectExceptions.java	
java/nio/channels/DatagramChannel/SendExceptions.java	
java/nio/channels/SocketChannel/ExceptionTranslation.java	
java/net/InetAddress/BadDottedIPAddress.java	
java/net/InetAddress/CachedUnknownHostName.java	
java/net/InetAddress/IPv4Formats.java	
java/net/SocketPermission/SocketPermissionCollection.java	
java/net/SocketPermission/Wildcard.java	
java/net/Socks/SocksV4Test.java	
java/net/URL/OpenStream.java	
java/net/httpclient/ConnectExceptionTest.java	

PPCLE:
java/nio/file/Files/InterruptCopy.java

sophia-guo:
PPCLE:
java/nio/file/Files/InterruptCopy.java `Error while cleaning up threads after test`
[InterruptCopy.jtr.txt](https://github.com/AdoptOpenJDK/openjdk-tests/files/2428496/InterruptCopy.jtr.txt)

sophia-guo:
all other nio failure;
[channels.zip](https://github.com/AdoptOpenJDK/openjdk-tests/files/2428576/channels.zip)

all other jdk_net failrue:
[net.zip](https://github.com/AdoptOpenJDK/openjdk-tests/files/2428580/net.zip)

sophia-guo:
Following tests intermittently fails on HS with unexpected errors. Looks like only happens on test-packet-ubuntu1604-x64-2 or with specific two sdk builds? The failures is:

```
21:06:26 # A fatal error has been detected by the Java Runtime Environment:
21:06:26 #
21:06:26 #  SIGSEGV (0xb) at pc=0x00007fb8245387ca, pid=28972, tid=29819
21:06:26 #
21:06:26 # JRE version: OpenJDK Runtime Environment (11.0.2+9) (build 11.0.2+9-201902021908)
21:06:26 # Java VM: OpenJDK 64-Bit Server VM (11.0.2+9-201902021908, mixed mode, tiered, compressed oops, g1 gc, linux-amd64)
[...]

PR #668, opened 2018-11-05
Update

ben-walsh:
Exclude tests which fail on OJ9 same way they fail on HS


smlambert:
@ben-walsh - can you please update the link to point to the open #602 issue? (instead of 593?)

ben-walsh:
@smlambert Done.

PR #670, opened 2018-11-05
Add Queue stage to test pipelines

AdamBrousseau:
Related eclipse/openj9#2372

### DIFF
--- a/buildenv/jenkins/openjdk_aarch64_linux
+++ b/buildenv/jenkins/openjdk_aarch64_linux
@@ -2,17 +2,18 @@
 LABEL='sw.os.linux&&hw.arch.aarch64'
 
 node ("master") {
-	checkout scm
-	def jenkinsfile = load "${WORKSPACE}/openjdk-tests/buildenv/jenkins/JenkinsfileBase"
-	jenkinsfile.setLabelParam()
-	cleanWs()
-}
-
-node("$LABEL") {
-    PLATFORM = 'aarch64_linux'
-    SDK_RESOURCE = 'upstream'
-    SPEC='linux_arm'
     checkout scm
     def jenkinsfile = load "${WORKSPACE}/openjdk-tests/buildenv/jenkins/JenkinsfileBase"
-    jenkinsfile.testBuild()
+    jenkinsfile.setLabelParam()
+    cleanWs()
+}
+
+stage('Queue') {
+    node("$LABEL") {
+        PLATFORM = 'aarch64_linux'
+        SDK_RESOURCE = 'upstream'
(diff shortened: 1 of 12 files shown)


AdamBrousseau:
Marking as WIP to allow for discussion. @llxia @ShelleyLambert 

Previously, we added a queue stage to the OpenJ9 pipelines which helps us visually determine
a) How long a job waits for a machine.
b) Separates queue time from the first 'real" stage of the pipeline.

I think this would be beneficial to have on the test pipelines as well.
See eclipse/openj9#2372 or an example job: https://ci.eclipse.org/openj9/view/Sandbox/job/Adam-Test-sanity.functional-JDK11-linux_x86-64_cmprssptrs/

smlambert:
Thanks @AdamBrousseau - seems like a good enhancement and will more clearly reveal where we can improve our addition and scheduling of machine resources.  

Once you are ready to shift this from WIP, we can do final review/merge.

AdamBrousseau:
Removed WIP. Updated only the version agnostic files as I understand from Lan that the version specific files will be removed in time.

llxia:
Also tested with openjdk regression test.

AdamBrousseau:
Just noticed this "breaks" the Blue Ocean view by only showing the Queue stage for the build. I'm surprised we hadn't noticed or received complaints before now because we've had this on the Build jobs since July.

Found open issue to add queue time by default
https://issues.jenkins-ci.org/browse/JENKINS-49179

It looks like nested stages was supposed to be resolved by
https://issues.jenkins-ci.org/browse/JENKINS-49050
according to
https://issues.jenkins-ci.org/browse/JENKINS-38442
So maybe its coming soon.

I suppose this isn't a deal breaker as classic view still works and it appears it will eventually work in Blue Ocean.

karianna:
Hmmm, not ideal but I personally only track blue ocean when doing a release and I've only really got 11.0.1 for hotspot to go and then it's quiet for a bit.

AdamBrousseau:
@karianna keep in mind this would only be at the test job level not the overall pipeline level.
ex
Before:
https://ci.adoptopenjdk.net/blue/organizations/jenkins/openjdk11_j9_systemtest_x86-64_linux/detail/openjdk11_j9_systemtest_x86-64_linux/96/pipeline
image

After:
https://ci.adoptopenjdk.net/blue/organizations/jenkins/openjdk11_j9_systemtest_x86-64_linux/detail/openjdk11_j9_systemtest_x86-64_linux/98/pipeline
image

karianna:
Ah OK, it's a bit trickier to debug but in all honesty, I just look for an overall pass/fail.

PR #676, opened 2018-11-07
Update label ppc to ppc64

AdamBrousseau:
- Label should be ppc64 to be more correct

Related Eclipse/OpenJ9#3593


AdamBrousseau:
Marking as WIP until machine labels are updated. Will need to add new label, merge, then remove old label.

AdamBrousseau:
@ShelleyLambert can you update the Adopt aix test machines?

smlambert:
Well, we do not have an Adopt AIX test machine.  We were using one of the build machines on the weekend, but then it went offline (I shall check if its back).  The other build machine does not have the test prereqs installed.

AdamBrousseau:
I have unmarked WIP and updated the machines at OpenJ9 so this is ready when you are.

PR #681, opened 2018-11-12
Update JDK11 jdk test groups

sophia-guo:
More jdk tests groups are defined in JDK11. Initially add as extended
Update using number as jdk_version


sophia-guo:
Added as extended. If needed will update to sanity per group. Group jdk_native_sanity added as disable as  `-nativepath to specify the location of native code ` is needed. depends on https://github.com/AdoptOpenJDK/openjdk-build/issues/248

smlambert:
After having talked with @sophia-guo, running time on Adopt machines is longer, so we will tag jdk_security1 as sanity, leaving others as extended for now.

PR #698, opened 2018-11-22
Add manual steps to run systemtest

Mesbah-Alam:
Resolves https://github.com/AdoptOpenJDK/openjdk-tests/issues/697


Mesbah-Alam:
Documented tests have been tested locally.

llxia:
We already have wiki for this:
https://github.com/eclipse/openj9/wiki/Reproducing-Test-Failures-Locally#system-tests

Mesbah-Alam:
@llxia - oh.. cool. Then we could use that wiki you mentioned, but we need to add the missing exports there for it to work.

llxia:
Why we need extra info? I thought we already set it base on JAVA_BIN
https://github.com/AdoptOpenJDK/openjdk-tests/blob/1114a6936698ee0cd4adb86beb95b866837ebd6d/systemtest/build.xml#L10

Mesbah-Alam:
Actually you are right! I didn't remember we set it in the build file. Just tried the steps locally and they worked. Will close this PR without committing.

Issue #700, opened 2018-11-23
Remove SEXX in playlist.xml


llxia:
We need to cleanup for all playlist.xml in the repo to remove reference to SEXX (i.e., SE80, SE90, SE100, SE110). It should be replaced by the JDK version (i.e., 8, 9, 10, 11). Once the clean up is completed, we can stop supporting SEXX.

For example:
https://github.com/AdoptOpenJDK/openjdk-tests/blob/1114a6936698ee0cd4adb86beb95b866837ebd6d/systemtest/playlist.xml#L41

https://github.com/AdoptOpenJDK/openjdk-tests/blob/1114a6936698ee0cd4adb86beb95b866837ebd6d/performance/idle_micro/playlist.xml#L38-L40

PR #703, opened 2018-11-27
Include back modularity tests to jdk11 build

Mesbah-Alam:
Include back modularity tests to jdk11 build

### DIFF
--- a/systemtest/playlist.xml
+++ b/systemtest/playlist.xml
@@ -38,8 +38,8 @@
 	$(TEST_STATUS)</command> 
 		<tags>
 		<subsets>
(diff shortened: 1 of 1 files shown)


sophia-guo:
Could ref: #700 be also addressed as both are working in same file?

Mesbah-Alam:
I'd like to keep them separate if that's fine, as #700 involves changes in other playlists too.  It'd make them to keep them together in one PR.

Mesbah-Alam:
On a second thought, it doesn't hurt to repalce the SExx with numbers in this PR for the systemtest playlist! I've made that update.

PR #701, opened 2018-11-23
Enable jre tests for jdk11

sophia-guo:
Signed-off-by: Sophia Guo <sophiag@ca.ibm.com>

PR #704, opened 2018-11-27
Skip regression jre tests

sophia-guo:
No jre image for jdk11u build and 'JRE' target will be away anytime from 12+. Temporarily skip the test for 11+.


sophia-guo:
relates with #701

PR #708, opened 2018-11-27
Temporarily exclude security enabled JLM tests on Windows

Mesbah-Alam:
Temporarily exclude security enabled JLM tests on Windows due to https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/198

### DIFF
--- a/systemtest/playlist.xml
+++ b/systemtest/playlist.xml
@@ -952,6 +952,7 @@
 	</test>
 	
 	<!-- Excluding the following test for openj9 for eclipse/openj9/issues/1520  -->
+	<!-- Excluded from Windows due to https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/198 -->
 	<test>
 		<testCaseName>TestJlmRemoteClassAuth</testCaseName>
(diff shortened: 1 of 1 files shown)


systemtest/playlist.xml
```diff
@@ -1451,7 +1467,7 @@
 		<impls>
 			<impl>openj9</impl>
 		</impls>
-		<platformRequirements>os.linux,^arch.ppc</platformRequirements>
+		<platformRequirements>os.linux,^arch.ppc,^os.win</platformRequirements>
```
renfeiw: you don't need to exclude this test, as it's linux only

systemtest/playlist.xml
```diff
@@ -1619,7 +1641,7 @@
...
```
renfeiw: you don't need to exclude this test, as it's linux only

systemtest/playlist.xml
Mesbah-Alam: Rght! Thanks for spotting. There were few more of these. Updated them all.

PR #710, opened 2018-11-28
Temporarily exclude ExplicitModulesTest

Mesbah-Alam:
Temporarily exclude ExplicitModulesTest


Mesbah-Alam:
The test fails across platforms in the eclipse.ci system test builds. Should be disabled until https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/200 is resolved.

PR #712, opened 2018-11-28
Do not use sshatent when on zos

Mesbah-Alam:
Do not use sshatent when on zos


Mesbah-Alam:
Tested using Grinder in internal Jenkins.

sophia-guo:
FYI @llxia

PR #716, opened 2018-11-29
Temporarily exclude PatchModule*Test_PlatformModPatchModule on AIX

Mesbah-Alam:
Temporarily exclude PatchModule*Test_PlatformModPatchModule on AIX

### DIFF
--- a/systemtest/playlist.xml
+++ b/systemtest/playlist.xml
@@ -2136,6 +2136,8 @@
 			<group>system</group>
 		</groups>
 	</test>
+	
+	<!-- Temporarily excluding this test from AIX due to : https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/202 -->
 	<test>
 		<testCaseName>PatchModuleTest_PlatformModPatchModule</testCaseName>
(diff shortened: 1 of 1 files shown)


Mesbah-Alam:
Note: this PR includes exclusion of JlinkTest_RequiredMod test from osx due to https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/203 as well.

systemtest/playlist.xml
```diff
@@ -2161,6 +2163,7 @@
 		<groups>
 			<group>system</group>
 		</groups>
+		<platformRequirements>^arch.ppc</platformRequirements>
```
sophia-guo: ppcle will also be excluded. Is that expected?

Mesbah-Alam: yes, need to exclude from all AIX builds.

smlambert: os.aix ?

Mesbah-Alam: hmm.. that's better, changing to os.aix

PR #721, opened 2018-12-01
Add aarch32 testing and doc update

### LINKED ISSUE #717: Create openjdk11_hs_openjdktest_arm_linux
And maybe openjdk11_hs_perftest_arm_linux too if possible.

We now have https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk11u/job/jdk11u-linux-arm-hotspot producing builds that seem to be operational, so would be good to be able to run the standard test suite against them and find out how good they really are :-)

smlambert:
Fixes #717 


README.md
```diff
@@ -36,7 +36,7 @@ See the different test jobs here:
 
 The directory structure in this openjdk-tests repository is meant to reflect the different types of test we run (and pull from lots of other locations).  The diagram below shows the test make target for each of the types, along with in-plan, upcoming additions (denoted by dotted line grey boxes).
 
-![Overview of Tests](doc/overviewOfAdoptTests.svg)
+![Overview of Tests](doc/TDoverviewOfAdoptTests.svg)
```
llxia: Since we are using `TDoverviewOfAdoptTests.svg`, do we still need `overviewOfAdoptTests.svg` below?

smlambert: I had not decided which of the 2 styles I prefer, do you have a preference? TD/topdown or LR/leftright?

llxia: I think TD looks prettier. But the user needs to zoom quite a bit due to the small font. 

It is better to separate it into multiple graphs? One graph shows all supported test components/types. The others show detail about each test component/type - i.e., external, perf, etc.

llxia:
Should we add toplevel `openjdk-tests` in [externaltests.svg](https://github.com/AdoptOpenJDK/openjdk-tests/blob/ad98c571c7a82ae8a1679f5f87b8844b6233349c/doc/diagrams/externaltests.svg)? So that [externaltests.svg](https://github.com/AdoptOpenJDK/openjdk-tests/blob/ad98c571c7a82ae8a1679f5f87b8844b6233349c/doc/diagrams/externaltests.svg) has the same structure as [perftests.svg](https://github.com/AdoptOpenJDK/openjdk-tests/blob/ad98c571c7a82ae8a1679f5f87b8844b6233349c/doc/diagrams/perftests.svg)

smlambert:
thanks @llxia, I've updated externaltests.svg, to have the openjdk-tests block and also to include geode, hbase, etc, as there are open issues to add those applications (so they show in diagram as dotted-line/next-up).

Issue #660, opened 2018-10-24
Functional tests do not compile on z/OS


Mesbah-Alam:
When functional teat compilation is attempted on z/OS, we get the following type of errors: 
```
11:35:49     [javac] /home/jenkins/workspace/Test-sanity.functional-JDK11-zos_390-64_cmprssptrs/openjdk-tests/functional/IllegalAccessError_for_protected_method/src/org/openj9/test/Agent.java:1: error: illegal character: '\u0015'
```
I suspect this is due to the fact that we are not supplying the correct encoding value in the build.xml files of the functional tests. 

We need to define what file encoding to be used for Java source files for tests as we do in case of [system tests](https://github.com/AdoptOpenJDK/openjdk-systemtest/blob/86a7276fd983ea732e2dca3d0d56143bb6b36fa4/openjdk.build/include/top.xml#L377).

**Build link:** https://hyc-runtimes-jenkins.swg-devops.com/view/Test_functional/job/Test-sanity.functional-JDK11-zos_390-64_cmprssptrs/9/console

Mesbah-Alam:
FYI @sophia-guo @vsebe

Mesbah-Alam:
FYI @llxia - this is the issue i was speaking about earlier.

Mesbah-Alam:
For system tests we have a parameter defined in top.xml that sets source encoding to `IBM-1047` when on z/OS, and tp `UTF-8` otherwise.  

```
<condition property="src-encoding" value="IBM-1047" else="UTF-8">
		<isset property="isZOS"/>
</condition>
```
https://github.com/AdoptOpenJDK/openjdk-systemtest/blob/2fb744f0c847bb31f6b4aa39173a3977f523008c/openjdk.build/include/top.xml#L381

The `src-encoding` property is subsequently used in the test projects' build files in the `javac` task: 
```
<javac srcdir="${src_dir}"
			   destdir="${bin_dir}"
			   fork="true"
			   executable="${java8_compiler}"
			   debug="true"
			   classpathref="project.class.path"
			   encoding="${src-encoding}"
			   includeantruntime="false"
			   failonerror="true">
			<include name="**/*.java"/>
</javac>
```

https://github.com/AdoptOpenJDK/openjdk-systemtest/blob/2fb744f0c847bb31f6b4aa39173a3977f523008c/openjdk.test.jlm/build.xml#L82

[...]

PR #722, opened 2018-12-03
Convert tap file from ebcdic to ascii on zos

renfeiw:
- convert tap file before parsing


llxia:
Related: #660

Issue #726, opened 2018-12-05
Add Multi-Byte Character Set locale tests 


smlambert:
An issue for the upcoming contribution of MBCS tests to AdoptOpenJDK.  These tests verify that JDK implementations behave appropriately under different locales.

PR #729, opened 2018-12-06
Add Multi-Byte Character Set locale tests enhancement

takiguc:
#726


smlambert:
Thanks @takiguc to you and your team for preparing this contribution!  

As discussed, the plan is to co-ordinate this PR with https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/618, so that we can test it in Grinders, then bring it into the master branch for inclusion in automated test runs (which will show up in this [Test_functional view](https://ci.adoptopenjdk.net/view/Test_functional/)).

smlambert:
@takiguc, after making updates and pushing them to your branch, you can re-run the Grinder job that @llxia ran via:

https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/516/rebuild/parameterized

takiguc:
@llxia @smlambert 
I moved from functional to functional-AdoptOpenJDK and changed get.sh.
Also, I applied some fixes into test cases.
See https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/527/console
Currently, the machine does not have MBCS locales, like Japanese.
Some of test cases for MBCS_Tests_CLDR_11_ja_JP_linux_0 worked fine with en_US locale.
So it said that "Looks like you failed 29 tests of 42."
Could you check "get.sh" change is acceptable solution ?

get.sh
```diff
@@ -235,6 +235,9 @@ getTestKitGenAndFunctionalTestMaterial()
 	mv openj9/test/Utils Utils
 	mv openj9/test/functional functional
 	rm -rf openj9
+    if [ -d functional-AdoptOpenJDK ]; then
+        mv functional-AdoptOpenJDK/* functional
+    fi
```
llxia: To have a root dir `functional-AdoptOpenJDK` is very confusing. It does not go with our overall structure. Please keep it `functional`.
We just need two functional folders to overwrite each other.

llxia:
Thank you so much for updating @takiguc 

> Currently, the machine does not have MBCS locales, like Japanese.

Does it mean we need to wait until we have machine support, then merge this PR? Any idea when we will get machine updated? Or temporarily exclude the tests that do not have support yet?

The thing is as long as this PR is in, it will be part of our auto builds that run multiple tests a day.

smlambert:
We are waiting for machine support (2 issues: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/627 and https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/618).  Those requests are for Adopt machines.  If we want to run these tests at openj9, there would likely need to be an additional request.  I do not believe we need to run these tests at both projects (let's avoid duplication).

@llxia - I am actually thinking it may be time to introduce a "3rd level".  Originally had imagined sanity run nightly and extended on releases (which is the plan for Adopt), but openj9 project runs extended nightly and PRs too.  I want a level that is selectively run on a schedule, in non-PR test jobs.  We have other test jobs that fit into this category as well.  In this way we can bring in these functional tests and not have them run at openj9, and run them on a schedule at Adopt where machine usage is lower.  

Thinking about generic naming for a new 'level', wondering about calling it 'scheduled' or 'special', so we have sanity, extended, scheduled.

get.sh
takiguc: * Rename from functional-AdoptOpenJDK to functional 
* get.sh was edited.
See https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/536/console

takiguc:
@smlambert @llxia 
Fixes were applied. Could you check them ?

get.sh
```diff
@@ -233,7 +233,13 @@ getTestKitGenAndFunctionalTestMaterial()
 
 	mv openj9/test/TestConfig TestConfig
 	mv openj9/test/Utils Utils
-	mv openj9/test/functional functional
+    if [ -d functional ]; then
+        cd openj9/test/functional
+        find . -print | cpio -pdm ../../../functional
```
llxia: Could we just directly use `cp` cmd? There is no need to print.

get.sh
```diff
@@ -233,7 +233,13 @@ getTestKitGenAndFunctionalTestMaterial()
...
+        cd ../../..
+    else
+	    mv openj9/test/functional functional
+    fi
```
llxia: Once we have `functional` in Adopt, this case will never happen.

takiguc: To keep directory structure, we should use rsync instead of cp or mv .
But AIX does not have rsync on default, so I selected cpio.

llxia:
@takiguc We have 3rd level `special` added https://github.com/eclipse/openj9/pull/4081, could you update to use it? That is, instead of using `<level>extended</level>`, use `<level>special</level>` for all tests.

In this way, these tests will only get compiled, but will not run in our auto builds. Once we have machine support, we can adjust the test level if needed. Thanks.

PR #730, opened 2018-12-06
Update excludes in systemtest playlist

smlambert:
Reinclude HCRLateAttachWorkload on jdk8 
Exclude DirectByteBufferLoadTest on AIX

### DIFF
--- a/systemtest/playlist.xml
+++ b/systemtest/playlist.xml
@@ -306,6 +306,23 @@
 			<group>system</group>
 		</groups>
 	</test>
+
+	<test>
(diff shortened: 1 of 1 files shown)


smlambert:
HCRLateAttachWorkload test - verified it is now working on both openj9 and hotspot (with grinders, https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/525 and https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/513 respectively)
Will continue to exclude for jdk11, as still fails with known issue https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/7

PR #734, opened 2018-12-07
Temporarily exclude system tests

llxia:


llxia:
This is for `AdoptOpenJDK:extraVariations` branch only

PR #737, opened 2018-12-07
Deploy ODM Benchmark

piyush286:
• Enabled ODM benchmark since benchmark package has been staged (https://github.com/AdoptOpenJDK/openjdk-tests/issues/457) and numactl has been installed (https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/394) on the perf machines
• Fixed various environment variables
• Updated the playlist to copy the benchmark config file in the correct directory before running it
• Changed tar to zip to package the benchmark output files
• Cleaned up some redundant code
• Other fixes

Signed-off-by: Piyush <piyush@ca.ibm.com>


llxia:
Thanks @piyush286 for updating.

performance/odm/playlist.xml
```diff
@@ -28,6 +30,5 @@
 		<groups>
 			<group>perf</group>
 		</groups>
-		<disabled>https://github.com/AdoptOpenJDK/openjdk-tests/issues/457</disabled>
```
llxia: ODM should not run in AdoptOpenJDK. If we remove `<disabled>`, please leave a comment if test material does not exist.

performance/odm/playlist.xml
```diff
@@ -16,8 +16,10 @@
 	xsi:noNamespaceSchemaLocation="../../TestConfig/playlist.xsd">
 	<test>
 		<testCaseName>ODM</testCaseName>
-		<command>chmod a+x ./scripts/benchmark.sh; \
-		./scripts/benchmark.sh $(TEST_RESROOT); \
+		<command>cd $(TEST_RESROOT)/ilog_wodm881; \
+		cp $(TEST_RESROOT)/../../../openjdk-tests/performance/odm/scripts/benchmark.sh .; \
```
llxia: I am not sure the copy is necessary.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -289,12 +289,13 @@ def post(test_target) {
...
 			//for performance test, achive regardless the build result
+			//TODO: Need to soft-code the output dir path
 			def benchmark_test_output_dir = 'jvmtest/performance/odm/ilog_wodm881/leftoverResults';
 			if (fileExists(benchmark_test_output_dir)) {
-				sh "tar -zcf ${testTarName} benchmark_test_output.tar.gz"
+				sh "zip -r benchmarkOutput.zip ${benchmark_test_output_dir}"
 				if (!params.ARTIFACTORY_SERVER) {
 					echo "ARTIFACTORY_SERVER is not set. Saving artifacts on jenkins."
-					archiveArtifacts artifacts: 'benchmark_test_output.tar.gz', fingerprint: true, allowEmptyArchive: true
+					archiveArtifacts artifacts: 'benchmarkOutput.zip', fingerprint: true, allowEmptyArchive: true
```
llxia: Any reason for changing tar to zip. All the `*_test_output.tar.gz` will later upload to artifactory. This change breaks file upload later.

performance/odm/playlist.xml
piyush286: Copying of this file is needed since it uses the current dir for setting the working dir (i.e. `WORKDIR`). That working dir var is used in the ODM script. We usually put all our scripts under the benchmark package folder so that everything is under one place and we don't have to worry about paths that much.

performance/odm/playlist.xml
piyush286: I thought we can run ODM in AdoptOpenJDK as long as the perf machines have the benchmark material. From this comment (https://github.com/AdoptOpenJDK/openjdk-tests/issues/457#issuecomment-401170759), seems like 2 perf machines already have the material. Please let me know if that's not the case and I can disable this benchmark.

buildenv/jenkins/JenkinsfileBase
piyush286: I didn't notice this since `ARTIFACTORY_SERVER` wasn't set for my pipeline. Thanks for catching that. It's fixed now!

performance/odm/playlist.xml
llxia: No, this is not the case. We should run ODM internally only, at least for now.
I did not get chance to remove the perf material on these two machines. If you can help to delete them, that would be great. Thanks.
We should still enable the test, so we can run internally.

smlambert: ODM benchmark zip files have been removed from both machines.

piyush286: @llxia Sounds good! I'll disable it! Btw in order to disable it for AdoptOpenJDK and enable it internally, I guess we'll need to have an internal repo with `odm/build.xml` instead `odm/exclude_build.xml` and no `<disabled>` element in `odm/playlist.xml` right? Not sure if there's a better way of doing this.

performance/odm/playlist.xml
llxia: - TKG will change the dir into `$(REPORTDIR)`, so that all test output will be archived if there is any failure. If we `cd` into another dir, test output will not be archived. If we really have to change dir, then we need to copy all scripts into `$(REPORTDIR)`
- I still do not really get why we cannot pass in `$(TEST_RESROOT)`. The original script worked before.

performance/odm/playlist.xml
llxia: no, we do not need to have an internal repo.
In all perf builds in Adopt, we just need to set `BUILD_LIST=performance/idle_micro` for now.
Your current change for enabling test is fine.

piyush286:
@llxia Sorry, I haven't been getting much time to work on this change due to other higher priority issues. I'll get back to this on Monday if not earlier.

llxia:
That's ok. Thanks for working on this. :)

performance/odm/playlist.xml
piyush286: @llxia I see that `openjdk-tests/performance/odm/autoGen.mk` creates a dir (i.e. something like this `/java/jenkins/workspace/Piyush_ODM_x86-64_linux/openjdk-tests/TestConfig/test_output_15436147118144/ODM_0`). I believe that's the `$(REPORTDIR)` dir? So that `autoGen.mk` changes the dir to that before calling `openjdk-tests/performance/odm/playlist.xml`. Without copying this `benchmark.sh` file, I was getting this error `chmod: cannot access './scripts/benchmark.sh': No such file or directory` since `benchmark.sh` is in `openjdk-tests/performance/odm/scripts/` and not `test_output_15436147118144/ODM_0`. I believe you should have got the same error. Anyway, what test output are we archiving if there's a failure? I believe we want to archive java core/dump files if those files are generated in that directory.

piyush286: A more general Q: What's the idea behind using `$(REPORTDIR)` (i.e. `/java/jenkins/workspace/Piyush_ODM_x86-64_linux/openjdk-tests/TestConfig/test_output_15436147118144/ODM_0`)? I believe TKG will be creating a new dir for each iteration (i.e. test_output_15436147118144/ODM_0, ODM_1, ODM_2...`) if we decide to run it with multiple iterations. Since we're are downloading the benchmark package here `${BUILD_ROOT}/performance/odm` (i.e. `/java/jenkins/workspace/Piyush_ODM_x86-64_linux/jvmtest/performance/odm`), we'll be running the java command from one of the scripts (i.e. `run_ilog_with_gcmv.sh`) in that directory. I'm planning to remove `run_ilog_with_gcmv.sh` from here in order to avoid having duplicate scripts and add any minor changes to our ODM perf repo.

performance/odm/playlist.xml
llxia: Perf jobs are updated to use `BUILD_LIST=performance/idle_micro`
Tested manually:
https://ci.adoptopenjdk.net/view/Test_perf/job/openjdk11_hs_perftest_x86-64_linux/85/

performance/odm/playlist.xml
llxia: > Without copying this `benchmark.sh` file, I was getting this error `chmod: cannot access './scripts/benchmark.sh': No such file or directory` 


I think $(REPORTDIR) was introduced after ODM was added, so we have some path issue that needs to be fixed.


> What's the idea behind using $(REPORTDIR)

The idea is to archive test output (including core/dump files) if there is a failure (without any user intervention). Each test has its own $(REPORTDIR). The build will automatically archive test_output_XXX (parent dir of $(REPORTDIR)) if the build is unstable (test failure) for debugging purpose.

In the existing code, ODM may have two archives:
- $(REPORTDIR) if the test fails
- `jvmtest/performance/odm/ilog_wodm881/leftoverResults` regardless of the test results ===> In order to compare measurements in ODM test, we also keep output files regardless of the test results. Ideally all output should be in $(REPORTDIR). We can refine this later if needed.

If we `cd` into another dir, core/dump files will not be archived in failure case. And to avoid cp test material around, I would prefer to use `$(TEST_RESROOT)`.

piyush286: @llxia All our benchmark scripts assume that we're inside the benchmark package directory while running the benchmark scripts. For example, if we aren't inside the `ilog_wodm881` dir (i.e. inside $(REPORTDIR)), `run_ilog_with_gcmv.sh` will throw this error `grep: bin/ra.xml: No such file or directory` since it assumes that it's in the same level dir as `bin`. Also, all our benchmark scripts take care of managing the dump/core files. Currently, for ODM, we're using `PERFFARM_DUMP="${WORKSPACE}/dump` to put dumps there but I'll change that to put it somewhere else (maybe `ilog_wodm881/leftoverResults`). Btw I'll be removing `run_ilog_with_gcmv.sh` from this repo in another PR in order to avoid having duplicates since we keep that script along with other ODM material. I'll make the necessary changes to manage the dump in our original ODM script so I won't make any changes to `run_ilog_with_gcmv.sh` in this repo for now. So for now, we'll be fine changing the dir from $(REPORTDIR) to `$(TEST_RESROOT)/ilog_wodm881` if you're okay with it. Please let me know if you have any other concerns that might be stopping us from merging this.

PR #751, opened 2018-12-17
Exclude jlm remote thread no auth test from Windows

Mesbah-Alam:
Exclude jlm remote thread no auth test from Windows
Signed-off-by: Mesbah_Alam@ca.ibm.com


llxia:
@Mesbah-Alam Could you port this change into extraVariations branch as well? Thanks

Issue #783, opened 2019-01-08
testEnvTeardown target is called before compiling and running tests


sophia-guo:
testEnvTeardown is called before compiling and running tests. Originally idea is we need to do testEnvTeardown before running.

@renfeiw any concern to remove it from compiling process?
Talked with @llxia including remove it from compiling process another step would be moving it to setup stage though it will be another issue.

renfeiw:
@sophia-guo I agree that it should be removed from compiling.

sophia-guo:
Closed by #785

PR #785, opened 2019-01-08
Stop calling testEnvTeardown target before building.

sophia-guo:
Only calling it before running tests
Remove bogus comment


sophia-guo:
Close #783

sophia-guo:
grinder https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/86/console

Issue #49, opened 2017-05-31
Enable jdk_awt tests and x11 DISPLAY env var set properly


smlambert:
There are currently a large number of failures in the jdk_awt group related to the x11 display not being set properly, in order to run those tests.  Expect the bulk of awt failures to go away after setting correctly.

23:42:59 ACTION: main -- Failed. Execution failed: `main' threw exception: java.awt.HeadlessException: No X11 DISPLAY variable was set, but this program performed an operation which requires it.

sophia-guo:
Double checked this failure only happens to linux(Windows is fine. In Windows, there is always a display).

For Linux we need set the DISPLAY environment variable to the address of X11 server and export that. What is the address of X11 server @gdams @bblondin

allandequeiroz:
Hello, I've been trying to reproduce it using a VM, I went to console mode only on linux and looked for the X server address.

![screen shot 2017-06-08 at 19 51 38](https://user-images.githubusercontent.com/629076/26945696-7d87f924-4c84-11e7-92d2-05268aaff240.png)

Then doing ```make test DISPLAY=:0``` apparently worked fine.

![screen shot 2017-06-08 at 19 48 30](https://user-images.githubusercontent.com/629076/26945734-9f9f1736-4c84-11e7-9484-da8a63f6e676.png)

What do you think if I try to get this address dynamically and set to DISPLAY env var?

sophia-guo:
Not sure if X11 forwarding is enabled on test machine? If yes we can either update jdk_awt test command to

<command>**export DISPLAY=localhost:0.0;**$(JAVA_COMMAND) -Xmx512m -jar $(Q)$(JTREG_DIR)$(D)lib$(D)jtreg.jar$(Q) \
	$(JTREG_BASIC_OPTIONS)  -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS)$(Q) \
	-w $(Q)$(TEST_RESROOT)$(D)work$(Q) \
	-r $(Q)$(TEST_RESROOT)$(D)report$(Q) \
	-jdk:$(Q)$(JDK_HOME)$(Q) \
	-exclude:$(Q)$(JTREG_TEST_DIR)$(D)ProblemList.txt$(Q) \
	$(Q)$(JTREG_TEST_DIR):jdk_awt$(Q); \
	$(TEST_STATUS)</command>
	
or export in Makefile should be ok.

allandequeiroz:
[...]

PR #789, opened 2019-01-09
Enable jdk_awt tests on UNIX-like OS platforms

sophia-guo:
1. Start Xvfb before test running on UNIX-Like OS platforms
2. Enable jdk_awt tests on UNIX-like OS platforms
3. Remove jck specific redundant code
4. Delete obsolete macos jenkins file


sophia-guo:
related #49

sophia-guo:
@smlambert correct. Should we limit it to jck and jdk test in case some special machine may not have it installed? Also other tests(function, system,etc) do not need this plugin invoked.

smlambert:
@sophia-guo - I considered to restrict to jck/openjdk test groups, as there have been some issues at the openj9 project (I had to quickly install jenkins plugin last night, and some linux machines appear to be missing xvfb).  I will check today if it looks like we should restrict, or can leave as is.

Issue #714, opened 2018-11-29
Include openj9 sanity functional tests to run in thirdparty containers


chandrams:
Include openj9 sanity functional tests to run in third party containers

Steps:

- Include a docker file to clone eclipse/openj9 repo
- Include a script to set the env variables and compile the test material and run sanity functional tests

chandrams:
Gave a run of openj9 sanity functional tests in docker container and see the below failures, compared it failures outside docker container, 14 of them are due to native libs not found. A couple of them highlighted fail only in docker container. 

SKIPPED test targets due to capabilities:
        cmdLineTester_pltest_hyp_HV_vmware_0 - capabilities (hypervisor:VMWare)
        cmdLineTester_pltest_hyp_HV_kvm_0 - capabilities (hypervisor:KVM)

FAILED test targets:
        algotest2_0
        ctest_0
        vmtest_0
        jsigjnitest_0
        gc_rwlocktest_0
        shrtest_linux_SE80_0
        thrstatetest_0
        cmdLineTester_jvmtitests_1
        cmdLineTester_jvmtitests_2
        cmdLineTester_jvmtitests_5
        cmdLineTester_jvmtitests_7
        cmdLineTester_jvmtitests_8
        **cmdLineTester_callsitedbgddrext_openj9_0**
        cmdLineTester_pltest_0
        cmdLineTester_pltest_numcpus_bound_linux_0
        **cmdLineTester_SCCommandLineOptionTests_1**

TOTAL: 188   EXECUTED: 111   PASSED: 95   FAILED: 16   SKIPPED: 77

chandrams:
Investigated the below two failures in docker containers and found the first one failing as it uses env variable LOGNAME to validate if the cache is created with the name of the user logged in, this doesn't work in a docker container as LOGNAME is blank.  Requires a test case change. To resolve the second failure, we will need to modify the content of /proc/sys/kernel/core_pattern on the host machine to just j9core and this would need root access.   

1) cmdLineTester_callsitedbgddrext_openj9_0

 [OUT] Could not load dump file and/or could not load XML file: Image file '/openj9/test/TestConfig/test_output_15433322797739/cmdLineTester_callsitedbgddrext_openj9_0/j9core.dmp' not found.
 [OUT] For a list of commands, type "help"; for how to use "help", type "help help"
 [OUT] > DDR is not enabled for this core file, '!' commands are disabled

[...]

PR #790, opened 2019-01-10
Included scripts to run openj9 sanity functional tests in docker containers

chandrams:
Included scripts to run openj9 sanity functional tests in docker containers
issue #714 
Signed-off-by: Chandrakala <chandra-ms@in.ibm.com>

### DIFF
--- a/thirdparty_containers/functional-test/build.xml
+++ b/thirdparty_containers/functional-test/build.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0"?>
+<project name="functional-test" default="build" basedir=".">
+	<taskdef resource="net/sf/antcontrib/antlib.xml" />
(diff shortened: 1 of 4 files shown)


llxia:
Thanks @chandrams 

I will approve this PR and we can refine this in a separate PR. 

Issues that need to be addressed:
- environment variables (such as SPEC, JDK_VERION, JAVA_IMP, etc) should be set. They should be passed in as parameters.
- test target (i.e., _sanity.functional) should be passed in a parameter, so we can run other targets (i.e., extended.functional, special.functional)

smlambert:
I will merge this, knowing it increases execution time of our external tests.  I will monitor if we are hitting our limits on machine time, and follow with actions to resolve if so.

PR #794, opened 2019-01-11
Re-include LambdaLoadTest_ConcurrentScavenge on xLinux

smlambert:
Verify fix for https://github.com/eclipse/openj9/issues/3491 and re-include test


smlambert:
Verified via Grinders at the openj9 project that this can be re-included (fix is in & test passes on xLinux).

PR #797, opened 2019-01-11
Restrict running xvfb to xlinux platform and jdk&jck test group only

sophia-guo:


sophia-guo:
Grinder: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/112/
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/113/

Issue #68, opened 2017-06-15
jdk_security3: sun/security/pkcs11/rsa/TestCACerts.java fails (SignatureException SHA256withRSA)


smlambert:
sun/security/pkcs11/rsa/TestCACerts.java and sun/security/rsa/TestCACerts.java fails with 
java.security.SignatureException: Signature does not match.  (Signature algorithm: SHA256withRSA), output from https://ci.adoptopenjdk.net/view/OpenJDK%20tests/job/openjdk_test_x86-64_linux/34/consoleFull:

/home/jenkins/workspace/openjdk_test_x86-64_linux/openjdk-test/OpenJDK_Playlist/openjdk-jdk8u/jdk/test/sun/security/pkcs11/rsa/TestCACerts.java
20:43:39 
20:43:39 ACTION: main -- Failed. Execution failed: `main' threw exception: java.security.SignatureException: Signature does not match.
20:43:39 REASON: Assumed action based on file name: run main TestCACerts 
20:43:39 TIME:   0.813 seconds
20:43:39 messages:
20:43:39 command: main TestCACerts
20:43:39 reason: Assumed action based on file name: run main TestCACerts 
20:43:39 Mode: agentvm
20:43:39 Agent id: 1
20:43:39 elapsed time (seconds): 0.813
20:43:39 configuration:
20:43:39 Boot Layer
20:43:39   class path: /home/jenkins/workspace/openjdk_test_x86-64_linux/openjdkbinary/j2sdk-image/jre/bin/../../lib/tools.jar
20:43:39               /home/jenkins/workspace/openjdk_test_x86-64_linux/openjdk-test/TestConfig/lib/jtreg/lib/javatest.jar
20:43:39               /home/jenkins/workspace/openjdk_test_x86-64_linux/openjdk-test/TestConfig/lib/jtreg/lib/jtreg.jar
20:43:39 
20:43:39 Test Layer
20:43:39   class path: /home/jenkins/workspace/openjdk_test_x86-64_linux/jvmtest/OpenJDK_Playlist/SE80/work/classes/0/sun/security/pkcs11/rsa
20:43:39               /home/jenkins/workspace/openjdk_test_x86-64_linux/openjdk-test/OpenJDK_Playlist/openjdk-jdk8u/jdk/test/sun/security/pkcs11/rsa
[...]

PR #798, opened 2019-01-11
Problem list SecretKeysBasic.sh for non-solaris platforms

jerboaa:
The test lists in its notes:
"test only runs on solaris at the moment"

It's failing on Linux. Thus, problem list it for all
platforms except Solaris.

Signed-off-by: Severin Gehwolf <sgehwolf@redhat.com>

### DIFF
--- a/openjdk_regression/ProblemList_openjdk8-openj9.txt
+++ b/openjdk_regression/ProblemList_openjdk8-openj9.txt
@@ -208,6 +208,7 @@ javax/crypto/CryptoPermission/CryptoPolicyFallback.java	125	generic-all
 javax/crypto/CryptoPermission/TestUnlimited.java	125	generic-all
 sun/security/rsa/TestCACerts.java	125	generic-all
 sun/security/ssl/SSLSocketImpl/SSLSocketCloseHang.java	125	generic-all
+sun/security/pkcs11/KeyStore/SecretKeysBasic.sh	JDK-8189603	linux-all macosx-all windows-all
 
 ############################################################################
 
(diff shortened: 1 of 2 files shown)


smlambert:
Nice spotting that @jerboaa!  

We should also update the openj9 problem list file.  

Is it actually related to https://github.com/AdoptOpenJDK/openjdk-tests/issues/68 (the TestCACerts.java test failure), may be more accurate to relate exclusion to https://bugs.openjdk.java.net/browse/JDK-8189603 ?

jerboaa:
> Nice spotting that @jerboaa!

:) Thanks for the review!

> Is it actually related to #68 (the TestCACerts.java test failure)

Aha! That's what those magic numbers mean. I didn't know what to do with that so just copied the line above. I'll fix it.

> may be more accurate to relate exclusion to https://bugs.openjdk.java.net/browse/JDK-8189603 or some other bugDB issue?

Would referring to it via `JDK-8189603` be accurate? Those exclude lists are new territory for me.

jerboaa:
@smlambert I've updated the PR. Let me know whether that looks right to you.

smlambert:
Essentially the exclude files go by this format (3 parts):
"test to exclude"   "link to issue that when fixed, indicates re-include test"  "platforms to exclude"

So, we should link to issues in the upstream repo where the fix must be put, when possible.  In some cases we may find we have to report/create issues in the upstream repos.  In other cases, we created issues in openjdk-tests repo, because it was not determined if the failure was because of test, product or machine issue, so we created a placeholder, and hope someone will take the time to properly triage it.  
  
Interestingly, looks like https://bugs.openjdk.java.net/browse/JDK-8186098 was fixed (which is mentioned as cause of JDK-8189603), which means that we should check if we have that fix in our test material...

PR #800, opened 2019-01-11
Use mv instead of cpio

llxia:
cpio does not preserve file encoding. This causes problem on zos. Change
to use mv cmd.


llxia:
Grinder: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder/796/

Issue #821, opened 2019-01-22
Intermittent failures on Windows


sophia-guo:
**No issues on JBS: report to openj9?**
Test java/rmi/server/RemoteServer/AddrInUse.java.AddrInUse failed with 'TEST FAILED: export on already-bound port succeeded' on all platforms with openj9 java8
https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk8_j9_openjdktest_x86-64_windows/178/testReport/junit/java_rmi_server_RemoteServer_AddrInUse/java/AddrInUse/
https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk8_j9_openjdktest_x86-64_linux_linuxXL/92/testReport/junit/java_rmi_server_RemoteServer_AddrInUse/java/AddrInUse/
java/io/File/DeleteOnExitNPE.java failed intermittently with ' No output should have been received from the process' on win64 with openj9 java8
https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk8_j9_openjdktest_x86-64_windows/177/testReport/junit/java_io_File_DeleteOnExitNPE/java/DeleteOnExitNPE/

java/util/zip/ZipFile/DeleteTempJar.java intermittently failed on Win32 with openj9 java8
https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk8_j9_openjdktest_x86-32_windows/115/consoleFull

**Others has issues in JBS, however status is resolved or closed. Asking for reopen?**
java/nio/channels/ServerSocketChannel/Basic.java : intermittent timeout failure on win32|64 with hotspot&openj9 java8

https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk8_hs_openjdktest_x86-32_windows/110/
https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk8_hs_openjdktest_x86-64_windows/182/testReport/java_nio_channels_ServerSocketChannel_Basic/java/Basic/
 
Exactly same issue reported and resolved in JBS. May need to re-open? 
https://bugs.openjdk.java.net/browse/JDK-8143100

Test java/nio/channels/ServerSocketChannel/AdaptServerSocket.java failed with Accept did not time out on win64 with hotspot java8. 
[...]

PR #822, opened 2019-01-22
Exclude intermittent time out failure tests on windows

sophia-guo:


sophia-guo:
Related with #821

PR #825, opened 2019-01-23
Exclude intermittent time out failure tests on windows

sophia-guo:


sophia-guo:
All are intermittent failures. 5 of them are reported in JBS with status 'resovled' or 'closed'. May need help to re-open it. 3 of them are openj9 specific, need to report to openj9. Details in #821 , exclude for now.

smlambert:
I have removed the label for now.  I think I know why they have that label.  Docker got installed on them, and so they would be available for application tests, but they may be missing something else for running other testing.  

We should look at what is missing and have it added, since when I remove these 2 machines from testing, it leaves us only 1 for this platform.

PR #838, opened 2019-01-28
Exclude failures on openj9 with java8

sophia-guo:


sophia-guo:
Related with #830

Issue #747, opened 2018-12-14
Support system test execution in parallel


llxia:
With [extra variations enabled ](https://github.com/AdoptOpenJDK/openjdk-tests/blob/extraVariations/systemtest/playlist.xml), sanity and extended system build can run for 20+ hrs.
- Test-sanity.system-JDK8-linux_x86-64 - 18hrs
- Test-sanity.system-JDK8-aix_ppc-64_cmprssptrs,Test-extended.system-JDK8-linux_x86-64 and Test-extended.system-JDK11-linux_x86-64 got killed after 20hrs

We should support system test execution in parallel. This could help us to shorten system test build time. Our goal is to complete the system test and functional test builds within 3hrs.

llxia:
Ideally, we can create separate folders and break the [playlist.xml](https://github.com/AdoptOpenJDK/openjdk-tests/blob/extraVariations/systemtest/playlist.xml) into multiple smaller sets. Similar as jck test: https://github.com/AdoptOpenJDK/openjdk-tests/tree/master/jck

I am thinking of separte playlist based on features. i.e., * LoadTest*, SharedClasses*, TestJlm*, modularity, others). If LoadTest set is too big, we can further break it into MathLoadTest*, DaaLoadTest*, LambdaLoadTest*, Mauve*, etc.

@ShelleyLambert @Mesbah-Alam any suggestions? Thanks

smlambert:
That seems like the most logical and direct approach, and is in keeping with existing subdirs approach used by other testing (so leverages what TKG already knows how to deal with this scenario).

llxia:
Thanks @ShelleyLambert . I will give it a try.

llxia:
system tests are running `special` level tests in parallel in internal Jenkins. Close this issue.

PR #857, opened 2019-02-06
Support systemtest parallelization

llxia:
- common.xml is created to have common targets (git clone stf, system
test material, etc)
- it is included in systemtest/build.xml and each systemtest subdirs
build.xml
- flag common_build.executed is created to make sure common.xml will
only be invoked once. That is, if systemtest/build.xml invokes
common.xml, subdirs build.xml should not invoke it (non-parallel runs).
Otherwise, each subdirs build.xml should invoke it (parallel runs).
- once IS_PARALLEL is selected in Grinder, child job will be triggered
per `BUILD_LIST/<subdir>`. The child job will inherit all parameters from
its parent job with modified BUILD_LIST (set to original
`BUILD_LIST/<subdir>`) and IS_PARALLEL (set to false). That is, each child
job is an independent job. It does its own git clone, compilation, test
execution and test output archive.

Issue: #747

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -10,20 +10,6 @@ def getBuildList() {
 	return TESTPROJECT
 }
 
-def archiveTestBinaries() {
-	dir("$WORKSPACE") {
-		sh "tar -zcf test-binaries.tar.gz ./openjdk-tests ./jvmtest ./openjdkbinary"
-	}
-	archiveArtifacts artifacts: '**/test-binaries.tar.gz', fingerprint: true, allowEmptyArchive: false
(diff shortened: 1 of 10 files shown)


llxia:
Due to the lack of machines, parallel testing is done on the internal farm.

openjdk regression jdk_math: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/120/
functional floatSanityTests: https://ci.eclipse.org/openj9/view/all/job/Lan_Test-Grinder/4/

llxia:
Below is a screenshot for parallel runs in blue ocean view:

screen shot 2019-02-06 at 9 02 02 am

PR #862, opened 2019-02-06
Exclude tests failure on Max_XL

sophia-guo:
The first time running on Max_XL and failed


sophia-guo:
related with #830

PR #867, opened 2019-02-07
Exclude tests incompatible with current test infra

ben-walsh:


ben-walsh:
Fair point, I just followed the exclusions that issue led to for JDK8 - https://github.com/AdoptOpenJDK/openjdk-tests/blame/master/openjdk_regression/ProblemList_openjdk8-openj9.txt

smlambert:
Yes, I did not look closely at the previous one, but should have asked then.  Thanks for putting the question re: core issue and how to properly resolve in https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/143.

sophia-guo:
Probably need to update **linux-s390** to **linux-s390x**. Better to backport to ProblemList_openjdk8-openj9.txt.

PR #874, opened 2019-02-08
Update exclude

sophia-guo:
Limit exclude to s390x specific
Add exclude failure on Mac with HS & Openj9


sophia-guo:
Backport to java8, relates with https://github.com/AdoptOpenJDK/openjdk-tests/pull/867

PR #878, opened 2019-02-11
Exclude failing jdk/nio/zipfs/ZipFSTester.java test

ben-walsh:
@sophia-guo I'm adding in an extra entry here for test jdk/nio/zipfs/ZipFSTester.java to reference the OpenJ9 issue that I've raised.

### DIFF
--- a/openjdk_regression/ProblemList_openjdk11-openj9.txt
+++ b/openjdk_regression/ProblemList_openjdk11-openj9.txt
@@ -226,7 +226,6 @@ java/nio/file/attribute/BasicFileAttributeView/UnixSocketFile.java https://githu
 java/nio/file/Files/CopyAndMove.java https://github.com/AdoptOpenJDK/openjdk-tests/issues/795 macosx-all
 java/nio/channels/AsyncCloseAndInterrupt.java	https://github.com/AdoptOpenJDK/openjdk-tests/issues/602	linux-all
 jdk/nio/zipfs/ZipFSTester.java https://github.com/eclipse/openj9/issues/4679 linux-x86-64,macosx-all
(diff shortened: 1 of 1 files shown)


smlambert:
@ben-walsh - please just replace the more general entry with yours, lets keep it to one entry per exclusion, deferring to the most specific issue raised (so in this case the issue raised upstream at openj9 repo).

ben-walsh:
Requested change made.

sophia-guo:
I'm not sure if we need the issue raised in Openj9. Test intermittently failed with HS Mac though with different behavior.

https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk11_hs_openjdktest_x86-64_macos/122/testReport/junit/jdk_nio_zipfs_ZipFSTester/java/ZipFSTester/

If there is the issue with test itself maybe no need to investigate other failures before test is fixed?

smlambert:
If its failing differently for hotspot it could be 2 different problems, but yes, ultimately as you all exclude tests, the intent is to point to the best description of the problem that can be given, upstream as far as it seems correct to report it.  

Generally, for openjdk tests there are really only a few options to be determined:
1) its a machine issue                             --> report to AdoptOpenJDK/openjdk-infrastructure
2) its a test setup issue at adopt           --> report to AdoptOpenJDK/openjdk-tests
3) its a test issue at openjdk                  --> report to openjdk
4) its a product issue at openjdk           --> report to openjdk
5) its a product issue at openj9             --> report to openj9 (which could get redirected to extensions repo)

did I miss anything?  for 3) and 4) it may already have been reported, and we need to check JBS.  If not, we need authors/contributors to raise those issues in the JBS (using the template provided to us in the [guidance wiki](https://github.com/AdoptOpenJDK/openjdk-tests/wiki/Guidance-for-Creating-OpenJDK-Test-Defects))

PR #884, opened 2019-02-12
Update referenced issue

ben-walsh:


sophia-guo:
Looks like the updated references is not correct.

ben-walsh:
Issue repo wrong but issue number correct. Fixed now.

sophia-guo:
Thanks Ben. Would you mind also update ProblemList_openjdk8-openj9.txt?

PR #898, opened 2019-02-14
Add ARTIFACTORY_VERSION variable

AdamBrousseau:
- Only attach buildInfo for artifacts being uploaded
  to an enterprise Artifactory server instance. OSS version
  doesn't support buildInfo

Related eclipse/openj9#4986

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -1,4 +1,7 @@
 #!groovy
+
+import groovy.json.JsonSlurper;
+
 def getBuildList() {
 	def TESTPROJECTS = [system:'systemtest', perf:'performance', jck:'jck', external:'thirdparty_containers', functional: 'functional', openjdk:'openjdk_regression', jdk:'openjdk_regression', runtest:'', sanity:'', extended:'']
 	String fullTarget="${TARGET}"
@@ -14,7 +17,7 @@ def makeTest(testParam) {
 	String tearDownCmd = "if [ `uname` = AIX ]; then MAKE=gmake; else MAKE=make; fi; \$MAKE -f ./openjdk-tests/TestConfig/testEnv.mk testEnvTeardown"
 	try {
 		sh "$tearDownCmd"
-		sh "./openjdk-tests/maketest.sh $testParam" 
+		sh "./openjdk-tests/maketest.sh $testParam"
 	} finally {
 		sh "$tearDownCmd"
 	}
@@ -29,7 +32,7 @@ def setupEnv() {
 		if (params.ADOPTOPENJDK_REPO.startsWith('git') && params.USER_CREDENTIALS_ID && !SPEC.startsWith('zos')) {
 			remoteConfigParameters.put('credentialsId', "${params.USER_CREDENTIALS_ID}")
 		}
-		
+
 		checkout([
 			$class: 'GitSCM', branches: [[name: "*/${params.ADOPTOPENJDK_BRANCH}"]],
 			extensions: [[$class:'CloneOption', shallow:true, depth:1],[$class: 'CleanBeforeCheckout'],[$class: 'RelativeTargetDirectory', relativeTargetDir: 'openjdk-tests']],
@@ -451,6 +454,8 @@ def uploadToArtifactory() {
 		def artifactoryRepo = params.ARTIFACTORY_REPO ? params.ARTIFACTORY_REPO : "sys-rt-generic-local"
 		def jenkinsDomain = params.JENKINS_DOMAIN ? params.JENKINS_DOMAIN : getJenkinsDomain()
 		def artifactoryUploadDir = "${artifactoryRepo}/${jenkinsDomain}/${JOB_NAME}/${BUILD_ID}/"
+		def artifactoryUrl = server.getUrl()
+		def artifactoryCreds = server.getCredentialsId()
 
 		def uploadSpec = """{
 			"files":[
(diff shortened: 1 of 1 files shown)


smlambert:
Would it not be simpler/better to have the same policy for free artifactory instances, as in still clean up and enforce a deletion policy, just with a bigger value than 20?

AdamBrousseau:
The issue is the free version does not support `buildInfo`. Trying to do `server.publishBuildInfo buildInfo` causes it to fail because the server does not support it.

smlambert:
Thanks for the explanation.  I will like to raise an issue against the TaaS free server to request support for it, in the event that it could be on offer and we could collapse this code into 1 case.

AdamBrousseau:
The TaaS one is paid, it's the one we're setting up at OpenJ9 that is free.
There were a few [options](https://github.com/eclipse/openj9/issues/2637#issuecomment-460444497) but at the moment we've headed down the free self hosted road.

AdamBrousseau:
@ShelleyLambert or @llxia this is ready for review and can be merged once the OpenJ9 PR is merged.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -461,13 +461,18 @@ def uploadToArtifactory() {
...
 			}"""
 
-		def buildInfo = Artifactory.newBuildInfo()
-		// if ARTIFACTORY_NUM_ARTIFACTS is not set, set to keep 20 build artifacts
-		def numArtifacts = params.ARTIFACTORY_NUM_ARTIFACTS ? params.ARTIFACTORY_NUM_ARTIFACTS : 20
-
-		buildInfo.retention maxBuilds: numArtifacts, deleteBuildArtifacts: true
-		server.upload spec: uploadSpec, buildInfo: buildInfo
-		server.publishBuildInfo buildInfo
+		if (params.ARTIFACTORY_VERSION == 'ent') {
```
llxia: - Could we make `ent` to be the default value? The current code force all existing jobs that use artifactory to provide an extra parameter - ARTIFACTORY_VERSION.
- Do we have a cleanup story for `oss`?

AdamBrousseau: We can, up to you. That is how I structured the OpenJ9 version of the `if`, but that doesn't mean we have to do the same thing for test. If you prefer I can do `if oss, else`.
For the cleanup, Sam has a pr eclipse/openj9#4996 which I have yet to review.

llxia: On second thought, is it possible to set `ARTIFACTORY_VERSION` based on `ARTIFACTORY_SERVER` value? As we set the Artifactory `Server ID` in [Jenkins artifacoty plugin](https://ci.eclipse.org/openj9/configure), so we can decide on how to name the server. Other alternatives could be finding artifactory version through artifactory API or using try catch to see if we can get `buildInfo`. But the first option seems simpler.

It would be great if we could do it without introducing a new parameter as we are already experiencing parameters explosion https://github.com/AdoptOpenJDK/openjdk-tests/issues/902

AdamBrousseau: Maybe this will help (isArtifactoryPro)
https://javadoc.jenkins.io/plugin/artifactory/org/jfrog/hudson/ArtifactoryServer.html#isArtifactoryPro-org.jfrog.hudson.DeployerOverrider-hudson.model.Item-
Will follow up tomorrow.

AdamBrousseau: Well that didn't work. The server object created is actually a different class org.jfrog.hudson.pipeline.common.types.ArtifactoryServer
not 
org.jfrog.hudson.ArtifactoryServer
Which doesn't have a method we could use to get the version afaict. I looked through the Doc and cannot find a way to get what we need.

The alternative would be to do a REST API call over https.
https://www.jfrog.com/confluence/display/RTF/Artifactory+REST+API#ArtifactoryRESTAPI-VersionandAdd-onsinformation
For OSS version returns
```
{
  "version" : "6.5.2",
  "revision" : "60502900",
  "addons" : [ ],
  "license" : "Artifactory OSS"
}
```
The Pro version has a large license key in the license field. So I could do the call and do `if OSS <1>, else <2>`.

AdamBrousseau: Have a workaround that curls the API. Will test tomorrow. Also opened a feature request on the plugin
https://issues.jenkins-ci.org/browse/JENKINS-56512

llxia: Thanks a lot for trying different options, Adam. :)

AdamBrousseau: I was trying to get the ART credentials via the `server` object
`server.getCredentialsId()` but that doesn't seem to be working (always returns `null`).
I opened https://issues.jenkins-ci.org/browse/JENKINS-56531 to address that issue. As a workaround, for our workaround, I thought about using `CUSTOMIZED_SDK_URL_CREDENTIAL_ID` since we already pass that for the get.sh. The problem with that would be, for example, a user running a grinder on taas, pulling an sdk from adopt. The get.sh wouldn't require the creds but the Artifactory upload step would.

I'm going to see if we can try/catch the upload to always do buildinfo and if it fails, upload without it. 
```
try {
	server.upload spec: uploadSpec, buildInfo: buildInfo
	server.publishBuildInfo buildInfo
} catch {
	server.upload spec: uploadSpec
}
```
Will update once I test it. hopefully the combined try/catch doesn't add alot of extra time.

AdamBrousseau: I'm glad we tried the try/catch. The original code is working now. I can publish buildInfo to the OSS server. The artifact properties and build retention data is basically ignored but it doesn't fail. I'm not sure why it failed when I attempted it last week. I will confirm on the test side and if it works I will close this PR.

PR #909, opened 2019-02-19
Replace special character in playlist.xml

renfeiw:
- replace with make variable or character reference


llxia:
This PR is related to TKG changes to use XML::Parser module, so special characters in playlist.xml need to be replaced. https://github.com/eclipse/openj9/pull/4785

llxia:
Could you please provide Grinder links if possible? Thanks

renfeiw:
Tested both level sanity and extended for group openjdk_regression and system internally.

Issue #915, opened 2019-02-21
Leftover images and containers issue


sophia-guo:
There are some images and containers are left over by external test jobs. Need to clean  up and figure out why it happens.

sophia-guo:
- Images are built when doing test compile and then left over. They will only be deleted when external test job doing compile again. 

- Container is automatically removed when it exits. All `docker cp` after `docker run --rm` actually do nothing.

sophia-guo:
images left on test-packet-ubuntu1604-x64-1, which can't be deleted with jenkins user.

<img width="809" alt="screen shot 2019-02-22 at 4 14 22 pm" src="https://user-images.githubusercontent.com/18030625/53271647-1759f100-36bd-11e9-8eaf-e1c380371a9d.png">

sophia-guo:
There is another case we cann't handle right now. Test build is aborted either manually or due to timeout during stage `test running`. In this case either containers or image will be left on slave, which has to be manually clean up.

smlambert:
Likely we can ask infra to run a cron job to clean up machines, to mitigate the cases that we can not handle via Jenkins scripts / Jenkins user id.

smlambert:
did your recent PR https://github.com/AdoptOpenJDK/openjdk-tests/pull/1104 address this fully @sophia-guo ?

sophia-guo:
Yes, this was addressed by a few recent PRs. Close it.

Issue #919, opened 2019-02-22
function and system tests under third party container tests are grouped as either functional or system instead of external


sophia-guo:
Function and system test under third party container tests are grouped as either functional or system instead of external, is it expected?

smlambert:
@llxia and I are aware of this.  We have not corrected it, because once we do these tests will run, and since all tests run in sequence, we will hit the time out limit (10hrs) and abort.

This should get fixed once we break up these external tests to run individually (each subdir, like scala, elasticsearch, derby, etc running in its own job).

smlambert:
I see this has been [corrected](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/external/functional-test/playlist.xml#L27) for all targets, so closing.

PR #920, opened 2019-02-22
Force delete containers and images when test finished

sophia-guo:
Force delete containers and images when test finished
Delete containers only when docker cp finished
Remove leftover volume map and add EXTRA_DOCKER_ARGS

related #915 


sophia-guo:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1113/console

system fails may need to be exclude. [openjdk8_hs_externaltest_x86-64_linux](https://ci.adoptopenjdk.net/view/Test_external/job/openjdk8_hs_externaltest_x86-64_linux/331/) passed as those tests are group as system so they won't run in the Test_external. See #919

PR #930, opened 2019-02-28
Modified SPEC for aarch64_linux

Yuehan-Lin:
Signed-off-by: Yuehan-Lin <Yuehan.Lin@ibm.com>


Yuehan-Lin:
@smlambert for review
I modified the SPEC value for openjdk_aarch64_linux into linux_aarch64_cmprssptrs, since corresponding buildspec is enabled in https://github.com/eclipse/openj9/tree/master/buildspecs

PR #938, opened 2019-03-04
Set failIfNoResults to true in nonparallel build

renfeiw:
- the build should be failed if there's no test


renfeiw:
related PR: https://github.com/eclipse/openj9/pull/4967

PR #961, opened 2019-03-07
Add wildfly dependent maven

sophia-guo:


sophia-guo:
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/162/console

Install dependent maven and there are other test failures. Keep it as disable.

PR #966, opened 2019-03-11
Create Test Triage Rotas

adamfarley:
These rotas will enable us to determine which areas of testing
are being monitored and triaged on a regular basis, and which
areas are not, for the purpose of ensuring consistent quality.

This enables individuals to recognise gaps in areas of testing
that are important to them, so they can most efficiently allocate
their time.

It also prevents effort duplication between triagers actively working
on the same area of testing.

Review requested: @andrew-m-leonard @ben-walsh @smlambert 

### DIFF
--- a/triage/README.md
+++ b/triage/README.md
@@ -1,0 +1,38 @@
+# AdoptOpenJDK Test Triage Rotas
+
+### OpenJ9 Rotas:
+- [OpenJDK8](./openj9_test_triage_rotas.md#8)
+- [OpenJDK11](./openj9_test_triage_rotas.md#11)
+- [OpenJDK12](./openj9_test_triage_rotas.md#12)
+- [OpenJDK Head](./openj9_test_triage_rotas.md#head)
+
+### Hotspot Rotas:
+- [OpenJDK8](./hotspot_test_triage_rotas.md#8)
+- [OpenJDK11](./hotspot_test_triage_rotas.md#11)
+- [OpenJDK12](./hotspot_test_triage_rotas.md#12)
+- [OpenJDK13](./hotspot_test_triage_rotas.md#head)
+
+### Details:
+
+Triage is defined here as "Exclude the failing test, raise a bug against the relevant party, and put the bug URL in the exclude file.".
+
+Triagers named in the rotas linked above are committed to "best effort" triage on test runs which are:
+- That test group.
+- That platform.
+- That release.
+- That Virtual Machine.
+- Run against release streams.
+- Run as part of a predetermined test schedule (e.g. the automated nightlies).
+
+Triagers named above are *not* committed to triaging test runs which are:
+- Run by an individual.
+- Run against private forks/branches.
+- Run outside of the predetermined schedule.
+- Run using any test material other than the primary source for that platform/release/test group.
+- Run without the triager's knowledge.
+
+(Note: These are guidelines only, and do not restrict triagers from committing to perform triage outside of these limits.)
+
+Finally, do note that the triagers in the rotas do not have a monopoly on triaging. 
+
+Non-rota individuals are explicitly welcome to perform triage as and when they can.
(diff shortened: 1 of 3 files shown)


adamfarley:
I was planning to get the bare rota approved before asking people what they want to sign up for.

I can include peoples' names as part of this PR, once the layout, test groups, etc, have been green lit.

adamfarley:
Will ask around in the meantime and get some names.

smlambert:
That will be good for you to gather names and intended missions, as you may find that the categories you have created in your tables are very granular, and a single name would fill an entire row, so may be more sensible to keep the rota simpler / less granular.  

Personally, I reiterate that this feels like a heavy corporate overlay of process onto an open and agile project, but as enough seem keen on it, we can try it out and see if we can tell whether it has any effect on the state of test jobs.  I do not wish to dampen your enthusiasm, but I also feel responsible to not bring in excess process where it may not be needed.

Your description mentions the concern that triagers may be actively working on the same area.  Have we hit scenarios like that often?

There has already been much good work to sanitize the 8 test targets that are "sanity.openjdk" (by Ben & Sophia), and around 62 test targets from "sanity.system" and "extended.system" (by Mesbah), and now working on the 11 targets that make up "sanity.external" and "extended.external" (Sophia and Mesbah).

Anyway, all that to be said, we can give this a try.  If it looks like its not making any noticeable & positive impact, we can always revise or remove it.

adamfarley:
Granularity: Another way to handle this might be to replace platforms with days of the week. Thus allowing the rota to, well "rotate". Enables cross-skilling, prevents burnout, etc. How does that sound?

Process:  A fair comment. I genuinely wish there was a more agile way of doing this that allowed reliable triage coverage, and I directly invite people to suggest such a method (perhaps a rota of squad names instead of people names, with group responsibility and dynamic triage allocation?). 

In the meantime, I feel this is a good way to gain coverage and a list of skilled triagers.

Overlap: Not on Adopt, yet, to my knowledge. This is more of an incidental benefit.

Impact: Very true. Lets see how it goes.

smlambert:
Days of week sound even more hard to administer...  :)

The concept of skilling people up is good.  On that note, perhaps you should rename your top-level dir to triage rather than triage_rotas, so that we can additionally add some "how-to triage" guidance doc in there (in a later PR).

On this side of the ocean, we don't really have spare folks to rotate into the schedule, we are also helping triage openj9 and internal builds (which is actually a much greater number of total test jobs than at this project presently), but let's see what we can cover, and if we can find some additional assistance.

adamfarley:
Days: I'm not sure what "administer" means in this context. We can monitor triage throughput via git commits if we need to, but I don't plan to monitor peoples' triage efforts on a daily basis. It'd just get the triagers' backs up, and it doesn't feel very trusting.

Skills: The folder has been renamed.

Busyness: Thank you for telling me. I'm sure each team will contribute resource as and when they feel they can, and I'm hoping this rota will help us allocate iteration time appropriately as people become available.

As you say; lets see what we can cover.

smlambert:
I should have used the term, self-administer, as that is what I meant, and the whole concept of a rota gets my back up, so I get what you are saying... it is unpleasant to feel that people do not trust in or disregard ones expertise or advice.

ben-walsh:
@adamfarley Looks good to me. Thanks for putting this together.

andrew-m-leonard:
Looks good to me Adam

adamfarley:
Thanks everyone. :)

karianna:
I think this is missing a link from the main README - no way new folks would know about this...

smlambert:
added here: https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/README.md#how-can-you-help

sxa555:
Once this is live it might be a good thing to post in #general and maybe even as a tweet from adoptopenjdk to see if we can get more people involved in the testing side of the project as whole. It could well be a nice low-barrier entry into the project to get people to help out with triage!

adamfarley:
My plan was to do that this morning. Didn't. Once Shelley's moved it to the wiki, I'll put a thing into various slack channels alerting people to the rota, and inviting people to use it.

PR #968, opened 2019-03-11
Changed AUTO_DETECT into boolean variable

Yuehan-Lin:
Signed-off-by: Yuehan-Lin <Yuehan.Lin@ibm.com>


llxia:
Tested on internal Grinder

PR #979, opened 2019-03-13
Stop openjdk-systemtest being built multiple times

lumpfish:
Fixes https://github.com/AdoptOpenJDK/openjdk-tests/issues/978


smlambert:
Thanks for this @lumpfish !  Can I assume you've run a few tests to verify it works as described (add grinder job links if applicable)?

llxia:
Due to https://stackoverflow.com/questions/4014094/in-what-order-are-ant-targets-if-and-depends-evaluated, I think we need to update all depended targets, not just `build-dependencies`? For example, `build-dependencies` depends on `configure_systemtest`, `configure_systemtest` depends on `check_systemtest`, etc.

lumpfish:
Grinder run from my branch here: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1246/consoleFull

Using the occurrence of one of the ant targets as an indication of how many times the build is being run I get this difference:

1. Latest nightly system test run:
print-ant-1.10.2-error: occurs 158 times
Time to build:  currently seems to be between about 8 mins and 20 mins: https://ci.adoptopenjdk.net/job/openjdk11_j9_systemtest_x86-64_linux/

2. Grinder run with the change:
print-ant-1.10.2-error: occurs 46 times
Time to build: 3 minutes 8 seconds

I surprised the different isn't greater - on my local machine it was about 3 mins vs 17 mins.

Mesbah-Alam:
> I think we need to update all depended targets, not just build-dependencies. For example, build-dependencies depends on configure_systemtest, configure_systemtest depends on check_systemtest, etc. 

Due to  https://stackoverflow.com/questions/4014094/in-what-order-are-ant-targets-if-and-depends-evaluated, I think we only need to update dependent targets where we have an "if" condition paired with "depends".

lumpfish:
Regarding "I think we need to update all depended targets" - it probably depends on the context.  If a target is just checking something or doing something trivial it probably doesn't matter much in terms of run time.

llxia:
If there a particular reason to use `depends`? Could we change the code to use regular ant call? 

```
<target name="common_build" unless="common_build.executed">
	<echo message="target common_build is called....." />
        <antcall target="build-dependencies" inheritall="true" />
</target>
```

In this way, all predecessors should be guarded by the flag.

llxia:
Grinder with above change https://github.com/AdoptOpenJDK/openjdk-tests/pull/979#issuecomment-472467941: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1247/

print-ant-1.10.2-error: occurs 30 times
Time to build: 3min50sec

lumpfish:
Yes, that would work too.  Raise a different PR if you prefer.

smlambert:
If I understand this correctly, @llxia 's proposed solution may be preferential, as it would cover us for any new subdirectories of system test material that we may add?

lumpfish:
There is a chain of dependencies.  This PR will stop the 'big' target being executed multiple times, @llxia 's suggestion will stop the whole chain being executed multiple times (though executing the other dependencies multiple times doesn't affect the total run time greatly).

llxia:
Sure, I will create a PR for the change.

smlambert:
Thanks for identifying the issue and your explanation @lumpfish - based on it let's go with #986 and can close this one.

PR #986, opened 2019-03-14
Switch to use regular ant call instead of depends

### LINKED ISSUE #978: openjdk-tests build builds openjdk-systemtest multiple times
I have noticed that a build of openjdk-system test takes much longer when built via the openjdk-tests builds scripts than via the openjdk-systemtest build script (make build).

This seems to be why:
1. In openjdk-tests/systemtest/build.xml contains a target
```
	<target name="build" depends="dist,common_build">
		<property name="common_build.executed" value="true" />
		<subant target="" inheritall="true">
			<fileset dir="." includes="*/build.xml" />
		</subant>
	</target>
```
2. The dependent target common build is in openjdk-tests/systemtest/common.xml and says:
```
	<!-- To make sure common_build target only excute once, the common_build target runs only if common_build.executed value is not set -->
	<target name="common_build" depends="build-dependencies" unless="common_build.executed">
		<echo message="target common_build is called....." />
	</target>
```
3. The dependent target build-dependencies is also in openjdk-tests/systemtest/common.xml
```
	<!-- target to build all projects in the repository.  -->
	<!-- dir must be set on the ant task otherwise the basedir property is not set to a new value in the subant task. -->
	<!-- Also make sure stf is built -->
	<target name="build-dependencies" depends="configure_systemtest">
		<if>
			<contains string="${JVM_VERSION}" substring="openj9"/>
			<then>
				<ant antfile="${SYSTEMTEST_ROOT}/openj9-systemtest/openj9.build/build.xml" dir="${SYSTEMTEST_ROOT}/openj9-systemtest/openj9.build" inheritAll="false">
					<property name="java.home" value="${env.JAVA_HOME}"/>
				</ant>
			</then>
			<else>
				<ant antfile="${SYSTEMTEST_ROOT}/openjdk-systemtest/openjdk.build/build.xml" dir="${SYSTEMTEST_ROOT}/openjdk-systemtest/openjdk.build" inheritAll="false">
					<property name="java.home" value="${env.JAVA_HOME}"/>
				</ant>
			</else>
		</if>
		<antcall target="common_dist" inheritall="false" />
	</target>
```
Running the openjdk-systemtest/openjdk.build/build.xml builds all the tests in the openjdk-systemtest repository.

4. After executing build-dependencies the original build target executed any other build.xml files it can find under openjdk-tests/systemtest.  Each on of those also has a dependency on the common-buil;d target - e.g. 
```
	<target name="build" depends="init, common_build">
		<echo message="xxxxxxxxx running build in daaLoadTest/build.xml"/>
		<copy todir="${DEST}">
			<fileset dir="." includes="*.mk,*.xml"/>
		</copy>
	</target>
```

5. The common_build target has tried to guard itself from multiple executions by setting the common_build.executed property.  However, if / unless attributes are only evaluated after dependencies have been executed, so this does not work (e.g. see https://stackoverflow.com/questions/4014094/in-what-order-are-ant-targets-if-and-depends-evaluated).  As a result build-dependencies executes for each build.xml file under openjdk-tests/systemtest. making the build about 10 times slower than it should be.
```
	<!-- To make sure common_build target only excute once, the common_build target runs only if common_build.executed value is not set -->
	<target name="common_build" depends="build-dependencies" unless="common_build.executed">
		<echo message="target common_build is called....." />
	</target>
```

6. This can be fixed by guarding the build-dependencies target with the common_build.executed property directly.

7. As an aside, it is not clear to me what the 'subant' calls to the build.xml files are meant to achieve, since the common build already builds all the openjdk-systemtest test cases. @Mesbah-Alam - any insights?

llxia:
Since if / unless ant attributes are only evaluated after dependencies
have been executed, switch to use regular ant call instead of depends.

Fixes #978

### DIFF
--- a/systemtest/common.xml
+++ b/systemtest/common.xml
@@ -217,8 +217,9 @@
 	</target>
 	
 	<!-- To make sure common_build target only excute once, the common_build target runs only if common_build.executed value is not set -->
-	<target name="common_build" depends="build-dependencies" unless="common_build.executed">
+	<target name="common_build" unless="common_build.executed">
 		<echo message="target common_build is called....." />
+		<antcall target="build-dependencies" inheritall="true" />
 	</target>
 		
 	<target name="clean">


smlambert:
Discussion and link to grinder found in #979

PR #987, opened 2019-03-14
Update tck tests from extended to sanity

sophia-guo:

### DIFF
--- a/thirdparty_containers/derby/playlist.xml
+++ b/thirdparty_containers/derby/playlist.xml
@@ -19,7 +19,7 @@
 		docker rmi -f adoptopenjdk-derby-test; \
 		$(TEST_STATUS)</command>
 		<levels>
-			<level>sanity</level>
+			<level>extended</level>
 		</levels>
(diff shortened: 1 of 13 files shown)


smlambert:
@sophia-guo thanks!  will you switch the other external tests over to extended in a separate PR (in this swap that we are doing)?

sophia-guo:
@smlambert did you mean all others should be switched to extended?  Just to confirm.

smlambert:
Yes, I was thinking to do that (at least shift some the others to extended).  Mostly this is an exercise in measuring how much available machine time we have nightly.  

My suggestion would be:

functional -> special
system -> special

the 3 mp-tck targets -> sanity

scala -> extended
elasticsearch -> extended
derby -> extended
lucene-solr -> extended
jenkins -> extended
wildfly -> extended
tomcat -> extended

What do you think?

sophia-guo:
@smlambert updated. The jobs under test_external will also be updated.

PR #1006, opened 2019-03-20
Enable mauve load test on jdk11 and 12

Mesbah-Alam:
Enable mauve load test on jdk11 and 12
Signed-off-by: Mesbah Mesbah_Alam@ca.ibm.com

### DIFF
--- a/systemtest/mauveLoadTest/playlist.xml
+++ b/systemtest/mauveLoadTest/playlist.xml
@@ -2,9 +2,9 @@
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
 	<include>../systemtest.mk</include>
 
-	<!-- exclude on openj9 on osx https://github.com/eclipse/openj9/issues/4091 -->
+	<!-- exclude on JDK8 openj9 on osx https://github.com/eclipse/openj9/issues/4091 -->
 	<test>
-		<testCaseName>MauveSingleThreadLoadTest</testCaseName>
+		<testCaseName>MauveSingleThreadLoadTest_OpenJ9_JDK8</testCaseName>
 		<variations>
 			<variation>NoOptions</variation>
 		</variations>
@@ -30,7 +30,32 @@
 		</impls>
 		<platformRequirements>^os.osx</platformRequirements>
 	</test>
-	
+	<test>
+		<testCaseName>MauveSingleThreadLoadTest_OpenJ9</testCaseName>
+		<variations>
+			<variation>NoOptions</variation>
+		</variations>
+		<command>export JAVA_HOME=$(JDK_HOME)$(D)  $(AND_IF_SUCCESS) \
+	perl $(SYSTEMTEST_RESROOT)$(D)stf$(D)stf.core$(D)scripts$(D)stf.pl  \
+	-test-root=$(Q)$(SYSTEMTEST_RESROOT)$(D)stf;$(SYSTEMTEST_RESROOT)$(D)openjdk-systemtest$(Q) \
+	-systemtest-prereqs=$(Q)$(SYSTEMTEST_RESROOT)$(D)systemtest_prereqs$(Q)  \
+	-java-args=$(Q)$(JVM_OPTIONS)$(Q) \
+	-results-root=$(REPORTDIR)  \
+	-test=MauveSingleThreadLoadTest; \
+	$(TEST_STATUS)</command>
+		<levels>
+			<level>sanity</level>
+		</levels>
+		<groups>
+			<group>system</group>
+		</groups>
+		<subsets>
+			<subset>11+</subset>
+		</subsets>
+		<impls>
+			<impl>openj9</impl>
+		</impls>
+	</test>	
 	<test>
 		<testCaseName>MauveSingleThreadLoadTest_HS</testCaseName>
 		<variations>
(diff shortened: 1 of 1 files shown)


Mesbah-Alam:
Mauve load tests should be enabled on JDK11 and JDK12 and only be excluded on OpenJ9 JDK8 due to https://github.com/eclipse/openj9/issues/4091.

systemtest/mauveLoadTest/playlist.xml
```diff
@@ -107,16 +128,40 @@
 			<group>system</group>
 		</groups>
 		<subsets>
-			<subset>8</subset>
+			<subset>11</subset>
+			<subset>12</subset>
```
renfeiw: 11+ will be better in this case

systemtest/mauveLoadTest/playlist.xml
```diff
@@ -163,8 +207,32 @@
...
```
renfeiw: 11+

systemtest/mauveLoadTest/playlist.xml
Mesbah-Alam: Ah.. missed that, updated!

systemtest/mauveLoadTest/playlist.xml
Mesbah-Alam: Updated.

Mesbah-Alam:
Hi Shelley, I've noticed that None of the 3 Mauve load tests are actually running in JDK11 and JDK12 builds. They are only running on JDK8. I believe we should enable these tests on JDK11 and JDK12 and open appropriate new issues, add excludes if they fail. @ShelleyLambert

systemtest/mauveLoadTest/playlist.xml
```diff
@@ -30,9 +30,8 @@
 		</impls>
 		<platformRequirements>^os.osx</platformRequirements>
 	</test>
-	
 	<test>
-		<testCaseName>MauveSingleThreadLoadTest_HS</testCaseName>
+		<testCaseName>MauveSingleThreadLoadTest_OpenJ9</testCaseName>
```
smlambert: I think that MauveSingleThreadLoadTest_OpenJ9 and MauveSingleThreadLoadTest_HS can be collapsed into 1 target again, with no <impls> in the target, as for 11+ the tests run on both impls on all platforms.  The only reason to split them would be in the event we have to exclude on a platform for a particular impl (as is the case for osx jdk8 j9).  Whenever we get a chance, we should re-unite the targets in playlists (that get split temporarily to accommodate tmp excludes and such.

systemtest/mauveLoadTest/playlist.xml
```diff
@@ -86,9 +108,8 @@
 		</impls>
 		<platformRequirements>^os.osx</platformRequirements>
 	</test>
-
 	<test>
-		<testCaseName>MauveSingleInvocationLoadTest_HS</testCaseName>
+		<testCaseName>MauveSingleInvocationLoadTest_OpenJ9</testCaseName>
```
smlambert: Same note as above, can collapse into 1 testCaseName, MauveSingleInvocationLoadTest for 11+ (removing <impls>, so targets are generated for all impls).

systemtest/otherLoadTest/playlist.xml
```diff
@@ -21,6 +22,61 @@
 		<groups>
 			<group>system</group>
 		</groups>
+		<impls>
+			<impl>openj9</impl>
+		</impls>
+	</test>
+	<test>
+		<testCaseName>ClassLoadingTest_HotSpot_Non_zLinux</testCaseName>
```
smlambert: Somewhere there should be comments explaining why the need to temporarily separate into 2 targets, and a link to the defect/issue that caused this change.

systemtest/otherLoadTest/playlist.xml
```diff
@@ -21,6 +22,61 @@
...
+			<level>sanity</level>
+		</levels>
+		<groups>
+			<group>system</group>
+		</groups>
+		<impls>
+			<impl>hotspot</impl>
+		</impls>
+		<subsets>
+			<subset>11</subset>
```
smlambert: 11+

systemtest/mauveLoadTest/playlist.xml
Mesbah-Alam: > Whenever we get a chance, we should re-unite the targets in playlists (that get split temporarily to accommodate tmp excludes and such.

Agreed. 

`MauveSingleThreadLoadTest_OpenJ9` is jdk11+ only, whereas `MauveSingleThreadLoadTest_HS` is not.

I will collapse `MauveSingleThreadLoadTest_HS` and `MauveSingleThreadLoadTest_OpenJ9` (and also the other two mauve test pairs) - once we have it working on osx jdk9 on openj9.

systemtest/otherLoadTest/playlist.xml
Mesbah-Alam: I think it's a bit hard to see this in the change view. 

We have 3 Mauve load tests - single thread, single invocation and multi-thread. Since we need the exclude for each one of them, the split has taken places in all 3 of them. So we ended up with three blocks of mauve tests, where in each block we have 3 `<test>` nodes. 

There is a comment with the link to the issue at the beginning of each such block.

smlambert: Yes, I see now when I view the full file, thanks...  

this PR is tricky to review when just looking at the default review mode.  :)

Mesbah-Alam: Yes... I understand .. since there are 3 pairs of tests whose names are very similar ... sorry about that..

smlambert: no worries, I know its temporary :)

PR #1014, opened 2019-03-21
Add artifactory info into the test job template

llxia:
- add ARTIFACTORY_REPO and ARTIFACTORY_ROOT_DIR
- rename JENKINS_DOMAIN to ARTIFACTORY_ROOT_DIR in Jenkins base to avoid
confusion. This value defaults to jenkins domain, but it can be set to
other value.

Issue: https://github.com/eclipse/openj9/issues/5198


llxia:
Tested in the internal Jenkins.

PR #1020, opened 2019-03-22
Support elastic test with different jdk version

sophia-guo:
Elastic requires different gradle version and gradle version requires
different JDK version


sophia-guo:
Grinder:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1346/consoleFull
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1345/consoleFull

Fix the issue mentioned in https://github.com/AdoptOpenJDK/openjdk-tests/issues/974#issue-420285218 'jdk11 Elasticsearch tests fail with  Could not determine java version from '11.0.2'.'

PR #1022, opened 2019-03-22
Add Artifactory link to build description

AdamBrousseau:

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -14,7 +14,7 @@ def makeTest(testParam) {
 	String tearDownCmd = "if [ `uname` = AIX ]; then MAKE=gmake; else MAKE=make; fi; \$MAKE -f ./openjdk-tests/TestConfig/testEnv.mk testEnvTeardown"
 	try {
 		sh "$tearDownCmd"
-		sh "./openjdk-tests/maketest.sh $testParam" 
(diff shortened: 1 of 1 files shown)


AdamBrousseau:
Test Build
https://ci.eclipse.org/openj9/view/Test/job/Test-sanity.system-JDK11-linux_x86-64_cmprssptrs/220/

llxia:
Could you provide a Grinder link? Thanks

AdamBrousseau:
https://ci.eclipse.org/openj9/view/Test/job/Test-sanity.functional-JDK11-linux_x86-64_cmprssptrs/641/

llxia:
What is the difference between this `Artifacts` link and the default artifactory link?
Screen Shot 2019-03-25 at 9 12 33 PM

AdamBrousseau:
The default (green line and circle) is a link to the webui for the Build artifact. On the OSS version, you can't easily navigate from there to the artifacts. The new one links to the file browser level directly to the folder where the artifacts are stored. In this case nothing was uploaded because the build was green so you get a 404. We should probably wrap the upload in a condition and check if there is anything to upload first. 

Related eclipse/openj9#5212. Note that in the OpenJ9 change we have the full URLs linked for a direct download. Where as here we just link to the folder because we don't know what artifacts, if any, exist.

llxia:
ah, I see. This is for Artifactory OSS version.

PR #1026, opened 2019-03-23
Exclude MauveMultithreadLoadTest from failing platforms and JDK versions

Mesbah-Alam:
Exclude MauveMultithreadLoadTest from failing platforms and JDK versions

### DIFF
--- a/systemtest/mauveLoadTest/playlist.xml
+++ b/systemtest/mauveLoadTest/playlist.xml
@@ -192,10 +192,11 @@
 		<impls>
 			<impl>openj9</impl>
 		</impls>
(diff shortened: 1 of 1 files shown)


pshipton:
It seems overkill to exclude Mauve. I think we should be excluding the problematic sub-tests like `gnu.testlet.java.text.MessageFormat.format14`

Mesbah-Alam:
I'll investigate how to exclude subtests.

Mesbah-Alam:
@pshipton - I suggest we exclude mauve for now, and I'll re-include it with sub-test exclude in the next PR. It will take a little time for me to figure out how to exclude sub-tests.

PR #1036, opened 2019-03-25
Exclude failing tests

adamfarley:
This change covers the exclusion of multiple failing openjdk
regression tests, along with the link to the issues that document
the failures.


adamfarley:
I can do that, sure. I was under the impression that the openj9 tests use the default problem list as well as the openj9 specific one. Can you tell me why tests excluded from both test sets are duplicated in the problem files?

smlambert:
Hotspot uses ProblemList_openjdk11.txt
OpenJ9 uses ProblemList_openjdk11-openj9.txt 

There is no base file.  

We could just as well rename ProblemList_openjdk11.txt -> ProblemList_openjdk11-hotspot.txt, but we were matching the api v1 path options naming convention at the time (https://github.com/AdoptOpenJDK/openjdk-api/blob/master/README.v1.md#path-options).

adamfarley:
Ah, I see. Saw two exclude files in the test output and didn't read closely enough.

Thanks for explaining Shelley. :)

PR #1045, opened 2019-03-29
Enable sun/rmi/rmic/manifestClassPath/run.sh

JasonFengJ9:
Enable `sun/rmi/rmic/manifestClassPath/run.sh` for `Java 8 mocosx-all`

The failure can't be reproduced in grinders, re-enable the test for `Java8 macosx-all`.

Related: https://github.com/eclipse/openj9/issues/5061

Reviewer: @ShelleyLambert 


JasonFengJ9:
@ShelleyLambert good catch! maybe it didn't and that's why we were able to grinder on it.

Issue #1042, opened 2019-03-28
Issues with Elasticssearch test builds


sophia-guo:
By runing elasticsearch test builds we found there are some issues related with automatic running

- [x] 1. https://github.com/AdoptOpenJDK/openjdk-tests/pull/1087

 If there are errors tests will stop. [Java Testing Framework](https://www.elastic.co/guide/en/elasticsearch/reference/current/testing-framework.html)  only supports running all tests without stopping on failures by option [-Dtests.haltonfailure=false](https://github.com/elastic/elasticsearch/blob/master/TESTING.asciidoc) . The underlying framework is Junit so that may be the main reason. 

```
21:51:49     [junit4] JVM J3:     1.83 ..    12.11 =    10.28s
21:51:49     [junit4] JVM J4:     1.82 ..    23.67 =    21.85s
21:51:49     [junit4] JVM J5:     1.88 ..    26.94 =    25.06s
21:51:49     [junit4] JVM J6:     1.85 ..    24.17 =    22.32s
21:51:49     [junit4] JVM J7:     1.81 ..    19.91 =    18.10s
21:51:49     [junit4] Execution time total: 34 seconds
21:51:49     [junit4] Tests summary: 10 suites, 55 tests, 1 error
21:51:49  
21:51:49  
21:51:49  FAILURE: Build failed with an exception.
```

https://ci.adoptopenjdk.net/view/Test_external/job/openjdk11_hs_external_extended_elasticsearch_x86-64_linux/10/
Stop with:

```
21:51:49  BUILD FAILED in 28m 5s
21:51:49  223 actionable tasks: 114 executed, 109 up-to-date
```

https://ci.adoptopenjdk.net/view/Test_external/job/openjdk11_hs_external_extended_elasticsearch_x86-64_linux/9/
Successful with:

```
22:33:58  BUILD SUCCESSFUL in 1h 28m 3s
22:33:58  589 actionable tasks: 343 executed, 246 up-to-date
```

- [x] 2. https://github.com/AdoptOpenJDK/openjdk-tests/pull/1055

 We are currently running test from the root directory, which is telling gradle to run all test tasks. If we'd like to run a single test task we can either cd into the subproject where the test exists, or use the explicit gradle task like :server:test and with option of 'Test case filtering'. However it doesn't have an option to skip a test task|test class| test method unless tests code is with certain annotations.  

- [x] 3. https://github.com/AdoptOpenJDK/openjdk-tests/issues/1057

[...]

PR #1057, opened 2019-04-04
Update external elasticsearch tests junit cp story

sophia-guo:
Find existing junit reports and cp instead of hard-coded as paths could
always be changing.


sophia-guo:
Related with issue 3 of #1042 

Grinder with hotspot :https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/216/
Grinder with openj9:
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/215/

PR #1059, opened 2019-04-05
Reenable HCR testing on JDK11

adamfarley:
We've updated the ASM we're using for this test, so now
it will run and pass on jdk11.

Removing the lambda playlist entry because it was an
accidental duplicate.

### DIFF
--- a/systemtest/lambdaLoadTest/playlist.xml
+++ b/systemtest/lambdaLoadTest/playlist.xml
@@ -74,32 +74,6 @@
 		<platformRequirements>os.linux,vm.cmprssptrs</platformRequirements> 
 	</test>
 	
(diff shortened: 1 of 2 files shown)


adamfarley:
Testing:
- Linux JDK11 on OpenJ9: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder-systemtest/103/
- Linux JDK11 on Hotspot: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder-systemtest/104
- Linux JDK8 on OpenJ9: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder-systemtest/98
    (Note: No windows testing due to issue #236 )

PR #1075, opened 2019-04-16
Run parallel tests out of node block

### LINKED ISSUE #994: Testing in parallel holds a Jenkins node executor unnecessarily
Possibly related #747 

I noticed some test jobs running that were orchestrating downstream parallel jobs. The issue is that the orchestrator portion is holding onto a node unnecessarily. I took a quick glace at the code but it does not seem like a 2 minute fix.

https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L154
https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L406

AdamBrousseau:
- When running tests in parallel it is not necessary for the
  top level build to hold onto a node while the downstream
  jobs execute. In order to exit out of the node block we need
  to return back to the top level pipeline file and then call
  a new parallel function. Some definitions need to be global
  in order for variables to propagate around.

Fixes #994

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -109,7 +109,7 @@ def setupParallelEnv() {
 				testSubDirSize = testSubDirs.size()
 			}
 			echo "testSubDirSize is ${testSubDirSize}, testSubDirs is ${testSubDirs}, running test in parallel mode"
-			def parallel_tests = [:]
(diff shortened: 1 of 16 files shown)


AdamBrousseau:
@llxia for review

llxia:
will merge after the release

Issue #1076, opened 2019-04-16
openjdk tests builds on aix got intermittent error 'Error: Unable to access jarfile jtreg.jar'


sophia-guo:
openjdk tests builds on aix got intermittent error 'Error: Unable to access jarfile /ramdisk0/build/workspace/openjdk8_hs_openjdktest_ppc64_aix/openjdk-tests/TestConfig/scripts/testKitGen/../../../../jvmtest/openjdk_regression/jtreg/lib/jtreg.jar'

It happened to both hotspot and openj9. And it's not machine-related as same build on same machines may not get the same error.

https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk8_hs_openjdktest_ppc64_aix/47/console

Suspect is the jtreg jar is not downloaded correctly. However if it's the case ant would give some error information.

sophia-guo:
Happened again,
https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk8_hs_openjdktest_ppc64_aix/70/console

llxia:
We don't have the job history anymore, but the above error seems to be the same as the one that happened internally. The real error in the internal build was the `getJtreg ` target failed (wget is not installed properly) so that jtreg did not get downloaded.
 
PR https://github.com/AdoptOpenJDK/openjdk-tests/pull/1512 changes wget to curl.

llxia:
Close this issue for now as we changed to use curl. Please reopen if the issue still exists.

PR #1077, opened 2019-04-16
Update JRE_IMAGE location

sophia-guo:
Force build fail if jtreg jar is not installed successfully


sophia-guo:
Relates #1076

PR #1078, opened 2019-04-16
Add Wycheproof tests to the external test pool

Mesbah-Alam:
Resolves https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/250

### DIFF
--- a/thirdparty_containers/wycheproof/build.xml
+++ b/thirdparty_containers/wycheproof/build.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0"?>
+<project name="wycheproof" default="build" basedir=".">
+	<taskdef resource="net/sf/antcontrib/antlib.xml" />
+	<description>
+		Build wycheproof Docker image
+	</description>
+
+	<!-- set properties for this build -->
+	<property name="DEST" value="${BUILD_ROOT}/thirdparty_containers/wycheproof" />
+	<property name="src" location="." />
+
+	<target name="init">
+		<mkdir dir="${DEST}"/>
+	</target>
+
+	<target name="clean_image" depends="init" description="clean wycheproof tck docker image if there is one">
+		<exec executable="docker">
+			<arg line="rmi -f adoptopenjdk-wycheproof" />
+		</exec>
+	</target>
+
+	<target name="build_image" depends="clean_image" description="build wycheproof tck docker image">
+		<exec executable="docker"  failonerror="true">
+			<arg line="build -t adoptopenjdk-wycheproof -f dockerfile/Dockerfile --pull . --build-arg SDK=${JVM_VERSION}" />
+		</exec>
+	</target>
+
+	<target name="dist" depends="build_image" description="generate the distribution">
+		<copy todir="${DEST}">
+			<fileset dir="${src}" includes="*.xml, *.mk"/>
+		</copy>
+	</target>
+
+	<target name="build">
+		<antcall target="dist" inheritall="true" />
+	</target>
+</project>
--- a/thirdparty_containers/wycheproof/dockerfile/Dockerfile
+++ b/thirdparty_containers/wycheproof/dockerfile/Dockerfile
@@ -1,0 +1,66 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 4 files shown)


Mesbah-Alam:
Tested here: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1510

thirdparty_containers/wycheproof/dockerfile/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+#Install Bazel v0.14.0
+RUN wget -q https://github.com/bazelbuild/bazel/releases/download/0.14.0/bazel-0.14.0-installer-linux-x86_64.sh
+RUN chmod 755 $WYCHEPROOF_HOME/bazel-0.14.0-installer-linux-x86_64.sh
+RUN $WYCHEPROOF_HOME/bazel-0.14.0-installer-linux-x86_64.sh
+
+# Clone Wycheproof 
+RUN pwd
+RUN git clone https://github.com/google/wycheproof
+WORKDIR $WYCHEPROOF_HOME/wycheproof
+RUN git checkout 189df209602bda4262034a1e19e84b1fe4330fd8
```
sophia-guo: May add the comment about this tag. If this could be different for later jdk version it might be better to set as a variable.

sophia-guo:
There are some failures is it easy to exclude the failure tests and open the issue to track it?

thirdparty_containers/wycheproof/dockerfile/Dockerfile
Mesbah-Alam: On a second thought, I don't think we need to use it. I'll remove it and start another test checking out wycheproof from master.

Mesbah-Alam:
> There are some failures is it easy to exclude the failure tests and open the issue to track it?

I haven't been able to figure out how to exclude tests in Bazel yet. They do talk about excluding "paths" in the [documentation](https://docs.bazel.build/versions/master/user-manual.html) - but it's still not clear. Need to figure this out later.

thirdparty_containers/wycheproof/build.xml
```diff
@@ -0,0 +1,38 @@
...
+<project name="wycheproof" default="build" basedir=".">
+	<taskdef resource="net/sf/antcontrib/antlib.xml" />
+	<description>
+		Build wycheproof Docker image
+	</description>
+
+	<!-- set properties for this build -->
+	<property name="DEST" value="${BUILD_ROOT}/thirdparty_containers/wycheproof" />
+	<property name="src" location="." />
+	<property name="jvm_version" location="${JVM_VERSION}" />
```
llxia: Do we need property  `jvm_version`? I do not see it used anywhere.

Mesbah-Alam: Thanks for the catch! I've removed `jvm_version`. 

This needs to be removed from the rest of the external tests as well -- as it's not used in any one of them, even though we've been copy-pasting it over in all.

thirdparty_containers/wycheproof/dockerfile/Dockerfile
Mesbah-Alam: Confirm: We don't need the tag: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1539/

Note: after checking out Wycheproof from head, we have more tests running now, and also more failing. There seems to be a "timeout" even happening. Need to figure out how to set timeout. 

Documentation : https://github.com/google/wycheproof, suggests that we may want to exclude the "long running" tests by executing OpenJDKTest and BouncyCastle targets (instead of running *all*) . I've started a new build with these 2 targets : https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1547/console. However, we may want to run all the tests.

Mesbah-Alam: Here's the result of the run when we execute OpenJDKTest and BouncyCastle targets (instead of OpenJDKAllTests and BouncyCastleAllTests target: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1547/#showFailuresLink

Mesbah-Alam:
@sophia-guo : I suggest we deliver this PR and get Wycheproof tests running in its job and add the step to exclude the failed tests as a second step.

PR #1086, opened 2019-04-22
Fix openjdk11-openj9 elastic test setup failure

sophia-guo:
Build failed msg 'Project 'core' not found in root project at 
'elasticsearch' at the beginning of the tests, which is tests setup issue.

### DIFF
--- a/thirdparty_containers/elasticsearch/playlist.xml
+++ b/thirdparty_containers/elasticsearch/playlist.xml
@@ -14,7 +14,7 @@
 -->
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
 	<test>
-		<testCaseName>elasticsearch_test_openj9</testCaseName>
+		<testCaseName>elasticsearch_test_openj9_jdk8</testCaseName>
 		<command>docker run --name elasticsearch-test $(EXTRA_DOCKER_ARGS) adoptopenjdk-elasticsearch-test:latest \
 		 "--exclude-task :core:test \
 		 --exclude-task :client:rest:test \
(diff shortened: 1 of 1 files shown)


sophia-guo:
https://ci.adoptopenjdk.net/view/Test_external/job/openjdk11_j9_external_extended_elasticsearch_x86-64_linux/29/console

```
15:48:31  FAILURE: Build failed with an exception.
15:48:31  
15:48:31  * What went wrong:
15:48:31  Project 'core' not found in root project 'elasticsearch'.
15:48:31  
15:48:31  * Try:
15:48:31  Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
15:48:31  
15:48:31  * Get more help at https://help.gradle.org
15:48:31  
15:48:31  BUILD FAILED in 57s
```

Grinder : https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/293/console

Tests were running and got same test failures as hotspot

```
00:11:36  FAILURE: Build failed with an exception.
00:11:36  
00:11:36  * What went wrong:
00:11:36  Execution failed for task ':plugins:ingest-attachment:unitTest'.
00:11:36  > There were test failures: 4 suites, 16 tests, 1 suite-level error, 1 error [seed: 2471400E6036ED36]
```
Will invest in a different issue.

smlambert:
Hi @sophia-guo, what issue does this address (use Fixes or Closes)?  I want to make sure we can relate this change back to details, as future, I imagine we will want to re-include those tests.

sophia-guo:
Hi @ShelleyLambert, this is a test setup issue. Different elastic version has different folder structures. There is actually no folder|project 'core' in newer versions. That is when doing exclude '--exclude-task :core:test' (failures with openj9 jdk8) job will fail. The fix is to set job command per version by separating testcases in playlist.xml. If we do have many different versions the options could be setting those in a mk file.

sophia-guo:
The new failure with this PR in is indeed test failure itself. Will open an issue to record and then exclude them. Will also update|open a issue to record the failures and excludes with openj9 jdk8.

PR #1092, opened 2019-04-23
Temporarily exclude jdk_jdi tests on openj9

smlambert:
Exclude during investigation on why jdi processes are not being killed at completion of tests.


smlambert:
Related to https://github.com/AdoptOpenJDK/openjdk-tests/issues/1071

Issue #834, opened 2019-01-25
Run external testing in parallel (1 Jenkins job per application)


smlambert:
Instead of running all the application tests in a single Jenkins job, have a parent job fire off child jobs per app.  This is the same approach we use for system tests.  

Benefits of this approach are:
- easier to triage (smaller console output and junit output doesn't get garbled with others)
- won't hit the timeout as we add new testing (with recent additions a single job can take up to 10hrs, which is our default timeout on jobs)

sophia-guo:
Closed by #991

PR #1096, opened 2019-04-25
Enable wildfly & kafka

### LINKED ISSUE #855: Exclude failing sanity.external test targets
Issue to clean sanity.external test targets that are failing and require triage.  Until #834 is completed, triage is difficult.

- [ ] kafka & wildfly tests fail on both implementations and all versions, exclude on all

sophia-guo:
As #834 is done, we should enable those tests

Also run kafka test against default scala version instead of all scala
versions

Close #855 

### DIFF
--- a/thirdparty_containers/kafka/dockerfile/Dockerfile
+++ b/thirdparty_containers/kafka/dockerfile/Dockerfile
@@ -35,10 +35,6 @@ FROM $IMAGE_NAME:$IMAGE_VERSION
 RUN apt-get update \
 	&& apt-get -y install \
 	ant \
-	apt-transport-https \
-	ca-certificates \
-	dirmngr \
-	curl \
 	git \
 	make \
 	unzip \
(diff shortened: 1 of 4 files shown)


sophia-guo:
Grinder https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/338/
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/316/

The only concern about enable as a parallel job is the availability of the machines.

sophia-guo:
Both kafka test build against 8 and 11 seems hanging. Cancelled.
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/341/console
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/338/console

Current test setup as running with all scala versions. Will limit to default scala versions as we do have separate test build running scala.

PR #1099, opened 2019-04-26
Add sleep & retry logic around the curl on the SDK

### LINKED ISSUE #1033: Add sleep & retry logic around the curl on the SDK
In the even we have a network blip while trying to curl an SDK, we should introduce a sleep and retry in an attempt to recover before failing outright. I can think of 2 options of where to add the fix.
1. In JenkinsfileBase around this if/else block
https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L249
1. In the get.sh script around the curl
https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/get.sh#L159
The bash may be more difficult but it would wrap only the curl command we are addressing with this issue (which has pros and cons itself). It would also work for users running the get.sh directly (not on Jenkins).

Thoughts @llxia ?

sophia-guo:
Set sleep 30 second and retry 5 times
Turn off 'set -e' for curl command to get useful error message as the
command is more likely to fail due to all different reasons.

Fix #1033 

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -155,14 +155,16 @@ getBinaryOpenjdk()
 	if [ "${download_url}" != "" ]; then
 		for file in $download_url
 		do
-			echo "curl -OLJks ${curl_options} $file"
-			curl -OLJks ${curl_options} $file
+			set +e
+			echo "curl -OLJks --retry 5 --retry-delay 30 ${curl_options} $file"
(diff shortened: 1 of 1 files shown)


sophia-guo:
Tried with server is down or during the download process the network is down, i.e. curl return code is 6, 56.

llxia:
fyi @pshipton

Issue #1103, opened 2019-04-29
All external test builds failed with 'No space left on device' on test-scaleway-ubuntu1604-x64-1


sophia-guo:
All container test builds failed on test-scaleway-ubuntu1604-x64-1 with message 

```
20:27:05  ERROR: [ssh-agent] Could not register BouncyCastle on the remote agent. 20:27:05  java.nio.file.FileSystemException: /tmp/resource-1012398343128323010: No space left on device 20:27:05  	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:91)  
```
https://ci.adoptopenjdk.net/view/Test_external/
https://ci.adoptopenjdk.net/view/Test_external/ https://ci.adoptopenjdk.net/view/Test_external/job/openjdk11_hs_external_extended_elasticsearch_x86-64_linux/28/console

 We had exactly same issue before on different machine  test-softlayer-ubuntu1604-x64-1 https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/761

If dangling docker volumes are the real reason of this issue we may need to consider the story of docker volume clean up.

sophia-guo:
'If your container generates non-persistent state data, consider using a tmpfs mount to avoid storing the data anywhere permanently, and to increase the container’s performance by avoiding writing into the container’s writable layer.'  https://docs.docker.com/storage/volumes/

sophia-guo:
I will log in the docker machines and do a cleanup manually cleanup and then the issue will be closed by #1104 , which will avoid to create any new dangling docker volumes from test builds.

As mentioned in https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/777#issuecomment-487678617 those dangling docker volumes should not be the reason of 'no space left'. The further investigation will be continued in https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/761

PR #1104, opened 2019-04-29
Remove the mount point /java

sophia-guo:
The VOLUME instruction creates a mount point with the specified name and
marks it as holding externally mounted volumes from native host or other
containers. It is used to get performance without externally mounting
any volumes.The current VOLUME ["/java"] doesn't do anyting but occupy
the Docker area and create as dangling volume, which need docker
management to deal with.

### DIFF
--- a/thirdparty_containers/derby/dockerfile/Dockerfile
+++ b/thirdparty_containers/derby/dockerfile/Dockerfile
@@ -43,10 +43,7 @@ RUN apt-get update \
 	wget \
 	ant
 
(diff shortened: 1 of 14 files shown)


sophia-guo:
This change set avoid to create any dangling docker volumes. Close #1103

sophia-guo:
Grinder: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/349/

PR #1114, opened 2019-05-03
Support displaying error code if downloading sdk fails

LongyuZhang:
- If downloading sdk fails, the error code will be displayed.

Signed-off-by: Longyu <longyu.zhang@ibm.com>

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -158,7 +158,9 @@ getBinaryOpenjdk()
 			set +e
 			echo "curl -OLJks --retry 5 --retry-delay 300 ${curl_options} $file"
 			curl -OLJks --retry 5 --retry-delay 300 ${curl_options} $file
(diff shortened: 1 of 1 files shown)


LongyuZhang:
reviewers: @llxia @renfeiw

LongyuZhang:
Tested internally
```
12:34:08  curl -OLJks --retry 5 --retry-delay 300  testtest
12:59:09  Error code: 6
12:59:09  Failed to retrieve testtest, exiting. This is what we received of the file and MD5 sum:
12:59:09  ls: cannot access testtest: No such file or directory
12:59:09  md5sum: testtest: No such file or directory
```

llxia:
FYI @pshipton , we will print out the error code, so we can get the actual error from downloading SDK.

PR #1115, opened 2019-05-03
Set failIfNoResults to true at all time

renfeiw:
- remove parallel condition


llxia:
On the second thought, we may still need this check for parent parallel builds. Close the PR.

Issue #172, opened 2017-11-15
Add more container-based 3rd party application tests


smlambert:
As an example of this type of testing, we have added scala tests: https://github.com/AdoptOpenJDK/openjdk-tests/tree/master/thirdparty_containers/scala

We have derby and solr_lucene tests running in the same "dockerized" way, so we should also add those test sub-folders.

Once we have those 3 examples of 3rd party application tests added, we should consider the list of other applications we would like to add and prioritize them (eclipse, cassandra, etc).  It will also be good to engage with these communities so that they are aware that we will be running their community test suites against the different flavours of JVM at AdoptOpenJDK (openjdk8, openjdk9, openjdk+openj9...).

suman-mitra:
Add Elasticsearch under thirdparty container tests
https://github.com/AdoptOpenJDK/openjdk-tests/issues/360

tellison:
In additional to a broad set of application tests to prove the quality of the binaries, it would be useful to have a suite of [MicroProfile](http://microprofile.io/) tests to demonstrate that our binaries are a good foundation for MicroProfile.

Microprofile doesn't have a reference implementation, rather there are a number of popular implementations that cover different proportions of the specification.

The following projects will provide a good test environment for our binaries: 
- [OpenLiberty](https://openliberty.io/),
- [WildFly Swarm](http://wildfly-swarm.io/), 
- [Payara](https://www.payara.fish/about), and 
- [Apache TomEE](http://tomee.apache.org/).

The tests would be taken from the MicroProfile TCKs, e.g. [fault tolerance APIs](https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/running_the_tck.asciidoc), [metrics APIs](https://github.com/eclipse/microprofile-metrics/blob/master/tck/running_the_tck.asciidoc), etc.  Note that not all the implementations implement every part of the MicroProfile spec, so we'd need to define the appropriate set of TCKs to run against each implementation.

As the MicroProfile projects are open communities, I expect we can get help in putting together the tests and running them on our test farm.

kwsutter:
[...]

PR #1118, opened 2019-05-06
Enable Apache Tomee MicroProfile TCKs

sophia-guo:
Limit to jdk8 for now as it mentioned that 

> Apache TomEE is built with Apache Maven >= 3.3.9 and Java 8 in https://github.com/apache/tomee

Related with #172 


sophia-guo:
Grider: 
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/367/

PR #1120, opened 2019-05-07
Add Renaissance Benchmark Suite

sophia-guo:
Enable akka-uct benchmarks with APACHE2 licenses


llxia:
Thanks @sophia-guo :)

PR #1131, opened 2019-05-14
Add support for customized sources URL option

### LINKED ISSUE #1124: Parameterize openjdk sources used for testing
This is somewhat related to https://github.com/AdoptOpenJDK/openjdk-tests/issues/1082

We'd like to automatically run the test pipeline on upstream OpenJDK hotspot Early Access (EA) builds so as to be able to automate consistent testing across platform specific builds and have test results available publicly.

While [get.sh](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/get.sh) has parameters for the JDK under test via `--sdk_resource customized` and `--customizedURL <url>`, the same is not possible for matching JDK sources. It is assumed the sources are in "sync" at AdoptOpenJDK/openjdk-jdkXu, where `X` is the JDK version like 8, 11, 12, etc. This assumption doesn't hold for EA upstream builds which have been built from sources *ahead* of the AdoptOpenJDK git mirrors. This is mostly an issue of running OpenJDK regression tests on customized builds.

In order to make this more flexible, I'd propose to implement something similar to `--sdk_resource` and `--customizedURL <url>`, but for JDK sources. Once that's implemented, Jenkins files could get enhanced to expose those new parameters. Then it should be possible to run the test pipeline on upstream builds automatically as soon as they become available at:
https://github.com/AdoptOpenJDK/openjdk8-upstream-binaries/releases and
https://github.com/AdoptOpenJDK/openjdk11-upstream-binaries/releases

jerboaa:
Closes #1124

Signed-off-by: Severin Gehwolf <sgehwolf@redhat.com>

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -20,6 +20,7 @@ PLATFORM=""
 JVMVERSION=""
 SDK_RESOURCE="nightly"
 CUSTOMIZED_SDK_URL=""
+CUSTOMIZED_SDK_SOURCE_URL=""
 OPENJ9_REPO="https://github.com/eclipse/openj9.git"
 OPENJ9_SHA=""
 OPENJ9_BRANCH=""
@@ -44,6 +45,7 @@ usage ()
 	echo '                [--sdkdir|-s binarySDKDIR] : if do not have a local sdk available, specify preferred directory'
 	echo '                [--sdk_resource|-r ] : indicate where to get sdk - releases, nightly , upstream or customized'
 	echo '                [--customizedURL|-c ] : indicate sdk url if sdk source is set as customized.  Multiple urls can be passed with space as separator'
+	echo '                [--customized_sourceURL|-S ] : indicate sdk source url if sdk source is set as customized.'
 	echo '                [--username ] : indicate username required if customized url requiring authorization is used'
 	echo '                [--password ] : indicate password required if customized url requiring authorization is used'
 	echo '                [--openj9_repo ] : optional. OpenJ9 git repo. Default value https://github.com/eclipse/openj9.git is used if not provided'
@@ -87,6 +89,9 @@ parseCommandLineArgs()
 
 			"--customizedURL" | "-c" )
 				CUSTOMIZED_SDK_URL="$1"; shift;;
+
+			"--customized_sourceURL" | "-S" )
+				CUSTOMIZED_SDK_SOURCE_URL="$1"; shift;;
 
 			"--username" )
 				USERNAME="$1"; shift;;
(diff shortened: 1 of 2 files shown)


jerboaa:
Works for me for local testing and verified via Grinder_Sandbox:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Sandbox/383/

Please review!

llxia:
Thanks @jerboaa 

Just a thought, `CUSTOMIZED_SDK_URL` does take multiple URLs (separated with space). Instead of creating a new parameter `CUSTOMIZED_SDK_SOURCE_URL `, it is possible to leverage the existing parameter `CUSTOMIZED_SDK_URL`?

get.sh
```diff
@@ -322,6 +349,9 @@ parseCommandLineArgs "$@"
 if [[ "$SDKDIR" != "" ]]; then
 	getBinaryOpenjdk
 fi
+if [[ "$TESTDIR" != "" ]]; then
+	getOpenJDKSources
+fi
```
sophia-guo: $TESTDIR should never be empty. I would suggest moving the condition `if [ "$SDK_RESOURCE" == "customized" ] && [ "$CUSTOMIZED_SDK_SOURCE_URL" != "" ]` out of function `getOpenJDKSources()` as

```
if [ "$SDK_RESOURCE" == "customized" ] && [ "$CUSTOMIZED_SDK_SOURCE_URL" != "" ]; then
         getOpenJDKSources
fi

jerboaa:
> Thanks @jerboaa
> 
> Just a thought, `CUSTOMIZED_SDK_URL` does take multiple URLs (separated with space). Instead of creating a new parameter `CUSTOMIZED_SDK_SOURCE_URL `, it is possible to leverage the existing parameter `CUSTOMIZED_SDK_URL`?

Perhaps. It feels wrong to "abuse" a parameter which is called "CUSTOMIZED_SDK_URL", though. That logic already does too much for my taste. To me a new parameter is cleaner.

get.sh
jerboaa: Thanks, fixed.

jerboaa:
I see a merge button. So not sure how you handle PRs being merged in terms of workflow. Is the PR submitter supposed to merge, once approved by a reviewer? Do you use a different workflow? Please let me know.

jerboaa:
Also: Thanks for the review, Sophia!

smlambert:
Let's proceed with the approach that the reviewer merges the PR.  We have been 'loose' with this as an approach to date.

This is barring the exceptional case that a change needs to be co-ordinated with another PR (potentially in another repo).  In the exceptional case, it should be stated clearly in the PR, so that it can be clearly communicated and co-ordinated.

@sophia-guo - if you are happy with @jerboaa 's changes, please merge them (and his other PR when ready).

jerboaa:
Thanks all!

PR #1132, opened 2019-05-14
Make JRE_IMAGE parameter workspace relative

jerboaa:
Prior to this patch it was already possible to pass `JRE_IMAGE` path via a
Jenkins parameter. However, string parameters won't get String-interpolated.
That means a param value of `${WORKSPACE}/foo/bar` results in a literal
`${WORKSPACE}` which isn't the user's intention.

This patch modifies the JRE_IMAGE environment variable to be workspace
relative if it comes from a Jenkins parameter. This solves a test issue we have
been seeing for upstream JDK builds, where math JRE tests would fail for no good
reason. After this patch one can set `JRE_IMAGE=openjdkbinary/j2sdk-image/jre`
and would then have the tests passing. Thoughts?

Tested with a sandbox grinder here:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Sandbox/387/

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -47,7 +47,7 @@ def setupEnv() {
 	}
 
 	env.TEST_JDK_HOME = "$WORKSPACE/openjdkbinary/j2sdk-image"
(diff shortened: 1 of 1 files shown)


jerboaa:
Thanks!

PR #1147, opened 2019-05-22
Shallow clone OpenJ9

fjeremic:
This change effectively reverts 285d19b9 to do a shallow clone of
OpenJ9. If a specific SHA is specified we will fetch all the SHAs
by a call to `git fetch --unshallow` which is put the repository in
a state identical to one that did not do a shallow clone.

This change is the best of both worlds as now for clones which do
not care about a particular SHA no longer incur the cost of cloning
the full repository history and clones which need a specific SHA
should always work.

Signed-off-by: Filip Jeremic <fjeremic@ca.ibm.com>

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -245,13 +245,14 @@ getTestKitGenAndFunctionalTestMaterial()
 		OPENJ9_BRANCH="-b $OPENJ9_BRANCH"
 	fi
 
-	echo "git clone $OPENJ9_BRANCH $OPENJ9_REPO"
-	git clone -q $OPENJ9_BRANCH $OPENJ9_REPO
+	echo "git clone --depth 1 $OPENJ9_BRANCH $OPENJ9_REPO"
+	git clone --depth 1 -q $OPENJ9_BRANCH $OPENJ9_REPO
 
 	if [ "$OPENJ9_SHA" != "" ]
 	then
 		echo "update to openj9 sha: $OPENJ9_SHA"
 		cd openj9
+		git fetch --unshallow
 		git fetch -q --tags $OPENJ9_REPO +refs/pull/*:refs/remotes/origin/pr/*
 		git checkout -q $OPENJ9_SHA
 		cd $TESTDIR


fjeremic:
@AdamBrousseau @llxia requesting review.

fjeremic:
> There is a small typo in your commit comment.

Fixed in c8850c0

llxia:
LGTM. Thanks @fjeremic.
Could we run a Grinder to verify?

fjeremic:
> Could we run a Grinder to verify?

Where can I find the external "Advanced Grinder" which lets me specify a SHA for OpenJ9?

get.sh
```diff
@@ -245,13 +245,14 @@ getTestKitGenAndFunctionalTestMaterial()
...
+	echo "git clone --depth 1 $OPENJ9_BRANCH $OPENJ9_REPO"
+	git clone --depth 1 -q $OPENJ9_BRANCH $OPENJ9_REPO
 
 	if [ "$OPENJ9_SHA" != "" ]
 	then
 		echo "update to openj9 sha: $OPENJ9_SHA"
 		cd openj9
+		git fetch --unshallow
 		git fetch -q --tags $OPENJ9_REPO +refs/pull/*:refs/remotes/origin/pr/*
 		git checkout -q $OPENJ9_SHA
```
AdamBrousseau: Is there a reason why we have `-q` on all the git commands?

fjeremic: Seems like a question for @llxia. The code was introduced here:
https://github.com/AdoptOpenJDK/openjdk-tests/pull/404/files#diff-bb552f35560c477dfc858910b86984f0R120

llxia:
I created `Grinder_Advanced` with `OPENJ9_SHA` parameter on AdoptOpenJDK Jenkins: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/build?delay=0sec

get.sh
llxia: I think the intention was to reduce the unnecessary console output.

fjeremic:
Thanks @llxia. Looks to be working fine against this change:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/2/console

PR #1150, opened 2019-05-25
use loop instead of curl retry option

llxia:
curl retry option does not work in the nightly builds. Use loop instead
of curl retry option


llxia:
Tested internally on windows, linux, aix and mac.

PR #1167, opened 2019-06-06
Turn off detailed error message for all level of external tests

sophia-guo:
External tests output is huge with all kinds of characters, which increase the occurrence of TAP parsing issue. 

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -90,7 +90,7 @@ def setupEnv() {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


sophia-guo:
We do have set 'outputTapToConsole: false'. However when exception happens it won't work, which make console output doubled.

https://ci.adoptopenjdk.net/view/Test_external/job/openjdk8_hs_external_sanity_payara-mp-tck_x86-64_linux/75/console

Grinder: https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/411/console

PR #1173, opened 2019-06-11
Added additional TIME_LIMIT parameter for Jenkins parallel running

LongyuZhang:
- added additional TIME_LIMIT parameter for pareallel running in JenkinsfileBase
[ci skip]


LongyuZhang:
reviewers @llxia

llxia:
Thanks @LongyuZhang

On second thought, we should avoid using `env` when unnecessary. In this case, we can just use `TIME_LIMIT`.

LongyuZhang:
> Thanks @LongyuZhang
> 
> On second thought, we should avoid using `env` when unnecessary. In this case, we can just use `TIME_LIMIT`.

The env has been removed as suggested. Thanks.

llxia:
Tested internally.

PR #1175, opened 2019-06-12
enable jck tests with Xvfb on all linux (including plinux and zlinux)

llxia:


llxia:
I have trouble to get a green run for `jdk_awt`. Enable jck test on all linux for now.

Issue #980, opened 2019-03-13
Re-name and clean up top-level directory structure


smlambert:
When we initially named some of our directories, we did not align them with our top-level test target names.  In order to make things clearer for those using this test resource, we will rename the following:

- [x] performance to perf

- [x] thirdparty_containers to external

- [x] systemtest to system

- [x] openjdk_regression to openjdk

These renaming changes will also require updates to scripts and documentation.

Additional clean-up, we have shifted triage doc to wiki pages, so we can remove the triage folder.

llxia:
ToDo after renaming systemtest to system:
- [ ] regenerate system test jobs in AdoptOpenJDK to set BUILD_LIST to `system`

- [ ] update BUILD_LIST description in Grinder on all sites

karianna:
Can this be closed now?

llxia:
yes, I think so.
wikis are updated: 
https://github.com/eclipse/openj9/wiki/Reproducing-Test-Failures-Locally
https://github.com/AdoptOpenJDK/openjdk-tests/wiki/How-to-Run-a-Grinder-Build-on-Jenkins

PR #1179, opened 2019-06-15
Rename to performance to perf to match top-level test target

smlambert:
- Related to #980 
- Remove odm 


smlambert:
Grinder (BUILD_LIST=perf, TARGET=akka-uct): https://ci.adoptopenjdk.net/view/all/job/Grinder_Sandbox/431/

buildenv/jenkins/JenkinsfileBase
```diff
@@ -406,9 +406,9 @@ def post(output_name) {
 					uploadToArtifactory()
 				}
 			}
-			//for performance test, achive regardless the build result
+			//for performance test, archive regardless the build result
 			//TODO: Need to soft-code the output dir path
-			def benchmark_test_output_dir = 'jvmtest/performance/odm/ilog_wodm881/leftoverResults';
+			def benchmark_test_output_dir = 'jvmtest/perf/odm/ilog_wodm881/leftoverResults';
 			if (fileExists(benchmark_test_output_dir)) {
```
sophia-guo: If odm is removed this archive for odm specific could also be removed?

sophia-guo:
I also see some liberty files are removed too. Just to confirm if that is expected? @smlambert

buildenv/jenkins/JenkinsfileBase
smlambert: Ah yes, ideally, we replace that with an archiving for all perf targets that get run.  Depending on how involved that is, I may defer to an upcoming PR (I will have another one prepped for later this week).

smlambert:
I did not intend to remove files so will correct and push changes shortly. I think the confusion is that those files are in a folder named bin, which is normally houses files that are built at execution time and not stored in repos (so in my gitignore list).

sophia-guo:
Perf test results may not be under TestConfig/test_output_*. The output location may be different we may need to define case by case later. That will be done later. Merge this first.

smlambert:
Thanks @sophia-guo - for the record, I ran a Grinder, and the json output from akka-uct benchmark is now archived as part of the tar.gz file (as I had hoped it would be), https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_SandboxExtra/2/artifact/

PR #1180, opened 2019-06-17
replaced/removed JAVA_BIN in JCK and OpenJDK folders of openjdk-tests

LongyuZhang:
- Replaced/removed JAVA_BIN in JCK and OpenJDK folders to avoid ambiguity in openjdk-test
[ci skip]

### DIFF
--- a/jck/build.xml
+++ b/jck/build.xml
@@ -195,18 +195,12 @@
 	</target>
 
 	<target name="configure_stf" depends="check_systemtest">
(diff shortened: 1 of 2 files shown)


LongyuZhang:
reviewers: @llxia @renfeiw

llxia:
I suspect `<property name="java.home"...` in build.xml is no longer needed. It was removed from systemtest ant. @Mesbah-Alam could you confirm this?

LongyuZhang:
> I suspect `<property name="java.home"...` in build.xml is no longer needed. It was removed from systemtest ant. @Mesbah-Alam could you confirm this?

java.home has been removed as suggested, tested sanity.jck. [https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder-systemtest/110/](url)

PR #1182, opened 2019-06-17
Fixes from perf PR 1179

smlambert:
Removing a submodule that was merged incorrectly as part of PR 1179


smlambert:
Launched https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_SandboxExtra/3/ to verify it fixes issues seen

PR #1185, opened 2019-06-18
Add more renaissance perf test targets

smlambert:
- Add 15 more renaissance benchmark targets

Related: https://github.com/AdoptOpenJDK/openjdk-tests/issues/1112


smlambert:
Initial test run: 
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_SandboxExtra/4/ 

Will run more variations before shifting this from draft.

PR #975, opened 2019-03-13
Add openjdk-jdk12u as default repo for jdk12 tests

smlambert:
Fixes #971 
Signed-off-by: smlambert <slambert@gmail.com>

PR #1186, opened 2019-06-18
Set auto map between JDK_VERSION and jdk source repo

sophia-guo:
As naming convention is stable we can set up this auto map instead of
adding a new condition property whenever a new JDK_VERSION is available.


sophia-guo:
Though ideally we should pull this information from upstream builds. See #975

sophia-guo:
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/452/

Issue #1168, opened 2019-06-06
Generate test jobs at ci.adoptopenjdk.net with new naming scheme


smlambert:
Test jobs have been auto-generated at other projects/Jenkins servers (see https://github.com/AdoptOpenJDK/openjdk-tests/issues/890).  We will do similar at AdoptOpenJDK (taking on a new naming convention, to make more consistent across projects and make tooling like TRSS easier to write.

We want to create more types of nightly and weekly test jobs (as part of the AQA v1.0 deliverables in time for the July release).

This will need to be co-ordinated with a change in the build pipelines, which currently rely on the test job names to trigger launching.

In this issue, we will need to do the following:

- generate the jobs 

- verify they run

- change build pipelines to trigger newly generated jobs

- delete old test jobs

smlambert:
Jobs generated:
- special.functional for openj9,hotspot jdk8
- sanity.perf, extended.perf for openj9,hotspot  jdk8,jdk11 x86-64_linux

llxia:
Regenerated the above jobs without `_Nightly` suffix.

llxia:
Once https://github.com/AdoptOpenJDK/openjdk-build/pull/1125 is merged, I will manually rename the top level external test job names to match with the new naming convention. For example, `openjdk11_hs_externaltest_x86-64_linux` will be renamed to `Test_openjdk11_hs_sanity.external_x86-64_linux`.

External test jobs cannot be auto-gened as they are the top-level test jobs and have special configuration.

llxia:
External test jobs are renamed and top-level jobs configure are updated.
https://ci.adoptopenjdk.net/view/Test_external/

smlambert:
Deleted the jobs that were generated with the unneeded suffixes.  I believe this can be closed now.

PR #1192, opened 2019-06-25
Add jenkins test pipeline for OpenJDK Project Builds

### LINKED ISSUE #1183: Add AdoptOpenJDK tests job for OpenJDK project builds
AdoptOpenJDK hosts upstream OpenJDK project builds for JDK 8 and JDK 11 at:
https://adoptopenjdk.net/upstream.html

These builds can be tested with the AdoptOpenJDK test suite by using the `SDK_RESOURCE=customized` parameter. Example param configs:

```
sanity.openjdk:
    SDK_RESOURCE=customized
    JDK_IMPL=hotspot
    BUILD_LIST=openjdk
    TARGET=sanity
    SDK_RESOURCE=customized
    CUSTOMIZED_SDK_URL="<jdk-url> <jre-url>"
    CUSTOMIZED_SDK_SOURCE_URL=<source-url>
    JenkinsFile=<jenkinsFile>
    AUTO_DETECT=true

sanity.system:
    SDK_RESOURCE=customized
    JDK_IMPL=hotspot
    BUILD_LIST=systemtest
    TARGET=sanity
    SDK_RESOURCE=customized
    CUSTOMIZED_SDK_URL="<jdk-url>"
    JenkinsFile=<jenkinsFile>
    AUTO_DETECT=true
```

Considering that OpenJDK project builds are being produced for JDK 8 for windows x64 and linux x64, JDK 11 for linux x64, linux aarch64, windows x64, and the test set should be `sanity.openjdk` and `sanity.system`, there would be **4 test jobs for each JDK 8 release** and **6 test jobs for each JDK 11 release**. All of them would run nicely in parallel given test system resources.

It would be great if there was a testing job running at the AdoptOpenJDK CI build farm which:

- Is a clone of [AdoptOpenJDK Grinder](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/) job
- Has an additional pre-build step which checks [upstream JDK 8 releases](https://github.com/AdoptOpenJDK/openjdk8-upstream-binaries/releases) and [upstream JDK 11 releases](https://github.com/AdoptOpenJDK/openjdk11-upstream-binaries/releases) and if there are any new releases, triggers said test jobs at AdoptOpenJDK CI passing relevant parameters (JDK URL, JRE URL, Source URL)
- Runs on a daily basis to check for new releases. The CI job would keep record of the `published_at` date of the latest released JDK 8/11 version so as to not trigger testing for releases which have already been tested.

The idea is to get automatic AdoptOpenJDK testing for releases built elsewhere in addition to other testing done prior the release of binaries.

jerboaa:
This pipeline performs Github API calls so as to determine whether or not new releases are available for OpenJDK Project Builds released at:

JDK 11: https://github.com/AdoptOpenJDK/openjdk11-upstream-binaries/
JDK 8:   https://github.com/AdoptOpenJDK/openjdk8-upstream-binaries

If there are new releases, `sanity.openjdk` and `sanity.system` test jobs are being scheduled for each {version, architecture, platform} triplet.

Example run on Grinder_Sandbox (blue ocean view):
https://ci.adoptopenjdk.net/blue/organizations/jenkins/Grinder_Sandbox/detail/Grinder_Sandbox/495/pipeline

The idea would be to clone `Grinder_Sandbox` or `Grinder` Jenkins job configs and set up a separate job which runs daily or some such. The pipeline has functionality built in so as to not trigger new test jobs when the `published_at` date of the releases hasn't changed. Example for this scenario:

https://ci.adoptopenjdk.net/job/Grinder_Sandbox/494/console

Thoughts?

Closes #1183

### DIFF
--- a/buildenv/jenkins/upstream_openjdk_tests
+++ b/buildenv/jenkins/upstream_openjdk_tests
@@ -248,18 +248,19 @@ node("ci.role.test") {
                         stage("prepare $r.OS-$r.ARCH") {
                             def jdk_url = r.JDK_URL
                             def source_url = r.SOURCE_URL
+                            def jenkins_file = r.getJenkinsFile()
                             def jdk_jre_url = "$r.JDK_URL $r.JRE_URL"
-                            parallelJobs["openjdk$version-$r.OS-$r.ARCH sanity.openjdk"] = {
+                            parallelJobs["Test_openjdk${version}_hs_sanity.openjdk_${r.ARCH}_${r.OS}"] = {
                                 // Run a _sanity.openjdk test job with appropriate parameters
                                 build job: TEST_JOB_NAME, parameters: [
                                     string(name: 'ADOPTOPENJDK_REPO', value: params.ADOPTOPENJDK_REPO ? params.ADOPTOPENJDK_REPO : "https://github.com/AdoptOpenJDK/openjdk-tests.git"),
                                     string(name: 'ADOPTOPENJDK_BRANCH', value: params.ADOPTOPENJDK_BRANCH ? params.ADOPTOPENJDK_BRANCH : "master"),
                                     string(name: 'OPENJ9_REPO', value: params.OPENJ9_REPO ? params.OPENJ9_REPO : ""),
                                     string(name: 'OPENJ9_BRANCH', value: params.OPENJ9_BRANCH ? params.OPENJ9_BRANCH : ""),
-                                    string(name: 'JenkinsFile', value: r.getJenkinsFile()),
+                                    string(name: 'JenkinsFile', value: jenkins_file),
                                     string(name: 'JDK_IMPL', value: "hotspot"),
                                     string(name: 'BUILD_LIST', value: "openjdk"),
-                                    string(name: 'TARGET', value: "sanity"),
+                                    string(name: 'TARGET', value: "sanity.openjdk"),
                                     string(name: 'SDK_RESOURCE', value: "customized"),
                                     string(name: 'CUSTOMIZED_SDK_URL', value: jdk_jre_url),
                                     string(name: 'CUSTOMIZED_SDK_SOURCE_URL', value: source_url),
@@ -267,17 +268,17 @@ node("ci.role.test") {
                                     booleanParam(name: 'AUTO_DETECT', value: true),
                                 ], propagate: true
                             }
-                            parallelJobs["openjdk$version-$r.OS-$r.ARCH sanity.system"] = {
+                            parallelJobs["Test_openjdk${version}_hs_sanity.system_${r.ARCH}_${r.OS}"] = {
                                 // Run a systemtest test job with appropriate parameters
                                 build job: TEST_JOB_NAME, parameters: [
                                     string(name: 'ADOPTOPENJDK_REPO', value: params.ADOPTOPENJDK_REPO ? params.ADOPTOPENJDK_REPO : "https://github.com/AdoptOpenJDK/openjdk-tests.git"),
                                     string(name: 'ADOPTOPENJDK_BRANCH', value: params.ADOPTOPENJDK_BRANCH ? params.ADOPTOPENJDK_BRANCH : "master"),
                                     string(name: 'OPENJ9_REPO', value: params.OPENJ9_REPO ? params.OPENJ9_REPO : ""),
                                     string(name: 'OPENJ9_BRANCH', value: params.OPENJ9_BRANCH ? params.OPENJ9_BRANCH : ""),
-                                    string(name: 'JenkinsFile', value: r.getJenkinsFile()),
+                                    string(name: 'JenkinsFile', value: jenkins_file),
                                     string(name: 'JDK_IMPL', value: "hotspot"),
                                     string(name: 'BUILD_LIST', value: "systemtest"),
-                                    string(name: 'TARGET', value: "sanity"),
+                                    string(name: 'TARGET', value: "sanity.system"),
                                     string(name: 'SDK_RESOURCE', value: "customized"),
                                     string(name: 'CUSTOMIZED_SDK_URL', value: jdk_url),
                                     booleanParam(name: 'IS_PARALLEL', value: false),
--- a/buildenv/jenkins/upstream_openjdk_tests
+++ b/buildenv/jenkins/upstream_openjdk_tests
@@ -62,7 +62,8 @@ class LatestReleaseInfo implements Serializable {
 
     public boolean isRetestNeeded() {
         ZonedDateTime previous = getLatestFromWorspace()
-        return previous == null || (previous != null && latestCurrent.compareTo(previous) > 0)
+        ZonedDateTime current = ZonedDateTime.parse(latestCurrent, DateTimeFormatter.ISO_DATE_TIME);
+        return previous == null || (previous != null && current.compareTo(previous) > 0)
     }
 
     public void writeLatestToWorkspace() {
--- a/buildenv/jenkins/upstream_openjdk_tests
+++ b/buildenv/jenkins/upstream_openjdk_tests
@@ -196,7 +196,9 @@ class ReleaseMetadata {
     }
 }
 
-node("ci.role.test") {
+Map parallelJobs = [:]
+//node("ci.role.test") {
+node() {
     parameters {
         string(name:'VERSIONS', defaultValue: '8 11', description: "Space separated list of versions to check")
         booleanParam(name:'INCLUDE_EA', defaultValue: true, description: "Whether or not to include EA builds")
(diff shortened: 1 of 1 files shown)


smlambert:
Also adding @sophia-guo as a reviewer.  Thanks for pulling this together @jerboaa.  

We can use the [TestAutoGen job](https://ci.adoptopenjdk.net/view/Test_grinder/job/Test_Job_Auto_Gen/) to generate the test jobs that this pipeline would kick off (and house those test jobs in a [Test_upstream view/tab in Jenkins](https://ci.adoptopenjdk.net/view/Test_upstream/).  (we may use a slightly different naming scheme than used here, in which case I will make some requests to update job names).

I will take a closer look later today, but am happy in principal with this initial step.  We can refine or update later (if need be).

jerboaa:
> Also adding @sophia-guo as a reviewer. Thanks for pulling this together @jerboaa.
> 
> We can use the [TestAutoGen job](https://ci.adoptopenjdk.net/view/Test_grinder/job/Test_Job_Auto_Gen/) to generate the test jobs that this pipeline would kick off

Sure. I'd need guidance as to how to kick off jobs there via some parameters. If you could point me to an example, that'd be great.

> (and house those test jobs in a [Test_upstream view/tab in Jenkins](https://ci.adoptopenjdk.net/view/Test_upstream/). (we may use a slightly different naming scheme than used here, in which case I will make some requests to update job names).

OK. That would be a follow-up once the job needs creating right?

> I will take a closer look later today, but am happy in principal with this initial step. We can refine or update later (if need be).

Thanks!

smlambert:
Yes, my suggestions would be follow-up work (though we may decide not to need to do any updates).  

I need to generate some other jobs today (relating to https://github.com/AdoptOpenJDK/openjdk-tests/issues/1168), so I don't mind to try my hand to generate the jobs you need.

jerboaa:
> Yes, my suggestions would be follow-up work (though we may decide not to need to do any updates).

I see. Looking a bit closer [TestAutoGen job](https://ci.adoptopenjdk.net/view/Test_grinder/job/Test_Job_Auto_Gen/)  seems to hard-code to `buildenv/jenkins/testJobTemplate` which doesn't seem flexible enough to generate those upstream test jobs. Perhaps you mean a new `buildenv/jenkins/upstreamTestJobTemplate`? I could perhaps update `buildenv/jenkins/testJobTemplate` to account for my use case too. Not sure.

> I need to generate some other jobs today (relating to #1168), so I don't mind to try my hand to generate the jobs you need.

Oh, cool. I won't stop you ;-) Well then, a clone of the `Grinder` or `Grinder_Sandbox` config, with name say `UpstreamAutotest` or some such, would be great. I'll add the rest of the configury once a job has been set up. That job could be in  [Test_upstream view/tab in Jenkins](https://ci.adoptopenjdk.net/view/Test_upstream/) as you suggest. Thanks!

buildenv/jenkins/upstream_openjdk_tests
```diff
@@ -0,0 +1,306 @@
...
+                        stage("prepare $r.OS-$r.ARCH") {
+                            def jdk_url = r.JDK_URL
+                            def source_url = r.SOURCE_URL
+                            def jdk_version = version
+                            def jenkins_file = r.getJenkinsFile()
+                            def jdk_jre_url = "$r.JDK_URL $r.JRE_URL"
+                            parallelJobs["Test_openjdk${version}_hs_sanity.openjdk_${r.ARCH}_${r.OS}"] = {
+                                // Run a _sanity.openjdk test job with appropriate parameters
+                                build job: TEST_JOB_NAME, parameters: [
+                                    string(name: 'ADOPTOPENJDK_REPO', value: params.ADOPTOPENJDK_REPO ? params.ADOPTOPENJDK_REPO : "https://github.com/AdoptOpenJDK/openjdk-tests.git"),
```
sophia-guo: I wondered if this is specific for upstream releases to run sanity.openjdk or sanity.system tests we should not need all those extra parameters to make it more clear. Those extra parameters are specific for grinder, i.e. for developers' personal build.  Parameters like BUILD_LIST, TARGET, JDK_IMPL, SDK_RESOURCE, CUSTOMIZED_SDK_URL &CUSTOMIZED_SDK_SOURCE_URL should be enough.

sophia-guo:
My understanding is current idea is to trigger all test jobs required as a clone of Grinder_sandbox.

```
TEST_JOB_NAME = "${env.JOB_NAME}"
...
 build job: TEST_JOB_NAME, parameters: [ ... ]
```

Going forward would be reuse file Jenkinsfilebase to trigger jobs autogen-ed  in  [Test_upstream view/tab in Jenkins](https://ci.adoptopenjdk.net/view/Test_upstream/), say some specific name as 'name related with os, arch, version, level'. And then 

```
 parallelJobs[**"specific name"**] = {
                                // Run a systemtest test job with appropriate parameters
                                build job: **specific name**, parameters: [
```

sophia-guo:
When it's merged the job will trigger by Grinder manually with extra parameters GH_USER GH_TOKEN FORCE_RETEST. Not sure if I got the whole idea?

buildenv/jenkins/upstream_openjdk_tests
jerboaa: Sure. I've kept them in as I wanted a way to test changes prior merge.

`JDK_VERSION` is needed too. As otherwise, it would inherit the default value from `Grinder` or `Grinder_Sandbox` which is `8` causing a conflict with `AUTO_DETECT=true` on JDK 11.

jerboaa:
> When it's merged the job will trigger by Grinder manually with extra parameters GH_USER GH_TOKEN FORCE_RETEST. Not sure if I got the whole idea?

The initial work would be a new job, which would look very similar to `Grinder` in terms of configuration. Plus extra (optional) parameters like you said. These are:

```
    parameters {
        string(name:'VERSIONS', defaultValue: '8 11', description: "Space separated list of versions to check")
        string(name:'GH_USERNAME', description: "Github API username to authenticate as")
        string(name:'GH_TOKEN', description: "Github API OAuth personalized access token to be used for authentication")
        booleanParam(name:'INCLUDE_EA', defaultValue: true, description: "Whether or not to include EA builds")
        booleanParam(name:'FORCE_RETEST', defaultValue: false, description: "Whether or not to force a re-test irrespective previous runs")
    }

```

jerboaa:
> My understanding is current idea is to trigger all test jobs required as a clone of Grinder_sandbox.
> 
> ```
> TEST_JOB_NAME = "${env.JOB_NAME}"
> ...
>  build job: TEST_JOB_NAME, parameters: [ ... ]
> ```
> 
> Going forward would be reuse file Jenkinsfilebase to trigger jobs autogen-ed in [Test_upstream view/tab in Jenkins](https://ci.adoptopenjdk.net/view/Test_upstream/), say some specific name as 'name related with os, arch, version, level'. And then
> 
> ```
>  parallelJobs[**"specific name"**] = {
> ```

That's already done, AFAIK.

>```
>                                 // Run a systemtest test job with appropriate parameters
>                                 build job: **specific name**, parameters: [
> ```

Wouldn't I need to know the job name ahead of time, though? I could probably do something like  [TestAutoGen job](https://ci.adoptopenjdk.net/view/Test_grinder/job/Test_Job_Auto_Gen/)  which is using `buildenv/jenkins/testJobTemplate`. Perhaps use a modified version of `testJobTemplate` which accommodates the `customized` JDK use-case. I'm confused, sorry.

Having said all that, perhaps we should get started somewhere. I propose this plan of action:

1. Get this merged with some review feedback of the jenkins file itself incorporated
2. Set up a new job in the `Test_Upstream` view, which will do the testing.
3. Then you can have a look at how it works and suggest a way forward to get it massaged into something we can all agree on. Right now, I'm a bit lost what exactly you want me to do next.

In step 2) a specific name will be used which makes it apparent that this is an **upstream JDK** test job. I'm guessing that's the biggest concern? Thoughts?

buildenv/jenkins/upstream_openjdk_tests
```diff
@@ -0,0 +1,306 @@
...
+                if (forceRetest || relInfo.isRetestNeeded()) {
+                    // trigger tests for each context
+                    ctxts.each{ r -> 
+                        stage("prepare $r.OS-$r.ARCH") {
+                            def jdk_url = r.JDK_URL
+                            def source_url = r.SOURCE_URL
+                            def jdk_version = version
+                            def jenkins_file = r.getJenkinsFile()
+                            def jdk_jre_url = "$r.JDK_URL $r.JRE_URL"
+                            parallelJobs["Test_openjdk${version}_hs_sanity.openjdk_${r.ARCH}_${r.OS}"] = {
```
jerboaa: @sophia-guo What parallel branch name do you suggest instead of this? `Test_upstream_openjdk...` perhaps?

buildenv/jenkins/upstream_openjdk_tests
sophia-guo: I see. The configuration of Grinder and Grinder_Sandbox is slightly different. No default JDK_VERSION value in Grinder and default value 8 in Grinder_sandbox. As AUTO_DETECT is default true JDK_VERSION should not set default value. I just removed default JDK_VERSION in Grinder_Sandbox.

sophia-guo: Same as JDK_IMPL, I've removed the default value.

smlambert:
As I understand this there are fundamentally 2 separate pieces of functionality:
1. the piece that monitors for changes and decides if tests should be triggered (which uses the 3 parameters GH_USER, GH_TOKEN and FORCE_RETEST)
2.  the test pipelines themselves (which can be handled by directly calling Jenkinsfilebase as is)

Ultimately for part 1, it would be great to be able to use existing Jenkins plugin functionality and do a post-commit trigger, however @jerboaa wants only to monitor changes to certain files in the GH repo he is monitoring.  Part 1 could eventually also be accomplished by monitoring changes via openjdk-api v3 when that work completes.

Part 2 should just call the test pipelines directly with as @sophia-guo suggests, only the needed parameters being passed.

My mention of using the testJobTemplate was merely to make it easier to do the initial creation of the jobs required for this initial pass, but as there are only 10 required, it can also be easily done by cloning existing jobs.  I have created one in the Test_upstream view, but not yet changed the defaults...

buildenv/jenkins/upstream_openjdk_tests
sophia-guo: @jerboaa if you see the job configuration https://ci.adoptopenjdk.net/view/Test_upstream/job/Test_upstream_openjdk8_hs_sanity.openjdk_x86-64_linux/configure, all those parameter has been set default value as you needed! Thanks @smlambert

buildenv/jenkins/upstream_openjdk_tests
```diff
@@ -0,0 +1,306 @@
...
+                                // Run a _sanity.openjdk test job with appropriate parameters
+                                build job: TEST_JOB_NAME, parameters: [
```
sophia-guo: This 'TEST_JOB_NAME' is exactly same as the branch name.

jerboaa: My understanding was that the key to the `parallelJobs` map is the "branch name". At least jenkins calls it that when it runs. "TEST_JOB_NAME" refers to the jenkins job name, no? Like "Grinder". Scheduling jobs that way assumes existing job config (with matching name referenced via `TEST_JOB_NAME`), does it not?

buildenv/jenkins/upstream_openjdk_tests
jerboaa: Ok, thanks!

buildenv/jenkins/upstream_openjdk_tests
sophia-guo: Branch name is not necessarily same as the test job name. It's nice to include the hint about the test jobs. For this case as long as the branch name including the information as '${version}, hotspot, sanity, arch, os ' it would be enough. So we know which test job will run.  

Yes, TEST_JOB_NAME refers to the jenkins job name, which is the test jobs we have generated. for example: job https://ci.adoptopenjdk.net/view/Test_upstream/job/Test_upstream_openjdk8_hs_sanity.openjdk_x86-64_linux/

For now we only have one job generated, we can either generate others by duplicate copy or by autogen. If we don't want the job failed by non-generated jobs we could also use @Library('openjdk-jenkins-helper@master') by check if (JobHelper.jobIsRunnable("${TEST_JOB_NAME}")) { build job }

sophia-guo: Hmm, I see you prefer to create a new job and all parallel test jobs are also clone of this new job?
Say the job number is 1, and then you have 6 test jobs. Those jobs will be job number 2,3,...,7.

sophia-guo: We have tried this strategy before for parallel jobs and think it's not clear/straightforward for triage as all test jobs seem similar. You can only go to console and know jobs targets. 
Screen Shot 2019-06-26 at 9 49 39 AM

See this picture it's not clear to know which job is running against 11|8. 

My original thought is to create a new job and this new job will trigger all jobs you want. Similarly to https://ci.adoptopenjdk.net/view/Test_external_weekly/job/openjdk_externaltest_extended_weekly/configure

But I'm ok with both.

jerboaa: @sophia-guo It makes sense what you are saying. I'll implement it so that the generated jobs have a good name as you suggest for discoverability. What I was trying to point out is that there seems to be a "branch name" and "job name" concept in jenkins (they're not necessarily the same). I understand the branch and job names should match for a revised version of this patch. It doesn't currently, which I'll fix. Thanks and sorry for the confusion.

smlambert:
I've created 8/10 jobs needed for this, as I assumed same as @sophia-guo , that we would proceed with pre-created jobs (in that way you have job history and if we want to reference jobs from TRSS they are easily findable...), but its fine to proceed as @jerboaa has now too.

jerboaa:
@smlambert The latest version now generates jobs in the pre-created set in the Test_Upstream view. There are missing 2 jobs, `Test_upstream_openjdk11_hs_sanity.system_aarch64_linux` and `Test_upstream_openjdk11_hs_sanity.openjdk_aarch64_linux`. If somebody could create them for me I can run a full test cycle with the latest update. Thanks. Copying from `Test_upstream_openjdk11_hs_sanity.openjdk_x86-64_linux` should work well.

Open to suggestions for the name of the new "Jenkinsfile" `upstream_openjdk_tests` currently. It still hooks onto various platform specific jenkinsfiles, since I'd otherwise need to duplicate the logic in there. Thoughts?

smlambert:
All 10 jobs are there now.  I am travelling, but will check in later today.  

Let's leave the name of the file as is for now and get this merged.  You can always update things later if you want.  Approving now, but leave any other details to you and Sophia.  Thanks!

jerboaa:
Thanks all!

PR #1196, opened 2019-06-27
Retain files that match a specific filename to help debug

### LINKED ISSUE #1193: Diagnostic files not being archived as expected with test_output
There have been 2 reports of core files generated from openjdk test crashes not getting archived.  We believe this functionality should work when ARCHIVE_TEST_RESULTS is selected in a Grinder job (https://github.com/AdoptOpenJDK/openjdk-tests/pull/1073), but we should investigate why it is not.

sophia-guo:
close #1193


sophia-guo:
Tested locally and on internal jenkins with ARTIFACTORY_SERVER.

PR #1199, opened 2019-06-28
Retrieve Github credentials from creds provider plugin

jerboaa:
This change removes retrieving credentials from jenkins job parameters and instead uses credentials stored via the jenkins credentials plugin. Credentials stored that way won't show up on build jobs, so that's better from a security perspective. In addition it makes it easier to run the job regularly to check for new releases.

@karianna was kind enough to add the needed credential to the AdoptOpenJDK jenkins instance.


jerboaa:
Tested with `Grinder_Sandbox` here:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Sandbox/519/console

jerboaa:
That was quick, thanks!

PR #1202, opened 2019-07-04
Set JDK_IMPL value to hotspot if hs is provided

llxia:

### DIFF
--- a/buildenv/jenkins/testJobTemplate
+++ b/buildenv/jenkins/testJobTemplate
@@ -78,6 +78,8 @@ if (JDK_IMPL == "openj9") {
 	JDK_IMPL_SN = "j9"
 } else if (JDK_IMPL == "hotspot") {
 	JDK_IMPL_SN = "hs"
+} else if (JDK_IMPL == "hs") {
+	JDK_IMPL = "hotspot"
 }
 
 println "LEVELS: ${LEVELS}"


llxia:
Current auto-gen job failed due to `JDK_IMPL` is set to `hs` https://ci.adoptopenjdk.net/job/Test_openjdk11_hs_sanity.openjdk_x86-64_linux/1/parameters/

This PR changes `testJobTemplate` to set  `JDK_IMPL` value to `hotspot` if `hs` is provided during auto-gen.

sophia-guo:
Wondered under what condition JDK_IMPL is set as 'hs'?  The variable passing in from upstream? In that case I would prefer something is done from upstream pipeline code. Though I'm ok with this change would prefer any unexpected value should be taken as fault value.

llxia:
No, it is not passed from upstream. It is set in `Test_Job_Auto_Gen` when generating the test jobs.
https://ci.adoptopenjdk.net/view/Test_grinder/job/Test_Job_Auto_Gen/

Ideally `JDK_IMPL` should set to `hotspot` in `Test_Job_Auto_Gen`. However, I would like to handle the case where `JDK_IMPL` is set to `hs`.

llxia:
On second thought, this change may not be necessary. Users should provide the correct value. I will add the description in the job configure to make it more clear.

PR #1215, opened 2019-07-11
Passing JenkinsFile parameter to children

mw918:
Previously, the JenkinsFile parameter
would not be passed onto children builds,
which would cause an issue if the user
wanted a JenkinsFile architecture that was
not the default. The children would spawn
with the default architecture
(openjdk_x86-64_linux), and as a result,
it would fail. This change passes the
parent's architecture to the children so
that the bug would be fixed.

Signed-off-by: Michael Wang <Ziqing.Wang@ibm.com>

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -46,6 +46,7 @@ def setupEnv() {
 		env.JDK_VERSION = (params.JAVA_VERSION.trim())[2..-2];
 	}
 
+	JENKINS_FILE = params.JenkinsFile ? params.JenkinsFile : ""			
 	env.TEST_JDK_HOME = "$WORKSPACE/openjdkbinary/j2sdk-image"
 	env.JRE_IMAGE = params.JRE_IMAGE ? "${WORKSPACE}/${params.JRE_IMAGE}" : "$WORKSPACE/openjdkbinary/j2re-image"
 	env.JVM_VERSION = params.JVM_VERSION ? params.JVM_VERSION : ""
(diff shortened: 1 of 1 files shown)


mw918:
@llxia Can you please take a look at this? There was an old issue involving passing the JenkinsFile architecture that we fixed while adding other features, and we decided that we should probably fix this minor bug first. We'd like to introduce it to the project.

llxia:
Please define a default value for `JenkinsFile` parameter. The current code will break the regular test builds as they do not have `JenkinsFile` parameter defined.

We need something as follow:
```
JENKINS_FILE = params. JenkinsFile ? params.JenkinsFile : ""
```

R-Santhir:
Would an empty string as a default value be sufficient for the regular test builds?

llxia:
> Would an empty string as a default value be sufficient for the regular test builds?

yes, i think so. `JenkinsFile` is not used in the regular build, but it cannot be undefined.

R-Santhir:
LGTM. @llxia what do you think?

buildenv/jenkins/JenkinsfileBase
```diff
@@ -126,6 +126,7 @@ def setupParallelEnv() {
 				//ToDo: need to find a better way to pass parameters to downstream builds.
 				//Ideally, we should only need to pass in modified parameters (i.e.,
 				//BUILD_LIST, IS_PARALLEL)
+				JENKINS_FILE = params. JenkinsFile ? params.JenkinsFile : ""
```
llxia: - Please remove the space in `params. JenkinsFile`
- Please put this line in `setupEnv()`
- Please use the grinder to verify the change and provide the links

mw918: The changes to the code were made as per your recommendations.
The test link for the launcher is the following: https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder_Sandbox3/951/
The test link for the first subtest is the following: https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder_Sandbox3/952/
As you can see in the parameters, the JenkinsFile field is not the default, and is the same for both.

llxia:
Please do not modify any unrelated files (i.e., external/payara-mp-tck/playlist.xml, openjdk/ProblemList_openjdk13-openj9.txt, etc )

mw918:
The old commits have been erased from this commit. It was re-tested with the Grinder, and here are the links with the launcher and first child:
https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder_Sandbox3/969/
https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder_Sandbox3/970/

mw918:
Hi @llxia , could you please have a look through the changes and see if we can complete the pull request? Thanks!

llxia:
Tested in the regular build.

PR #1237, opened 2019-07-24
Enable jck-runtime-api-signaturetest and jck-runtime-api-javax_naming-spi on AIX

llxia:
- Add cmd to associate XVFB with one display on aix for jck test only
- Enable jck-runtime-api-signaturetest and jck-runtime-api-javax_naming-spi on AIX

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -352,6 +352,11 @@ def runTest( ) {
 							echo "env.DISPLAY is ${env.DISPLAY}"
 							makeTest("${RUNTEST_CMD}")
 						}
+					} else if (env.SPEC.startsWith('aix')) {
+						sh "nohup /usr/bin/X11/X -force -vfb -x abx -x dbe -x GLX :0 &"
+						env.DISPLAY = ":0"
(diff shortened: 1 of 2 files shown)


llxia:
Waiting for XVFB to be installed on all aix machines

smlambert:
The xvfb prereq on AIX machines is also applicable to openjdk tests (jdk_awt and jdk_swing, etc), we should likely open an issue at openjdk-infrastructure, so that we can re-enable those tests on more than just x86 machines (see: https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/openjdk/playlist.xml#L139).

llxia:
Internal aix machines are updated. PR is ready for review.

llxia:
> Change looks good @llxia , we will need to verify that Adopt & OpenJ9 machines have xvfb installed before merging this, correct?

This PR only applies for jck tests on AIX. Since we only run jck tests internally atm, we do not need to wait for Adopt & OpenJ9 machines to have xvfb installed.

smlambert:
I needed to view the whole file to see it, thanks!  

Will merge now, and we can tidy up the setting of xvfb to encompass more types of tests later.

PR #1241, opened 2019-07-31
Parallel grinders

mw918:
Adding the ability to run multiple iterations of a test in parallel on multiple machines


buildenv/jenkins/JenkinsfileBase
```diff
@@ -22,6 +22,7 @@ def makeTest(testParam) {
 }
 
 def setupEnv() {
+	echo "In setupEnv"
```
llxia: Please remove debugging msg.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -103,22 +104,34 @@ def setupEnv() {
 		env.DOCKERIMAGE_TAG = params.DOCKERIMAGE_TAG
 	}
 	sh 'printenv'
+	echo "End of setupEnv"
```
llxia: Please remove debugging msg. Same for the rest of the file.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -160,17 +173,61 @@ def setupParallelEnv() {
...
+							string(name: 'LABEL', value: LABEL),
+							string(name: 'JCK_GIT_REPO', value: JCK_GIT_REPO),
+							string(name: 'SSH_AGENT_CREDENTIAL', value: SSH_AGENT_CREDENTIAL),
+							booleanParam(name: 'KEEP_WORKSPACE', value: KEEP_WORKSPACE),
+							string(name: 'ARTIFACTORY_SERVER', value: ARTIFACTORY_SERVER),
+							booleanParam(name: 'PERSONAL_BUILD', value: PERSONAL_BUILD),
+							booleanParam(name: 'IS_PARALLEL', value: false),
+							booleanParam(name: 'AUTO_DETECT', value: AUTO_DETECT),
+							string(name: 'RUNTIME_NAME', value: tempDirectory)
+						]
```
llxia: Please do not duplicate the code for creating `parallel_tests`. Only a couple of parameters are different. Please use the variables.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -328,9 +385,14 @@ def runTest( ) {
 				CUSTOM_OPTION = "${TARGET.toUpperCase()}_TARGET='${CUSTOM_TARGET}'"
 			}
 			RUNTEST_CMD = "./openjdk-tests _${params.TARGET} ${CUSTOM_OPTION}"
+			if (params.IS_PARALLEL == true)
+			{
+				iterations=1
```
llxia: Why `iterations` is set to 1? We should take the value from the users.

buildenv/jenkins/JenkinsfileBase
mw918: The reason why I have this is because the structure for loops, file names, etc. are different depending on if we're running subtests in parallel or multiple iterations of tests in parallel. So, I have this inside those loops and after the variable name changes. Should I still try to pass parameters by variable names?

llxia: yes, we should we variables. Duplicate the code is not necessary here and it will result in maintaining two very similar code paths.

llxia:
Please make sure to test this PR to cover all 4 cases mentioned in https://github.com/AdoptOpenJDK/openjdk-tests/issues/836

PR #1249, opened 2019-08-06
Removed tests from jdk11 and 12 openj9 problems lists

adam-thorpe:
Tests were passing on all systems, so should be fine to include again.


smlambert:
@adam-thorpe - can you resolve the conflicts in your PR, and I will review and merge, thanks

PR #1256, opened 2019-08-07
Enhancement when customize target is empty

sophia-guo:
Make sure if customized target is empty only default customized target will run in jenkins job. Otherwise it will run the whole test set.


sophia-guo:
Problematic build : https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/545/

Fix: https://ci.adoptopenjdk.net/view/work in progress/job/Grinder_Sandbox/549/

Issue #1225, opened 2019-07-17
openjdk junit results only showing for failing test targets 


smlambert:
We used to keep and display all junit results.  This behaviour seems to have changed (which is undesirable, in my opinion).  

Seen on mac & windows builds (where all 8 test targets ran, but only failed junit results showing):

windows example:
https://ci.adoptopenjdk.net/job/Test_openjdk8_hs_sanity.openjdk_x86-64_windows/7/tapResults/
jdk_util fails: https://ci.adoptopenjdk.net/job/Test_openjdk8_hs_sanity.openjdk_x86-64_windows/7/testReport/  shows only the 665 tests in that target

mac example:
jdk_rmi fails: https://ci.adoptopenjdk.net/job/Test_openjdk8_hs_sanity.openjdk_x86-64_mac/8/testReport/ shows only the jdk_rmi 137 tests junit results

testReport view used to show all targets (around 2500 for sanity.openjdk for jdk8, and around 3300 for jdk11)

sophia-guo:
The auto generated test jobs set KEEP_REPORTDIR as false https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/testJobTemplate#L160, which delete all passing test results and junit would say

```
...
05:44:25  KEEP_REPORTDIR=false
...
[Pipeline] junit
06:39:33  Recording test results
06:39:36  None of the test reports contained any result
[Pipeline] echo

```

The output of https://ci.adoptopenjdk.net/view/Test_openjdk/job/Test_openjdk8_hs_sanity.openjdk_x86-64_linux/21/, which passed and hence no any result at all.

Grinder passing build with test output as KEEP_REPORTDIR=true and get the passing test results https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/550/testReport/

If we'd like to keep and display all junit results we need to re-run auto gen with KEEP_REPORTDIR=true. @llxia

llxia:
I will regenerate the all jobs with `KEEP_REPORTDIR=true`

llxia:
`testReport` shows up for jdk8 openjdk test builds.

https://ci.adoptopenjdk.net/view/Test_grinder/job/Test_openjdk8_j9_sanity.openjdk_x86-64_linux/19/testReport/

I just regenerated jdk11 openjdk jobs. We should see `testReport` in next build in jdk11.
[...]

PR #1258, opened 2019-08-08
Set KEEP_REPORTDIR variable

llxia:
- if not provided, the default value is true

Issue: #1225


buildenv/jenkins/testJobTemplate
```diff
@@ -27,6 +27,7 @@ if (!binding.hasVariable('SUFFIX')) SUFFIX = ""
 if (!binding.hasVariable('JCK_GIT_REPO')) JCK_GIT_REPO = ""
 if (!binding.hasVariable('JCK_GIT_REPO_PREFIX')) JCK_GIT_REPO_PREFIX = ""
 if (!binding.hasVariable('SSH_AGENT_CREDENTIAL')) SSH_AGENT_CREDENTIAL = ""
+if (!binding.hasVariable('KEEP_REPORTDIR')) KEEP_REPORTDIR = "true"
```
smlambert: Are the openj9 builds using this testJobTemplate or have they concocted their own?  if they are using it directly, are we changing the default on them?

Just calling it out, in case we need to update settings of openj9 jobs

Otherwise, this LGTM

Issue #1261, opened 2019-08-09
Test failures with jdk8 hotspot on windows


sophia-guo:
Test category: openjdk
Testsuite name: jdk_net
Test name(s):
  
 java/util/Locale/LocaleProviders.sh

Product(s) tested: 
               openjdk version "1.8.0_222"
06:10:18  OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_222-201908080335-b10)
06:10:18  OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.222-b10, mixed mode)

OS/architecture: windows
Platform specific: yes

Reproducibility: yes

Exception/Error from Log:

```
Stacktrace

Execution failed: exit code 1    

Standard Output
```

JBS:
https://bugs.openjdk.java.net/browse/JDK-8013233

https://ci.adoptopenjdk.net/view/Test_openjdk/job/Test_openjdk8_hs_sanity.openjdk_x86-64_windows/18/testReport/java_util_Locale_LocaleProviders/sh/LocaleProviders/

sophia-guo:
Test category: openjdk
Testsuite name: jdk_net
Test name(s):

sun/net/www/http/ChunkedOutputStream/checkError.java

Product(s) tested:
openjdk version "1.8.0_222"
06:10:18 OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_222-201908080335-b10)
06:10:18 OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.222-b10, mixed mode)

OS/architecture: windows
Platform specific: yes

Reproducibility: yes

Exception/Error from Log:

```
ava.lang.RuntimeException: Test Failed: haven't received the expected IOException
	at checkError.testExit(checkError.java:121)
	at checkError.main(checkError.java:115)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.sun.javatest.regtest.agent.MainWrapper$MainThread.run(MainWrapper.java:127)
	at java.lang.Thread.run(Thread.java:748)
```

JBS:
https://bugs.openjdk.java.net/browse/JDK-8041924

[...]

PR #1269, opened 2019-08-12
Exclude failure test on windows with jdk8 hs|openj9

sophia-guo:
These tests have related JBS issues.


sophia-guo:
related #1261

PR #1270, opened 2019-08-12
Parallel tests of multiple iterations 

mw918:
Added the ability to run multiple iterations of the same test in parallel. Modified loop structure, as well as passing in changed parameters as variables.

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -108,26 +108,55 @@ def setupEnv() {
 def setupParallelEnv() {
 	stage('setupParallelEnv') {
 		timestamps{
-			setupEnv()
+			setupEnv()		
+			int iterate = 1
+			if (params.ITERATIONS) {
+				iterate = "${params.ITERATIONS}".toInteger()
+			}
+			//Initialization for things in common between both multiple iterations and many subtests
+			//repeatAmount is the number of times the loops go through
+			int repeatAmount = 0
+			int one = 1
+			
 			def testSubDirs = []
 			def testSubDirSize = 0
 			
-			//For parallel suppose BUILD_LIST is one specific category
-			dir("$WORKSPACE/openjdk-tests/${env.BUILD_LIST}") {
+			dir("$WORKSPACE/openjdk-tests/${env.BUILD_LIST}") 
+			{
 				testSubDirs = sh(returnStdout: true, script: "ls -d */").trim().tokenize()
 				testSubDirSize = testSubDirs.size()
 			}
-			echo "testSubDirSize is ${testSubDirSize}, testSubDirs is ${testSubDirs}, running test in parallel mode"
+			
+			//Setup for option-specific parameters
+			if (iterate == 1){
+				//For parallel suppose BUILD_LIST is one specific category
+				repeatAmount = testSubDirSize
+			}
+			else
+			{
+				repeatAmount = iterate
+			}
+			
+			//echo "testSubDirSize is ${testSubDirSize}, testSubDirs is ${testSubDirs}, running test in parallel mode"
 			parallel_tests = [:]
 
-			for (int i = 0; i < testSubDirSize; i++) {
-				def testSubDir = testSubDirs[i].trim().replace("/","");
+			for (int i = 0; i < repeatAmount; i++) {
+				def buildListName = ""
+				def currentDirectory = ""
+				if (iterate == 1){
+					currentDirectory = testSubDirs[i].trim().replace("/","");
+					buildListName = "${env.BUILD_LIST}/${currentDirectory}"
+				}
+				else{
+					currentDirectory = "test_iteration_${i}"
+					buildListName = env.BUILD_LIST
+				}
 				def TEST_JOB_NAME = "${JOB_NAME}"
 
 				//ToDo: need to find a better way to pass parameters to downstream builds.
 				//Ideally, we should only need to pass in modified parameters (i.e.,
 				//BUILD_LIST, IS_PARALLEL)
-				parallel_tests[testSubDir] = {
+				parallel_tests[currentDirectory] = {
 					build job: TEST_JOB_NAME, parameters: [
 						string(name: 'ADOPTOPENJDK_REPO', value: ADOPTOPENJDK_REPO),
 						string(name: 'ADOPTOPENJDK_BRANCH', value: ADOPTOPENJDK_BRANCH),
@@ -135,7 +164,7 @@ def setupParallelEnv() {
 						string(name: 'OPENJ9_BRANCH', value: OPENJ9_BRANCH),
 						string(name: 'JDK_VERSION', value: JDK_VERSION),
 						string(name: 'JDK_IMPL', value: JDK_IMPL),
-						string(name: 'BUILD_LIST', value: "${env.BUILD_LIST}/${testSubDir}"),
+						string(name: 'BUILD_LIST', value: buildListName),
 						string(name: 'TARGET', value: TARGET),
 						string(name: 'CUSTOM_TARGET', value: CUSTOM_TARGET),
 						string(name: 'SDK_RESOURCE', value: SDK_RESOURCE),
@@ -148,7 +177,7 @@ def setupParallelEnv() {
 						string(name: 'EXTRA_OPTIONS', value: EXTRA_OPTIONS),
 						string(name: 'JVM_OPTIONS', value: JVM_OPTIONS),
 						string(name: 'JenkinsFile', value: JENKINS_FILE),
-						string(name: 'ITERATIONS', value: ITERATIONS),
+						string(name: 'ITERATIONS', value: one.toString()),
 						string(name: 'LABEL', value: LABEL),
 						string(name: 'JCK_GIT_REPO', value: JCK_GIT_REPO),
 						string(name: 'SSH_AGENT_CREDENTIAL', value: SSH_AGENT_CREDENTIAL),
@@ -158,6 +187,7 @@ def setupParallelEnv() {
 						booleanParam(name: 'IS_PARALLEL', value: false),
 						booleanParam(name: 'AUTO_DETECT', value: AUTO_DETECT),
 						string(name: 'TIME_LIMIT', value: "${TIME_LIMIT}"),
+						string(name: 'RUNTIME_NAME', value: currentDirectory),
 					]
 				}
 			}


mw918:
@llxia Can you please check this out?

buildenv/jenkins/JenkinsfileBase
```diff
@@ -108,34 +108,63 @@ def setupEnv() {
 def setupParallelEnv() {
 	stage('setupParallelEnv') {
 		timestamps{
-			setupEnv()
+			setupEnv()		
+			int iterate = 1
+			if (params.ITERATIONS) {
+				iterate = "${params.ITERATIONS}".toInteger()
+			}
```
llxia: Please set variable in  setupEnv()
I think we do not need both  `repeatAmount` and `iterate`. One should be enough.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -108,34 +108,63 @@ def setupEnv() {
...
 			
-			//For parallel suppose BUILD_LIST is one specific category
-			dir("$WORKSPACE/openjdk-tests/${env.BUILD_LIST}") {
+			dir("$WORKSPACE/openjdk-tests/${env.BUILD_LIST}") 
+			{
 				testSubDirs = sh(returnStdout: true, script: "ls -d */").trim().tokenize()
 				testSubDirSize = testSubDirs.size()
 			}
-			echo "testSubDirSize is ${testSubDirSize}, testSubDirs is ${testSubDirs}, running test in parallel mode"
+
```
llxia: Only need to check sub dirs if `ITERATIONS` is 1.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -148,7 +177,7 @@ def setupParallelEnv() {
 						string(name: 'EXTRA_OPTIONS', value: EXTRA_OPTIONS),
 						string(name: 'JVM_OPTIONS', value: JVM_OPTIONS),
 						string(name: 'JenkinsFile', value: JENKINS_FILE),
-						string(name: 'ITERATIONS', value: ITERATIONS),
+						string(name: 'ITERATIONS', value: one.toString()),
```
llxia: I think we can just use `"1"`

buildenv/jenkins/JenkinsfileBase
```diff
@@ -158,6 +187,7 @@ def setupParallelEnv() {
 						booleanParam(name: 'IS_PARALLEL', value: false),
 						booleanParam(name: 'AUTO_DETECT', value: AUTO_DETECT),
 						string(name: 'TIME_LIMIT', value: "${TIME_LIMIT}"),
+						string(name: 'RUNTIME_NAME', value: currentDirectory),
```
llxia: Is this `RUNTIME_NAME ` needed?

buildenv/jenkins/JenkinsfileBase
```diff
@@ -108,34 +108,63 @@ def setupEnv() {
...
+			if (iterate == 1){
+				//For parallel suppose BUILD_LIST is one specific category
+				repeatAmount = testSubDirSize
+			}
+			else
+			{
+				repeatAmount = iterate
+			}
+			
+			//echo "testSubDirSize is ${testSubDirSize}, testSubDirs is ${testSubDirs}, running test in parallel mode"
```
llxia: I think we should keep this echo.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -108,34 +108,63 @@ def setupEnv() {
...
+				def buildListName = ""
+				def currentDirectory = ""
+				if (iterate == 1){
+					currentDirectory = testSubDirs[i].trim().replace("/","");
+					buildListName = "${env.BUILD_LIST}/${currentDirectory}"
+				}
+				else{
+					currentDirectory = "test_iteration_${i}"
+					buildListName = env.BUILD_LIST
+				}
```
llxia: `currentDirectory ` is misleading. `test_iteration_${i}` is not a directory. Could we call it something meaningful? For example, childTest.

llxia:
Please provide Grinder links for 4 test cases. Thanks.

buildenv/jenkins/JenkinsfileBase
mw918: Both variables are needed, because they have different functions. The variable iterate is the number of iterations as passed by the user. The variable repeatAmount is the number of times the loop runs; this is equal to ITERATIONS if it's greater than one, and the number of subdirectories if ITERATIONS is equal to one.

llxia: I think we can do something like below:
```
repeatAmount = "${params.ITERATIONS}".toInteger()
if (repeatAmount == 1) {
   // get testSubDirSize and set it
   repeatAmount = testSubDirSize
}
```

mw918: I just tried to move the variable in setupEnv(), and it didn't work. I also noticed in setupEnv() that there are no ints in there. The function is also used to setup situations both parallel as well as not, whereas iterate is only used for parallel situations. This is also a local variable that was there previously, so I'm not really sure if it's necessary to move it.

buildenv/jenkins/JenkinsfileBase
mw918: Yes, that's the feature that was needed to distinguish one iteration of the same test from another. That's what fixed the issue with the iterations building over top of each other on the same machine.

buildenv/jenkins/JenkinsfileBase
llxia: `iterations` is not only used in parallel situations. We also use it in `runTest()`:
https://github.com/AdoptOpenJDK/openjdk-tests/blob/b3ce67da9cf6c678eaffbbabb8abeb763f8482ed/buildenv/jenkins/JenkinsfileBase#L327

So it is better to set it once.

buildenv/jenkins/JenkinsfileBase
mw918: This is only useful in the situation where the iteration count is 1

mw918:
> 
> 
> Please provide Grinder links for 4 test cases. Thanks.

Here they are:
https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder_Sandbox3/1378/
https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder_Sandbox3/1377/
https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder_Sandbox3/1368/
https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder_Sandbox3/1363/

buildenv/jenkins/JenkinsfileBase
mw918: I'm not completely sure how to do this. I tried yesterday by defining the int in the setupEnv(), and there are no local variables being declared there. Should I make `iterations` a global variable, or is there some other way to do this?

llxia: We can set it in `env`.

llxia:
Please format the code using the IDE. For example, `{` should be in the same line of the code, not in the next line. This PR is using both cases.

```
if (iterate == 1){
```

```
else
{
```

buildenv/jenkins/JenkinsfileBase
```diff
@@ -58,6 +58,13 @@ def setupEnv() {
 	env.KEEP_REPORTDIR = params.KEEP_REPORTDIR ? params.KEEP_REPORTDIR : ''
 	SDK_RESOURCE = params.SDK_RESOURCE ? params.SDK_RESOURCE : "upstream"
 	env.AUTO_DETECT = params.AUTO_DETECT
+	
+	ITERATIONS = 1;
+	
+	if (params.ITERATIONS) {
+		ITERATIONS = "${params.ITERATIONS}".toInteger()
+	}
```
llxia: Please change to following:
```
ITERATIONS = params.ITERATIONS ? "${params.ITERATIONS}".toInteger() : 1
```

buildenv/jenkins/JenkinsfileBase
```diff
@@ -108,34 +115,53 @@ def setupEnv() {
...
-			setupEnv()
+			setupEnv()		
+			
+			//Initialization for things in common between both multiple iterations and many subtests
+			//repeatAmount is the number of times the loops go through
+			int repeatAmount = 0
 			def testSubDirs = []
-			def testSubDirSize = 0
+	
+			repeatAmount = ITERATIONS
```
llxia: Can we save one line?
```
int repeatAmount = ITERATIONS
```

buildenv/jenkins/JenkinsfileBase
```diff
@@ -108,34 +115,53 @@ def setupEnv() {
...
-			dir("$WORKSPACE/openjdk-tests/${env.BUILD_LIST}") {
-				testSubDirs = sh(returnStdout: true, script: "ls -d */").trim().tokenize()
-				testSubDirSize = testSubDirs.size()
+			//Setup for option-specific parameters
+			if (repeatAmount == 1){
+				//For parallel suppose BUILD_LIST is one specific category
+				dir("$WORKSPACE/openjdk-tests/${env.BUILD_LIST}") 
+				{
+					testSubDirs = sh(returnStdout: true, script: "ls -d */").trim().tokenize()
+					testSubDirSize = testSubDirs.size()
```
llxia: - `testSubDirSize` is not defined anywhere. Actually, we do not need it. We can do

```
repeatAmount = testSubDirs.size()
``` 
- Please keep the original echo msg

buildenv/jenkins/JenkinsfileBase
```diff
@@ -148,7 +174,8 @@ def setupParallelEnv() {
 						string(name: 'EXTRA_OPTIONS', value: EXTRA_OPTIONS),
 						string(name: 'JVM_OPTIONS', value: JVM_OPTIONS),
 						string(name: 'JenkinsFile', value: JENKINS_FILE),
-						string(name: 'ITERATIONS', value: ITERATIONS),
+						//string(name: 'ITERATIONS', value: one.toString()),
```
llxia: Please remove the comment

buildenv/jenkins/JenkinsfileBase
```diff
@@ -117,34 +124,53 @@ def setupEnv() {
...
-			echo "testSubDirSize is ${testSubDirSize}, testSubDirs is ${testSubDirs}, running test in parallel mode"
+			
 			parallel_tests = [:]
 
-			for (int i = 0; i < testSubDirSize; i++) {
-				def testSubDir = testSubDirs[i].trim().replace("/","");
+			for (int i = 0; i < repeatAmount; i++) {
+				def buildListName = ""
+				def childTest = ""
+				if (ITERATIONS == 1){
```
llxia: Please add a space between ) and {

buildenv/jenkins/JenkinsfileBase
```diff
@@ -117,34 +124,53 @@ def setupEnv() {
...
+					childTest = testSubDirs[i].trim().replace("/","");
+					buildListName = "${env.BUILD_LIST}/${childTest}"
+				}
+				else{
```
llxia: Please add a space between else and {

buildenv/jenkins/JenkinsfileBase
```diff
@@ -168,6 +195,7 @@ def setupParallelEnv() {
 						booleanParam(name: 'IS_PARALLEL', value: false),
 						booleanParam(name: 'AUTO_DETECT', value: AUTO_DETECT),
 						string(name: 'TIME_LIMIT', value: "${TIME_LIMIT}"),
+						string(name: 'RUNTIME_NAME', value: childTest),
```
llxia: Please add comment on why this param is needed.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -117,34 +120,52 @@ def setupEnv() {
...
-			def testSubDirSize = 0
+			int repeatAmount = ITERATIONS
 			
-			//For parallel suppose BUILD_LIST is one specific category
-			dir("$WORKSPACE/openjdk-tests/${env.BUILD_LIST}") {
-				testSubDirs = sh(returnStdout: true, script: "ls -d */").trim().tokenize()
-				testSubDirSize = testSubDirs.size()
+			//Setup for option-specific parameters
+			if (repeatAmount == 1){
+				//For parallel suppose BUILD_LIST is one specific category
```
llxia: Please remove this comment.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -117,34 +120,52 @@ def setupEnv() {
...
+			//repeatAmount is the number of times the loops go through
 			def testSubDirs = []
-			def testSubDirSize = 0
+			int repeatAmount = ITERATIONS
 			
-			//For parallel suppose BUILD_LIST is one specific category
-			dir("$WORKSPACE/openjdk-tests/${env.BUILD_LIST}") {
-				testSubDirs = sh(returnStdout: true, script: "ls -d */").trim().tokenize()
-				testSubDirSize = testSubDirs.size()
+			//Setup for option-specific parameters
```
llxia: Please add - If ITERATIONS is 1, then parallel based on subfolders.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -117,34 +120,52 @@ def setupEnv() {
...
+				//For parallel suppose BUILD_LIST is one specific category
+				dir("$WORKSPACE/openjdk-tests/${env.BUILD_LIST}") 
+				{
+					testSubDirs = sh(returnStdout: true, script: "ls -d */").trim().tokenize()
+					repeatAmount = testSubDirs.size()
+				}
 			}
-			echo "testSubDirSize is ${testSubDirSize}, testSubDirs is ${testSubDirs}, running test in parallel mode"
+			
+			echo "repeatAmount is ${repeatAmount}, testSubDirs is ${testSubDirs}, running test in parallel mode"
```
llxia: Please move the echo into if. Otherwise, for ITERATIONS!=1, the output is `testSubDirs is []`

buildenv/jenkins/JenkinsfileBase
```diff
@@ -117,34 +120,52 @@ def setupEnv() {
...
 			def testSubDirs = []
-			def testSubDirSize = 0
+			int repeatAmount = ITERATIONS
 			
-			//For parallel suppose BUILD_LIST is one specific category
-			dir("$WORKSPACE/openjdk-tests/${env.BUILD_LIST}") {
-				testSubDirs = sh(returnStdout: true, script: "ls -d */").trim().tokenize()
-				testSubDirSize = testSubDirs.size()
+			//Setup for option-specific parameters
+			//If ITERATIONS is 1, then parallel builds the different subfolders onto the machines
```
llxia: We do not `builds the different subfolders onto the machines`. We do parallel based on subfolders.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -117,34 +120,52 @@ def setupEnv() {
...
+			if (repeatAmount == 1){
+				//For parallel suppose BUILD_LIST is one specific category
```
llxia: This msg is misleading. Please remove it.

llxia:
Please also add git sign off.

Issue #1224, opened 2019-07-17
OpenJ9 fetch intermittently fails


cwesMills:
https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/get.sh#L267-L268
One of these 2 commands seems to intermittently fail, could something be added to retry the command(s) on failure?

llxia:
Could you provide more detail? For example, job output and/or job link?

cwesMills:
```
23:25:38  Eclipse OpenJ9 VM (build ibm_sdk-912c1a4f7c, JRE 11 Windows Server 2012 R2 amd64-64-Bit Compressed References 20190715_92 (JIT enabled, AOT enabled)
23:25:38  OpenJ9   - 912c1a4f7c
23:25:38  OMR      - 2967e8a8
23:25:38  JCL      - 411f32dbbb4 based on jdk-11.0.4+10)
23:25:38  get testKitGen and functional test material...
23:25:38  git clone --depth 1 -b ibm_sdk git@github.ibm.com:runtimes/openj9.git
23:28:02  update to openj9 sha: 912c1a4f7cbe776e9c4d3a0115c1b6c40ab5b936
23:28:49  From github.ibm.com:runtimes/openj9
23:28:49   * [new tag]             Java8_SR5_FP11_20180309_0723_B380776 -> Java8_SR5_FP11_20180309_0723_B380776
23:28:49   * [new tag]             Java8_SR5_FP21_20180829_1605_B395745 -> Java8_SR5_FP21_20180829_1605_B395745
23:28:49   * [new tag]             Java8_SR5_FP22_20180915_1256_B397057 -> Java8_SR5_FP22_20180915_1256_B397057
23:28:49   * [new tag]             Java8_SR5_FP6_20171122_2152_B371101 -> Java8_SR5_FP6_20171122_2152_B371101
23:28:49   * [new tag]             Java8_SR5_FP7_20171215_1234_B373586 -> Java8_SR5_FP7_20171215_1234_B373586
23:28:49   * [new tag]             openj9-0.0        -> openj9-0.0
23:28:49   * [new tag]             openj9-0.0M1      -> openj9-0.0M1
23:28:49   * [new tag]             openj9-0.0RC2     -> openj9-0.0RC2
23:28:49   * [new tag]             openj9-0.10.0-rc1 -> openj9-0.10.0-rc1
23:28:49   * [new tag]             openj9-0.11.0     -> openj9-0.11.0
[...]

PR #1278, opened 2019-08-14
WIP: Added retry loop on git fetch

cwesMills:
Some tests fail because of connection issues to github. This change will
retry the git fetch command if it fails, up to a maximum of 5 times,
with a 5 minute delay between each retry.
Fixes: #1224

Signed-off-by: Colton Mills <millscolt3@gmail.com>

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -265,7 +265,28 @@ getTestKitGenAndFunctionalTestMaterial()
 		echo "update to openj9 sha: $OPENJ9_SHA"
 		cd openj9
 		git fetch --unshallow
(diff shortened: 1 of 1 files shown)


llxia:
We would prefer to investigate why it failed. As far as I know, the intermittent failure only happens for the internal repo. I think we should remove `-q`, so we can know what goes wrong.

karianna:
@llxia and @cwesMills - do we need to still process this PR in that case?

smlambert:
@llxia and I discussed this yesterday.  I would like to close this PR and re-add the -q as our console output is unpleasant to navigate with the extra output.  

Suggest to continue internal investigations off of a branch and return to open a PR if it is needed once problem determination complete.

Issue #1280, opened 2019-08-15
Regular expression match =~ doesn't work on older versions of bash


Mesbah-Alam:
With older bash, regular expression match `=~` doesn't seem to work. For example, in a ci.eclipse AIX machine, we got the following failure: 

```
00:26:45  if [[ aix_ppc-64_cr =~ "osx" ]] ; then umask $original_umask_val && umask -p; fi; \
00:26:45  if [ $? -eq 0 ] ; then echo ""; echo "SharedClassesAPI_0""_PASSED"; echo ""; cd /home/u0020236/workspace/Test_openjdk11_j9_extended.system_ppc64_aix_Nightly/openjdk-tests/TestConfig/scripts/testKitGen/../../..; rm -f -r "/home/u0020236/workspace/Test_openjdk11_j9_extended.system_ppc64_aix_Nightly/openjdk-tests/TestConfig/scripts/testKitGen/../../../TestConfig/test_output_1565842161615/SharedClassesAPI_0"; else echo ""; echo "SharedClassesAPI_0""_FAILED"; echo ""; fi; } 2>&1 | tee -a "/home/u0020236/workspace/Test_openjdk11_j9_extended.system_ppc64_aix_Nightly/openjdk-tests/TestConfig/scripts/testKitGen/../../../TestConfig/test_output_1565842161615/TestTargetResult";
00:26:45  /bin/sh: 0403-057 Syntax error at line 4 : `=~' is not expected.
```

We need to replace `=~` with something else in the Shared Classes playlist.

PR #1281, opened 2019-08-15
Use regular expression that works on older bash to detect platform

Mesbah-Alam:
Related: https://github.com/AdoptOpenJDK/openjdk-tests/issues/1280

### DIFF
--- a/systemtest/sharedClasses/playlist.xml
+++ b/systemtest/sharedClasses/playlist.xml
@@ -24,14 +24,14 @@
 		<variations>
 			<variation>NoOptions</variation>
 		</variations>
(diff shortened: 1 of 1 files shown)


Mesbah-Alam:
Tested : 
Linux: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1914/
Mac: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1916/console

Internal AIX link will be shared with reviewer on Slack.

sophia-guo:
Did the grinder ( specifically AIX one) running on machine with older bash version?

sophia-guo:
@Mesbah-Alam tried and seemed it's not easy to find a machine with old bash to verify the change. However the change did work on newer old bash. That is saying it has no effect for machines with newer bash. We can probably let it in and keep an eye on #1280 to see how it works. @llxia your opinion?

PR #1286, opened 2019-08-16
Exclude JBS recorded tests

sophia-guo:
JBS issues fixed though the failures are still there. 


sophia-guo:
No win32 with openj9. Openj9 win64 failed for different issue and has been excluded with specific issue.

PR #1287, opened 2019-08-16
Removed quiet option from git fetch

cwesMills:
A command is failing during tests, as seen in #1224. Removed quiet option
(-q) from 2 commands to allow more output and hopefully determine what
causes the error.

Signed-off-by: Colton Mills <millscolt3@gmail.com>


cwesMills:
Related to #1278

PR #1298, opened 2019-08-22
WIP: Add jck label

Mesbah-Alam:
- Define LABEL as a parameter in testJobTemplate, with default value empty
- When group is JCK, set it to "&& !ci.role.test.no_jck"
- In the Jenkins template files for each platform, add logic append value found in LABEL, if it starts with "&&". This takes care of 3 scenarios - (a) when LABEL is empty (default) (b) When a custom value is provided for LABEL (e.g. node name given in grinders by users) (c) JCK builds where we'd have the JCK label supplied via generated jobs. 

FYI @AdamBrousseau @llxia @sophia-guo 

### DIFF
--- a/buildenv/jenkins/testJobTemplate
+++ b/buildenv/jenkins/testJobTemplate
@@ -105,8 +105,10 @@ ARCH_OS_LIST.each { ARCH_OS ->
 
 				// for all jck builds, we need -Xfuture option passed in
 				def EXTRA_OPTIONS = "";
+				def LABEL = "";
 				if (GROUP == "jck" && JDK_IMPL == "openj9") {
 					EXTRA_OPTIONS = "-Xfuture"
+					LABEL = "&& !ci.role.test.no_jck"
 				}
 				
 				// for jck builds, if JCK_GIT_REPO_PREFIX is set, set JCK_GIT_REPO if it is not set
@@ -136,7 +138,7 @@ ARCH_OS_LIST.each { ARCH_OS ->
 							stringParam('EXTRA_OPTIONS', EXTRA_OPTIONS, "Use this to append options to the test command")
 							stringParam('JVM_OPTIONS', "", "Use this to replace the test original command line options")
 							stringParam('ITERATIONS',"1", "Number of times to repeat execution of test target")
-							stringParam('LABEL', "", "Jenkins node label to run on, leave this blank to get sent to any machine matching the platform, set to node name for runs a particular machine")
+							stringParam('LABEL', LABEL, "Jenkins node label to run on, leave this blank to get sent to any machine matching the platform, set to node name for runs a particular machine")
 							stringParam('JCK_GIT_REPO', JCK_GIT_REPO, "For JCK test only")
 							stringParam('SSH_AGENT_CREDENTIAL', SSH_AGENT_CREDENTIAL, "Optional. Only use when ssh credentials are needed")
 							booleanParam('KEEP_WORKSPACE', false, "Keep workspace on the machine")
--- a/buildenv/jenkins/openjdk_x86-64_linux
+++ b/buildenv/jenkins/openjdk_x86-64_linux
@@ -6,8 +6,16 @@ if (params.DOCKER_REQUIRED) {
     LABEL='ci.role.test&&hw.arch.x86&&sw.os.linux'
 }
 
+if (params.LABEL) {
+	if (params.LABEL.startsWith('&&')) {
+		LABEL += params.LABEL
+	} else {
+		LABEL = params.LABEL
+	}
+} 
+
 stage('Queue') {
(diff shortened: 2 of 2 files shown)


sophia-guo:
FYI @llxia

Mesbah-Alam:
Testing: 

1) In order to test the template job, ran the generation job first, then ran the generated jobs with and without JCK. 
2) Also ran simple Grinder with JCK test with the branch of the PR to check the change made only in the template jenkins file. 

(Links to the testing shared with Sophia on Slack. )

Mesbah-Alam:
Background for this PR.. Copying Adam's comment from an internal issue: 

> Currently the only extra requirement for JCK tests is having the jck_root folder which means ~15-20G of space. Since we have a much smaller set of machines that won't satisfy this condition, it would be less work to have a negative label added to "small" machines. The plan going forward will be, when a machines runs out of space, add the label ci.role.test.no_jck and open an issue. Second step will be to examine the machines (in the issue) with this new label, and increase the size if possible. If the size can be increased then the label will be removed. Mesbah will add the label ... && !ci.role.test.no_jck to the test pipelines so that jck tests will not land on these machines.

buildenv/jenkins/testJobTemplate
```diff
@@ -136,7 +138,7 @@ ARCH_OS_LIST.each { ARCH_OS ->
 							stringParam('EXTRA_OPTIONS', EXTRA_OPTIONS, "Use this to append options to the test command")
 							stringParam('JVM_OPTIONS', "", "Use this to replace the test original command line options")
 							stringParam('ITERATIONS',"1", "Number of times to repeat execution of test target")
-							stringParam('LABEL', "", "Jenkins node label to run on, leave this blank to get sent to any machine matching the platform, set to node name for runs a particular machine")
+							stringParam('LABEL', LABEL, "Jenkins node label to run on, leave this blank to get sent to any machine matching the platform, set to node name for runs a particular machine")
 							stringParam('JCK_GIT_REPO', JCK_GIT_REPO, "For JCK test only")
```
sophia-guo: Instead of using existing LABEL I would suggest to add a new parameter EXTRA_LABEL. Hence if there is some other specific requirement it can just be passed in.

AdamBrousseau: Not a bad idea, it would make the logic more simple and more easy for the end user.

sophia-guo:
As this is an internal specific machine issue and there will be following up work(machine update and label removing) in internal I wondered if this specific LABEL be passed in by upstream? Now it looks like hard coded in Adopt?

AdamBrousseau:
If we pass it from upstream it would be changes to the OpenJ9 pipelines which are external as well. I could make the same argument that it should not be hardcoded there. Ideally there would be some kind of internal config file to make changes to defaults like this.

llxia:
I'm not sure I like setting jck label this way in testJobTemplate. This testJobTemplate should be general and is used by different Jenkins servers. Ideally, it should not handle special/temporary cases for the special site.

I agree with Sophia. This is an internal issue and the internal upstream job generates and triggers jck test jobs. If `EXTRA_LABEL` is added, the internal upstream job can set and change the value when needed.

But since this PR is a temporary solution and the `ci.role.test.no_jck` label will be removed once machines are updated, I am ok with the change in testJobTemplate.

AdamBrousseau:
Note that we have not guaranteed that all machines will be able to have their disks increased enough to support this testing. Furthermore, the requirements for these machines will be expanding in the coming weeks which may further limit the machines that are able to run these tests. We will likely need a different solution long term.

buildenv/jenkins/testJobTemplate
Mesbah-Alam: @sophia-guo - sure. I'll make that change!

Mesbah-Alam:
Valid points raised by all.

Hi, @ShelleyLambert , would you please leave your opinion on this? Especially regarding the use of the negative label `ci.role.test.no_jck`?

smlambert:
I am going to shift this discussion to backlog repo, as it pertains to internal testing.

smlambert:
but in short, my preference may be for a separate parameter called APPEND_LABEL or EXTRA_LABEL, as then it can be used for a variety of potential future requirements (and past requirements, a more generic way of handling sw.os.docker label, https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/openjdk_x86-64_linux#L3

I will like to think through some other possible use cases also.

smlambert:
Sounds like from the resolution of internal issue (machines with enough space, disabling the few without it), this can be closed, correct?

Mesbah-Alam:
Closing this PR without merging as we no longer have the need to schedule jck jobs on machines labelled for jck.

Issue #1294, opened 2019-08-22
jdk11u build pipeline extended system test runs fail


lumpfish:
All the extended system tests run by the build pipeline for the rerelease of jdk11u on MacOS failed with these messages:
```
Can't open perl script "/stf/stf.core/scripts/stf.pl": No such file or directory
```
Looks like something may be wrong with setting the test root directory which should prefix '/stf' in the playlist or testkitgen variables.
Job is https://ci.adoptopenjdk.net/job/Test_openjdk11_j9_extended.system_x86-64_mac/25/console

lumpfish:
The tests execute OK when I reran them as a Test_grinder job: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1925/console

smlambert:
@lumpfish you have shared a link to jdk11 builds (jdk8 looks ok, https://ci.adoptopenjdk.net/job/Test_openjdk8_j9_extended.system_x86-64_mac/22/), should update the title of this issue to match.

FYI @Mesbah-Alam @llxia

llxia:
The next jdk11 build is fine https://ci.adoptopenjdk.net/job/Test_openjdk11_j9_extended.system_x86-64_mac/26/

lumpfish:
Sorry - it should all have said jdk11, not jdk8.  Updated the title and comments.

Yes, follow on builds are ok, weird a parameter should go 'missing' though.

smlambert:
that is odd... wonder if there were temporary changes in build pipeline code (maybe should look)?  we can keep this open for the next few runs, to make sure its not some strange intermittent behaviour

sophia-guo:
Looks like build workspace is not clean  https://ci.adoptopenjdk.net/job/Test_openjdk11_j9_extended.system_x86-64_mac/25/console

```
10:40:46  clean:
10:40:46     [delete] Deleting: /Users/jenkins/workspace/Test_openjdk11_j9_extended.system_x86-64_mac/openjdk-tests/TestConfig/autoGenEnv.mk
10:40:46     [delete] Deleting directory /Users/jenkins/workspace/Test_openjdk11_j9_extended.system_x86-64_mac/openjdk-tests/TestConfig/src/bin
10:40:46     [delete] Deleting: /Users/jenkins/workspace/Test_openjdk11_j9_extended.system_x86-64_mac/openjdk-tests/TestConfig/src/EnvDetector.jar
```

[...]

PR #1299, opened 2019-08-26
Clean up the workspace and prune obsolete local branches

sophia-guo:
Clean up the workspace before every checkout to ensure build is not
affected by the files generated by the previous build


sophia-guo:
Related with #1294

PR #1320, opened 2019-09-04
Partial FreeBSD support

battleblow:
Add some support for FreeBSD:

* Determine the number of processors on FreeBSD
* GNU make is named gmake on *BSD

### DIFF
--- a/maketest.sh
+++ b/maketest.sh
@@ -11,11 +11,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ######
-if [ `uname` = AIX ]; then
-  MAKE=gmake
-else
-  MAKE=make
-fi
+case `uname` in
+  AIX | *BSD )
+    MAKE=gmake
+    ;;
+  *)
+    MAKE=make
+    ;;
+esac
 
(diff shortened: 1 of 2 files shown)


battleblow:
"Partial" because there are other necessary changes.

smlambert:
A few grinders to verify other platforms unaffected:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/382
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/384

If we had BSD machines at the project ATM, I would also run a few grinders on them, but this change is benign, and will merge.  

It is partial as there will need to be a few other changes to enable all the testing available from this project (systemtest update to use gmake to compile native tests, and separate jenkins file which can be added when/if we get test machines at the project).

battleblow:
Thanks Shelley!

Yes, partial because there are other changes needed.  I don't have the system test suite working yet, but I can run the openjdk and functional test suites with some other changes I have locally.  I'll push those for review as well, but wanted to separate these out since I hoped they wouldn't be controversial at all.

PR #1325, opened 2019-09-04
Portable interpreter invocation

battleblow:
* *BSD does not have /bin/bash.  Use /usr/bin/env bash for portability.

### DIFF
--- a/functional/SyntheticGCWorkload/concurrent_slack_sweep.sh
+++ b/functional/SyntheticGCWorkload/concurrent_slack_sweep.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/functional/SyntheticGCWorkload/run_abort_test.sh
+++ b/functional/SyntheticGCWorkload/run_abort_test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/functional/SyntheticGCWorkload/run_test.sh
+++ b/functional/SyntheticGCWorkload/run_test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/functional/SyntheticGCWorkload/run_tests.sh
+++ b/functional/SyntheticGCWorkload/run_tests.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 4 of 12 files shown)


battleblow:
Some notes on these changes:

There are a couple of minor whitespace changes involved here as well.  In particular, a few files didn't have a final \n and opening and saving them on a Unix box added this.  There was one file where I removed trailing whitespace on the last line too.

There are tradeoffs with this approach versus other options I considered.  I'll attempt to enumerate them here.

**Use "/usr/bin/env bash" vs "/bin/bash"**

**Pros**

* This approach should work on all platforms (standards compliant)
* It requires a relatively small number of changes.

**Cons**

* The first instance of 'bash' in PATH is selected, which may not be in the one in /bin and may not even be the Bourne Again Shell (any executable named bash counts).  Note that this is also the feature that allows this to work on *BSD, so it being a con is debatable.
* Other code that is downloaded as part of the tests doesn't get this fix.  E.g. the use of getSHA.sh from the OpenJ9 repo is not fixed by this change and will require an upstream change or another approach.

**Convert all of the scripts to be POSIX sh compliant**

**Pros**

* This approach should work on all platforms (standards compliant)

**Cons**

* This may be a considerable amount of work.  Based on checkbashisms(1), some scripts could already be moved over, but a script like /perf/liberty/scripts/bin/common_utils.sh showed a lot of hits for potential bashisms.
* Developers are restricted to POSIX sh features in scripts and can't use bashisms.

**Anyone working on *BSD creates a symlink**

**Pros**

* Requires no code changes
* Will work for other code pulled in by the tests

**Cons**

* Requires root access on *BSD, which is a requirement no other platform has.  This effectively rules out developers who either don't have root access or are unprepared to change their system in this fashion.  This has additional impact in that it complicates onboarding for new *BSD machines for either the project or developers who want to use them.
* Symlinks are potentially fragile.  Copying the bash executable could be used to work around this, but there are still potential problems during package or system update for those who might use /bin/bash as a shell once they see it exists.

Overall I thought using /usr/bin/env had the best set of trade offs.

smlambert:
Grinders:
openjdk/xlinux:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/390

openjdk/aix:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/391

openjdk/osx:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/393

battleblow:
See also,

https://github.com/AdoptOpenJDK/stf/pull/67
https://github.com/AdoptOpenJDK/openjdk-systemtest/pull/291
https://github.com/eclipse/openj9/pull/7024

smlambert:
All 3 related PRs are merged, non-BSD grinders passed (to verify existing testing unaffected).  Still waiting on a machine to actually be able to run testing, but can merge now, as it is a benign change from perspective of existing testing.

battleblow:
Thanks Shelley!

I believe @gdams was going to set up FreeBSD builds and mentioned that there are some FreeBSD machines that should be available for this.  I haven't seen him on Slack for a while though.  Maybe he is out of office?

smlambert:
Created https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/945 to capture the request more formally.  :)

PR #1328, opened 2019-09-05
Add vm/jdwp to sanity.jck

JasonFengJ9:
**Add vm/jdwp to sanity.jck (disabled)**

Adding `vm/jdwp` to `sanity.jck` test bucket to increase sanity daily run coverage (disabled for now).
Also make deleting previous built `natives` not fail if there is no such folder. (not all `JCK` test material come with `natives` folder)

Reviewer: @ShelleyLambert 
FYI: @Mesbah-Alam @DanHeidinga @pshipton 

### DIFF
--- a/jck/build.xml
+++ b/jck/build.xml
@@ -70,7 +70,7 @@
 			<!-- jck materials exist, update jck materials if needed-->
 			<else>
 				<echo message="${env.JCK_ROOT} exists, deleting previously built natives..." />
-				<delete includeemptydirs="true">
+				<delete includeemptydirs="true" quiet="true" failonerror="false">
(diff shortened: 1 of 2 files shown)


pshipton:
I don't agree with adding jdwp. I think it's just going to fail and make it harder to spot other failures.

smlambert:
Suggest adding it with a `<disabled>` tag, so that if someone wished to run it in a Grinder they could (target: disabled.jck-runtime-vm-jdwp), but it would not get executed in a regular run.

JasonFengJ9:
Okay, mark this PR WIP until the `vm/jdwp` test passes consistently.

PR #1329, opened 2019-09-05
Remove -p option from umask as it doesn't work on AIX

Mesbah-Alam:
Related to https://github.com/eclipse/openj9/issues/6840

Background: 

On AIX 7.1
```
umask -p
umask: Improper mask.
```

We should remove `-p` option as it doesn't work with all AIX versions, and simply use umask to output the updated umask value (for debug/logging purposes). 

FYI @hangshao0 


Mesbah-Alam:
Tested : Test_grinder/job/Grinder_Advanced/1034/console

PR #1335, opened 2019-09-09
Added documentation to show how to export environment variables

mw918:
Previously, there was no documentation on how to export environment variables. Now, there has been a section written on this topic in the userGuide, with the corresponding images in the diagrams folder.


mw918:
@ShelleyLambert can you please take a look at this?

smlambert:
I am taking a look, but will take a bit of time to make sure it makes sense at this spot in this doc location, because its extremely detailed, which may be too narrow for this location.

R-Santhir:
@mw918 can you make it so that this section can be directly linked to

doc/userGuide.md
```diff
@@ -33,6 +33,111 @@ You can set up your own Jenkins-based test builds using the AdoptOpenJDK openjdk
...
+
+ ![build_list_target](/doc/diagrams/buildListTarget.jpg)
+ 
+11.	Scroll to the bottom and hit the Build button
+
+ ![build](/doc/diagrams/build.jpg)
+
+
+#### Method 2: Put it in the systemtest.mk file
+
```
llxia: I am not sure why we put into `systemtest.mk`? This will work if you are going to run system test only. The correct place should be `testEnv.mk`.

smlambert: agree, this is part of our design and about granularity... in playlist for specific test target, in group mk file for application to group only (not a common use case), in testEnv.mk for application to all targets.

mw918: Where is the testEnv.mk file? I searched through the entire folder and couldn't find it.

smlambert: its part of TestKitGen so after one sets up the test environment on their machine (behind the scenes, test material and scripts are getting fetched from various locations), they would look for it in the TestConfig folder (https://github.com/eclipse/openj9/blob/master/test/TestConfig/testEnv.mk)

somewhat related note, soon we will be moving TestKitGen to its own repo (separate from openj9 and openjdk-tests)...

mw918: Should I keep in the instructions that keeping the line in the specific tests (systemtest.mk, openjdk.mk, etc) can remain an option, then just add a third option of putting it in testEnv?

smlambert: I think just put the option for testEnv for now, as there is no real use case for the middle option for now and it will increase confusion if we give too many choices.

mw918: Where would be a good place to put it in testEnv.mk? Right now I have that as the last line, and it isn't doing anything.

mw918:
@ShelleyLambert and @llxia I made the changes you suggested. Can you please take another look at this?

R-Santhir:
@mw918 you still have method 2, @ShelleyLambert mentioned that it would be less confusing to have only what you have described as method 1 and method 3.

> I think just put the option for testEnv for now, as there is no real use case for the middle option for now and it will increase confusion if we give too many choices.

PR #1342, opened 2019-09-12
Update hotspot jtreg tests

sophia-guo:
1. Add openj9 native path
2. Remove testcase hotspot_all, which is the whole set of hotspot jtreg
test set
3. Select hotspot_native_sanity and hotspot_compiler hotspot jtreg test
sets
4. Add openj9_native_sanity, openj9_compiler and runtime_nestmate test
set against OpenJ9


sophia-guo:
@smlambert Thanks Shelley. Have update runtime_nestmate for all imples. 

Also add a new variable JVM_NATIVE_OPTIONS  so no need to set separate native_test and compiler test against different impls as different testcases.  Hence also rename hotspot|openj9_native_test to jvm_native_test, hotspot|openj9_compiler to jvm_compiler.

sophia-guo:
Grinder: 
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/461/
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/462/
..
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/467/

sophia-guo:
Will merge as  ibmruntimes/openj9-openjdk-jdk11#208 won't block the build. It will just fail the native tests.

PR #1349, opened 2019-09-17
exclude some system test subfolders in parallel mode

llxia:
We currently only run special.system in parallel mode and some system
test subfolders do not have any system test in special level. In order
to save machine resources, exclude the system test subfolders that do
not contain special level in parallel.


llxia:
Grinder for a non-parallel test: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1996/

llxia:
fyi @pshipton

PR #1360, opened 2019-09-19
Enhance jdk_custom to also support hotspot and langtools tests

sophia-guo:
1.jdk_custom can also be used for hotspot and langtools tests
2.Simplify native test scenario. If needed use can simply export
TESTIMAGE_PATH. No need to specify different native path.
3.Replace JTREG_TEST_DIR with JTREG_JDK_TEST_DIR, which is more clear as
jdk, hotspot and langtools tests are at same level.

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -352,14 +352,8 @@ def buildTest() {
 				echo 'Cannot run copyArtifacts from systemtest.getDependency. Skipping copyArtifacts...'
 			}
 			
-			if (fileExists('openjdkbinary/openjdk-test-image/jdk/jtreg/native')) {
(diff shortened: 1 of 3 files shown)


sophia-guo:
Tested with 8, 11 with impl=openj9, hotspot and test=jdk,langtools and hotspot.
https://ci.adoptopenjdk.net/view/Test_grinder
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2043/
...
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2038/

sophia-guo:
Documents about running jdk_custom and jdk/hotspot native tests will need updates.

sophia-guo:
@llxia could you please update https://github.com/eclipse/openj9/wiki/Reproducing-Test-Failures-Locally with Note: if tests need native libraries, please export test image path. For example,

export TESTIMAGE_PATH = yourestImageLocation/jdk/jtreg/native

llxia:
Wiki is updated

PR #1363, opened 2019-09-20
Add Grinder link in job description

llxia:
- Rerun in Grinder link only shows up for failed test build
- add check to show display Artifactory link once

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -471,6 +471,8 @@ def post(output_name) {
 					def pattern = "${env.WORKSPACE}/*_test_output.*"
 					uploadToArtifactory(pattern)
 				}
+
+				addGrinderLink()
 			}
 			//for performance test, archive regardless the build result
 			if (env.BUILD_LIST.startsWith('perf')) {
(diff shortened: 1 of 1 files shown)


llxia:
- Grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/597/

image


In the Rerun in Grinder link, it takes the default target and user can change it if it is needed.

- Rerun Grinder link will not show for passed build: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/596/

buildenv/jenkins/JenkinsfileBase
```diff
@@ -596,12 +598,30 @@ def uploadToArtifactory(pattern) {
...
+	def	sdkUrlCredentialOpt = params.CUSTOMIZED_SDK_URL_CREDENTIAL_ID ? "&CUSTOMIZED_SDK_URL_CREDENTIAL_ID=${params.CUSTOMIZED_SDK_URL_CREDENTIAL_ID}" : ""
+
+	def url = "${HUDSON_URL}/job/Grinder/parambuild/?" \
+	+ "JDK_VERSION=${params.JDK_VERSION}" \
+	+ "&JDK_IMPL=${params.JDK_IMPL}" \
+	+ "&BUILD_LIST=${params.BUILD_LIST}" \
+	+ "&JenkinsFile=${params.JenkinsFile}" \
+	+ "&TARGET=${params.TARGET}" \
+	+ "${sdkSourceOpt}${sdkUrlOpt}${sdkUrlCredentialOpt}${customTargetOpt}"
+	currentBuild.description += "<br><a href=${url}>Rerun in Grinder</a> Note: TARGET field can be modified to run a failed test"
```
smlambert: I would drop the "Note:" part, as its a tight space, and here are some suggested minor wording options:
Or "TARGET should be changed to run only the failed test target"
Or "TARGET can be changed to run only the failed test target"
Or "TARGET can be modified to run the failed test target" , ["rather than the top-level test target"]...

llxia: updated

image
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/610/

llxia:
> though I suppose if we start using Artifactory at Adopt, I would rather the insertion of that link be in this location rather than where I am guessing it lives now (in some openj9 script?). Can correct that bridge when we get to it.

yes, the test script is fully customizable. We just need to set params for artifactory info (i.e., `ARTIFACTORY_SERVER`, `ARTIFACTORY_REPO `, etc) in our test jobs. For example, for configuring openj9 artifactory, we set default values in [Test_Job_Auto_Gen](https://ci.eclipse.org/openj9/view/Test/job/Test_Job_Auto_Gen/build?delay=0sec) in Openj9 Jenkins

PR #1366, opened 2019-09-23
Add sleep time between child parallel jobs

llxia:
- we have the SDK downloading issue when all child parallel jobs are
downloading at the same time. Add sleep time to delay starting each
child parallel job
- delay for each child job = SLEEP_TIME_BETWEEN_PARALLEL * index
- Set the default value of SLEEP_TIME_BETWEEN_PARALLEL to 300 secs in
script, so auto builds do not need to set the value
- Will set the default value for SLEEP_TIME_BETWEEN_PARALLEL in Grinder to 0.
User can modify SLEEP_TIME_BETWEEN_PARALLEL value in Grinder if needed.

Issue: https://github.com/eclipse/openj9/issues/7179

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -153,7 +153,8 @@ def setupParallelEnv() {
 				repeatAmount = testSubDirs.size()
 				echo "repeatAmount is ${repeatAmount}, testSubDirs is ${testSubDirs}, running test in parallel mode"
 			}
-			
+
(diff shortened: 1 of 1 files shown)


llxia:
Grinder test internally with default value in the script

Output:

image

Grinder with SLEEP_TIME_BETWEEN_PARALLEL =0 

Output:
image

smlambert:
I want to wait before resorting to this workaround until I see that the infra/releng options to address the problem are investigated/tried.

smlambert:
@llxia - please close it if you can confirm its no longer needed.

llxia:
I think the plan is to set up an Artifactory at UNB to solve the network issue. I will close this PR.

PR #1367, opened 2019-09-24
Enable grinder tests with different jdk|impl|platform.

sophia-guo:
1. Add testpipeline.groovy to create and launch personal builds
2. Add general JobTemplate.groovy to define test jobs.

### DIFF
--- a/buildenv/jenkins/JobTemplate.groovy
+++ b/buildenv/jenkins/JobTemplate.groovy
@@ -1,0 +1,87 @@
+#!groovy
+
+/*
+ Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


smlambert:
Will merge this now, so we can continue to refine it after seeing it in 'action' in WIP tab in Jenkins.

Adding a reminder note here, on upcoming revisions of this, there is an opportunity to take advantage of what @llxia did for the Deepsmith work, to choose to run on same machine or not, and to diff the differences in the run.  This will allow a triager to run same test for openj9 and hotspot impls on same machine and directly compare differences in output (using Test Compare in trss, or downloading the diff file which is archived as an artifact).

PR #1370, opened 2019-09-25
Fix the DSL script path

sophia-guo:


sophia-guo:
First run:
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Platforms/21/

PR #1372, opened 2019-09-26
Set JenkinsFile as a parameter

llxia:
In nightly test builds (auto-generated), JenkinsFile is directly set as
Script Path in test job configure, not as a parameter. As a result,
Rerun in Ginder link cannot get JenkinsFile param value. This PR sets
JenkinsFile as a parameter in test job configure.


llxia:
The issue is opened for approving the script in OpenJ9 Jenkins https://bugs.eclipse.org/bugs/show_bug.cgi?id=551529

PR #1377, opened 2019-09-30
Automatically generate folders and listview.

sophia-guo:
Organize test jobs automatically by adding auto-generated test jobs to
specific folders and listviews.

### DIFF
--- a/buildenv/jenkins/JobTemplate.groovy
+++ b/buildenv/jenkins/JobTemplate.groovy
@@ -19,15 +19,22 @@
  */
 
 if (!binding.hasVariable('ARCH_OS')) ARCH_OS = "x86-64_linux"
-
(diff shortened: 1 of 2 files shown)


sophia-guo:
This is open to discuss. The idea is similarly to build jobs structure, which could also be used for pr build, nightly( sanity), weekly (extended).
Screen Shot 2019-09-30 at 2 17 02 PM

llxia:
>  if we proceed with this type of structure, does trss support it, or how much work to change to support?

The current TRSS only has limited support for the folder structure. With the folder structure, the url structure changes. For example, the job cannot be accessed with `<domain>/job/<jobName>`

We can look into the url pattern and update TRSS code. Or check Jenkins rest API to see if we can get the child build links directly. It would be nice to have a build to check what needs to be changed.

PR #1380, opened 2019-10-01
Add BumbleBench

piyush286:
https://github.com/AdoptOpenJDK/openjdk-tests/issues/1379

- Add SIMD and string tests

Signed-off-by: Piyush Gupta <piyush286@gmail.com>

### DIFF
--- a/perf/bumbleBench/build.xml
+++ b/perf/bumbleBench/build.xml
@@ -1,0 +1,52 @@
+<?xml version="1.0"?>
+
+<!--
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+
+<project name="BumbleBench Suite" default="build" basedir=".">
+	<taskdef resource="net/sf/antcontrib/antlib.xml" />
+	<description>BumbleBench Suite</description>
+
+	<!-- Set global properties for this build -->
+	<property name="DEST" value="${BUILD_ROOT}/perf/bumbleBench" />
+	<property name="SRC" location="." />
+	
+	<target name="init">
+		<mkdir dir="${DEST}" />
+	</target>
+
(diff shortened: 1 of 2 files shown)


piyush286:
Requesting review from @llxia and @smlambert. Thanks!

I've just added 4 tests to get started. I'll be adding more tests and enabling more platforms in future PRs. For now, I've marked all of them as `special` since it is easier to run all of them using `special.perf` for functionality check for this PR. I was wondering whether there is any way to run all the tests under one build file. I didn't want to use `sanity` or `extended` since that would also launch other perf tests marked with those levels. I can change the levels to whatever you suggest.

Sanity Job with `special.perf`. All tests are passing.
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/718/console

```
16:01:44  TEST TARGETS SUMMARY
16:01:44  ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
16:01:44  PASSED test targets:
16:01:44  	bumbleBench-SIMDDoubleMaxMinBench_0
16:01:44  	bumbleBench-ExactBench_0
16:01:44  	bumbleBench-StringConversionBench_0
16:01:44  	bumbleBench-StringHashBench_0
16:01:44  	bumbleBench-StringIndexOfBench_0
16:01:44  
16:01:44  TOTAL: 5   EXECUTED: 5   PASSED: 5   FAILED: 0   DISABLED: 0   SKIPPED: 0
```

smlambert:
special is fine for now, we can reorganize at a later time should we want to.  

To run all tests in one build, since you can control what test targets are generated, what test material is built (and subsequently run) via BUILD_LIST, by merely setting BUILD_LIST=perf/bumblebench as you did in your Grinder, you run all of the tests... so in your case, because you set BUILD_LIST=perf/bumblebench, if you had tagged the tests as extended, and run TARGET=extended, it would have run all the extended tests within that directory.  In that same vein, 
if your BUILD_LIST=perf, and TARGET=extended, it would have picked up all the other tests tagged extended in the entire perf dir.

Grinder to show hotspot:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/726

Any specific reason to restrict these targets to jdk8?  (and to xlinux?)

piyush286:
Thanks Shelley for the tip! I'll try the suggestion for the build list and target.

> Any specific reason to restrict these targets to jdk8? (and to xlinux?)

I was thinking of adding all tests for BumbleBench starting with x86 Linux and then finally enabling them on the other platforms/jdk versions. Anyway, let me just test and enable them right now.

smlambert:
Eventually we should also update [perf doc for bumblebench](https://github.com/AdoptOpenJDK/openjdk-tests/tree/master/perf#bumblebench) to a link that describes these microbenches.

PR #1398, opened 2019-10-18
skip remaining ipv6 tests on macOS and skip TestZoneInfo310.java on all

gdams:
skips `java/net/MulticastSocket/SetLoopbackMode.java` on macOS
skips `java/net/MulticastSocket/Test.java` on macOS

skips `sun/util/calendar/zi/TestZoneInfo310.java` on everything (already skipped in the j9 problemList)


karianna:
Do these tests fail upstream?

gdams:
the two mac ones are because of machine config (we don't have ipv6 so have to skip). The other one does fail upstream

sophia-guo:
Those two jdk/net/MulticastSocket tests are actually testing IPv4 and IPv6 if only IPv6 is available. They should have no issue for cloud machines without ipv6.  They are enabled on mac with openj9 and passed. Also I have tried with grinder and works fine. 
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/726/testReport/
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/727/testReport/
https://ci.adoptopenjdk.net/view/Test_openjdk/job/Test_openjdk8_j9_sanity.openjdk_x86-64_mac/62/testReport/

@gdams do you have specific cloudy machine without IPV6? I'd like to try test on that. If it doesn't work we need to exclude the tests on openj9 and other jdk version too.

PR #1401, opened 2019-10-18
Reorganize JCK targets

Mesbah-Alam:
This PR reorganizes the JCK targets by breaking down large category of tests (vm, xml_schema, api, etc.)  into targets for each sub-category (sub-folder) in their own playlist.xml file as follows: 

- runtime/xml_schema - 4 targets 
- runtime/vm tests - 13 targets 
- runtime/api tests - 41 targets 
- runtime/lang tests - 21 targets 

- compiler/api tests - 5  targets 
- compiler/lang tests - 21 targets 

- devtools/java2schema - 2 targets 
- devtools/jaxws - 1 target 
- devtools/schema2java - 2 targets 
- devtools/schema_bind - 8 targets 
- devtools/xml_schema - 4 targets 

### DIFF
--- a/jck/compiler.lang/playlist.xml
+++ b/jck/compiler.lang/playlist.xml
@@ -15,7 +15,7 @@
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../TestConfig/playlist.xsd">
 	<include>../jck.mk</include>
 	<test>
-		<testCaseName>jck-runtime-lang-BINC</testCaseName>
+		<testCaseName>jck-compiler-lang-BINC</testCaseName>
 		<variations>
 			<variation>NoOptions</variation>
 		</variations>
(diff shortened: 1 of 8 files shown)


Mesbah-Alam:
Testing jck.sanity here: Test_grinder/job/Grinder/2801/console

Mesbah-Alam:
Testing extended.jck here: Test_grinder/job/Grinder/2813/console

llxia:
Grinder 2813  is aborted due to timeout (10hrs). Current sanity.jck only takes about 5hrs on Linux. It seems we have too many tests in sanity.

jck/compiler.lang/playlist.xml
```diff
@@ -0,0 +1,522 @@
...
+		</levels>
+		<groups>
+			<group>jck</group>
+		</groups>
+		<subsets>
+			<subset>11+</subset>
+		</subsets>
+	</test>
+	<test>
+		<testCaseName>jck-compiler-lang-CLSS-AIX</testCaseName>
```
llxia: Do we need to separate the test into `jck-compiler-lang-CLSS-NON-AIX` and `jck-compiler-lang-CLSS-AIX`? In the original file, we only have one test `jck-compiler-lang-CLSS`. And the subset is `11` in the original file, not `11+`. Not sure which one is correct.

https://github.com/AdoptOpenJDK/openjdk-tests/blob/a11b0bb8021706b61020f97c98dc0a420b89836f/jck/runtime.lang/playlist.xml#L543

jck/compiler.lang/playlist.xml
```diff
@@ -0,0 +1,522 @@
...
+	-test=Jck -test-args=$(Q)tests=lang/CONV,jckRoot=$(JCK_ROOT),jckversion=$(JCK_VERSION),testsuite=COMPILER$(Q); \
+	$(TEST_STATUS)</command>
+		<levels>
+			<level>extended</level>
+		</levels>
+		<groups>
+			<group>jck</group>
+		</groups>
+		<subsets>
+			<subset>12+</subset>
```
llxia: The original value is `11` https://github.com/AdoptOpenJDK/openjdk-tests/blob/a11b0bb8021706b61020f97c98dc0a420b89836f/jck/runtime.lang/playlist.xml#L585

llxia:
I noticed `subset` is set differently compare to the original subset. Some with higher value, some with lower value and some are missing. I listed some of them in above comments. What is the expected behavior?

Mesbah-Alam:
The only subset values changed should be similar to `<subset>11</subset>` being `<subset>11+</subset>` etc  - as likely the `11` value was added when we did not have jck 12,13,14 added and the test exists in those too and should be run. 

I will go over the subset values and double check .

jck/compiler.lang/playlist.xml
Mesbah-Alam: This is a copy from the way the test was broken in jck-runtime-lang-CLSS https://github.com/AdoptOpenJDK/openjdk-tests/blob/a11b0bb8021706b61020f97c98dc0a420b89836f/jck/runtime.lang/playlist.xml#L61

llxia: Is there a particular reason for doing the same for `jck-compiler-lang-CLSS` tests? From what I can see, `jck-compiler-lang-CLSS` did not get the chance to be executed on AIX due to build timeout. 

As this PR is about reorganizing jck targets, I would perfer not mix with other changes. This PR already contains lots of changes. If break `jck-compiler-lang-CLSS` into separate platforms is needed, we can do it another PR.

Mesbah-Alam: I agree. I've collapsed the two target to only one - `jck-compiler-lang-CONV`.

llxia: Thanks

jck/compiler.lang/playlist.xml
Mesbah-Alam: Updated all subset values in compiler tests to `11+`

Mesbah-Alam:
Did a comparison between the set of targets run as sanity currently, and the targets tagged as sanity in this PR - and the set is exactly the same apart from one target - `jck-compiler-api-signaturetest` (which takes ~10 m to run). 

However, the sanity.jck here: Test_grinder/job/Grinder/2805/ - runs for `10h` and times out. Investigating inside the build - we see that one of the compiler targets - `jck-runtime-lang-EXPR_0` took ~ 5h30m to complete. The same target takes ~ `20m` to complete in the regular sanity.jck runs (Test_openjdk11_j9_sanity.jck_s390x_linux_Nightly/135/tapResults/). 

The only difference is that in the Grinder, we are using an s390 jdk11 sdk from Adopt whereas the internal build is using an internal SDK. 

Started a new sanity.jck grinder with internal sdk here: Test_grinder/job/Grinder/2830/console

Mesbah-Alam:
Started a new sanity.jck grinder  : Test_grinder/job/Grinder/2830/console

Mesbah-Alam:
Started new extended.jck grinder: job/Grinder/2833/console

Mesbah-Alam:
FYI.. @llxia, sanity.jck Test_grinder/job/Grinder/2830/ - finishes in `1h 45m`.

jck/runtime.vm/playlist.xml
```diff
@@ -68,7 +48,7 @@
 	-test=Jck -test-args=$(Q)tests=vm/jdwp,jckRoot=$(JCK_ROOT),jckversion=$(JCK_VERSION),testsuite=RUNTIME$(Q); \
 	$(TEST_STATUS)</command>
 		<levels>
-			<level>sanity</level>
+			 /level>
```
llxia: This line is incompleted. Please fix it.

jck/devtools.java2schema/playlist.xml
```diff
@@ -0,0 +1,64 @@
...
+	-results-root=$(REPORTDIR)  \
+	-test=Jck -test-args=$(Q)tests=java2schema/CustomizedMapping,jckRoot=$(JCK_ROOT),jckversion=$(JCK_VERSION),testsuite=DEVTOOLS$(Q); \
+	$(TEST_STATUS)</command>
+		<levels>
+			<level>extended</level>
+		</levels>
+		<groups>
+			<group>jck</group>
+		</groups>
+			<subsets>
```
llxia: Please fix the indentation. Same for the rest of the files.

jck/runtime.vm/playlist.xml
Mesbah-Alam: Fixed! Thankns for the catch.

jck/devtools.java2schema/playlist.xml
Mesbah-Alam: Fixed! Thanks!

Mesbah-Alam:
Thanks for the detailed review, @llxia

PR #1403, opened 2019-10-21
Set correct test repo for aarch64 builds

johnoliver:


smlambert:
Some questions : is the repo permanently jdk-aarch64-jdk8u ?  or temporarily and will use jdk-jdk8u at some future time?  Out of curiosity, what is difference between the two?  

for future versions aarch64 jdk11, is it using common mirror?

I think you are able to contain the change to the build.xml file by checking PLATFORM/JDK_VERSION there, then overriding the regression repo, which would make it easier to find and remove later.

gdams:
> Some questions : is the repo permanently jdk-aarch64-jdk8u ? or temporarily and will use jdk-jdk8u at some future time? Out of curiosity, what is difference between the two?

It's a separate port that we will always remain where it is so this is permanent.

> for future versions aarch64 jdk11, is it using common mirror?

Yes the port got merged with upstream in JDK9 so only 8 needs this.

smlambert:
Grinder to verify looks good: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/863/

PR #1414, opened 2019-10-25
Add Jenkins files for Linux s390

vsebe:
Add Jenkins files for 31 bits and Large Heap platforms.

Signed-off-by: Violeta Sebe <vsebe@ca.ibm.com>

### DIFF
--- a/buildenv/jenkins/openjdk_s390_linux
+++ b/buildenv/jenkins/openjdk_s390_linux
@@ -1,0 +1,20 @@
+#!groovy
+
+
+if (params.DOCKER_REQUIRED) {
+    LABEL='ci.role.test&&hw.arch.s390x&&sw.os.linux&&sw.tool.docker&&hw.bits.32'
+} else {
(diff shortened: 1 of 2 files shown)


vsebe:
@llxia please review, thank you!
Please see internal pipelines.

llxia:
Could we follow a similar naming conversion and have`openjdk_s390x-64_linux_xl` and `openjdk_s390x-31_linux`?

vsebe:
s390x usually designates a 64bits architecture and s390 a 31 bits architecture but generally s390 refers to Linux on Z. Thus s390-31, s390-64 would probably be better. 
In order to be consistent, the existing script - [openjdk_s390x_linux](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/openjdk_s390x_linux) - would also have to follow the naming converntion. @llxia - What's the impact on changing it?

llxia:
Talked with @vsebe , I agree that `openjdk_s390x_linux` also needs to follow the naming convention. But changing the existing script means we need to regenerate the test jobs and change upstream scripts. We will keep the name as it is for now.

PR #1427, opened 2019-11-06
Set JTREG concurrency based on cores and memory size

jerboaa:
Upstream OpenJDK test framework, roughly, sets it to
min(NPROCS/2, MEM_IN_GB/2). Set it to a similar value
so as to avoid test reliability issues due to memory
limits.

### DIFF
--- a/openjdk/openjdk.mk
+++ b/openjdk/openjdk.mk
@@ -12,21 +12,46 @@
 # limitations under the License.
 ##############################################################################
 NPROCS:=1
+# Memory size in MB
+MEMORY_SIZE:=1024
+
 OS:=$(shell uname -s)
 
 ifeq ($(OS),Linux)
 	NPROCS:=$(shell grep -c ^processor /proc/cpuinfo)
+	MEMORY_SIZE:=$(shell \
+		expr `cat /proc/meminfo | grep MemTotal | awk '{print $$2}'` / 1024 \
+		)
 endif
 ifeq ($(OS),Darwin)
 	NPROCS:=$(shell sysctl -n hw.ncpu)
+	MEMORY_SIZE:=$(shell expr `sysctl -n hw.memsize` / 1024 / 1024)
 endif
 ifeq ($(OS),FreeBSD)
 	NPROCS:=$(shell sysctl -n hw.ncpu)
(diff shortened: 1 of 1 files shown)


jerboaa:
Grinder on the big Aarch64 box with this:
https://ci.adoptopenjdk.net/job/Grinder_Sandbox/884/console

It's rather conservative but certainly better than a concurrency value of NPROCS. I'll do some more testing on other boxes and may amend it. Hence, the WIP suffix.

jerboaa:
@smlambert Thoughts?

smlambert:
Thanks @jerboaa - have been meaning to look at what we were currently doing, based on the troubles seen running on the 'big box' / test-packet-armv8-ubuntu-16-04 reported in https://github.com/AdoptOpenJDK/openjdk-tests/issues/1203 and seen in a bunch of the recent release runs (seen more frequently now that there are less aarch64 machines in the pool.

Agree its an improvement, will be good to see it run through the other platforms to make sure it doesn't hose anything, thanks for looking at this !

jerboaa:
Windows sandbox grinder with this:
https://ci.adoptopenjdk.net/job/Grinder_Sandbox/886/

jerboaa:
Another run on the 96 core aarch64 box (on upstream aarch64 JDK 11.0.6+1 EA build):
https://ci.adoptopenjdk.net/job/Grinder_Sandbox/889/

Contrast this to flaky runs earlier on the same JDK: 11.0.6+1 EA. For example:
https://ci.adoptopenjdk.net/view/Test_upstream/job/Test_openjdk11_hs_sanity.openjdk_aarch64_linux_upstream/23/consoleFull

jerboaa:
This seems to build OK on Windows/Linux/Mac. I'm removing the `WIP` prefix.

jerboaa:
@smlambert More testing of this completed. Make file seems OK. I don't think any of the failures are related to this patch. Mac/Win (no JRE 8 got downloaded, hence failure). Aarch64 linux on the big box now passes (or is significantly better than before in terms of reliability). ppcle 64 bit linux failures for JDK 13 also seem unrelated.

I'd appreciate somebody reviewing this! Thanks!

jerboaa:
Thanks, @smlambert!

PR #733, opened 2018-12-07
Add additional information job description

llxia:
Following info is added:
- started by
- TARGET
- LABEL
- NODE

Fixes: #687

Signed-off-by: lanxia <lan_xia@ca.ibm.com>

Issue #687, opened 2018-11-15
Display additional info in Grinder job page


llxia:
With the increasing popularity of Grinder, it would be great if we can also display some additional information for users to find their own builds among 20-30 builds. Currently, we only display machine name. Additional information can be:
-  Started by user
- TARGET

<img width="230" alt="screen shot 2018-11-15 at 11 31 44 am" src="https://user-images.githubusercontent.com/19273206/48567000-249b5e80-e8ca-11e8-8513-72e4e77b450d.png">

PR #1431, opened 2019-11-07
Override LABEL variable when params.LABEL is passed

AdamBrousseau:
- When LABEL is passed as a param, the build
  description would still use LABEL rather
  than params.LABEL. Overiding the variable
  LABEL with params.LABEL will ensure the
  right LABEL is used for the description.
- Minor formatting and whitespace fixes

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -508,7 +508,7 @@ def addJobDescription() {
 	def description = (currentBuild.description) ? currentBuild.description + "<br>" : ""
 	currentBuild.description = description \
 		+ "TARGET: ${params.TARGET}<br/>" \
-		+ "LABEL: <a href=${JENKINS_URL}label/${LABEL}>${LABEL}</a><br/>" \
+		+ "<a href=${JENKINS_URL}label/${LABEL}>${LABEL}</a><br/>" \
 		+ "<a href=${JENKINS_URL}computer/${NODE_NAME}>${NODE_NAME}</a>"
 }
 
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -50,9 +50,9 @@ def setupEnv() {
 	env.KEEP_REPORTDIR = params.KEEP_REPORTDIR ? params.KEEP_REPORTDIR : ''
 	SDK_RESOURCE = params.SDK_RESOURCE ? params.SDK_RESOURCE : "upstream"
 	env.AUTO_DETECT = params.AUTO_DETECT
-	
+
 	ITERATIONS = params.ITERATIONS ? "${params.ITERATIONS}".toInteger() : 1
-	
+
 	if (params.JDK_IMPL) {
 		env.JDK_IMPL = params.JDK_IMPL
 	} else if (params.JVM_VERSION) {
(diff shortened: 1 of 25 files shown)


smlambert:
I am curious to hear if you hit a case where a wrong label was used?

AdamBrousseau:
For example
https://ci.adoptopenjdk.net/job/Grinder/1040/
I passed `LABEL:test-godaddy-win2016-x64-2` as a parameter but the description shows
```
TARGET: testOSMXBeanLocal_0
LABEL: ci.role.test&&hw.arch.x86&&sw.os.windows
test-godaddy-win2016-x64-2
```
So while in this case it is "ok" because I still get the NODE_NAME in the description, it is technically wrong because LABEL is incorrect. If I passed `LABEL:fooboo` and got the same machine above, I would expect the description to be
```
TARGET: testOSMXBeanLocal_0
LABEL: fooboo
test-godaddy-win2016-x64-2
```
But as is it will still print whatever the default is in the top level JenkinsFile.
It's a minor nit but still a bug at least imo.

smlambert:
Ah, I understand, and agree it may confuse if we call it LABEL, but perhaps something else, like DESC as funnily enough, just tonight, Lan and I used that bug as a feature, as it allowed us to click on it to get quickly to the list of other machines that we could rerun a test on... 

Screen Shot 2019-11-08 at 12 29 45 AM

I am worried that if we make the change you are suggesting:
1) it provides redundant information in the case when someone supplies the hostname
2) we lose the feature of being able to get quickly to the list of other machines to run on

So from a UX perspective, it seems less useful to change it, than to leave it as is, but I will leave this open to discussion.

AdamBrousseau:
1. Agree. I wonder how often does a user change the LABEL param? 
2. Agree, but only when LABEL is passed as a param. In any other build this wouldn't be the case. It would only be the Grinder use case where someone passes it.
I guess the main question is what was the original purpose or use case this was added for? Looks like #687 and #733 is the relevant issues but I don't see _why_ LABEL was added.
We could change the LABEL wording as you say but it might be more accurate to call it DEFAULT LABEL but ideally something shorter?

smlambert:
Or not call it anything, but add HOST: in front of hostname.
Those links are there to help testers/devs triage and quick views of things, they were added to serve us as we see fit.  I think its more important to have the quick link to other machines, than seeing what label was used (which I can see when I hit rebuild).  

The most common use case is to pass in a specific hostname as a label to rerun a failing test on a particular machine.  I do not know of any other common use cases.

AdamBrousseau:
Ok, so remove `LABEL` text in front of the default LABEL value. I think `HOST` was not added since the Adopt machines names are too long to fit on one line (https://github.com/AdoptOpenJDK/openjdk-tests/pull/733#issuecomment-445357414)

llxia:
I merged PR https://github.com/AdoptOpenJDK/openjdk-tests/pull/1434. Please rebase and update these PPC files too. Thanks

AdamBrousseau:
Rebased on #1434 and #1446. Removed `LABEL` text. Maybe hold off on this until a decision is made on #1445.

karianna:
@AdamBrousseau sorry will need a rebase again.

AdamBrousseau:
I believe this work has been inadvertently completed by #1521

PR #1432, opened 2019-11-07
Support running openjdk jtreg on zOS

andrew-m-leonard:
Enable openjdk sanity jtreg testing
- Enable zOS jtreg.tar.gz binary download via customized URL (temporary until zOS jtreg build available)
- Utilize zOS tar options
- getOpenjdk using git@ protocol supported on zOS
- use -othervm jtreg mode on zOS (-agentvm currently not working)

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -234,7 +234,7 @@ getBinaryOpenjdk()
 			elif [[ "$jar_dir_name"  =~ native-test-libs*  &&  "$jar_dir_name" != "native-test-libs" ]]; then
 				mv $jar_dir_name native-test-libs
 			#The following only needed if openj9 has a different image name convention
-			elif [[ "$jar_dir_name" != "j2sdk-image"  &&  "$jar_dir_name" != "native-test-libs" ]]; then
+			elif [[ "$jar_dir_name" != "j2sdk-image"  &&  "$jar_dir_name" != "native-test-libs" && "$jar_dir_name" != "jtreg" ]]; then
 				mv $jar_dir_name j2sdk-image
 			fi
 		done
--- a/openjdk/build.xml
+++ b/openjdk/build.xml
@@ -30,7 +30,7 @@
 			<property name="gitReleaseTag" value="--branch ${env.RELEASE_TAG}"/>
 		</then>
 		<else>
-			<property name="gitReleaseTag" value=""/>
+		        <property name="gitReleaseTag" value=""/>
 		</else>
 	</if>
 	
@@ -95,7 +95,7 @@
                                 <if>
                                     <contains string="${PLATFORM}" substring="zos"/>
                                     <then>
-                                        <property name="regressionRepo" value="git@github.com:ibmruntimes/${jdkName}.git" />
+                                        <property name="regressionRepo" value="git@github.ibm.com:runtimes/openj9-openjdk-jdk11-zos.git" />
                                     </then>
                                     <else>
 				        <property name="regressionRepo" value="https://github.com/ibmruntimes/${jdkName}.git" />
--- a/openjdk/build.xml
+++ b/openjdk/build.xml
@@ -86,16 +86,31 @@
 				<if>
 					<equals arg1="${env.ORIGIN_JDK_VERSION}" arg2="next" />
 					<then>
-						<property name="jdkName" value="openj9-openjdk-jdk" />
+                                            <if>
+                                                <contains string="${PLATFORM}" substring="zos"/>
+                                                <then>
+                                                </then>
+                                                <else>
+    						    <property name="jdkName" value="openj9-openjdk-jdk" />
+                                                </else>
+                                            </if>
 					</then>					
 					<else>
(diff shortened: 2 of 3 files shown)


openjdk/build.xml
```diff
@@ -75,7 +92,15 @@
 						<property name="jdkName" value="openj9-openjdk-jdk${JDK_VERSION}" />
 					</else>
 				</if>
-				<property name="regressionRepo" value="https://github.com/ibmruntimes/${jdkName}.git" />
+                                <if>
+                                    <contains string="${PLATFORM}" substring="zos"/>
+                                    <then>
+                                        <property name="regressionRepo" value="git@github.com:ibmruntimes/${jdkName}.git" />
```
smlambert: You don't need if platform zos here, you can just use ssh (https://github.com/smlambert/openjdk-tests/blob/zos/openjdk/build.xml#L83) for all platforms, as I believe it does work for all of them.

get.sh
```diff
@@ -234,7 +234,7 @@ getBinaryOpenjdk()
 			elif [[ "$jar_dir_name"  =~ native-test-libs*  &&  "$jar_dir_name" != "native-test-libs" ]]; then
 				mv $jar_dir_name native-test-libs
 			#The following only needed if openj9 has a different image name convention
-			elif [[ "$jar_dir_name" != "j2sdk-image"  &&  "$jar_dir_name" != "native-test-libs" ]]; then
+			elif [[ "$jar_dir_name" != "j2sdk-image"  &&  "$jar_dir_name" != "native-test-libs" && "$jar_dir_name" != "jtreg" ]]; then
```
smlambert: I am not a fan of this temporary approach (to change get.sh and upload via overloading CUSTOMIZED_SDK_URL), when I think the real/right solution is so close.  I was assuming you would not put in a PR but wait to get the real jtreg for zos build hosted in the open.  

Is there a reason to push this temporary work in now versus waiting the few days it will take for the final solution?

openjdk/build.xml
```diff
@@ -41,12 +41,29 @@
 				<available file="jtreg-4.2.0-tip.tar.gz" />
 			</not>
 			<then>
-				<exec executable="wget" failonerror="true">
-					<arg line="-q --no-check-certificate https://ci.adoptopenjdk.net/job/jtreg/lastSuccessfulBuild/artifact/jtreg-4.2.0-tip.tar.gz" />
-				</exec>
-				<exec executable="tar" failonerror="true">
-					<arg line="xf jtreg-4.2.0-tip.tar.gz -C ${DEST}" />
-				</exec>
+                            <if>
```
smlambert: Then this would end up looking more like what is in https://github.com/smlambert/openjdk-tests/blob/zos/openjdk/build.xml#L37-L57 once a patched jtreg executable can be available.

openjdk/openjdk.mk
```diff
@@ -28,7 +28,11 @@ ifeq ($(CYGWIN),1)
 endif
 EXTRA_JTREG_OPTIONS += -concurrency:$(NPROCS)
 
-JTREG_BASIC_OPTIONS += -agentvm
+ifeq ($(OS),OS/390)
+    JTREG_BASIC_OPTIONS += -othervm
```
smlambert: Could also add a comment if this is a temporary case due to -agentvm option not working on zos (that way there may be a chance we see it and change it later when it is working (curiosity, is there an issue captured somewhere describing what blocks it from working?)

openjdk/build.xml
andrew-m-leonard: I wasn't sure if anyone using git@ to build/test would need a GPG key setup? which not every user may do...?

get.sh
andrew-m-leonard: I don't think an open jtreg for zOS solution is days away, I suspect more like weeks....

openjdk/build.xml
andrew-m-leonard: yes

openjdk/openjdk.mk
andrew-m-leonard: we have no idea what the issue is yet, whether fundamental or a bug as there is no zOS agent diagnostics... i'm hoping a bug and once we fix all the sanity failing tests then maybe one of them fixes this....

openjdk/build.xml
AdamBrousseau: In order to use ssh you need an ssh key setup or in the case of Jenkins nodes an ssh-agent wrapped around whatever calls this build.xml. Further, I thought on z/os we needed to use the git:// protocol?

get.sh
AdamBrousseau: Is setting up a parameter for `jtreg_url` a bad idea, even if it's temporary and can be removed after said weeks? I know we don't really want more params. Is the end goal to host it at Adopt and have a simple switch between z/os and everything else?

andrew-m-leonard:
closing this, since considering different approach

Issue #1440, opened 2019-11-12
Build/test pipeline should use the same commits for each job


adam-thorpe:
jdk/nio/zipfs/DirectoryStreamTests.java
jdk/nio/zipfs/InvalidZipHeaderTests.java
jdk/nio/zipfs/jarfs/JFSTester.java

Failing on JDK11 on almost all platforms, both hotspot and openj9

These three tests seem to failing due to a recently backported change from JDK13:
https://github.com/ibmruntimes/openj9-openjdk-jdk11/commit/d98b2eb726599a50b6355ee4ee8016d172a51f1a
They seem to have problems when reading jar files (interacting with META-INF). As far as I'm aware this problem doesn't seem to be persistent on JDK13 so am not sure the cause here.

eg.
```
test JFSTester.testWalk(): failure
java.lang.AssertionError: Sets differ: expected [This is leaf 2.
, This is leaf 1.
, This is version 9 leaf 5.
, This is version 9 leaf 4.
, This is version 9 leaf 3.
, This is version 9 leaf 6.
] but got [This is version 9 leaf 4.
, This is leaf 1.
, This is version 9 leaf 3.
, This is leaf 2.
]
	at org.testng.Assert.fail(Assert.java:94)
	at org.testng.Assert.assertEquals(Assert.java:808)
	at org.testng.Assert.assertEquals(Assert.java:786)
	at JFSTester.testWalk(JFSTester.java:110)
```

sophia-guo:
Seems like those failures happened recently  https://ci.adoptopenjdk.net/view/Test_openjdk/job/Test_openjdk11_hs_sanity.openjdk_x86-64_linux/. May not be related with the seven months ago change ibmruntimes/openj9-openjdk-jdk11@d98b2eb. 

Test_upstream is fine https://ci.adoptopenjdk.net/view/Test_upstream/job/Test_openjdk11_hs_sanity.system_x86-64_linux_upstream/ though may not be the latest. If latest test_upstream have similar issues we can open the issue in JBS. @jerboaa

adam-thorpe:
@sophia-guo because this change was created in jdk next seven months ago and then backported to jdk11, its marked as the same date (both in mercurial and github).
Merged in openj9 2 days ago: https://github.com/ibmruntimes/openj9-openjdk-jdk11/commit/3cc03bc6fbbab6bbd54eecb860e007f44e14656e#diff-a78bc512ab9028e23111986846fd86bd

[...]

PR #1447, opened 2019-11-18
Set 'dev' as AdoptOpenJDK/openjdk-jdk{jdkversion}u default branch

sophia-guo:
AdoptOpenjdk nightly hotspot build is using 'dev' branch. This is to
make sure that the source and test consistent
 
### DIFF
--- a/openjdk/build.xml
+++ b/openjdk/build.xml
@@ -24,16 +24,7 @@
 	<property name="DEST" value="${BUILD_ROOT}/openjdk" />
 	<property name="src" location="." />
 	<property environment="env" />
-	<if>
-		<isset property="env.RELEASE_TAG"/>
-		<then>
-			<property name="gitReleaseTag" value="--branch ${env.RELEASE_TAG}"/>
-		</then>
-		<else>
-			<property name="gitReleaseTag" value=""/>
-		</else>
-	</if>
-	
+
 	<target name="getJtreg">
 		<mkdir dir="${DEST}"/>
 		<if>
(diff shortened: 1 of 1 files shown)


sophia-guo:
AdoptOpenJDK nightly hotspot build use dev as default branch.  https://github.com/AdoptOpenJDK/openjdk-build/blob/master/git-hg/README.mdPipelinejobs parameter description
https://github.com/AdoptOpenJDK/openjdk-build/blob/master/pipelines/jobs/pipeline_job_template.groovy#L59

The current test build is using master as default branch which is newer than 'dev' branch. The inconsistency could cause failures 
https://ci.adoptopenjdk.net/view/Test_openjdk/job/Test_openjdk11_hs_sanity.openjdk_x86-64_linux/85/#showFailuresLink

See #1440 
And

sophia-guo:
Grinder:
java/util
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/920/ 

jdk/nio
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/918/

openjdk/build.xml
```diff
@@ -93,13 +84,30 @@
...
 		</if>
-
-		<echo message="git clone --depth 1 -q ${gitReleaseTag} ${regressionRepo}" />
-		<exec executable="git" failonerror="false">
-			<arg line="clone --depth 1 -q ${gitReleaseTag} ${regressionRepo}" />
-		</exec>
+		<if>
+			<isset property="env.RELEASE_TAG"/>
+			<then>
+				<exec executable="git" failonerror="false">
```
smlambert: Hmm, wondering what are the consequences on not failing if there is an error to clone this repo?  I know it predates your changes, but should we change to failonerror="true" (and same for git clone command below)?

sophia-guo: Default failonerror is false. Definitely we should explicitly set it as true if git clone exits with error. Updated.
FYI @Mesbah-Alam should system-tests do the same update?

PR #1448, opened 2019-11-18
Support checkout specific openjdk commit

sophia-guo:
There is possibility that PR is merged into openjdk repo at the gap between jdk build and jdk test build, which makes source and test code inconsistent and might cause test failure.

This PR makes sure test build can checkout specific commit. Hence be consistent with source build.


sophia-guo:
Grinder:
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/915/

To make it work in pipeline build jobs 'OPENJDK_SHA' is needed to pass in by upstream( build jobs)

sophia-guo:
Grinder: 936-940
https://ci.adoptopenjdk.net/view/work%20in%20progress/job/Grinder_Sandbox/936/

PR #1450, opened 2019-11-19
Limit -nativepath:{location of the jdk test native library} to jdk10+

sophia-guo:
Explicitly Limit -nativepath options to jdk10+ as "make test-image" to
generate the test jdk libraries only works for jdk10+.

### DIFF
--- a/openjdk/openjdk.mk
+++ b/openjdk/openjdk.mk
@@ -102,16 +102,19 @@ endif
 JDK_NATIVE_OPTIONS :=
 JVM_NATIVE_OPTIONS :=
 CUSTOM_NATIVE_OPTIONS :=
-ifdef TESTIMAGE_PATH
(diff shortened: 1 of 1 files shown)


sophia-guo:
AdoptOpenJDK build don't build test images with openjdk8. However openj9 do have test images build for its own functional tests with same directory structure.  https://github.com/eclipse/openj9/pull/7733 .  So we need explicitly the feature to 10+. 

Grinder: internal Test_grinder/job/Grinder_Sandbox/36-41

sophia-guo:
FYI @vsebe. This should fix the error

```
01:08:08  Error: The -nativepath path does not exist: /home/jenkins/workspace/Test_openjdk8_j9_sanity.openjdk_x86-64_linux_xl_Nightly/openjdkbinary/openjdk-test-image/jdk/jtreg/native
```
on JDK8

PR #1452, opened 2019-11-20
Fix MBCS_Tests.Compiler, annotation and file for AIX

### LINKED ISSUE #1451: Fix MBCS_Tests.Compiler, annotation and file for AIX
I found three testcases failure(Compiler Ja_JP, annotation Ja_JP and file) on an AIX.

Because the sed command in Linux toolbox doesn't work, Compiler and annotation fail.
Also I can see a hung in the file case while running a grinder.  yes command in the test.sh may cause a hung issue.

KKES335:
Fixes #1451 
Signed-off-by: Kentaro Kuramochi kkes335@jp.ibm.com

### DIFF
--- a/functional/MBCS_Tests/Compiler/test.sh
+++ b/functional/MBCS_Tests/Compiler/test.sh
@@ -32,11 +32,20 @@ cd tmp
 TS=`${JAVA_BIN}/java CheckValidData "${TEST_STRING}"`
 
 JAVAFILE=${TS}.java
-sed "s/TEST_STRING/${TS}/g" ${BASE}/class_org.java > ${JAVAFILE}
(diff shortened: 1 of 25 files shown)


KKES335:
About Compiler and annotation, here document is used instead of sed command.
Correct method have been replaced in DateFormatTest.java in  i18n. 
Results of grinder are as follows.

AIX openjdk version "11.0.5" openj9
177 tests, 84 ok, 0 not ok, 93 skipped, 0 ToDo, 0 Bail Out!

Linux openjdk version "11.0.5" openj9
177 tests, 45 ok, 0 not ok, 132 skipped, 0 ToDo, 0 Bail Out!

PR #1461, opened 2019-11-25
Add cleanWs() calls in zos jenkins metadata files

Mesbah-Alam:
- Add cleanWs() calls in zos jenkins metadata files, as on z/OS, the `checkout scm` command doesn't clean workspace before checkout, unlike other platforms. 

FYI @AdamBrousseau 

### DIFF
--- a/buildenv/jenkins/openjdk_s390_zos
+++ b/buildenv/jenkins/openjdk_s390_zos
@@ -14,6 +14,7 @@ stage('Queue') {
         def gitConfig = scm.getUserRemoteConfigs()[0]
         def SCM_GIT_REPO = gitConfig.getUrl()
         def SCM_GIT_BRANCH = scm.branches[0].name
+        cleanWs()
         sh "git clone -b ${SCM_GIT_BRANCH} ${SCM_GIT_REPO} openjdk-tests"
         jenkinsfile = load "${WORKSPACE}/openjdk-tests/buildenv/jenkins/JenkinsfileBase"
         jenkinsfile.testBuild()
--- a/buildenv/jenkins/openjdk_s390x_zos
+++ b/buildenv/jenkins/openjdk_s390x_zos
@@ -14,6 +14,7 @@ stage('Queue') {
         def gitConfig = scm.getUserRemoteConfigs()[0]
         def SCM_GIT_REPO = gitConfig.getUrl()
         def SCM_GIT_BRANCH = scm.branches[0].name
+        cleanWs()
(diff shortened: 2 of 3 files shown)


Mesbah-Alam:
Test: Test_grinder/job/Grinder/3966/console

sophia-guo:
We do use cleanWs at the end of test job.
https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L498

For jobs generated by job template we also have
https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/testJobTemplate#L170 to make sure wipe out repository and force clone.  Job **Test_grinder/job/Grinder** is manually setup and I just updated configuration with the step "wipe out repository and fore clone". Need to check if there are other grinder jobs configuration need to be updated. 

For other directory we use `fileExists` to make sure it's a clean workspace. https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L192   Maybe there are some cases are missing? I'm not sure what is the issue and what is the root cause?

Mesbah-Alam:
> Maybe there are some cases are missing? 

The case that caused issue for me is, after launching a grinder, i killed it in the middle of the build (it was likely at the stage of checking out from one of the git repos). And then when I re-launched the grinder, it failed stating folder already exists .

sophia-guo:
In that case the issue may not only happen to zos?

Mesbah-Alam:
> In that case the issue may not only happen to zos?

Quoting @AdamBrousseau , "for non-zos, checkout scm handles it for you cause it's smarter".

Mesbah-Alam:
Sophia has found out that, enabling "Wipe out repository and force clean" option in the Grinder job fixes the issue. So we do not need this PR. FYI @AdamBrousseau

AdamBrousseau:
>For jobs generated by job template we also have
https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/testJobTemplate#L170 to make sure wipe out repository and force clone.

Line says `cleanBeforeCheckout()` which is different than `wipeOutWorkspace()`. However that would take place on master anyways not the slave.  You must have killed a job hard enough that it didn't to the cleanWs() in the finally in order to hit this issue. Adding the cleanWs() at the begging shouldn't hurt anything as is. If thins are changed in the future we'll just have to be careful because this is a special case.

sophia-guo:
I see. I can reproduce the issue in Test_grinder/job/Grinder/4005/console.

PR #1470, opened 2019-11-27
Change jdk_math_jre to special level

sophia-guo:
Make sure jre test is available for jre JDKs.

### DIFF
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -247,7 +247,7 @@
 			<subset>10</subset>
 		</subsets>
 		<levels>
-			<level>sanity</level>
+			<level>special</level>
 		</levels>
 		<groups>
 			<group>openjdk</group>


sophia-guo:
Will also create necessary build for special target and setup the build trigger in upstream build( openjdk-build).

sophia-guo:
Originally we'd like to enable it for all jdk_version as default. At some point new JDK11 build do not have jre sdk. And was told jre image won't be built since JDK12+.

> As there is a growing requirement for JRE image we suppose that build will produce both jdk and jre image?

https://github.com/AdoptOpenJDK/openjdk-tests/pull/701#issuecomment-441775333

> The JRE will go away at some point. We're only actually able to build it for 11 because OpenJDK left a legacy 'make JRE' target. But that could get taken away anytime from 12+. Is it possible to skip those tests if no JRE is detected?

https://github.com/AdoptOpenJDK/openjdk-tests/pull/701#issuecomment-441779721

Just check it is build at least in JDK13 build. So probably we need to double check it again as the story may change?

smlambert:
@karianna @johnoliver - per @sophia-guo comment above, where do we stand with enabling jre testing on all jdk versions?  its enabled for jdk8-10 (not 11+ though we see jre is being built for jdk13).

smlambert:
I will merge, we can update subsets later once we get more information.

PR #1477, opened 2019-12-03
Fix thorntail mp tck

lumpfish:
Increases heap size when running thorntail_mp_tck tests on openj9 jdks.
Fixes https://github.com/AdoptOpenJDK/openjdk-tests/issues/1469
Fixes https://github.com/AdoptOpenJDK/openjdk-tests/issues/1161

### DIFF
--- a/external/thorntail-mp-tck/playlist.xml
+++ b/external/thorntail-mp-tck/playlist.xml
@@ -12,10 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 -->
+<!-- Tests require a larger java heap than the default when running with openj9 jdk.  -->
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
 	<test>
 		<testCaseName>thorntail_microprofile_tck</testCaseName>
-		<command>docker run --name thorntail-mp-tck $(EXTRA_DOCKER_ARGS) adoptopenjdk-thorntail-mp-tck:latest \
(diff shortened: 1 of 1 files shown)


external/thorntail-mp-tck/playlist.xml
```diff
@@ -12,10 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 -->
+<!-- Tests require a larger java heap than the default when running with openj9 jdk.  -->
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TestConfig/playlist.xsd">
 	<test>
 		<testCaseName>thorntail_microprofile_tck</testCaseName>
-		<command>docker run --name thorntail-mp-tck $(EXTRA_DOCKER_ARGS) adoptopenjdk-thorntail-mp-tck:latest \
+		<command>docker run --name thorntail-mp-tck --env OPENJ9_JAVA_OPTIONS=-Xmx1g $(EXTRA_DOCKER_ARGS) adoptopenjdk-thorntail-mp-tck:latest \
```
sophia-guo: This will also inject the environment variable into the docker environment running against Hotspot, which won't hurt anything. However  we could also pass through --build-arg values based on JDK_IMPL when running the docker build. And then ENV the passed in build-arg values.

lumpfish: Yes I did realise.  As you say, it will be ignored by hotspot.  Done this way it keeps the code change to a minimum.

smlambert:
I've merged it as it was (passing to both, as it gets ignored in the hs case even though agree with the observation that it can be done indirectly through build args).  

I am working in the external tests lately, as we are adding support for other images, may slightly change the structure of the external projects, hopefully simplify the setting of args and env vars even further.

PR #1486, opened 2019-12-04
Updated Derby Test Dockerfile

james-crowley:
I was running into this error on the old Dockerfile:

```
Using docker image default Java
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8
openjdk version "1.8.0_222"
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_222-b10)
OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.222-b10, mixed mode)
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8
openjdk version "1.8.0_222"
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_222-b10)
OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.222-b10, mixed mode)
/derby
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8
Buildfile: /derby/build.xml

clean:

cleangenerated:

cleanstate:

cleanreleasefiles:

clobber:

BUILD SUCCESSFUL
Total time: 0 seconds
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8
Buildfile: /derby/build.xml

checkCompilerLevel:

BUILD FAILED
/derby/build.xml:211: Compiler level must be Java 9 or later.

Total time: 0 seconds
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8
Buildfile: /derby/build.xml

version_check:

BUILD FAILED
/derby/build.xml:94: FATAL ERROR:  The running Ant version, 1.10.5, is too old.

Total time: 0 seconds
```

So I went ahead a revamped the Dockerfile to fix the errors. I bumped the default SDK to Java 9 and bumped the Ant version to `1.10.6`. On Ubuntu Ant did not have a package version of `1.10.6`, so I switched to manually installing it. 

I also removed a good number of unneeded packages from the container. 

The only other suggestion I wanted to make is not to pull from `https://github.com/apache/derby.git`. With pulling from this and *not* checking out a certain tag or branch we are tasking for issues down the line. We will continue to hit issues with Java version and Ant version. 

With that being said, I am not 100% sure with that you guys are testing here. So if the idea is to pull from always the latest code on Derby and see the errors maybe we should leave it. @ShelleyLambert can you weight in on this?

### DIFF
--- a/external/derby/dockerfile/Dockerfile
+++ b/external/derby/dockerfile/Dockerfile
@@ -29,7 +29,7 @@ ARG IMAGE_VERSION=latest
 FROM $IMAGE_NAME:$IMAGE_VERSION
 
 # Install test dependent executable files
-RUN    apt-get update; \
+RUN apt-get update; \
     apt-get install -y --no-install-recommends \
         git \
         wget \
--- a/external/derby/dockerfile/Dockerfile
+++ b/external/derby/dockerfile/Dockerfile
@@ -22,26 +22,35 @@
 #
 # Build example: `docker build --build-arg IMAGE_NAME=<image_name> --build-arg IMAGE_VERSION=<image_version >-t adoptopenjdk-derby-test .`
 
-ARG SDK=openjdk8
+ARG SDK=openjdk9
 ARG IMAGE_NAME=adoptopenjdk/$SDK
 ARG IMAGE_VERSION=latest
 
 FROM $IMAGE_NAME:$IMAGE_VERSION
 
 # Install test dependent executable files
-RUN apt-get update \
-	&& apt-get -y install \
-	apt-transport-https \
-	ca-certificates \
-	dirmngr \
-	curl \
-	git \
-	make \
-	unzip \
-	vim \
-	tar \
-	wget \
-	ant
+RUN    apt-get update; \
+    apt-get install -y --no-install-recommends \
+        git \
+        wget \
+        tar \
+    ; \
+    rm -rf /var/lib/apt/lists/*
+
+ENV ANT_VERSION=1.10.6
+ENV ANT_HOME=/opt/ant
+
+# Install Ant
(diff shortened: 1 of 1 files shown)


smlambert:
Thanks for digging into this @james-crowley !  

To answer your question, what are we trying to test... generally, we were trying to achieve a couple of goals with this work (which is loosely captured in the [external/README.md](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/external/README.md#third-party-container-tests)), but mainly:
- verify the docker images that we produce at the project
- verify that some large applications are able to run using the Adopt images

If the application runs with both jdk8 and jdk11 (the LTS releases), plus the current release being built (at present jdk13, soon to be jdk14), we should support running tests for all versions.  Looks like in this example, the tip/master of the derby repo is updated to use modularity (so needs jdk9 or higher), which means we either need to stop triggering it with a parameter openjdk8 (which we have already done in the playlist file that TKG uses, https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/external/derby/playlist.xml#L25), or use a tagged version of the derby repo for 8, and a separate tagged version or master for 9+.  Note that if you are [running these external tests via TKG](https://github.com/eclipse/openj9/wiki/Reproducing-Test-Failures-Locally#external-tests), no test target would be generated if you were using jdk8 (because the contents of that playlist file indicate only for 9+).

For ant version, if we move towards an extra test layer as mentioned in Slack conversation, we will pin on a particular version of ant that will work for most cases.  If there are outliers, they will have to set up with whatever version of ant they need (or we also parameterize ant url and version to pass in).

PR #1491, opened 2019-12-06
Fix default repositories and branches for z/OS test pipelines

vsebe:
Use http protocol for all platforms except z/OS. Use ssh protocol on
z/OS.

Signed-off-by: Violeta Sebe <vsebe@ca.ibm.com>


vsebe:
@ShelleyLambert please review, thank you!

PR #1495, opened 2019-12-09
Updated Jenkins Test File

james-crowley:
Removed unneeded dependencies and unneeded `mkdir`.

@smlambert I also removed `CMD [--version]`. I don't think this is needed since `jenkins-test.sh` does not take in any parameters.

### DIFF
--- a/external/jenkins/dockerfile/Dockerfile
+++ b/external/jenkins/dockerfile/Dockerfile
@@ -20,7 +20,7 @@
 # If you want to build image based on other images, please use
 # `--build-arg list` to specify your base image
 #
-# Build example: `docker build --build-arg IMAGE_NAME=<image_name> --build-arg IMAGE_VERSION=<image_version >-t adoptopenjdk-jenkins-test .`
+# Build example: `docker build --build-arg IMAGE_NAME=<image_name> --build-arg IMAGE_VERSION=<image_version > -t adoptopenjdk-jenkins-test .`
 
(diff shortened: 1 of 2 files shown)


external/jenkins/dockerfile/Dockerfile
```diff
@@ -29,33 +29,21 @@ ARG IMAGE_VERSION=latest
...
+    rm -rf /var/lib/apt/lists/*
+
 ENV  JAVA_TOOL_OPTIONS="-Dfile.encoding=UTF8"
 
 # This is the main script to run jenkins tests
 COPY ./dockerfile/jenkins-test.sh /jenkins-test.sh
 
-# Clone jenkins src to /jenkins
 WORKDIR /
-RUN mkdir testResults
```
sophia-guo: testResults is used to copy test junit result from the container to jobs host, which is used by jenkins junit plugin to show the test results running by docker. 

https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/external/jenkins/dockerfile/jenkins-test.sh#L51
https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/external/jenkins/playlist.xml#L20

james-crowley: Oh your right! My bad. I did not see it was used, must of read quickly.

james-crowley: @sophia-guo Fixed in latest push!

PR #1517, opened 2019-12-19
Unexcluding old passing tests from jdk_net

adam-thorpe:


junyzheng:
I think this commit removed tests & the `macosx-all` option from the PR #1390 and it is causing some issue in Azure DevOps pipeline

adam-thorpe:
Ah I see, the macosx-all tag wasn't documented for a different reason in the problem list, so I missed it's importance. I'll re-exclude those two tests and add a brief issue to track it

PR #1527, opened 2020-01-07
Build should fail if it cannot find match in PLATFORM_MAP

llxia:


llxia:
- invalid platform `x86-64_linux_123`: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1626/
- valid platform: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1625/

Issue #1436, opened 2019-11-12
Invalid JTReg Test: java/net/CookieHandler/B6791927.java


adam-thorpe:
This test sets a cookie with an expiry date of Sat 09-Nov-2019 (which has now expired lol), which is causing the test to fail. My guess is that whoever wrote the test wasn't expecting it to be used however many years later. Setting a future date allows this test to pass.

Test: https://github.com/ibmruntimes/openj9-openjdk-jdk13/blob/openj9/test/jdk/java/net/CookieHandler/B6791927.java
```
java.lang.RuntimeException: Expiration date shouldn't be 0
	at B6791927.main(B6791927.java:46)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.sun.javatest.regtest.agent.MainWrapper$MainThread.run(MainWrapper.java:127)
	at java.base/java.lang.Thread.run(Thread.java:831)
```

adam-thorpe:
Interesting to note that this is passing on jdk8 with the latest builds, not sure why this is the case but testing locally will cause the failure

smlambert:
https://bugs.openjdk.java.net/browse/JDK-8234007

smlambert:
Before raising issues in this repo when hitting a problem with openjdk / jtreg tests, its best practice to look in JBS to see if there has already been a report upstream (in the repo and bug reporting system from where the test material came).  (see https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/doc/Triage.md#check-for-an-existing-issue).

adam-thorpe:
To be fair, the issue wasn't created in the bugs database when I raised this issue. However, I don't think I checked and need to get into the habit of doing so anyway. I guess I'll keep this issue open until we see it starts passing tests again

smlambert:
Ha, fair enough, but in the case where you discover an openjdk test case is (clearly) wrong, it should get reported it upstream (you would have been ahead of the crowd).  

[...]

PR #1529, opened 2020-01-08
Triage

adam-thorpe:


openjdk/ProblemList_openjdk13-openj9.txt
```diff
@@ -181,6 +180,7 @@ java/net/httpclient/websocket/PendingPongBinaryClose.java https://github.com/ecl
 java/net/httpclient/websocket/PendingPongTextClose.java https://github.com/eclipse/openj9/issues/4298 macosx-all
 java/net/httpclient/websocket/PendingTextPingClose.java https://github.com/eclipse/openj9/issues/4298 macosx-all
 java/net/httpclient/websocket/PendingTextPongClose.java https://github.com/eclipse/openj9/issues/4298 macosx-all
+java/net/CookieHandler/B6791927.java	https://github.com/AdoptOpenJDK/openjdk-tests/issues/1436	generic-all
```
smlambert: Based on the comment in issue #1436 - the expectation is that this test is fixed, and the fix should be picked up in the tagged branch used by test.  Do you still wish to exclude it?

adam-thorpe: Yes, I excluded it so that it will be easier to see any failing tests coming up to the release (making the pipeline green). I believe the codebase for jdk13 is being held onto until the release so we won't get any new tags until then. I'll just need to remember to unexclude it after the release

Issue #1530, opened 2020-01-08
Archive test logs if Jenkins test build result is not SUCCESS


llxia:
Currently, we only archive test logs and core files (in post stage) when the build result is UNSTABLE. In some unexpected crashes, the build result is FAILURE, so nothing is archived. The same thing happens for the build that is ABORTED. I think we should archive test logs and core files if the build result is not SUCCESS.

In the following example, the build crashed and core was generated, but nothing was archived: Test_openjdk11_j9_sanity.functional_s390x_zos_Nightly/172/

```
00:09:00.601  BUILD FAILED
00:09:00.601  /home/jenkins/workspace/Test_openjdk11_j9_sanity.functional_s390x_zos_Nightly/openjdk-tests/TKG/scripts/build_test.xml:58: The following error occurred while executing this line:
00:09:00.601  /home/jenkins/workspace/Test_openjdk11_j9_sanity.functional_s390x_zos_Nightly/openjdk-tests/functional/build.xml:52: The following error occurred while executing this line:
00:09:00.601  /home/jenkins/workspace/Test_openjdk11_j9_sanity.functional_s390x_zos_Nightly/openjdk-tests/functional/SharedCPEntryInvokerTests/build.xml:55: Compile failed; see the compiler error output for details.
00:09:00.601  
00:09:00.601  Total time: 3 minutes 53 seconds
00:09:00.601  1
00:09:00.601  20200105-22:28:03 - Attempting to move the TDUMPS to '/home/jenkins/workspace/Test_openjdk11_j9_sanity.functional_s390x_zos_Nightly/openjdk-tests/TKG/../TKG/test_output_15782810424214/compile', using the log to identify the TDUMP name to be moved
00:09:00.601  20200105-22:28:03 - Attempting to retrieve dumps with names: 'JENKINS.JVM.JENKINS4.D200105.T222658.X&DS
00:09:00.601  20200105-22:28:03 - Naming of dump consistent with multiple dumps 
00:09:00.601  20200105-22:28:03 - Changed dump name to JENKINS.JVM.JENKINS4.D200105.T222658 
00:09:00.601  listcat
00:09:00.601  20200105-22:28:03 - Looking for JENKINS.JVM.JENKINS4.D200105.T222658.X001 
[...]

PR #1534, opened 2020-01-09
Move post stage to finally

llxia:
We want to capture test logs and core files, not only when the build
result is UNSTABLE, but also any unexpected failure (i.e., FAILURE,
ABORTED, etc). This PR moves post stage (stage that archives test files)
to finally, so that we can capture test logs and core files if build
result != SUCCESS

Resolve: #1530

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -467,7 +467,7 @@ def post(output_name) {
 				)
 			}
 
-			if (currentBuild.result == 'UNSTABLE' || params.ARCHIVE_TEST_RESULTS) {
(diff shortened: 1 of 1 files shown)


llxia:
- test when the build result is SUCCESS: Test_grinder/job/Grinder/5264/
- test when the build result is UNSTABLE: Test_grinder/job/Grinder/5266/
- test when the build result is FAILURE (test failed at compilation): Test_grinder/job/Grinder_Sandbox/65/ ===> post stage happened after compilation failure

llxia:
I am going to ask for a review, so the above Grinder links do not expire. But I will not mark this `Ready for review` so that this PR will not get merged. I would like to merge this PR after 0.18 release.

PR #1536, opened 2020-01-10
Add param LABEL_ADDITION

### LINKED ISSUE #1445: Add ability to pass ADDITIONAL jenkins label
Currently we can override the `LABEL` parameter with a non-default value which is useful. It would also be useful to be able to add a label in order to append to the default values. This would allow us to further narrow down the type of machine we want to pickup.

One option is to add another parameter called `LABEL_ADDITION`. My preference however would be to extend this ability to the current LABEL parameter. For example, if a user input `LABEL='&&<MY_LABEL>'`, the pipelines would recognize the special characters and append rather than replace. This could also be extended to `||` if desired.

AdamBrousseau:
- Additional label(s) to be added to the LABEL variable

Fixes #1445

### DIFF
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -113,7 +113,11 @@ if (PLATFORM_MAP.containsKey(params.PLATFORM)) {
     LABEL = params.LABEL ? params.LABEL : PLATFORM_MAP[params.PLATFORM]["LABEL"]
 
     if (params.DOCKER_REQUIRED) {
-        LABEL="${LABEL}&&sw.tool.docker"
+        LABEL += "&&sw.tool.docker"
+    }
+
+    if (params.LABEL_ADDITION) {
+        LABEL += "&&${params.LABEL_ADDITION}"
     }
 
     println "SPEC: ${SPEC}"
@@ -143,4 +147,3 @@ if (PLATFORM_MAP.containsKey(params.PLATFORM)) {
 } else {
 	assert false : "Cannot find key PLATFORM: ${params.PLATFORM} in PLATFORM_MAP: ${PLATFORM_MAP}."
 }
-
--- a/buildenv/jenkins/testJobTemplate
+++ b/buildenv/jenkins/testJobTemplate
@@ -90,7 +90,7 @@ ARCH_OS_LIST.each { ARCH_OS ->
  	def ADOPTOPENJDK_REPO = "https://github.com/AdoptOpenJDK/openjdk-tests.git"
 	def OPENJ9_REPO = "https://github.com/eclipse/openj9.git"
 	def TKG_REPO = "https://github.com/AdoptOpenJDK/TKG.git"
-	
+
 	// zos has to use ssh
 	if (ARCH_OS.contains("zos")) {
 		ADOPTOPENJDK_REPO = ADOPTOPENJDK_REPO.replace("https://github.com/", "git@github.com:")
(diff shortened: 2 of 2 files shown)


AdamBrousseau:
https://hyc-runtimes-jenkins.swg-devops.com/view/Test_functional/job/Test_openjdk11_j9_sanity.functional_x86-64_linux_Personal/581/console
```
[Pipeline] echo
LABEL: ci.role.test&&hw.arch.x86&&sw.os.linux&&ci.project.openj9
[Pipeline] stage
[Pipeline] { (Queue)
[Pipeline] node
Still waiting to schedule task
Waiting for next available executor on ‘ci.role.test&&hw.arch.x86&&sw.os.linux&&ci.project.openj9’
```

AdamBrousseau:
Note I haven't tested the template portion or completed any script approvals yet.

buildenv/jenkins/openjdk_tests
```diff
@@ -113,7 +113,11 @@ if (PLATFORM_MAP.containsKey(params.PLATFORM)) {
     LABEL = params.LABEL ? params.LABEL : PLATFORM_MAP[params.PLATFORM]["LABEL"]
 
     if (params.DOCKER_REQUIRED) {
-        LABEL="${LABEL}&&sw.tool.docker"
+        LABEL += "&&sw.tool.docker"
+    }
+
+    if (params.LABEL_ADDITION) {
+        LABEL += "&&${params.LABEL_ADDITION}"
```
smlambert: Do you want to keep it to only additive &&, or leave it to users to supply a string that includes the "&&" or "||"?  I am fine with this, as it ensures no one forgets to add this portion of the string, just wanted to check that it covers all the cases needed.

buildenv/jenkins/testJobTemplate
```diff
@@ -139,7 +139,8 @@ ARCH_OS_LIST.each { ARCH_OS ->
 							stringParam('EXTRA_OPTIONS', EXTRA_OPTIONS, "Use this to append options to the test command")
 							stringParam('JVM_OPTIONS', "", "Use this to replace the test original command line options")
 							stringParam('ITERATIONS',"1", "Number of times to repeat execution of test target")
-							stringParam('LABEL', "", "Jenkins node label to run on, leave this blank to get sent to any machine matching the platform, set to node name for runs a particular machine")
+							stringParam('LABEL', "", "Jenkins node label to run on. Leave this blank for the default value specified in PLATFORM_MAP. Set to node name to run on a particular machine.")
+							stringParam('LABEL_ADDITION', "", "Additional label(s) to append to LABEL. Usually when you want to add a label to the default value.")
```
smlambert: Likely good to add that this is an Optional param, and depending on answer to above comment, would need to mention that this string requires && or || ...

buildenv/jenkins/openjdk_tests
AdamBrousseau: Good question. Originally when I proposed using existing LABEL param I thought it would be up to the user to explicitly specify. We could leave it open and prefix it with `&&` if either `&&` or `||` is not present.

buildenv/jenkins/testJobTemplate
AdamBrousseau: Yea, I had that extra detail in originally then decided to remove it. I could add the optional part and explain && || more if we go with my comment above.

smlambert:
@AdamBrousseau - would you be able to request a script approval for openj9 Jenkins, and I will approve at the Adopt Jenkins server.

AdamBrousseau:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=559473

PR #1560, opened 2020-01-16
Update double mapping option for SyntheticGCWorkload_DoubleMap_J9

bragaigor:
Remove double mapping command line option since double mapping is 
enabled by default

Related to issue: https://github.com/eclipse/openj9/issues/8349

Signed-off-by: Igor Braga <higorb1@gmail.com>


bragaigor:
No problem, and sounds good!

amicic:
@llxia Is it hard to get this test parse output (verbose GC, which we already output) and confirm that we see that double mapping is indeed enabled. Expected lines are something like this:

  <attribute name="arrayletDoubleMappingRequested" value="true"/>
  <attribute name="arrayletDoubleMapping" value="true"/>

PR #1575, opened 2020-01-23
Move setup() out of try catch finally block

llxia:
In finally, we have post stage to archive the test logs and cores.
setup() is for downloading SDK and getting test material. In setup(),
there are no test logs or cores to archive. Move setup() out of try catch
finally block to avoid confusion.


llxia:
Grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1919/

PR #1578, opened 2020-01-24
Add correct _ENCODE_FILE_NEW environment to curl for zOS

andrew-m-leonard:


andrew-m-leonard:
curl downloads of binary artifacts need to be untagged on zOS
Setting the _ENCODE_FILE_NEW environment variable for the context of the curl call.
It has no effect on non-zOS platforms

smlambert:
Grinder to verify no effect on non-zos: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1906/

PR #1590, opened 2020-01-29
Ensure openj9 attach api is enabled for jck jvmti tests

andrew-m-leonard:

### DIFF
--- a/jck/runtime.vm/playlist.xml
+++ b/jck/runtime.vm/playlist.xml
@@ -74,7 +74,34 @@
 		<groups>
 			<group>jck</group>
 		</groups>
+                <impls>
+                        <impl>hotspot</impl>
+                </impls>
 	</test>
+        <test>
+                <testCaseName>jck-runtime-vm-jvmti-openj9</testCaseName>
+                <variations>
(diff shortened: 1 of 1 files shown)


jck/runtime.vm/playlist.xml
```diff
@@ -74,7 +74,34 @@
...
+        -test=Jck -test-args=$(Q)tests=vm/jvmti,jckRoot=$(JCK_ROOT),jckversion=$(JCK_VERSION),testsuite=RUNTIME$(Q); \
+        $(TEST_STATUS)</command>
+                <levels>
+                        <level>sanity</level>
+                </levels>
+                <groups>
+                        <group>jck</group>
+                </groups>
+                <impls>
+                        <impl>openj9</impl>
```
Mesbah-Alam: We run JCKs for only OpenJ9 and IBM sdks anyway, so adding these <impl> tags here seems a bit redundant.

andrew-m-leonard: @ShelleyLambert suggested them?
In theory someone else could try and run JCK for Hotspot if they were allowed...

PR #1599, opened 2020-02-03
Check currentBuild.result value specifically

llxia:
If there is no error, currentBuild.result value will not set until the
whole build is completed. In other words, if there is no error, we
can still get `currentBuild.result != 'SUCCESS'` during the build.
Therefore, change the logic to check currentBuild.result value
specifically.

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -477,7 +477,7 @@ def post(output_name) {
 				)
 			}
 
-			if (currentBuild.result != 'SUCCESS' || params.ARCHIVE_TEST_RESULTS) {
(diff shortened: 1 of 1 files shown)


llxia:
Grinder:
- UNSTABLE: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2123/
- ABORTED : https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2121/
- SUCCESS : https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2124/

llxia:
Related: https://github.com/eclipse/openj9/pull/8459#issuecomment-580854335

PR #1611, opened 2020-02-12
Fetch correct repository for jdk14

adam-thorpe:
Quick fix, I couldn't find anything to allow me to do versions greater than 14
https://github.com/AdoptOpenJDK/openjdk-tests/issues/1610

### DIFF
--- a/openjdk/build.xml
+++ b/openjdk/build.xml
@@ -81,7 +81,15 @@
 							<property name="jdkName" value="openjdk-aarch64-jdk8u" />
   						</then>		
 					<else>
-						<property name="jdkName" value="openjdk-jdk${JDK_VERSION}u" />
+						<if>
+							<equals arg1="${JDK_VERSION}" arg2="14" />
+							<then>
+								<property name="jdkName" value="openjdk-jdk${JDK_VERSION}" />
+							</then>
+							<else>
+								<property name="jdkName" value="openjdk-jdk${JDK_VERSION}u" />
+							</else>
+						</if>
 					</else>
 				</if>
 				<property name="regressionRepo" value="https://github.com/AdoptOpenJDK/${jdkName}.git" />
--- a/openjdk/build.xml
+++ b/openjdk/build.xml
@@ -82,12 +82,19 @@
   						</then>		
 					<else>
 						<if>
-							<equals arg1="${JDK_VERSION}" arg2="14" />
+							<or>
+								<equals arg1="${JDK_VERSION}" arg2="8" />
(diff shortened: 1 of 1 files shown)


openjdk/build.xml
```diff
@@ -81,7 +81,15 @@
 							<property name="jdkName" value="openjdk-aarch64-jdk8u" />
   						</then>		
 					<else>
-						<property name="jdkName" value="openjdk-jdk${JDK_VERSION}u" />
+						<if>
+							<equals arg1="${JDK_VERSION}" arg2="14" />
```
smlambert: see my note in the issue, can you try using one of the extra conditions provided by ant-contrib http://ant-contrib.sourceforge.net/tasks/tasks/more_conditions.html

would suggest this:
`<property name="jdkName" value="openjdk-jdk${JDK_VERSION}" />
<if>
 	<isLessThan arg1="${JDK_VERSION}" arg2="14" />
<then>
 	<property name="jdkName" value="openjdk-jdk${JDK_VERSION}u" />
 </then>
</if>`

adam-thorpe: Ah cool I didn't see this bit of documentation, it isn't in he standard antlib one

adam-thorpe: Also, do any of the other build.xml files use the openjdk (hs) repos? I couldn't see any that did but worth asking

smlambert: Correct, the other types of tests we run come from other repos.  build.xml file in the openjdk directory is the only place where we pull material from that openjdk-jdk${JDK_VERSION} repo.

adam-thorpe: From what I can tell, `islessthan` is not going to be helpful here. 
From: http://ant-contrib.sourceforge.net/tasks/tasks/more_conditions.html
> These conditions are suitable for use in the <bool> element. Unfortunately, they cannot be used in the <condition> task`

smlambert: If that is the case, then I will ask you to switch your solution around a bit to explicitly check if equals 8 through 13 append u, else do not append u so we do not have to come back in 15+ to update.

smlambert:
jdk14 grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2212
jdk8 grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2213

adam-thorpe:
v3 api doesn't seem to want to return latest jdk8, shouldn't be related to any of my changes. Kicked off one on 11: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2215/console

PR #1572, opened 2020-01-22
Temporarily increase verbosity on curl for SDK fetch

AdamBrousseau:
- Add check for env variable VERBOSE_CURL which
  will change the default -s option to -v.
- Set VERBOSE_CURL for Jenkins builds temporarily
  in order to debug issue below.

Issue eclipse/openj9#8343

Signed-off-by: Adam Brousseau <adam.brousseau88@gmail.com>

PR #1638, opened 2020-02-19
Revert back to silent curl output on SDK fetch

AdamBrousseau:
Leaving in the framework so it can be easily changed in the
future or by developers.

Related #1572 eclipse/openj9#8343


llxia:
Tested in Grinder:
Test_grinder/job/Grinder/6213/

PR #1656, opened 2020-02-27
Fix system test command line for hs

Mesbah-Alam:
- Resolves https://github.com/AdoptOpenJDK/openjdk-tests/issues/1654
- Creates two separate command templates - one with openj9-systemtest in path, and one without and uses the two templates appropriately in playlists. This solves the issue of system tests breaking in hotspot builds where we do not use openj9-systemtest.

### DIFF
--- a/system/systemtest.mk
+++ b/system/systemtest.mk
@@ -41,9 +41,14 @@ endif
 
 SYSTEMTEST_RESROOT=$(TEST_RESROOT)/../
 
+OPENJ9_PRAM=""
+ifeq (,$(findstring $(JDK_IMPL),hotspot))
+	OPENJ9_PRAM=;$(SYSTEMTEST_RESROOT)$(D)openj9-systemtest
+endif
+
 define SYSTEMTEST_CMD_TEMPLATE
 perl $(SYSTEMTEST_RESROOT)$(D)stf$(D)stf.core$(D)scripts$(D)stf.pl \
-	-test-root=$(Q)$(SYSTEMTEST_RESROOT)$(D)stf;$(SYSTEMTEST_RESROOT)$(D)openjdk-systemtest;$(SYSTEMTEST_RESROOT)$(D)openj9-systemtest$(Q) \
+	-test-root=$(Q)$(SYSTEMTEST_RESROOT)$(D)stf;$(SYSTEMTEST_RESROOT)$(D)openjdk-systemtest$(OPENJ9_PRAM)$(Q) \
 	-systemtest-prereqs=$(Q)$(SYSTEMTEST_RESROOT)$(D)systemtest_prereqs$(Q) \
 	-java-args=$(Q)$(JVM_OPTIONS)$(Q) \
 	-results-root=$(REPORTDIR)


llxia:
Instead of making two separate command templates and change every playlist target, could we just use one template and add `$(SYSTEMTEST_RESROOT)$(D)openj9-systemtest` when JDK_IMPL=openj9 or ibm?

Mesbah-Alam:
@llxia  I've already made the change.

llxia:
From the usability point of view, I would prefer to have one cmd instead of two. There is no playlist.xml change needed. And there is no confusion about which one to use for other users.

Mesbah-Alam:
> There is no playlist.xml change needed

The playlist was already changed. So, if I have to implement what you suggested, I will have to undo those tested changes, make the changes anew, and retest. 

> And there is no confusion about which one to use for other users.

For me, it's more confusing to have something like this: 

```
<command>$(OPENJDK_SYSTEMTEST_CMD_TEMPLATE);$(SYSTEMTEST_RESROOT)$(D)openj9-systemtest -test-args=$(Q)workload=daa1$(Q) -test=DaaLoadTest; \
	$(TEST_STATUS)</command>
```

in the playlist; than this: 

```
<command>$(OPENJ9_SYSTEMTEST_CMD_TEMPLATE) -test-args=$(Q)workload=daa1$(Q) -test=DaaLoadTest; \
	$(TEST_STATUS)</command>
```

--- 

Anyway, there are pros and cons in either approach and I don't want to deliver a change quickly as it's blocking Adopt builds. So, I will discard my solution and implement yours , test it, and ask for review again. 

Thanks.

Mesbah-Alam:
`sanity.system` with hotspot jdk 8 : Test_grinder/job/Grinder/6307/console
`sanity.system` with openj9 jdk 8 : /Test_grinder/job/Grinder/6306/console

llxia:
No, I did not mean to change it as in https://github.com/AdoptOpenJDK/openjdk-tests/pull/1656#issuecomment-592158491

I am suggesting to do the following:
- There is no change need in any of the playlist.xml
- the change is only needed in `systemtest.mk`
`SYSTEMTEST_CMD_TEMPLATE` value gets changed based on JDK_IMPL. We should have something like below (not tested):
```
OPENJ9_PRAM=""
ifeq (,$(findstring $(JDK_IMPL),hotspot))
	OPENJ9_PRAM = $(SYSTEMTEST_RESROOT)$(D)openj9-systemtest
endif

define SYSTEMTEST_CMD_TEMPLATE
perl $(SYSTEMTEST_RESROOT)$(D)stf$(D)stf.core$(D)scripts$(D)stf.pl \
	-test-root=$(Q)$(SYSTEMTEST_RESROOT)$(D)stf;$(SYSTEMTEST_RESROOT)$(D)openjdk-systemtest$(Q)$(OPENJ9_PRAM) \
	-systemtest-prereqs=$(Q)$(SYSTEMTEST_RESROOT)$(D)systemtest_prereqs$(Q) \
	-java-args=$(Q)$(JVM_OPTIONS)$(Q) \
	-results-root=$(REPORTDIR)
endef
```

Mesbah-Alam:
oh.. nice.. let me try that out.

llxia:
Thanks @Mesbah-Alam

Mesbah-Alam:
It seems to work just fine, thanks @llxia - I didn't get your point earlier! :D 

Testing with sanity.jck on hotspot jdk 8 : Test_grinder/job/Grinder/6315/console 
Testing with sanity.jck on openj9 jdk 8 Test_grinder/job/Grinder/6314/console

llxia:
Thanks @Mesbah-Alam . LGTM

PR #1619, opened 2020-02-14
external.sh and add _with_tag targets for external tests

smlambert:
This PR is meant to serve as an example to add/enable xxxxxx_with_tag targets for all external tests.

Grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2223/console
   [echo] Generate dockerfile!!! : build_image.sh derby openjdk8 hotspot ubi jdk full
     [exec] ERROR: Dockerfile generation for /dockerfile////Dockerfile.. failed.
     [exec] /Users/shelley_lambert@ca.ibm.com/work2/openjdk-tests/TKG/../../jvmtest/external/build_image.sh: line 17: ./common_functions.sh: No such file or directory
     [exec] /Users/shelley_lambert@ca.ibm.com/work2/openjdk-tests/TKG/../../jvmtest/external/build_image.sh: line 18: ./dockerfile_functions.sh: No such file or directory

Working out some issues finding/running the common_functions.sh / dockerfile_functions.sh files

Another issue, may need to set permissions differently 
(see output: mkdir: /dockerfile///: Read-only file system)

PR #1660, opened 2020-03-02
Use external.sh and build_image.sh for more docker image tag coverage

smlambert:
This PR (replacing #1619) is an initial switch over for some external tests to use the new scripts which allow the use of any of the docker image tags that are produced and pushed to Docker hub.  

This is ultimately to facilitate the testing of the unofficial images so that we can create a repeatable QA process for verifying them and getting them pushed into the official Docker hub repo.


smlambert:
Camel with ubi / jdk11:
Grinder #[2367](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2367)

Derby with alpine / jdk11:
Grinder #[2368](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2368)

Jenkins with centos / jdk11:
Grinder #[2369](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2369)

Kafka with nightly / jdk8:
Grinder #[2370](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2370)

smlambert:
Things that are still missing:

- external.sh should take 2 report parameters (1 for src and 1 for dst info as not all apps use surefire-reports as src, and need to pass the REPORTDIR from TKG as dst)

- the run should actually fail if there is a failure during the build or execute stage and right now it is not

external/external.sh
```diff
@@ -0,0 +1,134 @@
...
+if [ $command_type == "run" ]; then
+	docker run $docker_args adoptopenjdk-$test-test:${JDK_VERSION}-$package-$docker_os-${JDK_IMPL}-$build_type
+fi
+
+if [ $report != "false" ]; then
+	docker cp adoptopenjdk-$test-test:${JDK_VERSION}-$package-$docker_os-${JDK_IMPL}-$build_type:/testResults/surefire-reports $report/external_test_reports;
+fi
+
+if [ $command_type == "clean" ]; then
+	docker rmi -f adoptopenjdk-$test-test:${JDK_VERSION}-$package-$docker_os-${JDK_IMPL}-$build_type
```
sophia-guo: Might be better to also include removing containers.

external/external.sh
```diff
@@ -0,0 +1,134 @@
...
+
+if [ $command_type == "build" ]; then
+	source $(dirname "$0")/build_image.sh $test $version $impl $docker_os $package $build_type
+fi
+
+if [ $command_type == "run" ]; then
+	docker run $docker_args adoptopenjdk-$test-test:${JDK_VERSION}-$package-$docker_os-${JDK_IMPL}-$build_type
+fi
+
+if [ $report != "false" ]; then
```
sophia-guo: Could we nest this report to if [ $command_type == "run" ] condition? Personally I think that's a bit more clearer.

smlambert: yes, will do

external/external.sh
james-crowley: @sophia-guo Are you talking about removing the base containers? Like `adoptopenjdk:openjdk13-openj9`? Cause the above command will remove the test container built upon the `adoptopenjdk`, but not the base container itself.

james-crowley:
> Jenkins with centos / jdk11:
Grinder #2369 - this tries to do the right thing, which is to build a centos based image, but that fails due to missing manifest, see console output:
07:49:55 [exec] manifest for adoptopenjdk/openjdk11:centos-nightly not found 07:49:55 [exec] Result: 1

@smlambert Do you have an issue open against https://github.com/AdoptOpenJDK/openjdk-docker/? I think we had this issue opened, https://github.com/AdoptOpenJDK/openjdk-docker/issues/268, but I am unsure what came of it.

smlambert:
re: https://github.com/AdoptOpenJDK/openjdk-tests/pull/1660#issuecomment-593657452 I did not raise a new issue in openjdk-docker yet as the Jenkins server was unhealthy over the weekend, which may have caused a blip... was going to run Grinders again in the morning to see, then raise if still existed

external/external.sh
sophia-guo: Sorry, I didn't realize that we actually using --rm to delete anonymous test containers. That is being said all is good.

smlambert: ok sweet

smlambert:
I can update some of our test jobs tonight after a few more grinder runs before merging this, thanks for the review!

smlambert:
Still hitting the same errors in some Grinders 
20:07:32       [exec] manifest for adoptopenjdk/openjdk11:centos-nightly not found
20:07:32       [exec] Result: 1
so have reopened https://github.com/AdoptOpenJDK/openjdk-docker/issues/281

external/external.sh
```diff
@@ -0,0 +1,133 @@
...
+parseCommandLineArgs "$@"
+
+if [ $command_type == "build" ]; then
+	source $(dirname "$0")/build_image.sh $test $version $impl $docker_os $package $build_type
+fi
+
+if [ $command_type == "run" ]; then
+	docker run $docker_args adoptopenjdk-$test-test:${JDK_VERSION}-$package-$docker_os-${JDK_IMPL}-$build_type
+	if [ $report != "false" ]; then
+		docker cp adoptopenjdk-$test-test:${JDK_VERSION}-$package-$docker_os-${JDK_IMPL}-$build_type:/testResults/surefire-reports $report/external_test_reports;
```
sophia-guo: Should be `docker cp container-name` not image name?

external/external.sh
sophia-guo: Currently there are two different scenarios. For tests do not need copy the results we can use --rm to automatically remove the container when it exits. For tests need to copy the test results we can't remove the container when exists. Instead we assign the container name when do `docker run --name`.  And  the container name will be used when doing `docker cp containername:directory....`.  In this scenario when doing clean the test container should also be removed. Sorry I just realized this.

smlambert: ah that explains it.  Let me attempt to generalize this enough to make it possible.  Essentially I would like all test results from all apps made available... but I recognize some may not be in obvious locations

smlambert:
Next grinders:

Camel with ubi / jdk11:
Grinder [#2438](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2438)

Derby with alpine / jdk11:
Grinder [#2439](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2439)

Jenkins with centos / jdk11:
Grinder [#2440](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2440)

Kafka with nightly / jdk8:
Grinder [#2441](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2441)

smlambert:
I will create a follow-up PR for proper handling of reports (in which we may also want to include cores, which in some cases go to a different dir than test results, so will need special handling if we want them).

PR #1661, opened 2020-03-02
Fetch latest/release testimage in get.sh

adam-thorpe:
Should just work, as it would treat it like using `customized_sdk` with 2 parameters


adam-thorpe:
Hold on I've just realised the problem with this is that we don't generate test images for jdk8

smlambert:
Quick Grinder to verify sh script (and yes, we should just add shellcheck to PR checklist) :  https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2422 passes setup

PR #1666, opened 2020-03-05
Update exclude lists

lumpfish:
Signed-off-by: Simon Rushton <rushton@uk.ibm.com>


lumpfish:
Added a couple of excludes following analysis of intermittent issues

Issue #1667, opened 2020-03-05
SDK fetch: Curl retries refuse to overwrite file.


AdamBrousseau:
**Describe the bug**
There is an option added to the curl command which will retry if the fetch fails. The issue is if the file is created at all on the filesystem, the retry will refuse to overwrite it.
Ex job
https://ci.eclipse.org/openj9/job/Test_openjdk11_j9_sanity.functional_x86-64_linux_Nightly/291/console
```
00:02:17.821  + ./openjdk-tests/get.sh -s /home/****/workspace/Test_openjdk11_j9_sanity.functional_x86-64_linux_Nightly -t /home/****/workspace/Test_openjdk11_j9_sanity.functional_x86-64_linux_Nightly/openjdk-tests -p x86-64_linux -r customized -j 11 -i openj9 -c https://140-211-168-230-openstack.osuosl.org/artifactory/ci-eclipse-openj9/Build_JDK11_x86-64_linux_Nightly/306/OpenJ9-JDK11-x86-64_linux-20200303-030650.tar.gz https://140-211-168-230-openstack.osuosl.org/artifactory/ci-eclipse-openj9/Build_JDK11_x86-64_linux_Nightly/306/test-images.tar.gz https://140-211-168-230-openstack.osuosl.org/artifactory/ci-eclipse-openj9/Build_JDK11_x86-64_linux_Nightly/306/debug-image.tar.gz --openj9_repo https://github.com/eclipse/openj9.git --openj9_branch master --openj9_sha 9130918dd6cd886679403df0efa8f22870fd814e --tkg_repo https://github.com/AdoptOpenJDK/TKG.git --tkg_branch master
00:02:17.821  TESTDIR: /home/****/workspace/Test_openjdk11_j9_sanity.functional_x86-64_linux_Nightly/openjdk-tests
00:02:17.821  get jdk binary...
00:02:17.821  _ENCODE_FILE_NEW=UNTAGGED curl -OLJSks --user ****:**** https://140-211-168-230-openstack.osuosl.org/artifactory/ci-eclipse-openj9/Build_JDK11_x86-64_linux_Nightly/306/OpenJ9-JDK11-x86-64_linux-20200303-030650.tar.gz
00:04:55.438  curl: (56) GnuTLS recv error (-9): A TLS packet with unexpected length was received.
00:04:55.438  curl error code: 56. Sleep 300 secs, then retry 1...
[...]

PR #1685, opened 2020-03-18
Remove the downloaded file before retry

llxia:
If curl fails to download a file, it may leave a damaged file on disk
(i.e., file that is partially downloaded). This PR removes the file if
it exists, so that retry can re-download the file without File exists
error.

Issue: #1667


llxia:
Grinder:
- wins nightly: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2511/
- xlinux nightly: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2513/
- xlinux customized: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2512/

PR #1690, opened 2020-03-20
Convert functional-test to use external.sh and add quickstarts tests

smlambert:
Continue to convert remaining tests to use external.sh
Add quarkus_quickstarts tests
Update to latest tag for quarkus tests
Remove LogManager workaround from quarkus test


external/dockerfile_functions.sh
```diff
@@ -546,8 +547,8 @@ generate_dockerfile() {
     print_clone_project ${file} ${test} ${github_url};
     print_entrypoint ${file} ${script} ${os};
 
-    if [[ ! -z ${cmd} ]]; then
-        print_cmd ${file} ${cmd};
+    if [[ ! -z ${testtarget} ]]; then
```
james-crowley: Should have an `else` after the `if`, to handle the "default" `cmd` value in the Dockerfile.

```
 if [[ ! -z ${testtarget} ]]; then
       print_cmd ${file} ${testtarget};
 else
       print_cmd ${file} ${cmd};	        
```

external/external.sh
```diff
@@ -121,13 +125,16 @@ function parse_tag () {
...
 	source $(dirname "$0")/build_image.sh $test $version $impl $docker_os $package $build_type
 fi
 
 if [ $command_type == "run" ]; then
 	if [ $reportsrc != "false" ]; then
-		docker run $docker_args --name $test-test adoptopenjdk-$test-test:${JDK_VERSION}-$package-$docker_os-${JDK_IMPL}-$build_type;
+		docker run $docker_args --name $test-test adoptopenjdk-$test-test:${JDK_VERSION}-$package-$docker_os-${JDK_IMPL}-$build_type $testtarget;
 		docker cp $test-test:$reportsrc $reportdst/external_test_reports;
 		else
 		docker run $docker_args --rm adoptopenjdk-$test-test:${JDK_VERSION}-$package-$docker_os-${JDK_IMPL}-$build_type;
```
james-crowley: Also need to add `$testtarget` at the end of this line.

james-crowley:
Also on line 33 in `external.sh` you should probably update the usage to reflect the addition of `testtarget`

external/dockerfile_functions.sh
smlambert: for this one, I am essentially commandeering the $cmd variable and replacing it with $testtarget, so will no longer need cmd defined in common_functions.sh (which I will remove as part of this PR)

james-crowley: That works. The only thing I could see is someone calling `build_all.sh`, which would build all the images. If there is no default `cmd`, then when you go to run the images there might be an error. 

For example, the `functional-test.sh`, expects a variable passed in, https://github.com/smlambert/openjdk-tests/blob/05d939124540d3ac4e7ec4e3fb4743e587cac54e/external/functional-test/dockerfile/functional-test.sh#L46. If there is no `$1` variable passed in will the Docker container fail?

I think having the `testtarget` is good. It allows us to test a certain workload. But if no `testtarget` is passed in should there be a default `cmd`? If no default is needed, we can fully remove `cmd`.

smlambert: For the cases (camel/quarkus/functional-test/system-test) where we had a default CMD, none are needed.  I think its likely clearer and more correct to pass the testtarget via the playlist.xml explicitly (since so far, in all the cases, they are specific to the different test targets defined in playlist.xml).

For the functional-test example you point out, for each of the playlist entries, I now define an explicit target, _sanity.functional.regular, _extended.functional.regular and _testExample.

If someone adds new test targets to the playlist.xml file, we would run Grinder to verify they have explicitly set a target in the `<command>` (call to external.sh).

PR #1699, opened 2020-03-27
Add OpenJ9 -Xdump options to avoid system and heap dumps for OOM

pshipton:
Some tests intentionally create OOMs, which produces unnecessary system
and heap dumps. When failures occur, the dumps bloat the size of the
result artifact, consuming more space and increasing download times. If
there is a legitimate OOM failure in these tests, there won't be a
system dump created and it will have to be reproduced manually with
modified options, but this seems a fair tradeoff for an unusual case.

jdk_math_0/work/java/math/BigInteger/DivisionOverflow
jdk_util_0/work/java/util/ResourceBundle/Bug4168625Test
jdk_lang_0/work/java/lang/ProcessBuilder/Basic_id0 - there are 2 here,
likely from sub-processes
jdk_lang_0/work/vm/gc/ArraySize
jdk_lang_0/work/vm/gc/InfiniteList
jdk_lang_0/work/java/lang/ref/SoftReference/Pin

Test grinder running jdk_math_j9_0 only: https://ci.eclipse.org/openj9/view/Test/job/Grinder/725

### DIFF
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -210,6 +210,33 @@
 		<groups>
 			<group>openjdk</group>
 		</groups>
+		<impls>
+			<impl>hotspot</impl>
+		</impls>
+	</test>
+	<test>
+		<testCaseName>jdk_lang_j9</testCaseName>
+		<variations>
+			<variation>-Xdump:system:none -Xdump:heap:none -Xdump:system:events=gpf+abort+traceassert+corruptcache</variation>
+		</variations>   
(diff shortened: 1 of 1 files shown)


pshipton:
Updated to handle the OpenJ9 specific options.

smlambert:
Grinders:
hs/jdk8/jdk_math: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2627
j9/jdk11/jdk_lang: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2628
j9/jdk14/jdk_util: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2629

PR #1709, opened 2020-04-01
Added md5 command for different OS

### LINKED ISSUE #1152: OSX: md5sum command not found
When a curl in the get.sh for the SDK fails, an md5sum is run to help add some debug to the output.
Originally added by @sxa555 [here](https://github.com/AdoptOpenJDK/openjdk-tests/commit/e2309c06e97ef2bb9943afdc51279191aec23280).
The issue is on OSX the command is `md5` not `md5sum`. I'm not sure if it's safe enough to just hardcode the other command for OSX or if the command should be somehow "figured out"  on the fly.

Example output (URL masked by me)
```
Running on osx
...
00:00:34.805  get jdk binary...
00:00:34.805  curl -OLJks --retry 5 --retry-delay 300 --user ****:**** <URL>
00:01:44.932  curl -OLJks --retry 5 --retry-delay 300 --user ****:**** <URL>
00:01:52.967  Error code: 7
00:01:52.967  Failed to retrieve <URL>, exiting. This is what we received of the file and MD5 sum:
00:01:52.967  ls: <URL>: No such file or directory
00:01:52.967  ./openjdk-tests/get.sh: line 171: md5sum: command not found
```

Would say this is low priority as it doesn't block or break anything.

Niwedita-Rani:
Fixes #1152

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -248,16 +248,13 @@ getBinaryOpenjdk()
 				echo "curl error code: $download_exit_code"
 				echo "Failed to retrieve $file, exiting. This is what we received of the file and MD5 sum:"
 				ls -ld $file
-				#md5sum $file
 				
-				#Starts-Added for checking md5 checksum command on fly
 				if [[ "$OSTYPE" == "darwin"* ]]; then
 				    md5 $file
 				 else
 				    md5sum $file
 				fi
-				#Ends-Added for checking md5 checksum command on fly
-				
+						
 				exit 1
 			fi
 			set -e
--- a/get.sh
+++ b/get.sh
@@ -252,9 +252,10 @@ getBinaryOpenjdk()
 				
 				#Starts-Added for checking md5 checksum command on fly
 				if [[ "$OSTYPE" == "darwin"* ]]; then
-				md5 $file
-				else
-				md5sum $file
+				    md5 $file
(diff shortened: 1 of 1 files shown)


get.sh
```diff
@@ -248,7 +248,16 @@ getBinaryOpenjdk()
 				echo "curl error code: $download_exit_code"
 				echo "Failed to retrieve $file, exiting. This is what we received of the file and MD5 sum:"
 				ls -ld $file
-				md5sum $file
+				#md5sum $file
```
smlambert, line 251: Thanks for your contribution @Niwedita-Rani - were you able to test it locally (perhaps by disconnecting your laptop if you have a Mac during a test run)?

I only have some minor review comments, which is you can remove the 3 commented out lines as they do not add to the understanding of the code

smlambert:
@Niwedita-Rani - hoping you have seen the review comment above and can update your PR.  :)

get.sh
Niwedita-Rani: Thanks @smlambert for reviewing the changes.I have removed the 3 commented lines.

smlambert:
Grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2963/

PR #1713, opened 2020-04-02
Fixing issue with Debian Slim Generation 

james-crowley:
There is a know issue with Debian Slim that in the image it is missing some man directories. Thus some packages will fail to install like `maven`. 

You follow the issue progress here, https://github.com/debuerreotype/debuerreotype/issues/10. 

Once that issue is fixed, this patch will not be needed.


external/dockerfile_functions.sh
```diff
@@ -139,7 +139,11 @@ print_debianslim_pkg() {
     local file=$1
     local packages=$2
 
-    print_ubuntu_pkg ${file} "${packages}"
+    echo -e "RUN apt-get update \\" \
```
smlambert, line 142: Wondering if we should put a comment in here linking to the issue, so we will have some reminder to un-do this change when the core issue is fixed?

james-crowley: just added the comment. Linked to the issue.

PR #1715, opened 2020-04-03
OpenJ9 AArch64: Exclude some tests that depend on GC

knn-k:
This commit excludes the following 3 tests.
- java/util/logging/LogManager/Configuration/ParentLoggerWithHandlerGC.java eclipse/openj9#8897
- java/util/WeakHashMap/GCDuringIteration.java eclipse/openj9#9043
- java/util/concurrent/ArrayBlockingQueue/WhiteBox.java eclipse/openj9#9044

These tests depend on GC behavior as discussed in eclipse/openj9#8897.

Signed-off-by: KONNO Kazuhiro <konno@jp.ibm.com>

PR #1739, opened 2020-04-20
OpenJ9 AArch64: Exclude a test that depends on GC

knn-k:
This commit excludes the following test.

- java/lang/reflect/callerCache/ReflectionCallerCacheTest.java eclipse/openj9#9260

This test depends on GC behavior as discussed in eclipse/openj9#8897.


knn-k:
The reason of excluding this test is the same as that in #1715.

PR #1747, opened 2020-04-22
Update getJtreg task logic

sophia-guo:
Check file jtreg-4.2.0-tip.tar instead of jtreg-4.2.0-tip.tar.gz
as there is requirement not to require GNU tar options when
downloading jtreg. jtreg-4.2.0-tip.tar.gz file has been decompressed as
jtreg-4.2.0-tip.tar

Move untar task outside of check file exist function as the task should be done for each build step (make compile).

### DIFF
--- a/openjdk/build.xml
+++ b/openjdk/build.xml
@@ -29,37 +29,38 @@
 		<mkdir dir="${DEST}"/>
 		<if>
 			<not>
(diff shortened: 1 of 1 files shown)


sophia-guo:
Fix issue reported at https://adoptopenjdk.slack.com/archives/C5219G28G/p1587415117054100

PR #1752, opened 2020-04-24
WIP: Enable the Attach API for openjdk jobs on z/OS

adamfarley:
With the Attach API disabled by default on z/OS, some tests will
just fail. Enabling the API allows Attachments, and enables the tests
to pass.


adamfarley:
Wrong branch.

PR #1753, opened 2020-04-24
Enable the Attach API for openjdk jobs on z/OS

adamfarley:
With the Attach API disabled by default on z/OS, some tests will
just fail. Enabling the API allows Attachments, and enables the tests
to pass.

### DIFF
--- a/openjdk/openjdk.mk
+++ b/openjdk/openjdk.mk
@@ -38,6 +38,9 @@ ifeq ($(CYGWIN),1)
 		| cut -d "=" -f 2-` / 1024 / 1024 \
 		)
 endif
+ifeq ($(OS),OS/390)
+	EXTRA_OPTIONS += -Dcom.ibm.tools.attach.enable=yes
+endif
 # Upstream OpenJDK, roughly, sets concurrency based on the
 # following: min(NPROCS/2, MEM_IN_GB/2).
 MEM := $(shell expr $(MEMORY_SIZE) / 2048)


smlambert:
Before you move forward with this @adamfarley - I had intended to message @andrew-m-leonard to discuss moving this logic out of this jobTemplate and down into the [jck.mk](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/jck/jck.mk) file, and for your update, it would move into the [openjdk.mk](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/openjdk/openjdk.mk) file.  

These are better places to add the EXTRA_OPTIONS (and as a bonus, then do not require us to go to all the Jenkins servers, 6 that I know of to do a script approvals for template changes).

andrew-m-leonard:
sounds reasonable, I presume it would be a matter of adding extra_options to:

llxia:
As mentioned above, move the logic to mk, we should check for `JDK_IMPL ` and `SPEC`, then append the jvm option in `EXTRA_OPTIONS ` in openjdk.mk and jck.mk. The rest will be take care of by TKG. There is no playlist.xml change needed. 

Similar to the following:

https://github.com/AdoptOpenJDK/openjdk-tests/blob/ed05e61cd66b444a4125989fdc7006bae744ac81/openjdk/openjdk.mk#L120-L121

sophia-guo:
Probably should use EXTRA_OPTIONS+= attachcommand in case EXTRA_OPTIONS is exported explicitly by user.

adamfarley:
Ok, taking the above comments onboard, the PR is now changed. OpenJDK test run running now:

https://hyc-runtimes-jenkins.swg-devops.com/view/Test_openjdk/job/Test_openjdk11_j9_sanity.openjdk_s390x_zos_Personal/21/console

Will add the same change to jck.mk and also remove the relevant bit in testJobTemplate if this works.

andrew-m-leonard:
@adamfarley note via the Personal test job above it won't be using your openjdk-tests repo
```
[Pipeline] sh
+ git clone -b master git@github.com:AdoptOpenJDK/openjdk-tests.git openjdk-tests 
Cloning into 'openjdk-tests'...
```
You need to use Grinder instead

adamfarley:
Good spot Andrew. :)

A shame we still have that parameter if the personal build doesn't use it, but that's a seperate bug.

andrew-m-leonard:
I also suspect you need to set EXTRA_JTREG_OPTIONS
looking at the playlist.xml the jtreg command is:
```
	$(JTREG_BASIC_OPTIONS) $(CUSTOM_NATIVE_OPTIONS) -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS)$(Q) \
	-r $(Q)$(TEST_RESROOT)$(D)report$(Q) \
	-jdk:$(Q)$(TEST_JDK_HOME)$(Q) \
	$(Q)$(OPENJDK_DIR)$(D)$(JDK_CUSTOM_TARGET)$(Q); \
	$(TEST_STATUS)</command>
```

smlambert:
re: https://github.com/AdoptOpenJDK/openjdk-tests/pull/1753#issuecomment-619107649 I do not believe you need this suggestion, as @llxia alluded to above, TKG appends EXTRA_OPTIONS to JVM_OPTIONS so the rest is taken care of for you.

andrew-m-leonard:
@ShelleyLambert ah yes, just discovered this: https://github.com/AdoptOpenJDK/TKG/blob/49f0056df4cb876f9daf5d3d7e5de164ec219ed0/src/org/testKitGen/MkGen.java#L108

smlambert:
re: https://github.com/AdoptOpenJDK/openjdk-tests/pull/1753#issuecomment-619107649 - what the heck is a personal build?  ;)  If there is a bug to be raised, I guess raise that in the repo where those pipelines live and gripe to people who wrote that pipeline (links to internal server jobs, we typically do not include in open issues, as they are not accessible to everyone).

The point is that its not an open pipeline.  Open testing pipelines do use/honour the ADOPTOPENJDK_REPO and ADOPTOPENJDK_BRANCH, including all Grinder jobs.  

Thanks for updating your PR!  Given its for zOS support, I have launched Grinder/7248 on the internal server using your branch to verify this change.

adamfarley:
@smlambert - Fair comments. No more griping or linking, promise. :)

adamfarley:
Grinder_zos/20 failed due to a parameter issue unrelated to the change (it couldn't reach the build I pointed it at). 

Grinder_zos/21 has the right parameters, and the option was included correctly. It still failed due to other bugs in the build, but I'm not seeing failures of the type we set out to fix, so I'm calling this a win.

Removing WIP and requesting merge.

Issue #1322, opened 2019-09-04
Dynamically generate parameters for parallel child jobs


llxia:
In parallel, we pass in all variables to child jobs regardless if they are used/set or not. This results that we need to [initialize all the parameters](https://github.com/AdoptOpenJDK/openjdk-tests/pull/1317). We only need to dynamically generate used parameters and pass in them to the child jobs.

Issue #1563, opened 2020-01-17
Support for smart parallelization in TKG / TRSS


llxia:
We would like to support for smart parallelization in TKG / TRSS, so that we can break test build into any number of smaller builds (based on test during and the number of machines) and run them in parallel. The goal is to achieve a balanced load and optimal overall testing throughput.

ToDo:

- TRSS provides API to divide tests into a number of buckets (based on test
historical duration) to get optimal throughput overall  https://github.com/AdoptOpenJDK/openjdk-test-tools/pull/146
- TKG generates makefiles on demand https://github.com/AdoptOpenJDK/TKG/issues/25
- Parse TRSS API output and run TKG with the tests
- Handle the case where TRSS is not available (#1801 )

smlambert:
As we revise and update the test pipeline scripts to support this feature, I thought it would be good to capture any of the known specially configured test jobs (parameters different than expected defaults) in one spot, to make sure we account for them in the test pipeline revisions.  

| [platform](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/openjdk_tests)\testGroup| openjdk | external | system | perf | functional | jck | NOTES |
| --- | --- | --- | --- | --- | --- | --- | --- |
| x86-64_linux |  |  |  | LABEL=ci.role.perf&& | BUILD_LIST=functional/subdir |   |  functional subdirectories was defined to avoid issue of running tests on machines with older ant versions JIT_Test dir needs new ant |
|x86-64_mac  |  |  |  |  LABEL=ci.role.perf&& |  |   |
|s390x_linux  |  |  |  |  |  |   |
| ppc64_linux |  |  |  |  |  |   |
| s390_zos |  |  |  |  |  |  EXTRA_OPTIONS=attachAPI |


Some test jobs were defined as weekly and in those cases use `@weekly` in Jenkins scheduler and so must also have default SDK_RESOURCE=nightly .  We can simplify this by updating build pipeline scripts to run sanity level testing nightly, but on weekends switch to extended level tests.  For release builds, ultimately build pipelines should run both sanity and extended levels.

llxia:
PR https://github.com/AdoptOpenJDK/TKG/pull/67

smlambert:
It seems like all of the related changes are in, are we ready to close this one?

llxia:
Yes, this can be closed.

PR #1755, opened 2020-04-27
Set parameters dynamically for parallel child jobs

llxia:
- Instead of explicit setting all the parameters, loop params, modify
parameters as needed and pass parameters to downstream jobs
- With this PR, we do not need to update parameter array whenever we add
new parameters in the parent job
- Remove explicit settings to prepare for #1563

Resolve: #1322

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -144,52 +144,33 @@ def setupParallelEnv() {
 					childTest = "test_iteration_${i}"
 					buildListName = env.BUILD_LIST
 				}
+
+				def childParams = []
+				// loop through all the params and change the parameters if needed
+				params.each { param ->
+					if (param.key == "ITERATIONS") {
+						childParams << string(name: param.key, value: "1")
+					} else if (param.key == "BUILD_LIST") {
+						childParams << string(name: param.key, value: "${buildListName}")
+					} else if (param.key == "IS_PARALLEL") {
+						childParams << booleanParam(name: param.key, value: false)
+					} else {
+						def value = param.value.toString()
+						if (value == "true" || value == "false") {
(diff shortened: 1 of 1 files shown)


llxia:
- IS_PARALLEL= true: Test_openjdk8_j9_special.system_x86-64_linux_Nightly/1521/
- IS_PARALLEL= true: Grinder/7333/
- IS_PARALLEL= false: Grinder/7337

buildenv/jenkins/JenkinsfileBase
```diff
@@ -144,52 +144,33 @@ def setupParallelEnv() {
...
+				params.each { param ->
+					if (param.key == "ITERATIONS") {
+						childParams << string(name: param.key, value: "1")
+					} else if (param.key == "BUILD_LIST") {
+						childParams << string(name: param.key, value: "${buildListName}")
+					} else if (param.key == "IS_PARALLEL") {
+						childParams << booleanParam(name: param.key, value: false)
+					} else {
+						def value = param.value.toString()
+						if (value == "true" || value == "false") {
```
smlambert, line 159: I guess this works because we either have Boolean parameters or String parameters (as our Choice/Credential parameters end up being handled as Strings).  :)

llxia: yes, choice/credential parameters end up being handled as Strings.

llxia:
> are there any other default values that you want to handle here as you have done for ITERATIONS/BUILD_LIST/IS_PARALLEL, or are the other defaults for params handled elsewhere in the script?

yes, the are a couple more params that we will need to handle, such as `IS_PARALLEL` will be renamed to `PARALLEL` and it will become a choice `['None', 'Dynamic', 'Subdir', 'Iteration']`; `NUM_MACHINE`; `TARGET`, etc. I will have a sperate PR for it as they are closely tied to template change and openj9 script.

PR #1759, opened 2020-04-28
Disable actions for examples

### LINKED ISSUE #1757: workflows/actionExample.yml fails every time I push to my fork
`workflows/actionExample.yml` fails every time I push to my fork, even if I have actions disabled, and I get an email notifying me of that failure.

I suggest the example should at least be syntactically correct and should trivially succeed. Apparently a file with only comments doesn't qualify.

sophia-guo:
Ignore action in example files

Close #1757 


sophia-guo:
Action is ignored though the basic steps are still needed for syntactic check.

PR #1767, opened 2020-05-04
use local params in the loop

llxia:
use local params, so params get set in each iteration

### DIFF
--- a/buildenv/jenkins/testJobTemplate
+++ b/buildenv/jenkins/testJobTemplate
@@ -107,15 +107,9 @@ ARCH_OS_LIST.each { ARCH_OS ->
 	JDK_VERSIONS.each { JDK_VERSION ->
 		LEVELS.each { LEVEL ->
 			GROUPS.each { GROUP ->
-				if (!TEST_JOB_NAME) {
-					TEST_JOB_NAME = "Test_openjdk${JDK_VERSION}_${JDK_IMPL_SN}_${LEVEL}.${GROUP}_${ARCH_OS}${SUFFIX}"
-				}
-				if (!TARGET) {
-					TARGET = "${LEVEL}.${GROUP}"
-				}
-				if (!BUILD_LIST) {
-					BUILD_LIST = GROUP
-				}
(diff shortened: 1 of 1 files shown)


llxia:
Test_Job_Auto_Gen_sandbox/38
Test_Job_Auto_Gen_sandbox/39
Grinder/7690

llxia:
More details:

In nightly build, testJobTemplate takes params such as level, group, etc to figure out TEST_JOB_NAME (loop is used as we may generate multiple jobs based on level x group x jdk_version combination). In the recent change, we also allow user to set TEST_JOB_NAME (no loop in this case). 

TEST_JOB_NAME needs to set to empty after each iteration to support the loop case. Otherwise, with the same TEST_JOB_NAME, the script will only generate one job with wrong params (example below).

Only one job is generated even if we have `GROUPS = jck, functional`
```
LEVELS: [sanity]
JDK_VERSIONS: [11]
GROUPS: [jck, functional]
ARCH_OS_LIST: [s390x_linux_xl]
GROUP: jck
inside if GROUP: jck
EXTRA_OPTIONS: -Xfuture --enable-preview
GROUP: functional
EXTRA_OPTIONS: 
Existing items:
    GeneratedJob{name='Test_openjdk11_j9_sanity.jck_s390x_linux_xl_Nightly'}
Started calculate disk usage of build
Finished Calculation of disk usage of build in 0 seconds
Started calculate disk usage of workspace
Finished Calculation of disk usage of workspace in  7 second
Notifying upstream projects of job completion
```

llxia:
Grinder/7690
Test_Job_Auto_Gen_sandbox/38
Test_Job_Auto_Gen_sandbox/39

llxia:
Test_Job_Auto_Gen_sandbox/42
Test_Job_Auto_Gen_sandbox/43
Grinder/7691
Grinder/7692

llxia:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=562706

smlambert:
I've done script approval at ci.adoptopenjdk.net

PR #1771, opened 2020-05-06
Rename Renaissance Tests

piyush286:
• Renamed Renaissance tests according to TRSS parser convention so that the results of these tests can be parsed and viewed on TRSS
• Disabled renaissance-finagle-chirper for now since it fails

Issue: https://github.com/AdoptOpenJDK/openjdk-tests/issues/1770

Signed-off-by: Piyush Gupta <piyush286@gmail.com>


piyush286:
Sanity Run: 
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2986/console

PR #1774, opened 2020-05-07
Exclude CheckHandles.java on 32bit Windows

M-Davies:


M-Davies:
This is simply an exclude from nightly testing to reduce noise, not a fix. Please don't close the accompanying issue

PR #1798, opened 2020-05-26
Don't fail build if xUnit plugin fails to process

AdamBrousseau:
- There is an intermittent failure we are seeing
  where the xUnit plugin will throw an exception
  while processing certain JCK result files.
  Setting this to false will allow the build to
  not fail (RED) when this exception occurs and
  will allow the build to finish as desired.

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -118,14 +118,14 @@ def setupParallelEnv() {
 				// limit childJobNum
 				if (childJobNum > 20) {
 					echo "Due to the limited machines, ITERATIONS can only be set up to 20. Current ITERATIONS is ${ITERATIONS}."
-					echo "Reset ITERATIONS to 20..." 
+					echo "Reset ITERATIONS to 20..."
 					childJobNum = 20
 				}
 			} else if (params.PARALLEL == "Dynamic") {
 				int numOfMachines = getNumMachines()
 				String PARALLEL_OPTIONS = "TEST=${TARGET} NUM_MACHINES=${numOfMachines}"
 				if (params.TRSS_URL) {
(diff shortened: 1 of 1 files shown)


AdamBrousseau:
@Mesbah-Alam for review

buildenv/jenkins/JenkinsfileBase
```diff
@@ -118,14 +118,14 @@ def setupParallelEnv() {
 				// limit childJobNum
 				if (childJobNum > 20) {
 					echo "Due to the limited machines, ITERATIONS can only be set up to 20. Current ITERATIONS is ${ITERATIONS}."
-					echo "Reset ITERATIONS to 20..." 
+					echo "Reset ITERATIONS to 20..."
```
Mesbah-Alam, line 121: There seems to be a lot of changes which

AdamBrousseau:
Appears this change doesn't resolve our issue. Setting to WIP for now.

smlambert:
Is this PR still needed or relevant or can it be closed?

AdamBrousseau:
The root issue for this was never resolved and we are still seeing it but this PR didn't help so it can be closed.

smlambert:
@AdamBrousseau - how frequently is the issue seen?  (trying to determine how to prioritize)

AdamBrousseau:
At the moment, just looking at the past few builds, it seems to be 100% reproducible on ub16 zlinux machines. It does not appear to be failing on other platforms or on rhel7 on zlinux. This is odd as this is opposite of what I stated in the issue back in February. We were tracking this via internal infra issue 3209, not sure if you have access to that anymore.

smlambert:
ya, unfortunately I do not have access to GHE repos anymore.  Hmm, I suspect that there is content from the zlinux output that is causing the transform (the CSS that takes javatest formatted xml and transforms to junit format) to fail.

PR #1812, opened 2020-06-03
Convert parallelList.mk between ascii and ebcdic for zos

llxia:
During parallel runs, we generate, read and archive parallelList.mk in
master (linux) and read from child build. During zos testing, we need
covert parallelList.mk between ascii and ebcdic depends on which platform
we are on.

Related: runtimes/backlog/issues/329

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -133,6 +133,10 @@ def setupParallelEnv() {
 				def parallelList = "openjdk-tests/TKG/parallelList.mk"
 				int NUM_LIST = -1
 				if (fileExists("${parallelList}")) {
+					if (SPEC.startsWith('zos')) {
+						echo 'Converting parallelList.mk file from ebcdic to ascii...'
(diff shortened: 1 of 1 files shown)


llxia:
Parent: Grinder_zos/31
Child: Grinder_zos_testList_0/7/

In serial: Grinder/8214/

buildenv/jenkins/JenkinsfileBase
```diff
@@ -496,6 +500,12 @@ def runTest( ) {
 			}
 			if (!TARGET.startsWith('-f')) {
 				TARGET="_${params.TARGET}"
+			} else if (TARGET.startsWith('-f parallelList.mk') && SPEC.startsWith('zos')) {
```
smlambert, line 503: is TARGET order dependent?  as in, will it always startWith '-f parallelList.mk', or could it TARGET.contains() ?

llxia:
retested: Grinder_zos_testList_0/9

smlambert:
Thanks for the 2nd set of Grinder links, see that the mk file gets converted and is readable on that platform now.

PR #1722, opened 2020-04-08
Add support for Elasticsearch Testing in Docker Generation

james-crowley:
This PR will enable support for using elasticsearch as an external test with Docker generation. Currently elasticsearch is the only test not enabled in the external test directory.

PR #1826, opened 2020-06-09
Remove quotes if the test is not using external.sh

llxia:


llxia:
Tested: https://ci.adoptopenjdk.net/job/Test_openjdk11_j9_extended.external_x86-64_linux_elasticsearch/6/

The above test failed later during test execution (docker cmd runs fine)

smlambert:
related #1722 which converts elasticsearch tests to use external.sh, and will need to use quotes

PR #1832, opened 2020-06-15
Add RISC-V support to OpenJDK test suites

ChengJin01:
Signed-off-by: Cheng Jin <jincheng@ca.ibm.com>

### DIFF
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -45,6 +45,14 @@ def PLATFORM_MAP = [
         'SPEC' : 'linux_ppc-64_le',
         'LABEL' : 'ci.role.test&&hw.arch.ppc64le&&sw.os.linux',
     ],
+    'riscv64_linux' : [
+        'SPEC' : 'linux_riscv64_cmprssptrs',
+        'LABEL' : 'ci.role.test&&sw.os.linux&&hw.arch.riscv&&hw.bits.64',
+    ],
+    'riscv64_linux_xl' : [
+        'SPEC' : 'linux_riscv64',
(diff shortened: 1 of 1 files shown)


ChengJin01:
Note: the value of SPEC  uses `riscv-64` instead of `riscv64`. 
If we agree to change to riscv64, then https://github.com/AdoptOpenJDK/TKG/blob/master/src/org/openj9/envInfo/JavaInfo.java needs to be updated to support this.
```
public String getSPEC() {
 ...
        String model = System.getProperty("sun.arch.data.model");
        System.out.println("System.getProperty('sun.arch.data.model')=" + model + "\n");
        if (Integer.parseInt(model.trim())==64) {
            spec += "-64"; <----- to be updated for riscv64
            spec = spec.replace("arm-64", "aarch64");
        }
```

Reviewer: @llxia 

FYI: @ShelleyLambert , @sxa

ChengJin01:
the SPEC for `RISC-V` has changed to `riscv64 `to align with the naming of buildspec in OpenJ9.

llxia:
Could you test this PR with https://github.com/AdoptOpenJDK/TKG/pull/83/ using Grinder_TKG on any platform? Thanks

ChengJin01:
> 
> 
> Could you test this PR with [AdoptOpenJDK/TKG#83](https://github.com/AdoptOpenJDK/TKG/pull/83) using Grinder_TKG on any platform? Thanks

Launched at `Test_grinder/job/Grinder_TKG/632` with the changes at https://github.com/AdoptOpenJDK/openjdk-tests/pull/1832 & https://github.com/AdoptOpenJDK/TKG/pull/83/

llxia:
The above Grinder failed because of missing test image. Relaunched at Grinder_TKG/640

PR #1833, opened 2020-06-16
WIP: add extraDockerArgs to build.xml so we can override docker image name

jennli:
Proposing a change to read `EXTRA_DOCKER_ARGS` in external test build steps.

I am running external tests in Azure DevOps but getting the following error when executing `./maketest.sh`:

```
manifest for adoptopenjdk/openjdk11:linux--2020.6.0.8523-nightly not found: manifest unknown: manifest unknown
```
at this line:
https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/external/elasticsearch/build.xml#L41

~My guess is that we can't access the docker image specified in the Dockerfile, so my team has been overriding~
We would like to override the docker registry and image name with an image we created in AzDo.

The way this is done right now is to replace the following line 
 https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/external/elasticsearch/dockerfile/Dockerfile#L26:

with

```
ARG IMAGE_NAME=another-image-from-azure
...
```

This is clearly not the idea approach, so I'm proposing a change I'd like to upstream, which can be seen in my PR.

My concern with this approach is that I don't know all the use cases of `EXTRA_DOCKER_ARGS`, so I'm not sure about any potential side effects of having extra docker args in the build steps. 

Let me know what you think, thanks.

### DIFF
--- a/external/build.xml
+++ b/external/build.xml
@@ -24,7 +24,8 @@
 	<property name="DEST_EXTERNAL" value="${BUILD_ROOT}/external" />
 	<property environment="env" />
 	<property name="top" location="../" />
-	
+	<property name="extraDockerArgs" value="${env.EXTRA_DOCKER_ARGS}"/>
+
 	<if>
 		<isset property="env.DOCKERIMAGE_TAG"/>
 		<then>
@@ -50,7 +51,7 @@
 			<fileset dir="." includes="*.mk"/>
 		</copy>
 	</target>
-	
+
 	<target name="docker_prune" description="Remove all unused images from the machine before running new builds">
 		<exec executable="docker">
 			<arg line="system prune --all --force" />
@@ -58,7 +59,7 @@
 	</target>
 
 	<target name="clean_image" description="clean test docker image if there is one">
-		<echo message="Executing external.sh --clean --dir ${TEST} --tag ${dockerImageTag} --version ${JDK_VERSION} --impl ${JDK_IMPL} " />						
+		<echo message="Executing external.sh --clean --dir ${TEST} --tag ${dockerImageTag} --version ${JDK_VERSION} --impl ${JDK_IMPL} " />
 		<exec executable="bash">
 			<arg value="${DEST_EXTERNAL}/external.sh"/>
 			<arg value="--clean"/>
@@ -83,7 +84,7 @@
 	</target>
 
 	<target name="build_image" description="build test dockerfile">
-		<echo message="Executing external.sh --build --dir ${TEST} --tag ${dockerImageTag} --version ${JDK_VERSION} --impl ${JDK_IMPL} " />						
+		<echo message="Executing external.sh --build --dir ${TEST} --tag ${dockerImageTag} --version ${JDK_VERSION} --impl ${JDK_IMPL} " />
 		<exec executable="bash">
 			<arg value="${DEST_EXTERNAL}/external.sh"/>
 			<arg value="--build"/>
--- a/external/elasticsearch/build.xml
+++ b/external/elasticsearch/build.xml
@@ -38,7 +38,7 @@
 			</else>
 		</if>
 		<exec executable="docker"  failonerror="true">
-			<arg line="build -t adoptopenjdk-elasticsearch-test -f dockerfile/Dockerfile --pull . --build-arg SDK=${JVM_VERSION} --build-arg ELASTIC_VERSION=${ELASTIC_VERSION} --build-arg IMAGE_VERSION=${dockerImageTag}"/>
+			<arg line="build -t adoptopenjdk-elasticsearch-test -f dockerfile/Dockerfile --pull . --build-arg SDK=${JVM_VERSION} --build-arg ELASTIC_VERSION=${ELASTIC_VERSION} --build-arg IMAGE_VERSION=${dockerImageTag} ${extraDockerArgs}"/>
 		</exec>
 	</target>
 
--- a/external/example-test/build.xml
+++ b/external/example-test/build.xml
@@ -22,7 +22,7 @@
 
 	<target name="build_image" depends="clean_image" description="build example test docker image">
 		<exec executable="docker"  failonerror="true">
-			<arg line="build -t adoptopenjdk-example-test -f dockerfile/Dockerfile --pull . --build-arg SDK=${JVM_VERSION} --build-arg IMAGE_VERSION=${dockerImageTag}"/>
+			<arg line="build -t adoptopenjdk-example-test -f dockerfile/Dockerfile --pull . --build-arg SDK=${JVM_VERSION} --build-arg IMAGE_VERSION=${dockerImageTag} ${extraDockerArgs}"/>
 		</exec>
 	</target>
 


external/build.xml
```diff
@@ -24,7 +24,8 @@
 	<property name="DEST_EXTERNAL" value="${BUILD_ROOT}/external" />
 	<property environment="env" />
 	<property name="top" location="../" />
-	
+	<property name="extraDockerArgs" value="${env.EXTRA_DOCKER_ARGS}"/>
```
jennli, line 27: reads `EXTRA_DOCKER_ARGS `

external/elasticsearch/build.xml
```diff
@@ -38,7 +38,7 @@
 			</else>
 		</if>
 		<exec executable="docker"  failonerror="true">
-			<arg line="build -t adoptopenjdk-elasticsearch-test -f dockerfile/Dockerfile --pull . --build-arg SDK=${JVM_VERSION} --build-arg ELASTIC_VERSION=${ELASTIC_VERSION} --build-arg IMAGE_VERSION=${dockerImageTag}"/>
+			<arg line="build -t adoptopenjdk-elasticsearch-test -f dockerfile/Dockerfile --pull . --build-arg SDK=${JVM_VERSION} --build-arg ELASTIC_VERSION=${ELASTIC_VERSION} --build-arg IMAGE_VERSION=${dockerImageTag} ${extraDockerArgs}"/>
 		</exec>
```
jennli, line 42: In Azure DevOps we will populate `EXTRA_DOCKER_ARGS` with this string:
```
--build-arg IMAGE_NAME=a-docker-image-we-can-access
```

sophia-guo:
Could you try to set your azure build with setting, that should work.

```
DOCKERIMAGE_TAG = another-image-from-azure

```

sophia-guo:
Docker image specified in the Dockerfile is in the Docker anyone should be able to access it. But the image version should not like 'linux--2020.6.0.8523-nightly'. What's your job settings, i.e. `DOCKERIMAGE_TAG`?

jennli:
@sophia-guo thanks for your comment, now I realized that my guess about "cannot access Adopt's docker image" was completely wrong. 

As it turns out, my team chose to use images from azure, instead of pulling Adopt's images, because we want to run tests with the same images that we used for building OUR latest jdk binaries. This results in a need of changing the image name to something else.

Does this answer your question?

smlambert:
I believe @junyzheng proposed something off of a branch I had, for the same purpose.  I will look that up and look more closely at what you are proposing and give feedback tomorrow.  I am sure there are options for us here.  Cheers!  :)

junyzheng:
IMO to override docker image we need to set: docker registry name, image name, and tag name. I think right now the image name is generated through `version number + implementation` inside the bash script. So one way to implement this could be move docker registry name, image name, and tag name out of the bash script mainly this one: https://github.com/AdoptOpenJDK/openjdk-tests/blob/565191efde062aa5337befdd943611c75aff6d42/external/dockerfile_functions.sh#L76

jennli:
@smlambert hey Shelley, any thoughts about this approach?

smlambert:
Yes, apologies my thinking is a day late!!  ;)

As @junyzheng mentions, we make the assumption to auto-gen our Dockerfiles based FROM the adoptopenjdk ones, which means its not very flexible if others want to base from other registries/images.  Since we can not predict how other registries or image names are organized (but can assume its not the same as AdoptOpenJDK images), we introduce a decision point in one of our shell scripts that said, if particular env vars exist, then use those, else print_image_args() as per AdoptOpenJDK image name structure.

Can you tell me a bit more about how you are running?  Are you using TKG to run in AzDo?  As in, are you running something like these commands to trigger this external testing for particular tests?

git clone https://github.com/AdoptOpenJDK/openjdk-tests.git
cd openjdk-tests
./get.sh
export BUILD_LIST=external/example-test
export TEST_JDK_HOME=yourJDKUnderTestHome
make compile
make _example-test

If that is the case, then have you tested the use of EXTRA_DOCKER_ARGS to change IMAGE_NAME (as per your PR)?  
export EXTRA_DOCKER_ARGS="--build-arg IMAGE_NAME=another-image-from-azure"

ATM, we do not pass EXTRA_DOCKER_ARGS into the [docker build step](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/external/external.sh#L135) , but given your use case, we could determine what is set and decide which path to take, default build off of AdoptOpenJDK stuff, or from other info if its provided.

jennli:
Hey no worries Shelley, actually I feel bad taking time out of your day...

Yes we are running tests using TKG in AzDo following those steps you posted.

And yes I have tried passing in `EXTRA_DOCKER_ARGS` with "--build-arg IMAGE_NAME=another-image-from-azure" to override IMAGE_NAME, which works exactly as I have hoped. It subsequently brought me to make this draft PR and see if there's a chance to upstream it. I only added extra docker args to the elasticsearch build.xml, and example-test build.xml as an example, but I suspect the same change can be applied to all the tests' build step.

jennli:
another value I might want to override with `EXTRA_DOCKER_ARGS` is an ANT source URL, like this one:
https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/external/derby/dockerfile/Dockerfile#L47

Joseph informed me about some rules that prevents abusing apache services, from https://infra.apache.org/infra-ban.html

so my team currently downloads ANT from a cached source that we have uploaded to Azure, to avoid abusing Apache services. So I want to parameterized ANT URL with an argument, like

```
# Dockerfile

ARG ANT_URL
...
RUN wget --no-check-certificate --no-cookies $ANT_URL \
```
I hope this 2nd use case of EXTRA_DOCKER_ARGS in the build steps can strengthen the value of my proposal.

external/elasticsearch/build.xml
jennli:
```
--build-arg IMAGE_NAME=another-image-from-azure --build-arg ANT_URL=cached-ant-source-from-azure
```

if ANT_URL is parameterized in the Dockerfiles

smlambert:
Yes, I am interested in changes that would allow the flexibility to pull from a different registry and basing from a different image name.   I suspect we will also hit similar story (at least re: ANT_URL) when/if we run these tests in the AdoptOpenJDK public AzDo?  

I think we will need to add extra functionality in[ external.sh that parses EXTRA_DOCKER_ARGS](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/external/external.sh#L51) and sends --build_arg ones to [build_image.sh script](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/external/external.sh#L135), while others to docker run command.

In this PR you are editing build.xml files which should be updated to use the new style of generating Dockerfile on the fly, or are in draft to change to the new style (https://github.com/AdoptOpenJDK/openjdk-tests/pull/1722).  So we should plan to include the feature you propose into the new story of generating Dockerfiles on the fly (as I had hoped we could be able to clean up and no longer maintain the static dockerfiles in the openjdk-tests/external dir.

jennli:
OK good idea! Do you know when https://github.com/AdoptOpenJDK/openjdk-tests/pull/1722 will be merged? I'm trying to decide if I should just make a temporary fix in our mirrored code to unblock myself until upstream is updated with the new way of generating Dockerfiles.

smlambert:
yes, if you need something this week, likely good to temp fix your mirror, then we can follow with the change discussed.

smlambert:
Closing as stale, but created https://github.com/adoptium/aqa-tests/issues/2645 so we do not lose the request for such a feature.

PR #1834, opened 2020-06-16
Set a custom OpenJ9 timeout handler for OpenJDK testing

pshipton:
Modified the existing OpenJ9 targets. Add a openj9_custom target as
well. If we find timeouts in other targets we can customize them for
OpenJ9 in the future, rather than duplicating everything now.

The timeout handler uses an OpenJ9 `jcmd` command to create a core file
when a timeout occurs, as well as calling jstack as the default timeout
handler does.

Issue https://github.com/eclipse/openj9/issues/9651

Depends on https://github.com/AdoptOpenJDK/TKG/pull/85

### DIFF
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -34,6 +34,29 @@
 		</groups>
 	</test>
 	<test>
+		<testCaseName>openj9_custom</testCaseName>
+		<variations>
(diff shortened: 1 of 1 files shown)


pshipton:
@llxia @smlambert please review.

PR #1835, opened 2020-06-17
Start an X11 server for openjdk tests on AIX

Haroon-Khel:
Ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1277

Ive tidied the code up a bit, and reduced the X11 server start up code to 2 conditions:

1.  (`BUILD_LIST` is openjdk or jck) and `SPEC` is aix 
2. (`BUILD_LIST` is openjdk and `SPEC` is linux_x86-64) or (`BUILD_LIST` is jck and `SPEC` is linux)

jdk_awt tests are run only on x86-64 architectures, [see here](https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1277#issuecomment-613663200), hence why there are 2 different conditions for `SPEC` in the second point, else this could have been a bit tidier.

Finding a way to test this.

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -447,26 +447,6 @@ def get_sources() {
 	}
 }
 
-// Starts an X11 server and sets the DISPLAY variable. Currently only for linux and aix
-def setEnvDisplay(spec) {
-	if (spec.startsWith('aix')) {
-		sh "nohup /usr/bin/X11/X -force -vfb -x abx -x dbe -x GLX :0 &"
-		env.DISPLAY = ":0"
-		echo "env.DISPLAY is ${env.DISPLAY}"	
-	}
-	else if (spec.startsWith('linux')) {
-		wrap([$class: 'Xvfb', autoDisplayName: true]) {
-			def DISPLAY = sh (
-				script: 'ps -f  | grep \'[X]vfb\' | awk \'{print \$9}\'',
-				returnStdout: true
-			).trim()
-			env.DISPLAY = "${DISPLAY}"
-			echo "env.DISPLAY is ${env.DISPLAY}"
-			makeTest("${RUNTEST_CMD}")
-		}
-	}
-}
-
 def make_test() {
 	// use sshagent with Jenkins credentials ID for all platforms except zOS
 	if (!env.SPEC.startsWith('zos')) {
@@ -558,8 +538,26 @@ def runTest( ) {
 			RUNTEST_CMD = "./openjdk-tests ${TARGET} ${CUSTOM_OPTION}"
 			for (int i = 0; i < ITERATIONS; i++) {
 				echo "ITERATION: ${iterationCount}/${ITERATIONS}"
-				setEnvDisplay(env.SPEC)
-				// makeTest("${RUNTEST_CMD}")
+				if (spec.startsWith('aix')) {
+					sh "nohup /usr/bin/X11/X -force -vfb -x abx -x dbe -x GLX :0 &"
+					env.DISPLAY = ":0"
+					echo "env.DISPLAY is ${env.DISPLAY}"	
+					makeTest("${RUNTEST_CMD}")
+				}
+				else if (spec.startsWith('linux')) {
+					wrap([$class: 'Xvfb', autoDisplayName: true]) {
+						def DISPLAY = sh (
+							script: 'ps -f  | grep \'[X]vfb\' | awk \'{print \$9}\'',
+							returnStdout: true
+						).trim()
+						env.DISPLAY = "${DISPLAY}"
+						echo "env.DISPLAY is ${env.DISPLAY}"
+						makeTest("${RUNTEST_CMD}")
+					}
+				} 
+				else {
+					makeTest("${RUNTEST_CMD}")
+				}
 				iterationCount++
 			}
 		}
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -538,13 +538,13 @@ def runTest( ) {
 			RUNTEST_CMD = "./openjdk-tests ${TARGET} ${CUSTOM_OPTION}"
 			for (int i = 0; i < ITERATIONS; i++) {
 				echo "ITERATION: ${iterationCount}/${ITERATIONS}"
-				if (spec.startsWith('aix')) {
+				if (env.SPEC.startsWith('aix')) {
 					sh "nohup /usr/bin/X11/X -force -vfb -x abx -x dbe -x GLX :0 &"
 					env.DISPLAY = ":0"
 					echo "env.DISPLAY is ${env.DISPLAY}"	
 					makeTest("${RUNTEST_CMD}")
 				}
-				else if (spec.startsWith('linux')) {
+				else if (env.SPEC.startsWith('linux')) {
 					wrap([$class: 'Xvfb', autoDisplayName: true]) {
 						def DISPLAY = sh (
 							script: 'ps -f  | grep \'[X]vfb\' | awk \'{print \$9}\'',
@@ -569,8 +569,6 @@ def post(output_name) {
 		timestamps{
 			if (env.DISPLAY != null) {
 				env.DISPLAY = ""
-				// Debug code
-				echo "env.DISPLAY is ${env.DISPLAY}"
 			}
 			if (output_name.contains(',')) {
 				output_name = "specifiedTarget"
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -538,40 +538,25 @@ def runTest( ) {
 			RUNTEST_CMD = "./openjdk-tests ${TARGET} ${CUSTOM_OPTION}"
 			for (int i = 0; i < ITERATIONS; i++) {
 				echo "ITERATION: ${iterationCount}/${ITERATIONS}"
-				if (env.BUILD_LIST == 'openjdk') {
-					if (env.SPEC.startsWith('linux_x86-64')) {
-						wrap([$class: 'Xvfb', autoDisplayName: true]) {
-							def DISPLAY = sh (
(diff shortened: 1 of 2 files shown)


Haroon-Khel:
@ShelleyLambert I have to manually mention you since I dont think I have the permissions in this repo to add reviewers

buildenv/jenkins/JenkinsfileBase
```diff
@@ -538,40 +538,25 @@ def runTest( ) {
...
-							).trim()
-							env.DISPLAY = "${DISPLAY}"
-							echo "env.DISPLAY is ${env.DISPLAY}"
-							makeTest("${RUNTEST_CMD}")
-						}
-					} else if (env.SPEC.startsWith('aix')) {
-						sh "nohup /usr/bin/X11/X -force -vfb -x abx -x dbe -x GLX :0 &"
-						env.DISPLAY = ":0"
+				// Starting an X11 server
+				if ((env.BUILD_LIST == 'openjdk' || env.BUILD_LIST.startsWith('jck')) && env.SPEC.startsWith('aix')) {
```
smlambert, line 542: Actually wondering if we create a method called setEnvDisplay(env.BUILD_LIST, env.SPEC) as we will also want to set DISPLAY for osx.  We could potentially just set it no matter what BUILD_LIST is defined, as its presumably benign and does not interfere with tests that do not need it.  In that case, its essentially a check for SPEC and then running the appropriate commands to accomplish the setting of the DISPLAY variable on the machine.

```
setEnvDisplay (env.SPEC) {
    if (env.SPEC.startsWith('aix')) {
        sh "nohup /usr/bin/X11/X -force -vfb -x abx -x dbe -x GLX :0 &"
					env.DISPLAY = ":0"
    } else if (env.SPEC.startsWith('linux')) {
        wrap([$class: 'Xvfb', autoDisplayName: true]) {
						def DISPLAY = sh (
							script: 'ps -f  | grep \'[X]vfb\' | awk \'{print \$9}\'',
							returnStdout: true
						).trim()
						env.DISPLAY = "${DISPLAY}"
    } else if (env.SPEC.startsWith('osx') {
        // what ever commands needed there
    }
    echo "env.DISPLAY is ${env.DISPLAY}"
}
```
then call it at line 542 and follow with the call to makeTest

I believe the only reason that openjdk jdk_awt tests are not running on other platforms is this DISPLAY issue, so it will be good for us to run them against your branch changes to see if they look healthier.  (in other words, the check for && env.SPEC.startsWith('linux_x86-64') would no longer be needed, and we can re-enable targets on other platforms with DISPLAY set for all platforms).

Haroon-Khel: Okay so Ive removed the BUILD_LIST check and reduced the condition for SPEC to either aix or linux. It looks alot tidier and more manageable.

I havent added the commands needed to run an x11 server for osx as I do not have them. Is there already an open issue regarding running x11 tests on osx?

smlambert:
We can kick off some Grinder variations to test this, its looking good.

One other thing I would want to check is whether there is any issue on AIX, where we run back to back tests on the same machine... as with this approach, we are not unsetting the DISPLAY variable in the POST step (where if I am not mistaken, the xvfb Jenkins plugin used for the linux case does so).

Haroon-Khel:
I tried running jdk_awt on an aix machine [here](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3411/tapTestReport/) but the tests were skipped. Instead I ran a custom test, which I know fails when DISPLAY isnt set, and it seemed to pass, [here](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3415/tapTestReport/)

buildenv/jenkins/JenkinsfileBase
smlambert: For reference for what commands used by openjdk to set DISPLAY, see here: https://github.com/AdoptOpenJDK/openjdk-jdk8u/blob/master/jdk/test/start-Xvfb.sh (or [jdk11u version of start-Xvfb.sh](https://github.com/AdoptOpenJDK/openjdk-jdk11u/blob/master/test/jdk/start-Xvfb.sh)) though don't know the details of how they expect test machines to be configured.

smlambert:
Yes, jdk_awt and jdk_swing will have to have changes to allow for more platforms to be enabled, right now only x86 arch allowed, see [playlist.xml](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/openjdk/playlist.xml#L178).  Think that could be changed to ^os.osx,^os.zos and remove arch restriction now that all should be able to have DISPLAY set.

Haroon-Khel:
jdk_awt runs on AIX with no x11 error, see [grinder](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3431/), but there are alot of failures. By a quick inspection, these failures dont seem to be too different than a control [grinder](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3433/console) (jdk_awt run on x86_64 linux, **master** branch). Though once the tests have finished running, i'll have a better idea of the differences between the two.

Running jdk_awt tests with my changes on x86_64 linux, see [grinder](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3432/console), however gave the following x11 error:
```
STATUS:Failed.`main' threw exception: java.awt.AWTError: Can't connect to X11 window server using '-displayfd -displayfd' as the value of the DISPLAY variable.
```
This refers to this [command](https://github.com/AdoptOpenJDK/openjdk-tests/pull/1835/files#diff-9da4c7b7710fb521a3c011d52316f1f9R460), which retrieves the DISPLAY variable from a `ps -f` command. I havent changed this command from the master branch so I am not sure why this seems to be failing. This same error does not occur on the control grinder

Haroon-Khel:
> One other thing I would want to check is whether there is any issue on AIX, where we run back to back tests on the same machine... as with this approach, we are not unsetting the DISPLAY variable in the POST step (where if I am not mistaken, the xvfb Jenkins plugin used for the linux case does so).

Ahh, I think the above error may be related to this

smlambert:
re: https://github.com/AdoptOpenJDK/openjdk-tests/pull/1835#issuecomment-647738232 I think your hunch may be right, can try to add a unsetDISPLAY in post step and see if resolves it.

smlambert:
strike the [above comment](https://github.com/AdoptOpenJDK/openjdk-tests/pull/1835#issuecomment-649028856), I read your comment more closely, your changes fail on xlinux, env.DISPLAY does not get set properly any more for xlinux, moving the code into the setDisplay method may have caused a side effect.

Haroon-Khel:
After going through the console logs of the failing linux grinder, it seems that the DISPLAY variable may be set correctly if `$9` is changed to `$10` in the following line `script: 'ps -f  | grep \'[X]vfb\' | awk \'{print \$10}\'',`. Ive kicked off a grinder to test this. Ive also added aline which unsets the DISPLAY variable in post, so lets see if that works

Haroon-Khel:
The [grinder](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3466/console) for the test mentioned above still seems to give `STATUS:Failed.`main' threw exception: java.awt.AWTError: Can't connect to X11 window server using '2' as the value of the DISPLAY variable.` Its now detecting a number, but still failing.

To debug this error more meticulously, ive created a seperate branch on my fork, https://github.com/Haroon-Khel/openjdk-tests/tree/experimental.x11, to which im adding the same changes as on the enable.x11 branch, but im adding them bit by bit and testing at each change so that I can see exactly which change of code seems to cause the error

Haroon-Khel:
I think ive found what causes the error. It seems the [makeTest command](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L550) needs to be inside the `wrap` method during the xvfb script for linux

Haroon-Khel:
For this reason, ive removed the setEnvDisplay method entirely and placed its code back into runTest. I feel like the makeTest command, which runs the test for all platforms/spec, shouldnt be in a method dedicated to setting a variable for only 2 platforms

smlambert:
Rerun several grinders with plan to close based on their results (do not require all tests to pass, require them not to fail with x11 error).  


- [ ] 8/hotspot/xlinux/jdk_awt


- [ ] 11/hotspot/aix/jdk_awt


- [ ] 14/openj9/aix/jdk_awt

smlambert:
From grinders for this PR, now that we can run some of the jdk_awt tests on aix, we can see a new failure that causes many failures (which needs separate triage and possibly an openjdk-build issue raised):

```
Execution failed: `main' threw exception: java.lang.UnsatisfiedLinkError: 
/home/jenkins/workspace/Grinder/openjdkbinary/j2sdk-image/lib/libfontmanager.so: 
/home/jenkins/workspace/Grinder/openjdkbinary/j2sdk-image/lib/libfontmanager.so, 
LIBPATH=/home/jenkins/workspace/Grinder/openjdkbinary/j2sdk-
image/lib/server:/home/jenkins/workspace/Grinder/openjdkbinary/j2sdk-
image/lib:/home/jenkins/workspace/Grinder/openjdkbinary/j2sdk-
image/lib/jli:/home/jenkins/workspace/Grinder/openjdkbinary/j2sdk-image/../lib, LD_LIBRARY_PATH= : 0509-022 Cannot load 
module /usr/lib/libpng16.a(libpng16.so.16). 0509-150 Dependent module libgcc_s.a(shr.o) could not be loaded. 0509-022 
Cannot load module libgcc_s.a(shr.o). 0509-026 System error: A file or directory in the path name does not exist. 0509-022 
Cannot load module /home/jenkins/workspace/Grinder/openjdkbinary/j2sdk-image/lib/libfontmanager.so. 0509-150 Dependent 
module /usr/lib/libpng16.a(libpng16.so.16) could not be loaded.  
```
(see https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3497/#showFailuresLink as example jdk11/hotspot 417 failures, versus 220 failures on [xlinux grinder](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3500/))

PR #1871, opened 2020-07-08
Revert "Temp exclude sun/util/calendar/zi/TestZoneInfo310.java"

pshipton:
Reverts AdoptOpenJDK/openjdk-tests#1793

This test is passing now. I believe it was failing earlier because the tests don't necessarily fetch the test material which matches the JDK being tested.

https://ci.eclipse.org/openj9/view/Test/job/Grinder/943/

Fixes https://github.com/eclipse/openj9/issues/9642


pshipton:
@smlambert for review

PR #1883, opened 2020-07-13
enable jck-runtime-vm-instr for JDK8

llxia:
Related: runtimes/backlog/issues/357 and runtimes/backlog/issues/352

### DIFF
--- a/jck/runtime.vm/playlist.xml
+++ b/jck/runtime.vm/playlist.xml
@@ -157,9 +157,6 @@
 		<groups>
 			<group>jck</group>
 		</groups>
-		<subsets>
-			<subset>10+</subset>
-		</subsets>
 	</test>
 	<test>
 		<testCaseName>jck-runtime-vm-cldc</testCaseName>


Mesbah-Alam:
@llxia : on a second thought, I recall @JasonFengJ9 would have preferred to have some of these vm tests in sanity instead of extended (which is why `jck-runtime-vm-instr` was put in sanity at the first place, albeit 10+). So, please keep `jck-runtime-vm-instr` as sanity. Also, we could simply not have any <subset> tag for this target -- so that it can run in 8, 11, 15.. all of them (instead of just 8)-- as the test exits in all.

JasonFengJ9:
> have preferred to have some of these vm tests in sanity instead of extended (which is why jck-runtime-vm-instr was put in sanity at the first place, albeit 10+).

That's correct, there are some vm/lang tests running nightly to catch JCK breakage earlier.

llxia:
Sure, that will make this change simpler.

llxia:
Updated. Waiting for internal Jenkins to be available so I can run a Grinder.

llxia:
Grinder/8927

llxia:
Close this PR as the change is in #1884

Issue #1887, opened 2020-07-14
quarkus_quickstarts_test_0 docker image free space too small


andrew-m-leonard:
https://ci.adoptopenjdk.net/job/Test_openjdk14_j9_sanity.external_x86-64_linux_quarkus_quickstarts/27/console
```
22:23:58  build_image:
22:23:58       [echo] Executing external.sh --build --dir quarkus_quickstarts --tag nightly --version 14 --impl openj9 
22:23:58       [exec] 
22:23:58       [exec] Writing /home/jenkins/workspace/Test_openjdk14_j9_sanity.external_x86-64_linux_quarkus_quickstarts/jvmtest/external/quarkus_quickstarts/dockerfile/14/jdk/ubuntu/Dockerfile.openj9.full ... done
22:23:58       [exec] 
22:23:58       [exec] #####################################################
22:23:58       [exec] INFO: docker build --no-cache -t adoptopenjdk-quarkus_quickstarts-test:14-jdk-ubuntu-openj9-full -f /home/jenkins/workspace/Test_openjdk14_j9_sanity.external_x86-64_linux_quarkus_quickstarts/jvmtest/external/quarkus_quickstarts/dockerfile/14/jdk/ubuntu/Dockerfile.openj9.full /home/jenkins/workspace/Test_openjdk14_j9_sanity.external_x86-64_linux_quarkus_quickstarts/jvmtest/external/
22:23:58       [exec] #####################################################
22:23:58       [exec] Sending build context to Docker daemon  58.88kB
22:23:58       [exec] Step 1/22 : ARG SDK=openjdk14-openj9
22:23:58       [exec] Step 2/22 : ARG IMAGE_NAME=adoptopenjdk/$SDK
22:23:58       [exec] Step 3/22 : ARG OS=ubuntu
22:23:58       [exec] Step 4/22 : ARG IMAGE_VERSION=nightly
22:23:58       [exec] Step 5/22 : ARG TAG=$OS-$IMAGE_VERSION
22:23:58       [exec] Step 6/22 : FROM $IMAGE_NAME:$TAG
22:23:58       [exec]  ---> 2fa2054c958d
22:23:58       [exec] Step 7/22 : ARG QUARKUS_QUICKSTARTS_TAG=1.3.2.Final
22:23:58       [exec]  ---> Running in 248d7144454a
22:23:58       [exec] Removing intermediate container 248d7144454a
22:23:58       [exec]  ---> 033a1897fe47
[...]

PR #1890, opened 2020-07-14
Add docker_prune as depends in clean_image target

llxia:
When running using subdir, subdir/build.xml will be triggered directly.
In this case, we will not run parent dir init, which results not running
docker_prune target. To fix this, add docker_prune as depends in
clean_image target. clean_image target is called in each subdir
build.xml

Resolves: #1887

### DIFF
--- a/external/build.xml
+++ b/external/build.xml
@@ -34,7 +34,7 @@
 			<property name="dockerImageTag" value="nightly"/>
 		</else>
 	</if>
-	<target name="init" depends="docker_prune">
+	<target name="init">
 		<mkdir dir="${DEST_EXTERNAL}" />
 	</target>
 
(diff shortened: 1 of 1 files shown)


llxia:
Grinder/3583/ runs `quarkus_quickstarts_test` on the previous failed machine `test-softlayer-ubuntu1604-x64-​1`

```
11:52:44  docker_prune:
11:54:14       [exec] Deleted Images:
11:54:14       [exec] untagged: adoptopenjdk-quarkus-test:8-jdk-ubuntu-openj9-full
11:54:14       [exec] deleted: sha256:4e6950f377aec9824bb7559164deb3be23ecab07164257bc53ae2102d74ab834
11:54:14       [exec] deleted: sha256:5a3b5db0a6e1c2e5b121e383cdbb9b56c3ed81ee963335c0c6732fb2e46d7cac
11:54:14       [exec] deleted: sha256:97b1491ac4dd5edcfad2678808f95e425dcb6f32f695241748fe07063b4d2341
11:54:14       [exec] deleted: sha256:4403f9bcad93abf3fa8df0ed83ab0257bb07d7b770dd36448cb42f29f61ee88d
...
```

sophia-guo:
@llxia What's the expected behaviour if buildlist = external/ ?

llxia:
> What's the expected behaviour if buildlist = external/ ?

If buildlist = external,  `docker_prune ` will be called for each subdir/build.xml. I think it is fine as we do not have the case where we run buildlist = external in one build. Even if we do, `docker_prune ` should not take too long.

sophia-guo:
I tried and no test was running at all. Though it's another issue.

PR #1896, opened 2020-07-16
Excluding Zip64SizeTest on linux-ppc64le until we can run it

adamfarley:
One of our primary ppc64le machines (build-osuosl-aix71-ppc64-1) has
a ramdisk with a 2gb file size limit, which appears to be the reason
that this test keeps failing on this machine.

An infra issue has been raised, and it will be resolved some time
after GA. Excluding the test until then.


adamfarley:
Thanks @ShelleyLambert

Issue #1895, opened 2020-07-16
Test builds leave Git workspace on master


AdamBrousseau:
**Describe the bug**
A few recent changes have caused our disk usage on master to increase substantially.
1. #1838 **Use ADOPTOPENJDK_REPO and ADOPTOPENJDK_BRANCH param in job configure**. This changed the behaviour of the test job launches to no longer use lightweight checkout. The effect is that we no longer pull the single pipeline script [JenkinsfileBase](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/JenkinsfileBase) onto master. But rather we clone the entire test Git repo which totals around 175M (as of today).
2. The compounding factor is the way the parallel tests run. Now that test jobs spawn multiple child jobs, each with a new name, they each leave another copy of the test repo in a workspace on master.

This is where we're at today on one of our internal servers.
PR1838 was merged on June 25. Parallel testing had been running partially for a little while before this if I recall correctly.
June 25th:
2426 Test jobs
Totaling 208GB disk usage on master. This includes both the workspaces and the console log history which we had set at 30.

Today July 15:
1565 Test jobs (we cleaned up alot of stuff)
Totaling 275GB of disk usage **only** for the workspaces on master (no logs included in calculation).

Given that this instance currently is capped at 500GB storage and we have other builds on it, we have very little room right now.

What are our options?
1. Revert back to lightweight checkout. This would break being able to use a parameter variable for the scm checkout (see 1838)
2. Use a custom refspec for the pipeline clones. I'm not sure this would work as we may have multiple use cases which may not all fit into 1 refspec. TBD
3. Use shallow clone. Ie. Don't clone the history of the repo, just the latest of whatever branch we need. This seems like the best option to start testing in my opinion.

smlambert:
Thanks for this investigation and synopsis @AdamBrousseau - I agree that option 3 looks like the best starting point.

llxia:
Thanks @AdamBrousseau . I will update the script to do shallow clone.

llxia:
I am going reopen this issue as git clone shallow does not seem to save much space. I am not sure what exact cmd Jenkins use, but my local git clone only uses 87M vs Jenkins 174M.

[...]

PR #1898, opened 2020-07-16
Set shallow to true for git clone

llxia:
Fixes: #1895

### DIFF
--- a/buildenv/jenkins/testJobTemplate
+++ b/buildenv/jenkins/testJobTemplate
@@ -201,6 +201,7 @@ ARCH_OS_LIST.each { ARCH_OS ->
 										cleanBeforeCheckout()
 										pruneStaleBranch()
 										cloneOptions {
+											shallow(true)
 											timeout(60)
 										}
 									}


llxia:
Grinder/9004
-> generated child job Grinder_testList_1/91
```
 > git fetch --tags --force --progress --prune --depth=1 -- https://github.com/llxia/openjdk-tests.git +refs/heads/*:refs/remotes/origin/* # timeout=60
```

llxia:
Note: we will need to approve testJobTemplate file at Adopt Jenkins once this PR is merged.

smlambert:
As we have completed all releases (for all versions and all implementations), lockdown of repos is lifted, and I will merge now (and approve script at Adopt).

smlambert:
Script approval at Adopt completed and verified via https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3603

AdamBrousseau:
It's probably not default because the default checkout is lightweight which is lighter and faster than a shallow clone.

Did we test to see what the disk usage is for the workspace using a shallow clone?

AdamBrousseau:
My test shows the workspace is reduced only by a few MB to 172M so I think this solution is out :(

AdamBrousseau:
You might be able to get this to work with a custom refspec as well as the shallow and maybe also setting the "honour refspec on initial clone" but you'd have to play around with it. The problem may be that that combination of options may not fit everyone's use case for the pipelines.

llxia:
> It's probably not default because the default checkout is lightweight which is lighter and faster than a shallow clone.

Unfortunately, the l[ightweight default](https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.workflow.CpsScmContext.lightweight) is false.

> My test shows the workspace is reduced only by a few MB to 172M so I think this solution is out :(

what cmd did you run?

AdamBrousseau:
I tested with a test job generated using the new shallow option. I removed the workspace on master then ran the test job and then checked the workspace usage on master.

llxia:
ok, thanks. I do not have access to master and not able to run cmd on it through a job, so I was wondering what you did. I can play with other options, but I may need your help to check the space on the master.

PR #1910, opened 2020-07-22
Add triggers based on regular intervals

sophia-guo:
Add triggers based on regular intervals, which gives the choice to generate built periodically jobs.

Signed-off-by: Sophia Guo sophiag@ca.ibm.com


sophia-guo:
This won't affect builds which would like to be triggered either manually or by upstream builds.

sophia-guo:
Jobs under https://ci.adoptopenjdk.net/view/Test_openjdk_weekly/ are not running due to the trigger is missing. If we do like those jobs to be running we can re-generate them by setting Parameter : TRIGGER_SCHEDULE.

PR #1918, opened 2020-07-27
Retry if curl return code is 18 indicating partial download

adamfarley:
Sometimes, when curl tries to download Jtreg, we get return code 18.

This indicates the download only got a partial file.

With this fix, in the event of a partial download we wait 20 minutes
and attempt to download the file again.

This provides additional tolerance of transitory network
issues, etc.

### DIFF
--- a/openjdk/build.xml
+++ b/openjdk/build.xml
@@ -39,9 +39,40 @@
 					</then>
 				<else>
 					<echo message="Getting Jtreg from https://ci.adoptopenjdk.net/job/jtreg/lastSuccessfulBuild/artifact/jtreg-4.2.0-tip.tar.gz"/>
-					<exec executable="curl" failonerror="true">
+					<exec executable="curl" failonerror="false" output="curl.download.jtreg.output.tmp.file" resultproperty="curl.download.jtreg.return.code">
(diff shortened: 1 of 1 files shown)


adamfarley:
Note: I wasn't able to reproduce the issue after adding the last set of echo messages, but we know that the core functionality works because it successfully solved the problem before said messages were added.

https://ci.adoptopenjdk.net/job/grinder_sandbox_iteration_5/2/console
https://ci.adoptopenjdk.net/job/grinder_sandbox_iteration_9/2/console

sophia-guo:
Looks like the logic is a little bit complex. The main idea is if download is partial sleep 20 mins and retry. May we just check if the return code is not zero then delete the file, sleep 20 mins and retry, which is brutal but simple?

smlambert:
My leaning on this one is to close it without delivering for several reasons:

1. The download issue was only seen for a few days and has not returned

1. This does not address but only masks the core issue, which is not understood

1. If the problem returns, this may not actually properly work around the problem (as in after 60 minutes of waiting we may still fail to download the complete file)

1. A different approach may be to have a back-up location to pull jtreg from in the event the initial URL fails.  This could be accomplished by pushing the jtreg job successful artifacts elsewhere and see if it can be pulled from there, if the direct job will not serve it up properly.

adamfarley:
Closing as advised.

As for Shelley's comments, if the problem doesn't occur again, then it's not a problem. If it does, I'm in favor of option 4.

PR #1921, opened 2020-07-27
Setting java/foreign/TestNative.java as native type

### LINKED ISSUE #1920: Set java/foreign/TestNative.java as a separate test case
java/foreign/TestNative.java depends on the testImages. For cases there is no testimages available it should be labeled as <type>native</type>.

sophia-guo:
Exclude it in Problemlist and add as separate native type. So it will be
run by sanity and will not run by sanity.regular.

Fix #1920 


sophia-guo:
related with https://github.com/AdoptOpenJDK/run-aqa/issues/24

PR #1922, opened 2020-07-28
Set JDK_REPO for zos as is if it's using ssh.

sophia-guo:


sophia-guo:
fix runtimes/backlog/issues/365

PR #1925, opened 2020-07-29
Re-include hidden classes related testing

pshipton:
Originally excluded via #1796

This testing needs to pass for the Java 15 release, or be excluded as
not-applicable to OpenJ9.

Although these tests don't all yet pass, I think they should be running since they need to be addressed for the Java 15 release. The OpenJDK testing already has 21 or more other failures, adding these won't impact the overall status.

A grinder shows the current status https://ci.eclipse.org/openj9/job/Grinder/1017


pshipton:
@smlambert although only 2 of the tests pass, I think this is ready to be merged.

pshipton:
@hangshao0 fyi

PR #1934, opened 2020-08-11
Excluding some hidden class tests on jdk15 for j9

adamfarley:
Because the JEP required for these tests to work (371) has not
yet been fully implemented in OpenJ9.

See here for more details:
https://github.com/eclipse/openj9/issues/9328


adamfarley:
Thanks

PR #1939, opened 2020-08-19
Add ProblemList_openjdk16.txt based on jdk15

milderhc:
Add JDK 16 problem list based on the content of ProblemList_openjdk15.txt.


milderhc:
@smlambert added!

PR #1942, opened 2020-08-26
Revert "Excluding some hidden class tests on jdk15 for j9"

pshipton:
Reverts AdoptOpenJDK/openjdk-tests#1934

Hidden classes is mostly implemented, and expected to be completed soon. We're coming up on Milestone 2 at the end of the week. We shouldn't be hiding failures, otherwise we'll end up shipping with problems.

### DIFF
--- a/openjdk/ProblemList_openjdk15-openj9.txt
+++ b/openjdk/ProblemList_openjdk15-openj9.txt
@@ -91,12 +91,6 @@ java/lang/invoke/ArrayConstructorTest.java	https://github.com/AdoptOpenJDK/openj
 java/lang/invoke/CallerSensitiveAccess.java	https://github.com/eclipse/openj9/issues/6768	generic-all
 java/lang/invoke/ClassSpecializerTest.java	https://github.com/AdoptOpenJDK/openjdk-tests/issues/1297	generic-all
 java/lang/invoke/DefineClassTest.java		https://github.com/AdoptOpenJDK/openjdk-tests/issues/1297	generic-all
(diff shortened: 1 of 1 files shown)


pshipton:
@smlambert pls review.

hangshao0:
Dan is suggesting excluding TypeDescriptorTest.java and BasicTest.java for Java 15.  

https://github.com/ibmruntimes/openj9-openjdk-jdk15/pull/21#issuecomment-682030111

These tests are failing simply because openj9 and RI choose a slightly different format for the identifier appended to hidden/anon class name. Nothing concerning.

smlambert:
re: https://github.com/AdoptOpenJDK/openjdk-tests/pull/1942#issuecomment-682039311 in that case, lets create another PR for those exclusions.  I wonder if the test material itself should be improved (suggest a fix to OpenJDK) otherwise, this exclusion would belong in the set of 'permanent exclusions' due to test case material not being applicable to an OpenJ9 implementation.

hangshao0:
> I wonder if the test material itself should be improved

Yes, it is actually a test issue. The JEP says the format of identifier is chosen by JVM implementation, but the test is check for the exactly same format used by hotspot. 

> this exclusion would belong in the set of 'permanent exclusions' due to test case material not being applicable to an OpenJ9 implementation.

For Java 15, we can exclude them, as we don't have much time left before the release. Dan suggests changing the format of OpenJ9 so that we can pass these tests in a later release, so they won't be in 'permanent exclusions'.

PR #1944, opened 2020-08-27
Exclude Hidden class BasicTest and TypeDescriptorTest

hangshao0:
These tests assert identifier appended to hidden class name is the 
same format as hotspot. OpenJ9 has a slightly different format.

Signed-off-by: Hang Shao <hangshao@ca.ibm.com>


hangshao0:
@ShelleyLambert

pshipton:
See also https://github.com/AdoptOpenJDK/openjdk-tests/pull/1942#issuecomment-682039311 and following comments.

PR #1970, opened 2020-09-17
update arm platform and arch

sophia-guo:


sophia-guo:
Both openjdk-build and openjdk-api3 using arm as arch property. Update from aarch32 to arm in this repo to make it consistent. This won't affect Openj9 SPEC. Grinder jobs and job generator configuration need to update platform description.

smlambert (approved):
thanks @sophia-guo - we will wait and merge after all the releases are out :) at which point we can update the Grinder parameter comments as you suggested.

smlambert:
I've updated all the Grinder and Auto_gen jobs https://ci.adoptopenjdk.net/view/Test_grinder/ with the new value.

PR #1988, opened 2020-10-01
Remove the logic of using master node to get user email

llxia:
We use master node to get user email in Grinder (PERSONAL_BUILD is
enabled), but some Jenkins may not have master. Remove the logic of
using master node to get user email. We will get user email later when
the job is scheduled on a node.

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -714,6 +714,12 @@ def getJDKImpl(jvm_version) {
 }
 
 def addJobDescription() {
+	if (params.PERSONAL_BUILD) {
+		// update build name if personal build
+		wrap([$class: 'BuildUser']) {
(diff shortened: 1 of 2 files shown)


llxia:
Grinder:
- openj9 jenkins: https://ci.eclipse.org/openj9/view/Test/job/Grinder/1082
- adopt jenkins: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4024/
- internal jenkins: Grinder/10990/

llxia:
The only drawback of this PR is that the parent job of the Grinder of Grinders will not have user email on the job description. The child jobs will still have user email. The reason is that the parent job of the Grinder of Grinders only triggers parallel runs. It does not run on a specific node.

For example: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4025/

smlambert (approved):

PR #2001, opened 2020-10-08
Remove some java/foreign excludes

pshipton:
See https://github.com/eclipse/openj9/issues/10828

These tests were excluded due to intermittent failures, which may have been machine issues. If there was a problem, it may be fixed now. We can remove the excludes and see how it goes before deciding what to do.


karianna (approved):
LGTM

PR #2007, opened 2020-10-08
if RELEASE_TAG is set checkout RELEASE_TAG

### LINKED ISSUE #2006: OpenJDK tests ignore RELEASE_TAG value
https://ci.adoptopenjdk.net/job/Test_openjdk8_j9_sanity.openjdk_aarch64_linux/62/rebuild/parameterized has RELEASE_TAG=openj9-0.23.0 passed in, but we git clone `openj9` branch

```
git clone --depth 1 -q -b openj9 https://github.com/ibmruntimes/openj9-openjdk-jdk8 openjdktemp
```
It looks like instead of using `RELEASE_TAG`  value, we are using `JDK_IMPL` value.

Suspect there is a bug in https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/openjdk/build.xml

sophia-guo:
Fix #2006 

### DIFF
--- a/openjdk/build.xml
+++ b/openjdk/build.xml
@@ -109,9 +109,12 @@
 						</exec>
 					</then>
 					<else>
-						<echo message="git clone --depth 1 -q -b ${env.JDK_BRANCH} ${JDK_REPO_SPECIFIC} openjdktemp" />
+						<condition property="branch" value="${env.RELEASE_TAG}" else="${env.JDK_BRANCH}">
+							<isset property="env.RELEASE_TAG"/>
+						</condition>
+						<echo message="git clone --depth 1 -q -b ${branch} ${JDK_REPO_SPECIFIC} openjdktemp" />
 						<exec executable="git" failonerror="true">
-							<arg line="clone --depth 1 -q -b ${env.JDK_BRANCH} ${JDK_REPO_SPECIFIC} openjdktemp" />
+							<arg line="clone --depth 1 -q -b ${branch} ${JDK_REPO_SPECIFIC} openjdktemp" />
 						</exec>
 					</else>
 				</if>


llxia:
Do we use `RELEASE_TAG ` for something else in the test pipeline? If not, should we set JDK_BRANCH to `RELEASE_TAG ` value `${buildConfig.SCM_REF}` at upstream script?
https://github.com/AdoptOpenJDK/openjdk-build/blob/a37d55290568c70fe07914609ddd639c3407bc7d/pipelines/build/common/openjdk_build_pipeline.groovy#L191

https://github.com/AdoptOpenJDK/openjdk-build/blob/a37d55290568c70fe07914609ddd639c3407bc7d/pipelines/build/common/openjdk_build_pipeline.groovy#L112

In this way, we will reduce one parameter from the test pipeline.

sophia-guo:
For upstream trigger story we don't need extra RELEASE_TAG( JDK_BRANCH is enough). It is only used when user want to run a grinder with specific RELEASE_TAG without explicitly set JDK_REPO and JDK_BRANCH. Though we may remove that part of logic later if JDK_REPO and JDK_BRANCH is changed to be required variable.

smlambert:
I would like to see JDK_REPO and JDK_BRANCH explicitly set by upstream jobs (that way we can remove convoluted logic used to guess which of very many options we should use based on JDK_IMPL and JDK_VERSION and/or other things).

sophia-guo:
closed by https://github.com/AdoptOpenJDK/openjdk-build/pull/2166

llxia:
PR AdoptOpenJDK/openjdk-build#2166 sets OPENJ9_BRANCH, but it does not fix this OpenJDK issue. The fix for this OpenJDK issue could be similar.

PR #2013, opened 2020-10-14
Reinstate DirectByteBufferLoadTest

lumpfish:
Signed-off-by: Simon Rushton <rushton@uk.ibm.com>

### DIFF
--- a/system/otherLoadTest/playlist.xml
+++ b/system/otherLoadTest/playlist.xml
@@ -98,8 +98,6 @@
 		</groups>
 	</test>
 	
-	<!-- Temporarily exclude from Linux aarch64 for: https://github.com/eclipse/openj9/issues/3065 -->
-	<!-- Temporarily exclude from AIX due to : https://github.com/eclipse/openj9/issues/3933 -->
 	<test>
 		<testCaseName>DirectByteBufferLoadTest</testCaseName>
 		<variations>
@@ -113,7 +111,6 @@
 		<groups>
 			<group>system</group>
 		</groups>
-		<platformRequirements>^arch.arm,^os.aix</platformRequirements>
 	</test>
 	
 	<test>


lumpfish:
Do not merge before  https://github.com/AdoptOpenJDK/openjdk-systemtest/pull/368 is merged.

Reinstating the test on ppc64_aix may cause a reoccurrance  of https://github.com/eclipse/openj9/issues/3933 - but since resolving that with the currently available diagnostics does not appear to be possible reinstating the test will mean we can assess whether that issue does still occur and if so attempt to get fresh diagnostics.  During the testing of the fix in https://github.com/AdoptOpenJDK/openjdk-systemtest/pull/368 the test ran successfully on ppc64_aix.

Mesbah-Alam:
Should this PR also un-exclude the [NioLoadTest](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/system/otherLoadTest/playlist.xml#L149)  ?

lumpfish:
> Should this PR also un-exclude the [NioLoadTest](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/system/otherLoadTest/playlist.xml#L149) ?

I haven't looked into NioLoadTest yet.  If it suffers from the same issue then the NioLoadTest test case would need the same fix as the DirectByteBufferLoadTest (introduce a 1ms delay between running tests) before reenabling in playlist.xml.  (So I'd create a separate PR.)

Mesbah-Alam:
Thanks, yes, NioLoadTest also failed for the same issue, as evident in the comment [here](https://github.com/AdoptOpenJDK/openjdk-systemtest/issues/180#issuecomment-425526029). Would be interesting to add the 1ms delay in that test, un-exclude the test and try.

Mesbah-Alam:
Since https://github.com/AdoptOpenJDK/openjdk-systemtest/pull/368 has already been merged, I'll go ahead and merge this one too.

PR #2018, opened 2020-10-15
Excludes for known failures in hs jdk15

smlambert:
Exclude java/foreign/TestMismatch on arm32
Exclude 2 jdk_jdi testcases on mac 
Signed-off-by: Shelley Lambert <slambert@gmail.com>

### DIFF
--- a/openjdk/ProblemList_openjdk11-openj9.txt
+++ b/openjdk/ProblemList_openjdk11-openj9.txt
@@ -144,7 +144,6 @@ sun/management/HotspotThreadMBean/GetInternalThreads.java	https://github.com/Ado
 # jdk_math
 
 java/math/BigInteger/LargeValueExceptions.java	https://github.com/eclipse/openj9/issues/9082	linux-aarch64
-java/math/BigInteger/PrimeTest.java https://github.com/AdoptOpenJDK/openjdk-tests/issues/440 linux_arm
 
 ############################################################################
 
--- a/openjdk/ProblemList_openjdk12-openj9.txt
+++ b/openjdk/ProblemList_openjdk12-openj9.txt
@@ -169,7 +169,6 @@ sun/management/HotspotThreadMBean/GetInternalThreads.java	https://github.com/Ado
 
 # jdk_math
 
-java/math/BigInteger/PrimeTest.java https://github.com/AdoptOpenJDK/openjdk-tests/issues/440 linux_arm
 
 ############################################################################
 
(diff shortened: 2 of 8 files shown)


smlambert:
Grinders:
jdk15 hs mac jdk_jdi: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4220
jdk15 hs arm32 jdk_foreign: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4221

karianna (approved):
lgtm

smlambert:
The grinder shows that TestMismatch is still being included in the linux_arm run, so need to understand why (2 possible reasons, 1 is that I am not using the right OS-ARCH tags, linux-aarch32 in the exclude file ... other is that this test has relatives/dependencies that also need to be excluded, for it to be properly excluded from jdk_foreign target).

On the mac note, I see that 3 other tests failed (while the 2 that were excluded are no longer showing up in the list as fails).  They all do seem to be related to the same intermittent attach api issue seen on a set of macos1012+ machines, so I am afraid that the temp hack of excluding all tests related to this infra issue is going to take us down a giant rabbit hole.  

I'll spend a bit more time on this, in case there are other things I should include in the excludes before its merged.

smlambert:
Tests need to be excluded under linux-arm not linux-aarch32 (thanks for checking @sophia-guo!) 
New Grinder shows this to be the case: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4236

Given this revelation, that means we have been running a bunch of tests on arm32 that I previously thought had been disabled.  Since they have been running and passing, I will go ahead and remove the entries from the problem lists (exclude files).

smlambert:
jdk8 sanity.openjdk on arm32 after removing refs to linux-aarch32 (which were not excluding tests anyway): https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4240

smlambert:
Will wait for the sanity.openjdk grinder to complete, and then close a bunch of the arm32 openjdk-tests issues upon results.

smlambert:
Some of the test cases are not in targets that are in sanity.openjdk (but rather in targets that are now in extended.openjdk):
jdk_nio grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4243/
jdk_net grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4242/

smlambert:
Additionally removed references to linux_arm in openj9 problem list files (for 2 reasons, there is no arm32 build for openj9, linux_arm would have had to have been linux-arm to be recognized by jtreg as a OS-ARCH to disable).

@sophia-guo - can you take one last look at this PR?

sophia-guo (approved):
LGTM

PR #2020, opened 2020-10-15
Exclude ClassLoader/deadlock/DelegateTest, using incompatible -Xlog

pshipton:
java/lang/ClassLoader/deadlock/DelegateTest.java is using a -Xlog option
which isn't understood by OpenJ9, resulting in a error to start the JVM.
Previous to https://github.com/eclipse/openj9/pull/10873 the option
caused a warning, but now it causes an error.

A future change may turn it back into a warning, however for now the
test needs to be excluded as it's failing across all platforms.

Issue https://github.com/eclipse/openj9/issues/10898

Signed-off-by: Peter Shipton <Peter_Shipton@ca.ibm.com>


pshipton:
@smlambert or @llxia for review pls.

llxia (approved):
LGTM

PR #2022, opened 2020-10-19
Do not wait for active docker nodes

adamfarley:
Docker nodes are often spun up as needed, so waiting for one to
become active will result in us waiting forever.

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -800,10 +800,12 @@ def run_parallel_tests() {
 	if (params.PARALLEL && params.PARALLEL != "None") {
 		stage ("Parallel Tests") {
 			def childJobs = parallel parallel_tests
-			if (params.ACTIVE_NODE_TIMEOUT && params.ACTIVE_NODE_TIMEOUT.isInteger()) {
-				waitForANodeToBecomeActive("$LABEL", params.ACTIVE_NODE_TIMEOUT)
-			} else {
-				waitForANodeToBecomeActive("$LABEL", "0")
+			if (!(params.DOCKER_REQUIRED || "$LABEL".contains("docker"))) {
+				if (params.containsKey("ACTIVE_NODE_TIMEOUT") && params.ACTIVE_NODE_TIMEOUT.isInteger()) {
+					waitForANodeToBecomeActive("$LABEL", params.ACTIVE_NODE_TIMEOUT)
+				} else {
+					waitForANodeToBecomeActive("$LABEL", "0")
+				}
 			}
 			node("$LABEL") {	
 				childJobs.each {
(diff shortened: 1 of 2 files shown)


adamfarley:
Creating this PR with the expectation that it will not be merged until after the October 2020 release.

karianna (approved):
LGTM - this will likely speed up the release

karianna:
@ShelleyLambert your call to land this or not

smlambert:
I do not actually think this PR is required, given default of ACTIVE_NODE_TIMEOUT is zero (and we do not intend to change the default) and DOCKER_REQUIRED does not mean spin up a docker container to run tests in (which we have not yet added as a 'feature').  

DOCKER_REQUIRED is appended to the overall LABEL to find machines to run external test group on (example: to then look for an online machine labelled 'ci.role.test&&hw.arch.x86&&sw.os.linux&&sw.tool.docker').

adamfarley:
@ShelleyLambert - Unfortunately, active node timeout of 0 doesn't mean we don't check for a node. It means that if we fail to find an active node that meets our needs, we wait 0 seconds before running this:

```
assert false : "Cannot find any machine that matches the LABEL: " + LABEL
```

adamfarley:
And my intention with the dual-prong docker check (which I could have done a clearer job explaining, I admit) was to try and cover all possibilities (e.g. a user using a docker label but failing to tick the box, and the box being ticked but the labekl's add-on not being added on for some reason).

If you feel that's over-complicating things, I'm content to remove the DOCKER_REQUIRED check, as the label should be in its final form by the queue stage.

smlambert:
re: https://github.com/AdoptOpenJDK/openjdk-tests/pull/2022#issuecomment-712266394 I am well aware.  

As I have stated on numerous occasions, that is how we have been doing things for almost 2 months now (without https://github.com/AdoptOpenJDK/openjdk-tests/pull/2012).  The code that asserts and test pipeline exits if no machine online with a label that matches. This approach has so far, worked without incident.  We have not seen a case where a timeout is needed (that a machine may be moments away from being online... the typical use case is that someone has provided an incorrect label and the right thing to do is exit immediately).  

However, it was proposed that there could somewhere be environments where machines go on and offline, that a few minute wait may buffer us from.  No one to date has pointed me at examples to these environments, or measured what amount of wait time will work in those environments.  So we have solved for a problem that may or may not exist.  What we may have done with these changes is created a condition where dev teams will complain that testing takes too long and off-loaded handling of infra instability into test scripts, instead of stabilizing the infra layer.  But here we are.

Yes, preference to have clean, clear, uncomplicated code, as it will be easier to maintain.

smlambert:
Also, re: https://github.com/AdoptOpenJDK/openjdk-tests/pull/2022#issue-505916607 

> waiting for one to become active will result in us waiting forever

I do not believe that is true, we will either not wait at all, or wait the amount of time set by ACTIVE_NODE_TIMEOUT

adamfarley:
> I do not believe that is true, we will either not wait at all, or wait the amount of time set by ACTIVE_NODE_TIMEOUT

True. Thanks for catching the mistake. Will correct the commit message.

As for your other comment, you are quite correct. If docker nodes haven't caused failures by now, we likely won't see it prior to the release.

Will retract the PR to clear the PR list, and hold off until after the release.

PR #2023, opened 2020-10-19
Add diff.sh to compare file contents 

Mesbah-Alam:
- Adds a script to compare two given JCK repos and find out if there are differences in test folders 


Mesbah-Alam:
Test with JCK 15 and 16 (with no difference) : runtimes-jenkins/Test_grinder/job/jckdiff/9/
Test with JCK 8 and 16 (with difference): runtime-jenkins/Test_grinder/job/jckdiff/10

jck/diff/diff.sh
```diff
@@ -0,0 +1,87 @@
...
+{
+	cd $WORKDIR
+	version=$1
+	repo=$2
+	
+	if [ -d "$WORKDIR/$version" ] ; then 
+		echo "Using existing test repo at: $WORKDIR/$version"
+	else 
+		echo "Git cloning test materials from $repo..."
+		git clone $repo $version
```
llxia, line 13: Please use `git clone --depth 1 -q ...` as we do not need git history and download progress.

jck/diff/diff.sh
```diff
@@ -0,0 +1,87 @@
...
+	if cmp -s $WORKDIR/$dirName-$VERSION_VALUE1.lst $WORKDIR/$dirName-$VERSION_VALUE2.lst; then 
+		echo "SAME : '$dirName' folder identical in both given versions: $VERSION1 & $VERSION2"
+	else 
+		diff $WORKDIR/$dirName-$VERSION_VALUE1.lst $WORKDIR/$dirName-$VERSION_VALUE2.lst > $WORKDIR/$dirName-diff.log 
+		echo "DIFFERENT : '$dirName' folder content are different in two given versions: $VERSION1 & $VERSION2"
+		echo "Please manually investigate the following differences in the two given repositories:"
+		cat $WORKDIR/$dirName-diff.log
+	fi 
+
+	rm -rf $WORKDIR/*.lst
```
llxia, line 45: We may want to keep it for debugging purposes.

llxia:
A couple of things for Jenkins job:
- workspace need to be cleaned before and after the job as the size of the test material is quite big
- build should fail if there is any difference
- in case of failure, we need to archive the results

jck/diff/diff.sh
Mesbah-Alam: Updated. Thanks for pointing out!

jck/diff/diff.sh
Mesbah-Alam: Moved that line inside setup(), so the *.lst files will remain in workspace after the process completes.

Mesbah-Alam:
Hi @llxia , 

Thanks for the recommendations you made for the Jenkins job.

> workspace need to be cleaned before and after the job as the size of the test material is quite big
Done! 

> build should fail if there is any difference
> in case of failure, we need to archive the results 
Done! 
Example of build failing when there are differences found in two given repositories (logs archived): Test_grinder/job/jckdiff/29/

I've added some more checks too: 
- Fail gracefully when wrong repository value provided: Test_grinder/job/jckdiff/30/console
- Fail when two identical repositories are provided: Test_grinder/job/jckdiff/32/console

Mesbah-Alam:
This is an example when we have two repos that don't have differences: Test_grinder/job/jckdiff/31/console

jck/diff/diff.sh
Mesbah-Alam: Actually, I've changed the code further after the above comment. The script is no longer generating any *.lst files. All text files it generates are *.txt (easily viewable from jenkins build page when archived).

llxia (approved):
Thanks @Mesbah-Alam . We will merge it after the release

smlambert (approved):

PR #2025, opened 2020-10-22
Allow option to skip TCK git update

jennli:
We have a use case in Azure DevOps where we store unzipped TCK material using AzDo Artifact instead of git repo, so we cannot run git commands against that folder. This PR allows the option to skip running git update commands if the `env.SKIP_JCK_GIT_UPDATE` env var is defined.


karianna (approved):
LGTM for merge after release

Mesbah-Alam (approved):
LGTM

PR #2038, opened 2020-11-04
Use disable variation for the mathLoadTest and otherLoadTest

renfeiw:
- replace comments with disable statements


renfeiw:
tested internally: Grinder_TKG/829

llxia (approved):
LGTM

llxia:
Will forgo the PR builds as these are `special` level changes and will not be tested in PR builds.

PR #2039, opened 2020-11-04
get.sh: allow tarball extraction with Solaris tar

sxa:
Solaris `tar` does not support `-C` on extraction. Note that while there's a similar line a bit earlier that looks like it should be similarly modified:
- I believe that's only applicable for OpenJ9 which has the debugimage packages, and we don't have OpenJ9 on Solaris
- It also has `--strip=1` which is a GNU specific option which is a little harder to workaround :-)


sxa:
I'm fairly sure the changes to `build.xml` could be impemented in a better way, so very happy to entertain better suggestions - if I could use the full path to the current directory then this would be preferable:
```xml
    <exec executable="tar" failonerror="true" diur="${DEST}>
        <arg line="xf ${PWD}/jtreg-4.2.0-tip.tar" />
```
It's also the case that the `failonerror="true"` here doesn't appear to work - it wasn't obvious to begin with what was failing as it wasn't obviously aborting on this failure - the jobs just failed to find the stuff extracted from the jtreg tarblall later on.

But for now, this change (and the associated TKG PR) allows Solaris to run the tests (machine currently chugging through https://ci.adoptopenjdk.net/job/Grinder/4567/console quite happily)

openjdk/build.xml
```diff
@@ -53,12 +53,12 @@
...
 			<then>
 				<exec executable="tar" failonerror="true">
-					<arg line="xfo jtreg-4.2.0-tip.tar -C ${DEST}" />
+					<arg line="xfo jtreg-4.2.0-tip.tar" />
 				</exec>
 			</then>
 		<else>
-			<exec executable="tar" failonerror="true">
-				<arg line="xf jtreg-4.2.0-tip.tar -C ${DEST}" />
+			<exec executable="sh" failonerror="true">
```
karianna, line 60: Worth commenting why we're not using the tar task

perf/liberty/scripts/bin/common_utils.sh
```diff
@@ -719,8 +719,8 @@ calculateFootprint()
       # now find the footprint figure from the awk output
       local FOOTPRINT=`awk '/JTC footprint/ { print $8 }' footprintAwk.tmp`
       # TODO: Once verified, remove variables not used
-      #local free=`vmstat -l | sed -e 's/  */ /g' |tr ' ' '\n' | tail -n 1 | tail -c 10`
-      #active=`vmstat -l | sed -e 's/  */ /g' |tr ' ' '\n' | tail -n 2 | head -n 1| tail -c 10`
+      #local free=`vmstat -l | sed -e 's/  */ /g' |tr ' ' '\n' | tail -1 | tail -c 10`
```
karianna, line 722: Can this commented out code be removed now?

perf/liberty/scripts/bin/common_utils.sh
```diff
@@ -954,8 +954,8 @@ getPreBenchmarkHugePagesInUse()
 "
   case ${PLATFORM} in
     Linux)
-      local HPPRETOTAL=`cat /proc/meminfo | grep HugePages_Total | sed 's/HugePages.*: *//g' | head -n 1`
-      local HPPREFREE=`cat /proc/meminfo | grep HugePages_Free | sed 's/HugePages.*: *//g' | head -n 2|tail -n 1`
+      local HPPRETOTAL=`cat /proc/meminfo | grep HugePages_Total | sed 's/HugePages.*: *//g' | head -1`
+      local HPPREFREE=`cat /proc/meminfo | grep HugePages_Free | sed 's/HugePages.*: *//g' | head -2|tail -1`
```
karianna, line 958: whitespace around `|`?

perf/liberty/scripts/bin/sufp_benchmark.sh
```diff
@@ -606,7 +606,7 @@ startupFootprint()
         local ServletSU=${SUTime}
         calcTime "${SERVER_DIR}/logs/${TRACE_FILE}" ${LIBERTY_START_TIME} "EJB_500" ${SERVER_NAME} NoPrint
         local HugeEJBSU=${SUTime}
-        local max=`echo "${ServletSU}\n${HugeEJB}"|sort -n|tail -n 1`
+        local max=`echo "${ServletSU}\n${HugeEJB}"|sort -n|tail -1`
```
karianna, line 609: whitespace `|`

karianna (changes requested):

openjdk/build.xml
sxa: I think we'd have to comment on it in lots of places :-) I'm trying to do the minimal required here to give me something that works without being inconsistent with the rest of the file in order to validate this change on it's own.

perf/liberty/scripts/bin/common_utils.sh
sxa: Quite probably, unless it's useful as a reference to someone.
Looks like it was put there by @piyush286 a [couple of years ago](https://github.com/AdoptOpenJDK/openjdk-tests/pull/1128)

sxa:
Updated following reviews. Re-running test on Solaris/SPARC at https://ci.adoptopenjdk.net/job/Grinder_TKG/36/console

openjdk/build.xml
sxa: Also, there's no indication in https://docs.ansible.com/ansible/latest/collections/ansible/builtin/unarchive_module.html that it can handle an equivalent of the strip operation

sxa:
@ShelleyLambert It doesn't look like we have an option in https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/openjdk_tests for Solaris/X64, only SPARC - is that an oversight or is there work to be done before it can be added?

perf/liberty/scripts/bin/common_utils.sh
piyush286: We can remove it. Thanks Stewart!

sxa: @piyush286 Are those comments still required or is it ok if I delete them now?

karianna (approved):

karianna:
Happy to land if @ShelleyLambert is good with it.

perf/liberty/scripts/bin/common_utils.sh
piyush286: @sxa We can remove those commented out lines.

smlambert (approved):
LGTM, and yes, we will need to add Solaris/x86 to the map in openjdk_tests.  

Was it an oversight, perhaps, if by oversight you mean, we did not realize there was a build for that platform, so it was never added...  we will likely also need to update auto-detect code in TKG and likely some update to systemtest test, but that can all be done after this PR lands.

PR #2056, opened 2020-11-16
Use JDK_IMPL_BASE to distinguish implementation

joeyleeeeeee97:
Use base JDK implementation to trigger tests.
Issue: https://github.com/AdoptOpenJDK/openjdk-tests/issues/2041

### DIFF
--- a/buildenv/jenkins/testJobTemplate
+++ b/buildenv/jenkins/testJobTemplate
@@ -18,6 +18,13 @@
  * A template that defines a test job.
  */
 
+def getBaseJDKImpl(JDK_IMPL) {
+	if (JDK_IMPL == "hotspot" || JDK_IMPL == "sap" || JDK_IMPL == "corretto" || JDK_IMPL == "dragonwell") {
+		return "hotspot"
+	}
+	return JDK_IMPL
+}
+
 if (!binding.hasVariable('JDK_IMPL')) JDK_IMPL = "openj9"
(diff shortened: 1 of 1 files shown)


smlambert:
Related: https://github.com/AdoptOpenJDK/TKG/issues/115 - there is a consideration to introduce JDK_VENDOR so as to not convolute JDK_IMPL (which was originally intended to be VM implementation, which would ultimately get used in this template. 

We need not delay this change, but rather revisit the template after a change for TKG 115 is in.

karianna:
> Related: [AdoptOpenJDK/TKG#115](https://github.com/AdoptOpenJDK/TKG/issues/115) - there is a consideration to introduce JDK_VENDOR so as to not convolute JDK_IMPL (which was originally intended to be VM implementation, which would ultimately get used in this template.
> 
> We need not delay this change, but rather revisit the template after a change for TKG 115 is in.

Agreed I'd rather that we use Vendor - but we need to do a pretty major refactor across build, test, API and website for this

llxia (approved):
LGTM

llxia:
Tested in internal Test_Job_Auto_Gen_Sandbox (61 and 62) as the script needs Admin approval for running at Adopt Jenkins.

llxia:
@ShelleyLambert I will leave it to you to merge because this script needs to be approved in Adopt Jenkins. Thanks

PR #2059, opened 2020-11-18
maketest.sh: Add process cleanup for Windows at the end of testing

sxa:
Hopefully fixes https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1573 by checking for, and attempting to stop, java processes on the machine that have been created since the time that the test job was started. This should prevent process leaks across jobs which cause a number of failures either of failed tests or locked directories that cannot be cleaned up. Equivalent functionality may well be moved into TKG at a later point.

~Draft until I've verified that the functionaliy doesn't cause any new breaks~

As per https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1669#issuecomment-727863096 and https://github.com/AdoptOpenJDK/openjdk-tests/issues/2058#issuecomment-729278319

### DIFF
--- a/maketest.sh
+++ b/maketest.sh
@@ -20,6 +20,12 @@ case `uname` in
     ;;
 esac
 
+if [ "$OSTYPE" = "cygwin" ]; then
+  export JOBSTARTTIME=$(date -R)
+  echo PROCESSCATCH: Java processes which are currently on the machine:
+  powershell -c "Get-Process java | select id,processname,starttime"
+fi
+
 if [ "$#" -eq 1 ];then
 	cd $1/TKG
 	$MAKE compile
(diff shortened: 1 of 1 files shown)


maketest.sh
```diff
@@ -20,6 +20,10 @@ case `uname` in
     ;;
 esac
 
+if [ "$OSTYPE" = "cygwin" ]; then
```
karianna, line 23: should that be `==` or `eq` to avoid assignment confusion?

maketest.sh
```diff
@@ -29,3 +33,24 @@ else
 	$MAKE $@
 fi
 
+# For now (while this code is being prototyped) I only want this to be
+# done while running in jenkins, to avoid a potentially destabilising
+# change if people are running other java processes on their local system
```
karianna, line 38: This reminds me that we need to split out what is pure build and test scripts vs what is Adopt Build farm functionality.  X-team discussion required methinks :-)

maketest.sh
sxa: `eq` is numeric so inappropriate.
Could be `==` but I prefer to use the more generic POSIX standard forms wherever possible.

sxa:
FYI @willsparker :-)

maketest.sh
```diff
@@ -29,3 +35,29 @@ else
...
+set -x
+if [ ! -z "$EXECUTOR_NUMBER" ]; then
+  if [ "$OSTYPE" = "cygwin" ]; then
+    # Windows code based on https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1669#issuecomment-727863096
+    echo PROCESSCATCH: Showing java processes - one will be the jenkins agent - I will remove any of these created since $JOBSTARTTIME:
```
smlambert, line 45: My preference is to not use first person in echo statements (drop 'I will', 'I am going to', etc) and just say 'Remove any...' and 'Attempt to stop...' etc

maketest.sh
```diff
@@ -29,3 +35,29 @@ else
...
+      powershell -c "Get-Process java | select id,processname,starttime | where {\$_.StartTime -gt (Get-Date -Date \"$JOBSTARTTIME\")}"
+      powershell -c "Get-Process java | select id,processname,starttime | where {\$_.StartTime -gt (Get-Date -Date \"$JOBSTARTTIME\")} | stop-process"
+      echo PROCESSCATCH: Removed processes - here is what is remaining:
+      powershell -c "Get-Process java"
+    fi
+  else
+    # Non-Windows code from https://ci.adoptopenjdk.net/job/SXA-processCheck
+    echo PROCESSCATCH: Checking for any leftover java processes on the machine
+    if ps -fu $USER | grep java | egrep -v "remoting.jar|agent.jar|grep"; then
+      echo PROCESSCATCH: There are rogue procesess still on the machine
```
smlambert, line 59: spelling procesess -> processes

maketest.sh
```diff
@@ -29,3 +35,29 @@ else
...
+      echo PROCESSCATCH: Not killing them for now as this script should be tested more thoroughly first
```
smlambert, line 60: Not sure you need the colour commentary of 'script should be tested more thoroughly first' ... essentially if you get into this block, it is an alert that should trigger someone to manually clean-up the rogue processes

maketest.sh
```diff
@@ -29,3 +35,29 @@ else
 	$MAKE $@
 fi
 
+# For now (while this code is being prototyped) I only want this to be
```
smlambert, line 38: Could mention this is temporary location for this code that cleans up stray processes, & potentially put a reference to https://github.com/AdoptOpenJDK/TKG/issues/45 where it will live in a future iteration.

smlambert:
Grinders: [4832](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4832) - [4835](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4835), plus PR github workflows pass (with one unrelated exception on jdk15, as expected, since this code is only invoked if on Jenkins / EXECUTOR_NUMBER defined).

smlambert (approved):
Some very minor comments, overall LGTM, will await Grinders to complete (and if you wanted to update echo statements/comments or not) before merging.

maketest.sh
sxa: Awww I quite like the idea that you're having a conversation with a sentient machine ... But I'll change it -)

maketest.sh
sxa: Was more of a "TODO" (Although SXA-processCheck covers that situation)_ ... May just edit it to say that as a comment along with the reference to the TKG issue

PR #2071, opened 2020-11-26
Change Mode140 to Mode145 in special.system tests

llxia:
Related: https://github.com/eclipse/openj9/issues/11105


llxia:
Internal Grinder/12259

llxia:
FYI @Mesbah-Alam @pshipton

smlambert (approved):
LGTM

pshipton:
lgtm. Hopefully this mode is added now, I don't think it existed before.

The reason we are doing this is because gencon is the default for many years and not optthruput. If we're only going to test one mode with -Xshareclasses it might as well be the default gc policy.

PR #2079, opened 2020-12-01
Add check for jck update

Mesbah-Alam:
- Adds a check to see if jck materials on disc are already up-to-date. If yes, we simply use them as is. 
- Related to backlog/issues/405

### DIFF
--- a/jck/build.xml
+++ b/jck/build.xml
@@ -59,44 +59,27 @@
 					<arg line="rev-parse"/>
 					<arg line="HEAD"/>
 				</exec>
-				<echo message="localSHA Original ---${localSHAClean}---"/>
 				<loadresource property="localSHATrimmed">
 					<propertyresource name="localSHA"/>
 					<filterchain>
 						<tokenfilter>
 							<trim/>
 						</tokenfilter>
+						<striplinebreaks/>
 					</filterchain>
 				</loadresource>
-				<echo message="localSHA after trim --${localSHATrimmed}--"/>
-				<loadresource property="localSHAClean">
-					<propertyresource name="localSHATrimmed"/>
-						<filterchain>
-							<striplinebreaks/>
-						</filterchain>
-				</loadresource>
-				<echo message="localSHA after striplinebreaks --${localSHAClean}--"/>
-				
 				<exec executable="git" dir="${JCK_ROOT_USED}" outputproperty="remoteSHA">
 					<arg line="ls-remote"/>
 					<arg line="${JCK_GIT_REPO_USED}"/>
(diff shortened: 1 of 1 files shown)


Mesbah-Alam:
Tests: 
- When test materials already exists on disc and no update is necessary: Grinder/12349
- When test materials did not exist on disc : Grinder/1234

jck/build.xml
```diff
@@ -51,6 +51,33 @@
 	</condition>
 	<target name="stage_jck_material">
 		<!-- Starting downloading or updating JCK materials based on JCK GIT REPO and JCK VERSION-->
+		<if>
+			<available file="${JCK_ROOT_USED}" type="dir" />
+			<then>
```
llxia, line 56: By adding this, we are checking `<available file="${JCK_ROOT_USED}" type="dir" />` twice. Also, we do not need to run this code if `env.SKIP_JCK_GIT_UPDATE` is set.
This code should be added after line 103

Mesbah-Alam: > This code should be added after line 106

 - If we do this then the hard reset logic, which is in an outer block, will not be able to use the property `local-material-uptodate`. 

> we are checking <available file="${JCK_ROOT_USED}" type="dir" /> twice.

Yes, unfortunately, since the logic depends on JCK_ROOT_USED. 

Just to explain a bit more, this is what we are doing: 
~~~~~~ 

if (JCK_ROOT_USED is available) { 
 1 Obtain local and remote sha 
 2. Compare them and based on that, set the property `local-material-uptodate`  
}  // <-- Note, if we move this inside the else block, we can not use `local-material-uptodate` in the outer block later on, where we are doing hard reset. Please also note: this hard reset is executed for both fresh clone and update scenarios - which is why it is in the outer block. 

If (JCK_ROOT_USED is NOT available) { // This is the fresh clone scenario - which involves a hard reset later on  
  3. Do fresh clone of JCK repo 
} else {. // This is the update scenario 
   if (env.SKIP_JCK_GIT_UPDATE) { 
      4.  skip update 
    } else { 
             5. Run `git clean` 
             if (local-material-uptodate == false) { 
                    6. Run git pull // to update JCK repo 
              }
    }  
} 

If (on z/OS)   { 
   if ( (local-material-uptodate == false) { 
       7. run `git reset hard` 
   }
}

llxia: Discussed with @Mesbah-Alam , since there is a plan to further optimize the code soon, we will not spend time to clean up this now.

llxia (approved):
LGTM

PR #2080, opened 2020-12-02
Exclude ProcessBuilder Basic_id0 test on aix

Haroon-Khel:
ref https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1644#issuecomment-737283413

This same test has been excluded on aarch64 due to https://github.com/eclipse/openj9/issues/9032. I am excluding it on aix for the same reason. 

@smlambert Is editing this file all that is needed to exclude a test?


smlambert (approved):

smlambert:
thanks @Haroon-Khel - I did not see that we need to exclude this test for other versions (if so, you'd need to update the version specific problemlist files.

PR #2087, opened 2020-12-04
Add mixed platforms

keithc-ca:
Needed by eclipse/openj9#11166.

### DIFF
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -1,12 +1,16 @@
 #!groovy
 
-def PLATFORM_MAP = [ 
+def PLATFORM_MAP = [
     'arm_linux' : [
         'SPEC' : 'linux_arm',
         'LABEL' : 'ci.role.test&&sw.os.linux&&hw.arch.aarch32',
     ],
     'aarch64_linux' : [
         'SPEC' : 'linux_aarch64_cmprssptrs',
+        'LABEL' : 'ci.role.test&&sw.os.linux&&hw.arch.aarch64',
+    ],
+    'aarch64_linux_mixed' : [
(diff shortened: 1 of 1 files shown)


llxia:
Thanks @keithc-ca 
Just a note, ottawa.csv also needs to be updated to support the newly added platforms. Otherwise, we may have tests running with incorrect modes. I am ok with merge this change and update ottawa.csv in the future PR.

buildenv/jenkins/openjdk_tests
```diff
@@ -1,121 +1,145 @@
 #!groovy
 
-def PLATFORM_MAP = [ 
+def PLATFORM_MAP = [
     'arm_linux' : [
         'SPEC' : 'linux_arm',
-        'LABEL' : 'ci.role.test&&sw.os.linux&&hw.arch.aarch32',
+        'LABEL' : 'ci.role.test&&sw.os.linux&&hw.arch.aarch32'
```
llxia, line 6: I would like to keep the trailing comma. This was added purposely 
- if there is a new entry, this line will not change
- it prevents the error in future

llxia: Leaving trailing commas simplifies the git history. If we merge this PR, all of those lines will be updated, and when we look for one line history, we will find this PR and not the reason for the line in the first place.
Please see the article https://time2hack.com/are-you-using-trailing-commas-in-your-javascript/
The git commits are much cleaner if we add an extra comma.
I understand that this is groovy code and not JavaScript, but the same logic applies.

keithc-ca:
Commas added and rebased.

llxia:
Thanks @keithc-ca 
Internal Grinder_TKG/1003/ on xlinux

llxia (approved):
LGTM

PR #2095, opened 2020-12-07
Update userGuide with new exclude feature

renfeiw:

### DIFF
--- a/doc/userGuide.md
+++ b/doc/userGuide.md
@@ -327,16 +327,28 @@ For example:
     <variations>
       <variation>NoOptions</variation>
       <variation>-Xmx1024m</variation>
-    <variations>
+    </variations>
     ...
 ```
 To exclude the entire suite:
 
 ```auto exclude test jdk_test```
 
-To exclude the test case with variation ```-Xmx1024m```:
+To exclude the 2nd variation listed which is assigned suffix_1 ```-Xmx1024m```:
 
 ```auto exclude test jdk_test_1```
+
+To exclude the test for openj9 only:
+
+```auto exclude test jdk_test impl=openj9```
+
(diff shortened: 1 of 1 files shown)


llxia (approved):
LGTM

doc/userGuide.md
```diff
@@ -338,6 +338,18 @@ To exclude the test case with variation ```-Xmx1024m```:
 
 ```auto exclude test jdk_test_1```
 
+To exclude the test for openj9 only:
+
+```auto exclude test jdk_test impl=openj9```
+
+To exclude the test for java 8 only:
+
+```auto exclude test jdk_test ver=8```
```
smlambert, line 347: We are using ver to describe subset, is the plan to eventually match it to what it is named in the playlist (as impl to impl)?  It will be easier for people to remember usage.

doc/userGuide.md
```diff
@@ -338,6 +338,18 @@ To exclude the test case with variation ```-Xmx1024m```:
...
+
+To exclude the test case with variation ```-Xmx1024m``` against openj9 java 8 only:
```
smlambert, line 349: It may be useful to say "To exclude the 2nd variation listed which is assigned suffix _1 ```-Xmx1024m```  against openj9 java 8 only:"

doc/userGuide.md
renfeiw: Yes, the plan to change subset to ver in the future.

doc/userGuide.md
renfeiw: Updated

smlambert (approved):
thanks @renfeiw (and for the demo in today's AQAvit meeting showing each of these features listed in the doc)

PR #2101, opened 2020-12-08
Update --system to --global to avoid Admin permission

### LINKED ISSUE #2094: Update to use --global versus --system in git config command on Windows
Based on findings in https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1732#issuecomment-739840847 it seems we need to update https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/openjdk/build.xml#L87 to use --global versus --system (and check other build files for other cases).  

This is causing the 'permission denied /etc/gitconfig' seen during triage.

sophia-guo:
Close #2094 

### DIFF
--- a/openjdk/build.xml
+++ b/openjdk/build.xml
@@ -84,7 +84,7 @@
 			<contains string="${SPEC}" substring="win" />
 			<then>
 				<exec executable="git" failonerror="false">
-					<arg line="config --system core.longpaths true" />
+					<arg line="config --global core.longpaths true" />
 				</exec>
 			</then>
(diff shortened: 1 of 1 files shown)


karianna (approved):
lgtm

smlambert:
For the record, since we do not include impl=j9, test group = openjdk as a combination in PR testing atm:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5261/ (J9/jdk11/jdk_math Grinder passes).

PR #2106, opened 2020-12-10
Exclude PlatformLoggerTest.java for all platforms

pshipton:
It wasn't excluded for AIX, and fails intermittently in AIX XL builds
for the same reason it was excluded on other platforms (doesn't keep a
strong reference to the loggers, which can be gc'ed).

Signed-off-by: Peter Shipton <Peter_Shipton@ca.ibm.com>


karianna (approved):
lgtm

PR #2113, opened 2020-12-14
Fix tomee server slow start due to SecureRandom entropy source

andrew-m-leonard:
tomee server may start slowly due to poor entropy source, use
```
-Djava.security.egd=file:/dev/./urandom
```
instead.

### DIFF
--- a/external/dockerfile_functions.sh
+++ b/external/dockerfile_functions.sh
@@ -391,7 +391,7 @@ print_bazel_install() {
 print_java_tool_options() {
     local file=$1
 
-    echo -e "ENV JAVA_TOOL_OPTIONS=\"-Dfile.encoding=UTF8\"\n" >> ${file}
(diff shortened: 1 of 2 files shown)


andrew-m-leonard:
Test Grinder on godaddy node now runs complete tomee in 56mins:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5474/

smlambert (approved):
Thanks @andrew-m-leonard !  

Note: we no longer use the static Dockerfile, so the change in that file is not needed.  I will leave the change you've done in line 41, with an action that we could go in and remove all static Dockerfiles in all external tests, now that we auto-generate them via external.sh (giving us the ability to run testing on most of the image tags that get produced and pushed to hub.docker.com).

sxa:
@andrew-m-leonard Why does this PR use `/dev/./urandom` instead of `/dev/urandom`? It looks rather odd

andrew-m-leonard:
@sxa not exactly sure, but that is the suggested solution from numerous places

PR #2124, opened 2020-12-16
Exclude native test CheckHandles.java in problemlist

sophia-guo:
Exclude test in problemlist and run as a separate testcase


sophia-guo:
Related with https://github.com/AdoptOpenJDK/run-aqa/issues/61.

Test has been tagged as native and excluded by jdk_lang group in jdk11 problem list.

This change will also exclude it by jdk_lang group in problem lists of jdk11+, which will not affect currently nightly builds. Current nightly build target is sanity, which includes both regular and native.

Test:
https://github.com/sophia-guo/runaqaTest/runs/1566237622?check_suite_focus=true

smlambert (approved):

PR #2140, opened 2021-01-04
WIP: Add vendor-specific tests for jdk builds

adamfarley:
These tests ensure that any vendor-specific build changes or fixes
are present and functional in the created builds on relevant
platforms, versions, etc.

The freetype and cuda tests are included in this change, as
documented.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ functional
 !functional/security
 !functional/SyntheticGCWorkload
 !functional/MBCS_Tests
+!functional/BuildTestsForSpecificVendors
 
 Utils
 systemtest_prereqs/*
--- a/functional/BuildTestsForSpecificVendors/README.md
+++ b/functional/BuildTestsForSpecificVendors/README.md
@@ -1,0 +1,69 @@
+# Vendor-Specific Testing
+
+## Build Tests
+
+### 1) Contents
+1) Contents
+2) Purpose
+3) Usage
+4) Current Tests
+5) Adding new tests
+6) FAQ
+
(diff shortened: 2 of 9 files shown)


adamfarley:
Running various tests now and resolving issues as they spring up. Will tag people for reviews and feedback (including Shelley) once the tests are passing.

adamfarley:
Tests still running. Thought of a few improvements as well, and they're in there too. Still a couple of issues left, including not knowing how to check for cuda enablement on Windows. For some reason the "cudart" string isn't present in the j9prt file even if cuda functionality is definitely enabled during the config stage of the build.

Will double-check I'm testing the right Windows build, and raise it with the OpenJ9 folks.

aahlenst:
Are you aware of https://github.com/AdoptOpenJDK/openjdk-tests/pull/2067? Or is this PR already the right way to solve the problems discovered in aforementioned PR?

Regardless, I find it useful if those tests are comprehensive. This means we should not only test for the presence of CUDA on Windows but also for the absence everywhere else (as far as possible). This would allow us to detect changes.

smlambert:
Will combine this PR, with #2067 and house in a test dir in the openjdk-build repo (to be run as an independent smoke test which must pass in order for the AQA test jobs to be triggered).  We will leave this open form until the new combined openjdk-build PR lands.

adamfarley:
Closing this, as the parts which are going to be used are being merged into [this PR](https://github.com/AdoptOpenJDK/openjdk-build/pull/2397).

Will leave the branch hanging around for a while, in case it's needed later.

PR #2153, opened 2021-01-11
Print error message if the adopt api returns one

adamfarley:
Sometimes we get an "adoptopenjdk" file back from the adopt api
when we request a binary via curl. This change assumes that's an
error message html file and prints the contents.

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -168,6 +168,7 @@ getBinaryOpenjdk()
 	cd $SDKDIR
 	mkdir -p openjdkbinary
 	cd openjdkbinary
+	uses_api_to_get_binary="false"
 
 	if [ "$SDK_RESOURCE" != "upstream" ]; then
 		if [ "$(ls -A $SDKDIR/openjdkbinary)" ]; then
@@ -220,6 +221,7 @@ getBinaryOpenjdk()
 		if [ "$SDK_RESOURCE" == "releases" ]; then
 			release_type="ga"
 		fi
+		uses_api_to_get_binary="true"
 		download_url="https://api.adoptopenjdk.net/v3/binary/latest/${JDK_VERSION}/${release_type}/${os}/${arch}/jdk/${JDK_IMPL}/${heap_size}/adoptopenjdk"
 
 		if [[ "$JDK_VERSION" -ge "11" ]]; then
@@ -267,6 +269,12 @@ getBinaryOpenjdk()
 				_ENCODE_FILE_NEW=UNTAGGED curl -OLJSk${curl_verbosity} ${curl_options} $file
 				download_exit_code=$?
 				count=$(( $count + 1 ))
+
+				# Checking for error messages returned from the API.
+				if [ -f "adoptopenjdk" ] && [ $uses_api_to_get_binary == "true" ]; then
+				    download_exit_code=-99
+				    break
+				fi
 			done
 
 			if [ $download_exit_code != 0 ]; then
@@ -278,6 +286,11 @@ getBinaryOpenjdk()
 				    md5 $file
 				 else
 				    md5sum $file
+				fi
+
+				if [ -f "adoptopenjdk" ] && [ $uses_api_to_get_binary == "true" ]; then
+				    echo "Since we used the API to fetch a file, and we now have an adoptopenjdk file, this may contain an error message. Printing..."
+				    cat "adoptopenjdk"
 				fi
 
 				exit 1


adamfarley:
Being discussed in https://github.com/AdoptOpenJDK/openjdk-api-v3/issues/341

Do not merge until the debate is complete, please.

sxa (commented):
What do we need `uses_api_to_get_binary` for here?

adamfarley:
That was so we only assume that adoptopenjdk is an api error message file is we're using the api to return the build.

I didn't want to assume adoptopenjdk was an error file unless we're trying to get the binary from the api. That way nobody else can use adoptopenjdk as a file name and accidentally trigger the logic.

sophia-guo:
If the SDK_RESOURCE is nightly or release it's using api to get binary.  If the return is file adoptopenjdk it means jar|zip file isn't downloaded correctly, which will fail the build by following unzip step. https://ci.adoptopenjdk.net/view/Failing%20Test%20Jobs/job/Test_openjdk11_j9_extended.external_ppc64le_linux_scala/24/console

What the purpose of this PR?

smlambert (changes requested):
Presume this is for debugging purposes only and not a PR that will be merged, and that healthcheck of API would not live in this Jenkinsfile used for AQA tests, but rather in API tests in the API repo.

adamfarley:
Sophia: The purpose of this PR was to retry the download if it failed the first time or two. Since the entire code chunk already catches curl failures and retries, I believed it wasn't too big of a leap to imagine other errors handled in the same way.

Shelley: This isn't an API Healthcheck, but rather some code to give us the right error message in the event of a failure, coupled with (what I previously believed) was some improved retry logic.

Let's focus on the improved error message, and remove the bit about the retry.

adamfarley:
Changes made. In the event that we fail to download the build, we provide a useful error message and handle it like a normal failure, except now we do not attempt to retry the download.

sophia-guo:
@adamfarley got it. Do you know why the download fails and return with 0? If it's the jar unavailable or it's the network issue? I've tried with a wrong url ( change correct one _curl -OLJSks  https://api.adoptopenjdk.net/v3/binary/latest/11/ea/linux/ppc64le/jdk/openj9/normal/adoptopenjdk_ to wrong one _curl -OLJSks  https://api.adoptopenjdk.net/v3/binary/latest/11/ea/linux/ppc64le/jdk/openj/normal/adoptopenjdk_ ) and got the same file with return code 0.   If it's network issue or server issue it makes sense to retry. If it's some other reasons it may not be helpful to retry.

adamfarley:
@sophia-guo - We're having the "why is this happening" conversation here: https://github.com/AdoptOpenJDK/openjdk-api-v3/issues/341

Right now I think the problem is that the curl command *is* successfully returning a data stream from the URL in all cases, so it declares success even if we didn't get the file we wanted. I'm not sure curl knows the difference between returning the file we're after, and returning a html file containing an error message, as in both cases it's just a data stream.

smlambert:
> Right now I think the problem is that the curl command is successfully returning a data stream from the URL in all cases

I do not care for this at all, for us to compensate in the test script for what appears to be incorrect behaviour from the API.  I would prefer the API to 'fail to serve' if it has no build artifacts to serve up based on the query (and I would think that redistributors using the API would have similar expectations).

adamfarley:
This specific behavior seems to be a CloudFlare thing rather than an API thing. The two http return codes we're getting seem to imply either: "The api you're trying to access didn't return anything intelligible." or "The service you're trying to access isn't available."

The headers we've identified express clear error messages. This appears to be what "fail to serve" currently looks like.

If we'd like to change that behaviour, I expect we'd need to modify CloudFlare settings if possible. I'm not sure you can fix this in the API, especially if the problem is a genuine "service is not available", as there's no difference between a modified and unmodified api if they're both genuinely unavailable.

smlambert:
Can I get you to resolve the conflict @adamfarley ? 

Also I am removing the 'bug' label from this, as it is not addressing a test bug, but printing out more details when the API and/or is not working as expected.

smlambert (commented):
@adamfarley - another question, in your investigation you note there are 3 error codes that you are seeing, are they terminal (not transient) for a particular API query?  Or do you see a case that supports the retry?

adamfarley:
They all appear to be transient, and during a big grinder (where we download over and over again, wiping the tar.gz each time) I checked and no more than 2 seemed to be right next to each other. 

However, Stewart proposed that we should fix the underlying issue rather than work around it, as download failures make us look bad. I can see the reasoning there.

adamfarley:
Merge conflict resolved.

smlambert:
@adamfarley - I think I am going to park this PR (without merging it).  It has however been useful, thank you.

It made us aware that we should actually be using SDK_RESOURCE=upstream (instead of SDK_RESOURCE=nightly) for external tests, to ensure we are testing the 'rea'l nightly build, (as noted, when the API misbehaves, it can end up serving stale builds as nightly (which can be out of date, when the API needs to be kicked ... this would mean we may run the same tests against the same build which is not a good use of resources).    Related: https://github.com/AdoptOpenJDK/openjdk-tests/issues/1802

I will make the necessary change to pass along the UPSTREAM params to the external test jobs that get launched.

adamfarley:
@smlambert - I don't understand the resistance to including this code change.

Currently, when we are given an error message this way, we hide/ignore the error message (so the user doesn't see it), and we continue until some other bit of code figures out it's not a tar.gz file and fails with an unrelated error message that doesn't help the debugging much.

Why would we not want to display this error message?

smlambert:
Some of the reasoning I chose not include this change:
1) _won't get exercised_: Our main automation will not use this code path (so it won't inform us at the project of anything, as the path won't be run from build pipelines)
2) _real problem lives elsewhere_: It is already stated the real issue exists not in the test code, a real fix/solution will happen elsewhere.  This was debugging code that helped locate where the real issue was.  If you want to continue to verify the proper functioning of the API, then perhaps useful to add smoke or operational tests elsewhere (re: https://github.com/AdoptOpenJDK/openjdk-tests/issues/1802#issuecomment-729499753)
3) _sets up a condition for confusion and issue bloat_: For the use case that exercises the code path, personal grinders and/or local runs... there is no useful action to take with knowing the http error code.  It would only serve to confuse a developer, and best case they raise an issue against test on a problem outside of our control to fix.  
4) _for debugging, not production_: This PR helped for debugging, but I do not really wish to clutter the code base with debugging code, especially does not lead to an actionable task. 
4) _test pipelines already do the 'right thing'_: The current code states the problem clearly enough ("hey, you did not download a tar.gz file as we expected, we can not unpack it and so we fail/exit immediately").
5) _extra noise in console_: This PR adds extra output to the console (runs cat on a file called adoptopenjdk... of unknown size/content which will get printed to and add clutter the console (depending on its content, which is 'unknown', though we make an assumption its got an http error code somewhere in it).

PR #2156, opened 2021-01-11
Exclude TestAESCipher tests that modify security permissions specific to hotspot

andrew-m-leonard:
Excludes problem hotspot tests from: https://github.com/eclipse/openj9/issues/11614


andrew-m-leonard:
Test job: https://ci.adoptopenjdk.net/view/Test_openjdk/job/Test_openjdk8_j9_extended.openjdk_x86-64_linux/17/console

smlambert:
Thanks @andrew-m-leonard - Is the plan to make the fix in the extensions repo (as openj9/11614 suggests) or permanently exclude?  If permanently exclude, we will use a [openjdk-tests/1297](https://github.com/AdoptOpenJDK/openjdk-tests/issues/1297) issue in the PR, if will fix under 11614 then can leave as is).

Is this also needed in later jdk versions, or just jdk8?

sophia-guo:
Looks like it's jdk8 only. And test need to be updated to work for openj9 or need to be excluded permanently.

andrew-m-leonard:
@smlambert I think Peter's view is to permanently exclude, so i'll update to point at https://github.com/AdoptOpenJDK/openjdk-tests/issues/1297

andrew-m-leonard:
@smlambert updated... and yes jdk8 only

smlambert (approved):
thanks @andrew-m-leonard !

Issue #2199, opened 2021-01-22
Comment triggered PR builds


smlambert:
Comment triggered PR builds (related to AdoptOpenJDK/TKG#109)
-- Trigger PR testing manually using comment in the git PR
-- Allow users to specify SDK (default is nightly). Users can specify releases/nightly
-- Allow users to specify test targets (sanity.system, extended.openjdk, etc)
-- Allow other values to be specified. For example:
```
action nightly sanity.functional xlinuxlargeheap jdk11
action releases testA,testB xlinux,osx jdk8,jdk11 hotspot,openj9
```

smlambert:
@patkarns @icohedron @yanshan24 @xfxcwynlc @xichen1

smlambert:
In our call, we referenced an example of comment triggered workflow ([autoTestPR.yml](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/.github/workflows/autoTestPR.yml) which ends up calling some [python code](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/.github/workflows/autoTestPR.yml#L33) in the [TKG repo](https://github.com/AdoptOpenJDK/TKG/blob/master/scripts/testBot/disable.py))

In a similar fashion, can we trigger some PR tests (like (PushOrPRtestbuild.yml)[https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/.github/workflows/PushOrPRtestbuild.yml] from a comment trigger instead of on [push](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/.github/workflows/PushOrPRtestbuild.yml#L4) or [pull_request](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/.github/workflows/PushOrPRtestbuild.yml#L14) trigger.

patkarns:
Update: we got the arguments using a python script.

Next up, we'll trigger the PR test and check for invalid parameters (like versions).

@Icohedron

Icohedron:
GitHub Workflow implementation for comment triggered PR builds should be complete in PR #2222

PR #2222, opened 2021-02-02
Add comment-triggered PR build action

Icohedron:
Implements a GitHub Workflow for triggering PR builds. Addresses #2199 

Known issues:
- ~~The builds checkout the master branch of the base repo rather than the head repo and branch of the PR. This must be fixed. Would need to change [these lines](https://github.com/AdoptOpenJDK/openjdk-tests/pull/2222/commits/e532bd50bb6c69b55c69b1a1869239fae6a1d627#diff-ab5b4efba7d032a7065a9d242e7e1bd92b59251ab0c5c95b4a3aa6f1d0b38db7R96-R101) of the workflow file.~~ Fixed!
- ~~When there is an error in the comment, the resulting error message mentions the `action_argparse.py` script used to parse the comment. A minor issue but I see no obvious fix aside from overriding the error messages of Python's `argparse` or writing a new parser from scratch. (See example error below)~~ Fixed!
- A "Build(s) failed" error message is displayed even when no builds were started. A minor issue but I see no obvious fix. (See example error below)
- ~~A newline at the end of the comment will result in an empty error message. Additionally, **one can suffix the comment with `|| <command>` to execute any arbitrary shell command on the GitHub runner**. Security issue? I do not see a way to prevent this at the time being.~~ Fixed!
- Depending on when the workflow was cancelled, the status report comment may say "Build(s) successful!" instead of "Build(s) cancelled!" because the runBuild part of the workflow was skipped while the parseComment part of the workflow was successful.

Discussions:
- What parameters should be required, and what should the default values for non-required parameters be?
- Instead of the keyword "action", perhaps we can use a keyword that is more descriptive of what it does?

Potential improvements:
- Instead of hard-coding valid parameter values or accepting all values, we can fetch the valid parameter values from some resource (e.g. the test repository's file structure and/or Jenkins Grinder options)
- Instead of reporting a vague "Build(s) failed" message, give a break-down of which builds failed and which builds succeeded within the comment itself rather than simply giving a hyperlinked workflow ID for the user to click for more details.
- Add additional parameters for `custom_target`, `openjdk_testRepo`, and `tkg_Repo` to pass to the [run-aqa action](https://github.com/AdoptOpenJDK/run-aqa) that runs the builds.

Example usage:
image

Example error:
image

### DIFF
--- a/.github/workflows/runAqa.yml
+++ b/.github/workflows/runAqa.yml
@@ -89,7 +89,6 @@ jobs:
       fail-fast: false
       matrix: ${{ fromJson(needs.parseComment.outputs.build_parameters) }}
     steps:
-    - uses: actions/checkout@v2
     - uses: AdoptOpenJDK/install-jdk@v1
       with:
         version: ${{ matrix.jdk_version }}
@@ -113,7 +112,7 @@ jobs:
             core.setFailed(`Request failed with error ${err}`)
           }
     - name: AQA
-      uses: AdoptOpenJDK/run-aqa@v1
+      uses: sophia-guo/run-aqa@PR
       with:
          build_list: ${{ matrix.build_list }}
          target: ${{ matrix.target }}
--- a/.github/workflows/runAqa.yml
+++ b/.github/workflows/runAqa.yml
@@ -5,16 +5,16 @@ on:
 jobs:
   parseComment:
     runs-on: ubuntu-latest
-    if: startsWith(github.event.comment.body, 'action') && github.event.issue.pull_request
+    if: startsWith(github.event.comment.body, 'run aqa') && github.event.issue.pull_request
     outputs:
-      build_parameters: ${{ steps.action_argparse.outputs.build_parameters }}
+      build_parameters: ${{ steps.argparse.outputs.build_parameters }}
     steps:
     - uses: actions/checkout@v2
     - name: Parse parameters
       env:
         args: ${{ github.event.comment.body }}
-      run: python3 .github/workflows/action_argparse.py $args 2> log.txt
-      id: action_argparse
+      run: python3 .github/workflows/runAqaArgParse.py $args 2> log.txt
+      id: argparse
     - name: Output log
       if: failure()
       # Store the contents of log.txt into an environment variable and escape characters to preserve newlines and other symbols.
@@ -57,12 +57,12 @@ jobs:
         script: |
           comment_body = `
           @${{ github.actor }} Build(s) started with the following parameters:
-          - sdk_resource: ${{ steps.action_argparse.outputs.sdk_resource }}
-          - build_list: ${{ steps.action_argparse.outputs.build_list }}
-          - target: ${{ steps.action_argparse.outputs.target }}
-          - platform: ${{ steps.action_argparse.outputs.platform }}
-          - jdk_version: ${{ steps.action_argparse.outputs.jdk_version }}
-          - jdk_impl: ${{ steps.action_argparse.outputs.jdk_impl }}
+          - sdk_resource: ${{ steps.argparse.outputs.sdk_resource }}
+          - build_list: ${{ steps.argparse.outputs.build_list }}
+          - target: ${{ steps.argparse.outputs.target }}
+          - platform: ${{ steps.argparse.outputs.platform }}
+          - jdk_version: ${{ steps.argparse.outputs.jdk_version }}
+          - jdk_impl: ${{ steps.argparse.outputs.jdk_impl }}
 
           Workflow Run ID: [${{ steps.workflow_run_info.outputs.id }}](${{ steps.workflow_run_info.outputs.url }})
           `;
@@ -74,13 +74,13 @@ jobs:
           })
     - name: Echo parameters
       run: |
-        echo build_parameters: ${{ steps.action_argparse.outputs.build_parameters }}
-        echo sdk_resource: ${{ steps.action_argparse.outputs.sdk_resource }}
-        echo build_list: ${{ steps.action_argparse.outputs.build_list }}
-        echo target: ${{ steps.action_argparse.outputs.target }}
-        echo platform: ${{ steps.action_argparse.outputs.platform }}
-        echo jdk_version: ${{ steps.action_argparse.outputs.jdk_version }}
-        echo jdk_impl: ${{ steps.action_argparse.outputs.jdk_impl }}
+        echo build_parameters: ${{ steps.argparse.outputs.build_parameters }}
+        echo sdk_resource: ${{ steps.argparse.outputs.sdk_resource }}
+        echo build_list: ${{ steps.argparse.outputs.build_list }}
+        echo target: ${{ steps.argparse.outputs.target }}
+        echo platform: ${{ steps.argparse.outputs.platform }}
+        echo jdk_version: ${{ steps.argparse.outputs.jdk_version }}
+        echo jdk_impl: ${{ steps.argparse.outputs.jdk_impl }}
 
   runBuild:
     runs-on: ${{ matrix.platform }}
--- a/.github/workflows/runAqa.yml
+++ b/.github/workflows/runAqa.yml
@@ -112,7 +112,7 @@ jobs:
             core.setFailed(`Request failed with error ${err}`)
           }
     - name: AQA
-      uses: sophia-guo/run-aqa@PR
+      uses: AdoptOpenJDK/run-aqa@v1
       with:
          build_list: ${{ matrix.build_list }}
          target: ${{ matrix.target }}
--- a/.github/workflows/runAqa.yml
+++ b/.github/workflows/runAqa.yml
@@ -111,11 +111,6 @@ jobs:
           } catch (err) {
             core.setFailed(`Request failed with error ${err}`)
           }
-    - name: Extract head repo and branch
-      id: extract_branch
-      run: |
-        echo ::set-output name=repo::${{ fromJSON(steps.get-pr.outputs.result).head.repo.full_name }}
-        echo ::set-output name=branch::${{ fromJSON(steps.get-pr.outputs.result).head.ref }}
     - name: AQA
       uses: AdoptOpenJDK/run-aqa@v1
       with:
@@ -123,7 +118,7 @@ jobs:
          target: ${{ matrix.target }}
          jdksource: 'install-jdk'
          version: ${{ matrix.jdk_version }}
-         openjdk_testRepo: '${{ steps.extract_branch.outputs.repo }}:${{ steps.extract_branch.outputs.branch }}'
+         openjdk_testRepo: '${{ fromJSON(steps.get-pr.outputs.result).head.repo.full_name }}:${{ fromJSON(steps.get-pr.outputs.result).head.ref }}'
     - uses: actions/upload-artifact@v2
       if: failure()
       with:
--- a/.github/workflows/runAqa.yml
+++ b/.github/workflows/runAqa.yml
@@ -89,6 +89,7 @@ jobs:
       fail-fast: false
       matrix: ${{ fromJson(needs.parseComment.outputs.build_parameters) }}
     steps:
+    - uses: actions/checkout@v2
     - uses: AdoptOpenJDK/install-jdk@v1
       with:
         version: ${{ matrix.jdk_version }}
--- a/.github/workflows/runAqa.yml
+++ b/.github/workflows/runAqa.yml
@@ -128,8 +128,8 @@ jobs:
     - uses: actions/upload-artifact@v2
       if: failure()
       with:
-        name: test_output_system
-        path: ./**/test_output_*/
+        name: test_output
+        path: ./**/output_*/
 
   reportStatus:
     runs-on: ubuntu-latest
--- a/.github/workflows/runAqa.yml
+++ b/.github/workflows/runAqa.yml
@@ -89,7 +89,6 @@ jobs:
       fail-fast: false
       matrix: ${{ fromJson(needs.parseComment.outputs.build_parameters) }}
     steps:
-    - uses: actions/checkout@v2
     - uses: AdoptOpenJDK/install-jdk@v1
       with:
         version: ${{ matrix.jdk_version }}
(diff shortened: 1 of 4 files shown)


llxia:
> What parameters should be required, and what should the default values for non-required parameters be?

I think we should have the default values for the following parameters:
--sdk_resource = nightly
--build_list = openjdk
--target = jdk_math
--platform = x86-64_linux
--jdk_version = 8
--jdk_impl = openj9

We can adjust this later to cover the common use case.

sophia-guo:
> The builds checkout the master branch of the base repo rather than the head repo and branch of the PR. This must be fixed. Would need to change these lines of the workflow file.

The event is `issue_comment`, which is different from the event `pull request`. Under the current status it need extra steps to get the PR's branch. 

There is an open issue in github action and hence have some examples of how to do it. Hope it helpful. https://github.com/actions/checkout/issues/331

Regarding the comment-trigger, it might be helpful to separate if it's a PR comment or an issue comment. If it's a PR comment   there will be extra steps to get the branch of PR. If it's an issue comment the branch can be set directly by parameters. ( My understand is for now we can first focus on PR comment)

sophia-guo:
For job runBuild you may use `needs: parseComment`. So if job `parseComment` fail ` job runBuild` will be skipped automatically.

Icohedron:
> > What parameters should be required, and what should the default values for non-required parameters be?
> 
> I think we should have the default values for the following parameters:
> --sdk_resource = nightly
> --build_list = openjdk
> --target = jdk_math
> --platform = x86-64_linux
> --jdk_version = 8
> --jdk_impl = openj9
> 
> We can adjust this later to cover the common use case.

These defaults appear to fail to build. See this run: [544298906](https://github.com/Icohedron/openjdk-tests/actions/runs/544298906)

Icohedron:
> * The builds checkout the master branch of the base repo rather than the head repo and branch of the PR. This must be fixed. Would need to change [these lines](https://github.com/AdoptOpenJDK/openjdk-tests/pull/2222/commits/e532bd50bb6c69b55c69b1a1869239fae6a1d627#diff-ab5b4efba7d032a7065a9d242e7e1bd92b59251ab0c5c95b4a3aa6f1d0b38db7R96-R101) of the workflow file.

This issue has been fixed!

Icohedron:
> * When there is an error in the comment, the resulting error message mentions the `action_argparse.py` script used to parse the comment. A minor issue but I see no obvious fix aside from overriding the error messages of Python's `argparse` or writing a new parser from scratch. (See example error below)

This issue has been fixed!

Icohedron:
New issue: A newline at the end of the comment will result in an empty error message. Additionally, **one can suffix the comment with `|| <command>` to execute any arbitrary shell command on the GitHub runner**. Security issue? I do not see a way to prevent this at the time being.

smlambert:
I think you have described how this vulnerability has to be handled.  Checking the comment and allowing only the set of parameters for run-aqa.  If anything else is present, then exit with an error message.

Icohedron:
> New issue: A newline at the end of the comment will result in an empty error message. Additionally, **one can suffix the comment with `|| <command>` to execute any arbitrary shell command on the GitHub runner**. Security issue? I do not see a way to prevent this at the time being.
> 
> Would need to alter [this line](https://github.com/AdoptOpenJDK/openjdk-tests/pull/2222/files#diff-ab5b4efba7d032a7065a9d242e7e1bd92b59251ab0c5c95b4a3aa6f1d0b38db7R14) in `action.yml` to get the comment body as argument to the Python script while stripping extra whitespace and disallowing additional shell commands and operations.

This issue has been fixed! See commit [7d7489a](https://github.com/AdoptOpenJDK/openjdk-tests/pull/2222/commits/7d7489adfaa7edb758833190439ec6bd874ee899)

The fix/remedy to the issue was to store the user input (the comment body) in an environment variable before passing it to the Python script, as is suggested in [this GitHub Security Lab article](https://securitylab.github.com/research/github-actions-untrusted-input).

Icohedron:
> Instead of the keyword "action", perhaps we can use a keyword that is more descriptive of what it does?

I have changed the keyword to "run aqa". Now instead of "action", use "run aqa".

> > > What parameters should be required, and what should the default values for non-required parameters be?
> > 
> > 
> > I think we should have the default values for the following parameters:
> > --sdk_resource = nightly
> > --build_list = openjdk
> > --target = jdk_math
> > --platform = x86-64_linux
> > --jdk_version = 8
> > --jdk_impl = openj9
> > We can adjust this later to cover the common use case.
> 
> These defaults appear to fail to build. See this run: [544298906](https://github.com/Icohedron/openjdk-tests/actions/runs/544298906)

The issue was that the target required an underscore prefix. Therefore I have made the argparse python script add an underscore prefix if one is not present.

smlambert:
@llxia @sophia-guo - looks like this PR is ready for a final review 👍

.github/workflows/runAqa.yml
```diff
@@ -0,0 +1,193 @@
...
+         build_list: ${{ matrix.build_list }}
+         target: ${{ matrix.target }}
+         jdksource: 'install-jdk'
+         version: ${{ matrix.jdk_version }}
+         openjdk_testRepo: '${{ steps.extract_branch.outputs.repo }}:${{ steps.extract_branch.outputs.branch }}'
+    - uses: actions/upload-artifact@v2
+      if: failure()
+      with:
+        name: test_output_system
+        path: ./**/test_output_*/
```
llxia, line 132: The folder is renamed. It should be ` path: ./**/output_*/`

.github/workflows/runAqa.yml
```diff
@@ -0,0 +1,193 @@
...
+      with:
+         build_list: ${{ matrix.build_list }}
+         target: ${{ matrix.target }}
+         jdksource: 'install-jdk'
+         version: ${{ matrix.jdk_version }}
+         openjdk_testRepo: '${{ steps.extract_branch.outputs.repo }}:${{ steps.extract_branch.outputs.branch }}'
+    - uses: actions/upload-artifact@v2
+      if: failure()
+      with:
+        name: test_output_system
```
llxia, line 131: The executed tests depend on users' input. We should use `name: test_output`

.github/workflows/runAqaArgParse.py
```diff
@@ -0,0 +1,67 @@
...
+    raw_args = sys.argv[1 + len(keywords):]
+
+    parser = argparse.ArgumentParser(prog=keyword, add_help=False)
+    # Improvement: Automatically resolve the valid choices for each argument populate them below, rather than hard-coding choices.
+    parser.add_argument('--sdk_resource', default=['nightly'], choices=['nightly', 'releases'], nargs='+')
+    parser.add_argument('--build_list', default=['openjdk'], choices=['openjdk', 'functional', 'system', 'perf', 'external'], nargs='+')
+    parser.add_argument('--target', default=['_jdk_math'], nargs='+')
+    parser.add_argument('--platform', default=['x86-64_linux'], nargs='+')
+    parser.add_argument('--jdk_version', default=['8'], nargs='+')
+    parser.add_argument('--jdk_impl', default=['openj9'], nargs='+')
```
sophia-guo, line 50: Would it be nicer to set jdk_impl and platform as choices, similarly to build_list and sdk_source?

Icohedron: It would be `openj9` and `hotspot` as the only two choices?

.github/workflows/runAqa.yml
```diff
@@ -0,0 +1,193 @@
...
+        echo jdk_impl: ${{ steps.argparse.outputs.jdk_impl }}
+
+  runBuild:
+    runs-on: ${{ matrix.platform }}
+    needs: parseComment
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJson(needs.parseComment.outputs.build_parameters) }}
+    steps:
+    - uses: actions/checkout@v2
```
sophia-guo, line 92: Is this checkout necessary?

sophia-guo, line 92: Wondered if this checkout is necessary?

Icohedron: It is a line that is also in [`PushOrTestPRBuild.yml`](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/.github/workflows/PushOrPRtestbuild.yml#L32) for running tests.

.github/workflows/runAqa.yml
```diff
@@ -0,0 +1,193 @@
...
+            pull_number: context.issue.number
+          }
+          core.info(`Getting PR #${request.pull_number} from ${request.owner}/${request.repo}`)
+          try {
+            const result = await github.pulls.get(request)
+            return result.data
+          } catch (err) {
+            core.setFailed(`Request failed with error ${err}`)
+          }
+    - name: Extract head repo and branch
```
sophia-guo, line 115: As repo and branch are only used once in AQA step, I would suggest remove this echo step. AQA can use `${{ fromJSON(steps.get-pr.outputs.result).head.repo.full_name }}` and `${{ fromJSON(steps.get-pr.outputs.result).head.ref }}` directly.

Icohedron: I have that initially but it refused to work for some reason. I can try again though.

.github/workflows/runAqaArgParse.py
sophia-guo: > It would be openj9 and hotspot as the only two choices?  For jdk_impl @smlambert

.github/workflows/runAqa.yml
sophia-guo: Yes, I thought about it some more and realized that this step is probably kind of redundancy for situations which doesn't use default repos.

.github/workflows/runAqa.yml
Icohedron: Leaving the echo step in would also let a person check that the correct repo and branch were checked out in the workflow job logs.

.github/workflows/runAqa.yml
Icohedron: The checkout is necessary, otherwise the job runs into an error.
`Error: Unable to locate executable file: /home/runner/work/openjdk-tests/openjdk-tests/get.sh. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.`

.github/workflows/runAqa.yml
Icohedron: Builds appear to run successfully after removing the echo step and using the variables directly.
However there isn't a way to tell which repo and branch it checked out in the logs, as far as I can tell.

.github/workflows/runAqaArgParse.py
smlambert: yes those 2 choices (and should we ever need it, in future, we can add jdk_vendor as an argument, but will not include it for now)

.github/workflows/runAqa.yml
Icohedron: I just tried sophia-guo/run-aqa@PR alongside removing the checkout (commit [
ea2f2e6](https://github.com/AdoptOpenJDK/openjdk-tests/pull/2222/commits/ea2f2e65340feda9d4ec7a9a24b225fdefea3a82)). I encounter the same error.
[https://github.com/Icohedron/openjdk-tests/runs/1914866085?check_suite_focus=true](https://github.com/Icohedron/openjdk-tests/runs/1914866085?check_suite_focus=true)

sophia-guo: Yes, I just updated run-aqa it should be fine. https://github.com/sophia-guo/openjdk-tests/runs/1919789165?check_suite_focus=true

.github/workflows/runAqa.yml
sophia-guo: This actually same issue with https://github.com/AdoptOpenJDK/openjdk-tests/pull/2222#discussion_r577228967. run-aqa used to check if openjdk-tests is checkout, if yes, will ignore the parameter 'openjdk_testRepo'. Updated run-aqa with the check removed. The console output would have something like `/usr/bin/git clone --depth 1 -b branch https://github.com/user/openjdk-tests.git`.

.github/workflows/runAqa.yml
sophia-guo: With or without the checkout should be fine when the PR https://github.com/AdoptOpenJDK/run-aqa/pull/70 is merged and v1 is updated.

.github/workflows/runAqaArgParse.py
sophia-guo: I have set platform as 'x86-64_linux ` with extra space by accident and got https://github.com/sophia-guo/openjdk-tests/runs/1919706546?check_suite_focus=true. It would be nice to set platform as choices so the job would fail when do the parseComment instead of runBuild and can't find the runner.

Icohedron: > I have set platform as 'x86-64_linux ` with extra space by accident and got https://github.com/sophia-guo/openjdk-tests/runs/1919706546?check_suite_focus=true. It would be nice to set platform as choices so the job would fail when do the parseComment instead of runBuild and can't find the runner.

That would be quite a long choices list. The choice allows github runner platforms, which include various versions of windows, mac, and linux.

.github/workflows/runAqa.yml
Icohedron: I see the PR has been merged. Is v1 updated?

sophia-guo: Just updated v1.

.github/workflows/runAqaArgParse.py
sophia-guo: `platform` is what jdk supports on. Here is the list https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/openjdk_tests. `windows-latest` or others are supported virtual-environments in github runners. https://github.com/actions/virtual-environments. 

That is no all supported platforms  are enabled on github runners.
There is a map between them. 

For example `x86-64_linux`, `x86-64_linux_mixed`, `x86-64_linux_xl` can map to any one of `ubuntu-16.04`, `ubuntu-latest or ubuntu-18.04`, `ubuntu-16.04`. 

So choices would be [`x86-64_linux`, `x86-64_linux_mixed`, `x86-64_linux_xl`, `x86-64_mac` , `x86-64_mac_mixed`, `x86-64_mac_xl`, `x86-64_windows`,`x86-64_windows_mixed`,`x86-64_windows_xl`]. Going forward some of them will be removed or combined so the list would be smaller. 

We can leave it as is and update later if the choices helps.

sophia-guo (approved):
LGTM

llxia (approved):
Thanks @Icohedron

PR #2233, opened 2021-02-04
Move JdiTest playlist entry from lambdaLoadTest to otherLoadTest

lumpfish:

### DIFF
--- a/system/otherLoadTest/playlist.xml
+++ b/system/otherLoadTest/playlist.xml
@@ -257,6 +257,25 @@
 		</subsets>
 	</test>
 	<test>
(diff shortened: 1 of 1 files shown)


lumpfish:
The JdiTest test case was removed from lambdaLoadTest,xml in https://github.com/AdoptOpenJDK/openjdk-tests/pull/2229 because it did not belong there (JdiTest tests the java debug interface, not lambdas).
This PR adds it to otherLoadTest.xml (there is no playlist specific to debugging tests).

Mesbah-Alam (approved):
LGTM

llxia (approved):

PR #2234, opened 2021-02-04
Add a custom playlist target to run system tests

Mesbah-Alam:
Resolves https://github.com/AdoptOpenJDK/openjdk-tests/issues/2212

- Adds `system_custom` playlist target which enables running of individual STF system tests in Grinders 
- Updates JenkinsfileBase logic for `system_custom` target to allow stf tests with extra options to work 

To run stf system tests using `system_custom` in a Grinder, please ensure the following is set: 

```
1) BUILD_LIST=system
2) TARGET=system_custom
3) CUSTOM_TARGET=-test=<stf_test_class_name>
     - Optionally, if test requires arguments, CUSTOM_TARGET=-test=<stf_test_class_name>  -test-args="x=a,y=b"
```

Note, `<stf_test_class_name>` should be the name of the STF class to run, e.g. [CassLoadingLoadTest](https://github.com/AdoptOpenJDK/openjdk-systemtest/blob/master/openjdk.test.load/src/test.load/net/adoptopenjdk/stf/ClassloadingLoadTest.java); not a [playlist target](https://github.com/AdoptOpenJDK/openjdk-tests/blob/517467de209aae47db938d4d2b58f45727912322/system/otherLoadTest/playlist.xml#L57). 

### DIFF
--- a/system/otherLoadTest/playlist.xml
+++ b/system/otherLoadTest/playlist.xml
@@ -18,6 +18,20 @@
 		</groups>
 	</test>
 	<test>
+		<testCaseName>system_custom</testCaseName>
+		<variations>
+			<variation>NoOptions</variation>
+		</variations>
+		<command>$(SYSTEMTEST_CMD_TEMPLATE) $(CUSTOM_TARGET);\
+	$(TEST_STATUS)</command>
+		<levels>
+			<level>special</level>
+		</levels>
+		<groups>
+			<group>system</group>
+		</groups>
+	</test>
+	<test>
 		<testCaseName>MiniMix_5m</testCaseName>
 		<variations>
 			<variation>Mode150</variation>
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -539,7 +539,11 @@ def runTest( ) {
 
 			TARGET = "${params.TARGET}";
 			if (TARGET.contains('custom') && CUSTOM_TARGET!='') {
-				CUSTOM_OPTION = "${TARGET.toUpperCase()}_TARGET='${CUSTOM_TARGET}'"
+				if (TARGET == 'system_custom') {
+					env.SYSTEM_CUSTOM_TARGET=CUSTOM_TARGET
+				} else {
+					CUSTOM_OPTION = "${TARGET.toUpperCase()}_TARGET='${CUSTOM_TARGET}'"
+				}
 			}
 			if (!TARGET.startsWith('-f')) {
 				TARGET="_${params.TARGET}"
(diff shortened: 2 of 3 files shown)


Mesbah-Alam:
- Test where the STF load test to run requires additional test specific parameters to be passed: (e.g. `CUSTOM_TARGET=-test=MathLoadTest -test-args="workload=math,timeLimit=5m"`): https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6451/

- Test where the STF load test to run requires no additional parameters (e.g. `CUSTOM_TARGET=-test=ClassloadingLoadTest`): https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6452/

- Test where extra JVM debug option is supplied with custom target: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6453/console

llxia:
Just for the record, we have tried to use `CUSTOM_OPTION`. This does not work for system tests as `CUSTOM_TARGET ` may contain space and `"` is required by STF. 

```
CUSTOM_TARGET=-test=MathLoadTest -test-args="workload=math,timeLimit=5m"
```

And `CUSTOM_OPTION` gets passed into `maketest.sh`. The actual value gets altered with quotes and make cannot recognize it. After several attempts, we think setting `env` to avoid the value passing to ` maketest.sh` is the best way so far.

Not related to this PR, but it may worth investigating if we should remove  `maketest.sh` and directly invoke make cmd in the future.

Mesbah-Alam:
> Not related to this PR, but it may worth investigating if we should remove maketest.sh and directly invoke make cmd in the future.

Raised https://github.com/AdoptOpenJDK/openjdk-tests/issues/2235 for this.

system/otherLoadTest/playlist.xml
```diff
@@ -17,6 +17,20 @@
 			<group>system</group>
 		</groups>
 	</test>
+	<test>
+		<testCaseName>system_custom</testCaseName>
+		<variations>
+			<variation>NoOptions</variation>
+		</variations>
+		<command>$(SYSTEMTEST_CMD_TEMPLATE) $(CUSTOM_TARGET);\
```
llxia, line 25: We need to set a default value for `CUSTOM_TARGET `. Otherwise, the build without `CUSTOM_TARGET ` (i.e., regular build) will fail.

Mesbah-Alam: Ah.. that was missed. 

Added! 

Ran some more tests with the default value: 

When `CUSTOM_TARGET` field is empty : https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6495/console 
When `CUSTOM_TARGET` has a value given by user: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6496/console

They look all good.

llxia (approved):
LGTM

smlambert (approved):

PR #2239, opened 2021-02-05
Update userGuide to include vendor disable feature

renfeiw:

### DIFF
--- a/doc/userGuide.md
+++ b/doc/userGuide.md
@@ -342,6 +342,10 @@ To exclude the test for openj9 only:
 
 ```auto exclude test jdk_test impl=openj9```
 
+To exclude the test for adoptopenjdk vendor only:
+
+```auto exclude test jdk_test vendor=adoptopenjdk```
+
 To exclude the test for java 8 only:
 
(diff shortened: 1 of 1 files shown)


smlambert (approved):
LGTM
Out of curiosity, in the case where we have not mapped vendor to a single word, would this mechanism still work or would one need to use quotes, or put the [mapping in place](https://github.com/AdoptOpenJDK/TKG/blob/master/src/org/openj9/envInfo/JavaInfo.java#L130)?

renfeiw:
> LGTM
> Out of curiosity, in the case where we have not mapped vendor to a single word, would this mechanism still work or would one need to use quotes, or put the [mapping in place](https://github.com/AdoptOpenJDK/TKG/blob/master/src/org/openj9/envInfo/JavaInfo.java#L130)?

In the case, quotes are required. I would do a local test with that to make sure it works.

smlambert:
Its okay, I think if some distributor was using AQA and TKG enough that they wanted to disable tests, we'd add the mapping.  I was just wondering what the behaviour would be.

llxia (approved):
LGTM

PR #2251, opened 2021-02-09
Always set lightweight children

### LINKED ISSUE #2250: Jenkins: Test children should have the same SCM config settings as parent
**Describe the bug**
Recently, the option for lightweight checkout was added (#2216 ). This setting, and all scm settings, should propagate down to any child jobs that are generated. This will further reduce disk space burden on the Jenkins node. It will also eliminate any surprise differences between the parent and child jobs.

AdamBrousseau:
- Except for grinder jobs
- Also allow Repo and Branch to be passed to DSL
- Whitespace cleanup including replacing some spaces
  with tabs to match the rest of the file

Fixes #2250

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -25,7 +25,7 @@ def setupEnv() {
 	env.JOBSTARTTIME = sh(script: 'LC_TIME=C date +"%a, %d %b %Y %T %z"', returnStdout: true).trim()
 
 	// Check what existing java processes are running
-        echo "PROCESSCATCH: Java processes which are currently on the machine:"
+		echo "PROCESSCATCH: Java processes which are currently on the machine:"
 	if (PLATFORM.contains("windows")) {
 		sh(script:'powershell -c "Get-Process java | select id,processname,starttime"', returnStatus:true)
 	} else {
@@ -135,7 +135,7 @@ def setupParallelEnv() {
 				// limit childJobNum
 				if (childJobNum > 20) {
 					echo "Due to the limited machines, NUM_MACHINES can only be set up to 20. Current NUM_MACHINES is ${NUM_MACHINES}."
-					echo "Reset NUM_MACHINES to 20..." 
+					echo "Reset NUM_MACHINES to 20..."
 					childJobNum = 20
 				}
 			} else if (params.PARALLEL == "Dynamic") {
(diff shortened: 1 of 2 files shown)


AdamBrousseau:
@llxia for review. Hide whitespace changes for the real changes. Untested at this point.

AdamBrousseau:
I added an exception for Grinder jobs as this would be the likely place where we would hit an edge case.

llxia:
Thanks @AdamBrousseau yes, Grinder is for debugging purposes, and users are very likely to change Adopt repo and branch when using Grinder. If the Grinder child jobs are generated with hardcoded Adopt repo and branch, there is a chance to rerun a Grinder child job based on the other poeple's repo and branch. Once this happens, it will very hard to detect. Due to this reason, it is better to make an exception case for Grinder child jobs.

llxia:
On the second thought, this change should not be needed. We just need to add `LIGHT_WEIGHT_CHECKOUT` as a job parameter in https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/testJobTemplate#L188 and any child job will inherit this parameter value from its parent.

Please see the example in the following jobs:
Grinder/13840/
Grinder_testList_1/configure

AdamBrousseau:
I suppose that could work too. It does add yet another parameter to the test job. The only caveat I can think of is if someone has their own grinder job with hardcoded custom repo/branch and lightweight=true. That would generate a child job with lightweight=true and hardcoded to Adopt repos.

llxia:
`LIGHT_WEIGHT_CHECKOUT ` should always be false for Grinder. We can add a note for this or remove this parameter from the Grinder. By default `LIGHT_WEIGHT_CHECKOUT` is set to false in testJobTemplate.

Personally, I would prefer less code in Jenkins script if possible.

AdamBrousseau:
Ok. I've updated the change, please review.

llxia (commented):
Thanks @AdamBrousseau LGTM. Could you also remove buildenv/jenkins/JenkinsfileBase? We can have another PR for fixing the indentation.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -333,6 +333,8 @@ def createJob( TEST_JOB_NAME, ARCH_OS ) {
 
 	def templatePath = 'openjdk-tests/buildenv/jenkins/testJobTemplate'
 
+	jobParams.put('LIGHT_WEIGHT_CHECKOUT', LIGHT_WEIGHT_CHECKOUT)
```
AdamBrousseau, line 336: @llxia this change is still here. Do you want the whitespace changes in a separate pr?

llxia: ah, sorry, I missed this line. Yes, please keep the whitespace changes.

llxia (approved):
Thanks @AdamBrousseau for testing this internally.

smlambert (approved):
thanks @AdamBrousseau !

Issue #2272, opened 2021-02-12
Directories trigger for PR testing should run regular tests


llxia:
Directories trigger for PR testing does not have test image, so it should only trigger regular tests. For example, 
sanity.openjdk.regular
extended.openjdk.regular

Otherwise, native tests will fail in PR testing.
```
FAILED test targets:
	jdk_lang_native_0
	jdk_foreign_native_0
```
https://github.com/AdoptOpenJDK/openjdk-tests/pull/2266/checks?check_run_id=1882453674

llxia:
The fix is delivered in PR #2283

PR #2283, opened 2021-02-17
add suffix to ignore native tests

xichen1:
According to #2272, directories changed trigger for PR testing should only run the regular tests, 
by adding the suffix `.regular` to the target, native tests will be ignored.

### DIFF
--- a/.github/workflows/directoriesFilesChangePR.yml
+++ b/.github/workflows/directoriesFilesChangePR.yml
@@ -48,7 +48,7 @@ jobs:
         uses: AdoptOpenJDK/run-aqa@v1
         with:
           build_list: ${{ matrix.build_list }}
-          target: '_sanity.${{ matrix.build_list }}'
+          target: '_sanity.${{ matrix.build_list }}.regular'
           jdksource: 'install-jdk'
           version: ${{ matrix.version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/directoriesFilesChangePR.yml
```diff
@@ -48,7 +48,7 @@ jobs:
         uses: AdoptOpenJDK/run-aqa@v1
         with:
           build_list: ${{ matrix.build_list }}
-          target: '_sanity.${{ matrix.build_list }}'
+          target: '_sanity.${{ matrix.build_list }}.regular'
```
sophia-guo, line 51: I believe target `_sanity.regular` should be enough. Parameter `build_list` has already limited the tests category.

smlambert: Yes, though there is nothing wrong with being very explicit here (for clarity to others who do not know that if build_list is set, that TKG will only build and execute the targets within that directory).

xichen1: updated using `target: '_sanity.regular'`.

llxia (approved):
LGTM

karianna (approved):
LGTM

Issue #2204, opened 2021-01-25
Workflow job triggered by a change to the buildenv directory


smlambert:
Similar to #2200 this issue is to Investigate and trial a github workflow that determines if openjdk-tests/buildenv dir was changed, and then take a certain type of action. 

If buildenv dir changed, investigate how to trigger a Jenkins job in a github workflow to trigger a Grinder (via https://github.com/marketplace/actions/trigger-jenkins-job or similar) . 

If both test dirs and buildenv, dir changes, can a Grinder be triggered with targets from the playlists within the test dirs.

Essentially, this in combination with comment triggered PR testing (see https://github.com/AdoptOpenJDK/openjdk-tests/issues/2199) can allow us to reduce the amount of testing that happens as part of PR testing.  

A prototype workflow would trigger a Jenkins Grinder to test Jenkins changes.

patkarns:
Update: formatted and passed targets from changed test dirs (output of Xichen's code #2200 ) as parameters into Grinder_Simple

- See: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Simple/209/ (added empty files to the functional and system folders)

Going from here:

- [ ] Add test dir target(s) to the parameters only when **both** test dirs and buildenv dir changes (right now the targets are added to the parameters regardless of whether buildenv changes, also handle the case where test dirs is unchanged)
- [ ] Using sanity for efficiency/speed right now, will have to change or make the test groups dynamic later
- [ ] Change from yanan's username to a general user

smlambert:
Closing as won't pursue at this time.

Issue #2200, opened 2021-01-22
Certain test directories/files change trigger for PR testing


smlambert:
Investigate and trial a github workflow that determines which openjdk-tests dir was changed, and then based on a defined mapping, takes different types of action. 

Start by looking at top-level directories and doing something basic like, if something changed in:
perf -> run sanity.perf test target
system -> run sanity.system test target
functional -> run sanity.functional test target
openjdk -> run sanity.openjdk test target

This approach could then be tuned to look at which test targets in a playlist changed and run the target or the level its contained within.  For now, we will ignore changes to doc, external and jck directories.

We can investigate what action to take if the buildenv dir changes in https://github.com/AdoptOpenJDK/openjdk-tests/issues/2204.  

If both test dirs and buildenv, dir changes, can a Grinder be triggered with targets from the playlists within the test dirs.

Essentially, this in combination with comment triggered PR testing (see https://github.com/AdoptOpenJDK/openjdk-tests/issues/2199) can allow us to reduce the amount of testing that happens as part of PR testing.  

A prototype workflow would output a result, which is a change-based test target, or a comma-separated list of test targets to be ingested by run-aqa github action to run a build with specific test target to run against changes in test category or trigger specific jenkins build to test jenkins changes.

sophia-guo:
Related with #2292 
Probably we need to confirm:

> If buildenv dir changed, investigate how to trigger a Jenkins job in a github workflow to trigger a Grinder (via https://github.com/marketplace/actions/trigger-jenkins-job or similar)

Current #2292 is trying the grinder with default parameters. Should we set the grinder with platform which is not supported on github runner?

`If both test dirs and buildenv, dir changes, can a Grinder be triggered with targets from the playlists within the test dirs.`
Similar to the former one, Should we set the grinder with platform which is not supported on github runner?

smlambert:
re: https://github.com/AdoptOpenJDK/openjdk-tests/issues/2200#issuecomment-785426288

We can pick a platform where we have lots of machines attached to Jenkins, the purpose here is to check that changes to the Jenkins scripts are not broken.  

yes, I think if both buildenv and some test group dir changes, like system, the target/build_list could be matched up... that can be done as an iterative improvement to this initial contribution.

Another enhancement relating to the comment triggered workflow would be if someone typed "run aqa --platform=SomeNonGithubRunnerPlatform"... would be to trigger a Grinder on that platform, but for now lets leave it simple...

[...]

PR #2292, opened 2021-02-22
Trigger jobs on dir changes

patkarns:
Combined #2200 with #2204.

Grinder_Simple is triggered with parameter TARGET from the changed test directories. If only buildenv is changed, we run a job with default parameters.

Currently running sanity for all, we will have to talk about which test groups to run for each as we go forward.


.github/workflows/getBuildLists.py
```diff
@@ -15,7 +16,19 @@ def main():
...
+            print('::set-output name=build_env::true')
+            flag = True
+    
+    if not flag:
+        print('::set-output name=build_env::false')
+    
+    if len(result) > 0:
+        print('::set-output name=test_dirs_changed::true')
+    else:
+        print('::set-output name=test_dirs_changed::false')
```
sophia-guo, line 29: This could be combined with 

```
 if not result:
        result.append('skip')	        result.append('skip')
```

and  update the if condition in https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/.github/workflows/directoriesFilesChangePR.yml#L32 ?

patkarns: Good point! I've revised it now.

llxia:
Do we still need `TriggerGrinderTest.yml` and `buildenvChangesInPR.yml`?

.github/workflows/directoriesFilesChangePR.yml
```diff
@@ -57,3 +57,53 @@ jobs:
...
+    env:
+        TARGET_LIST:  ${{ needs.getBuildLists.outputs.buildLists }}
+    needs: getBuildLists
+    if: ${{ needs.getBuildLists.outputs.buildEnv }}
+    steps:
+    - uses: actions/checkout@v2
+    - name: format test targets str
+      id: format_test_targets_str
+      run: |
+        python3 ./.github/workflows/createTestList.py ${{ needs.getBuildLists.outputs.buildLists }}
```
sophia-guo, line 74: Looks like `buildLists` are passed in as arguments. However there is no code dealing with the arguments in `createTestList.py `.  `createTestList.py ` is actually using environment `TARGET_LIST` setting the target tests. My suggestion would be parsing the arguments instead of using environment variable.

.github/workflows/directoriesFilesChangePR.yml
```diff
@@ -57,3 +57,53 @@ jobs:
         with:
           name: output_changed_based_build_list
           path: ./**/output_*/
+  triggerGrinderTests:
+    name: Build
+    runs-on: ubuntu-latest
+    outputs:
+      test_targets_str: ${{ steps.format_test_targets_str.outputs.test_targets_str }}
+    env:
+        TARGET_LIST:  ${{ needs.getBuildLists.outputs.buildLists }}
```
sophia-guo, line 66: By parsing the arguments `needs.getBuildLists.outputs.buildLists` we will not need this environment variable.

.github/workflows/createTestList.py
```diff
@@ -0,0 +1,22 @@
+import sys
+import os
+import ast
+
+def main():
+    
+    sanity_format = "sanity.{}"
+    formattedTests = []
+    
+    targetNames = ast.literal_eval(os.getenv('TARGET_LIST'))
```
sophia-guo, line 10: Instead of using the environment variable I would suggest parse the program arguments.

.github/workflows/directoriesFilesChangePR.yml
```diff
@@ -57,3 +57,53 @@ jobs:
...
+        job: "Grinder_Simple"
+    - name: trigger single job
+      uses: stigmelling/jenkins-action@master
+      if: ${{!contains( fromJson(needs.getBuildLists.outputs.buildLists), 'skip')}}
+      with:
+        jenkinsUrl: 'https://ci.adoptopenjdk.net/'
+        username: 'xfxcwynlc'
+        token: ${{ secrets.TOKEN }}
+        job: "Grinder_Simple"
+        params: '{"TARGET": "${{ steps.format_test_targets_str.outputs.test_targets_str}}" }'
```
sophia-guo, line 91: Looks like the params is something `TARGET: TESTLIST=sanity.functional,sanity.system`, which won't work. Target should be something like `TARGET:testList TESTLIST=jit_jitt,jit_recognizedMethod,testSCCMLTests2_1` https://github.com/eclipse/openj9/blob/master/test/docs/OpenJ9TestUserGuide.md#run-a-list-of-tests-
 and testlist won't work with groups. 


``

.github/workflows/createTestList.py
```diff
@@ -0,0 +1,22 @@
...
+        formattedTests.append(sanity_format.format(argument))
+    
+    allTests = ','.join(formattedTests)    
+    testTargets = 'TESTLIST={}'.format(allTests)
+    
+    print(testTargets)
+    print('::set-output name=test_targets_str::{}'.format(testTargets))
+
+if __name__ == "__main__":
+    main()
```
sophia-guo, line 22: This file may not need if we are setting the target as something like `sanity.openjdk`. We could simply set job 'trigger single job'  as params: `{"TARGET": "sanity", "BUILD_LIST": 'openjdk'}`

.github/workflows/directoriesFilesChangePR.yml
sophia-guo: To make the grinder job you will probably need to pass in params as a Json with "name":"value" params'. Something like {"TARGET": "sanity", "BUILD_LIST": 'openjdk'}. As we don't know if the size of `needs.getBuildLists.outputs.buildLists` it would be better to set up the job by matrix.

.github/workflows/createTestList.py
sophia-guo: `If both test dirs and buildenv, dir changes, can a Grinder be triggered with targets from the playlists within the test dirs.` That is instead of `sanity` we would like a testcase in playlist.xml. For example if build_list is `openjdk`, we can set target as `jdk_custom`. If build_list is `system` we could set target as `ClassLoadingTest`.

.github/workflows/directoriesFilesChangePR.yml
smlambert: I am wondering if we can just leverage the same story we use for "Rerun in Grinder" link, which is essentially a curl command with all parameters in the URL?

Short Example (with subset of params):  https://ci.adoptopenjdk.net/job/Grinder/parambuild/?JDK_VERSION=11&JDK_IMPL=hotspot&JDK_VENDOR=adoptopenjdk&BUILD_LIST=openjdk&PLATFORM=s390x_linux&TARGET=jdk_other_0


Long Example (with all params): https://ci.adoptopenjdk.net/job/Grinder/parambuild/?TARGET=extended.openjdk&SDK_RESOURCE=upstream&TEST_FLAG=&UPSTREAM_TEST_JOB_NAME=&DOCKER_REQUIRED=false&ACTIVE_NODE_TIMEOUT=0&EXTRA_DOCKER_ARGS=&VENDOR_TEST_DIRS=&TKG_OWNER_BRANCH=AdoptOpenJDK%3Amaster&PLATFORM=s390x_linux&KEEP_REPORTDIR=true&PERSONAL_BUILD=false&ADOPTOPENJDK_REPO=https%3A%2F%2Fgithub.com%2FAdoptOpenJDK%2Fopenjdk-tests.git&LABEL=&EXTRA_OPTIONS=&CUSTOMIZED_SDK_URL=&BUILD_IDENTIFIER=&ADOPTOPENJDK_BRANCH=master&ARTIFACTORY_SERVER=&KEEP_WORKSPACE=false&JDK_VERSION=11&USER_CREDENTIALS_ID=&ITERATIONS=1&VENDOR_TEST_REPOS=&JDK_REPO=https%3A%2F%2Fgithub.com%2Fadoptopenjdk%2Fopenjdk-jdk11u&RELEASE_TAG=&OPENJ9_BRANCH=master&OPENJ9_SHA=&JCK_GIT_REPO=&VENDOR_TEST_BRANCHES=&UPSTREAM_JOB_NAME=build-scripts%2Fjobs%2Fjdk11u%2Fjdk11u-linux-s390x-hotspot&OPENJ9_REPO=https%3A%2F%2Fgithub.com%2Feclipse%2Fopenj9.git&CUSTOM_TARGET=&VENDOR_TEST_SHAS=&JDK_BRANCH=dev&LABEL_ADDITION=&ARTIFACTORY_REPO=&ARTIFACTORY_ROOT_DIR=&UPSTREAM_TEST_JOB_NUMBER=&DOCKERIMAGE_TAG=&JDK_IMPL=hotspot&SSH_AGENT_CREDENTIAL=&AUTO_DETECT=true&TKG_SHA=&CUSTOMIZED_SDK_URL_CREDENTIAL_ID=&NUM_MACHINES=&OPENJDK_SHA=&BUILD_LIST=openjdk&UPSTREAM_JOB_NUMBER=822&TIME_LIMIT=10&JVM_OPTIONS=&PARALLEL=None

patkarns: @smlambert I will make the changes below and then investigate the curl command

.github/workflows/createTestList.py
patkarns: @sophia-guo Hi Sophia! The issue seems to imply we're running all the targets in playlist.xml, but your example shows only one. Just wanted to double check that we're running all the tests in the playlist.xml file?

patkarns: Folders like perf also has subdirectories with their own playlist.xml files. 

If something in the perf folder changes, should we 
- Run tests from all the playlist.xml files in all the subfolders of perf? Regardless of which subdirectory changed (e.g. `perf/bumbleBench/playlist.xml`, `perf/dacapo /playlist.xml`, and so on)  
- Or just bring in targets from the the subdirectories that changed, as well as targets from perf/playlist.xml (the root folder) if one exists (e.g. if I only changed things in the `perf/dacapo` subdirectory, to only run `perf/dacapo /playlist.xml` and `perf/playlist.xml` if one exists)

sophia-guo: https://github.com/AdoptOpenJDK/openjdk-tests/pull/2292#discussion_r582749502 I think an easy test case for each category should be enough as it is to check if the grinder still work.  @ShelleyLambert your opinion?

smlambert: Yes, agree, a single test target is sufficient (no need to run a full sanity or extended suite).

patkarns: Updated the params!
image

.github/workflows/directoriesFilesChangePR.yml
patkarns: @sophia-guo Is there a preference towards using a curl command with parameters in the URL versus using the parameters passed as JSON?

sophia-guo: Does this rerun link mean we don't need `stigmelling/jenkins-action` ?

xfxcwynlc: After reading the documentation, I think a generic account is not able to fix permission issues...
* When a forked repo (like our branch) sends a pull request to the AdobeOpenJDK:master, the secret token stored in our repo are NOT passed to Action. The action running in AdobeOpenJDK:master will see an empty secret. 
Screen Shot 2021-03-08 at 7 28 24 PMReference: https://docs.github.com/en/actions/reference/encrypted-secrets

* Suppose now AdobeOpenJDK:master create the secret and in our PR we refers to the name of the secrets stored in your repo, the secret is still not passed as well. 

Screen Shot 2021-03-08 at 7 34 50 PM

By having a generic account and store the token in the repo, we can only trigger the Jenkins job when accepting and merging the PR from a fork.  At the PR stage, a Jenkins job can't be properly triggered due to the token issue. 

Related Forum Discussion:
https://github.community/t/make-secrets-available-to-builds-of-forks/16166
https://github.community/t/make-fork-secrets-available-to-fork-prs/17762
https://discuss.circleci.com/t/pull-requests-not-triggering-build/1213

smlambert:
Thanks @patkarns and @xfxcwynlc for your work here, it was very informative and valuable. 

Based on all the excellent investigation, digging and trials that occurred for this, I think we have arrived at a much better understanding and recognize the limitations.  Given the findings, we will not proceed with this approach and I will close this PR and the associated enhancement issue, https://github.com/AdoptOpenJDK/openjdk-tests/issues/2204.

PR #2295, opened 2021-02-22
Use any node for copying TAP files in parallel test builds

llxia:
Resolve: https://github.com/AdoptOpenJDK/openjdk-tests/issues/2262

Signed-off-by: lanxia <lan_xia@ca.ibm.com>

PR #2312, opened 2021-02-26
Remove waitForANodeToBecomeActive post-parallel

AdamBrousseau:
This step is unnecessary now that we run
on any node.

Follow up to #2295


AdamBrousseau:
I think it's safe to assume we don't need this anymore.
@llxia

sophia-guo:
Probably the function `def waitForANodeToBecomeActive(def label, String activeNodeTimeoutString) `  should also be removed as it is only used in this case.

llxia (approved):
LGTM

sophia-guo (approved):

Issue #2289, opened 2021-02-18
Reduce verbosity of parseComment job in runAqa.yml workflow 


smlambert:
For cases where parseComment finds there is nothing to do, the bot does not need create a comment.  In fact it is preferable to only comment if there is an action that has happened (be quiet when skipped).

(But excited to see this workflow in place, yay!)

Icohedron:
@smlambert Could you give an example of when parseComment should not create a comment?

Icohedron:
oh that is very odd

Icohedron:
Working on fixing this ASAP

sophia-guo:
Might be this line https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/.github/workflows/runAqa.yml#L130

PR #2313, opened 2021-02-27
reportStatus job enhancements (runAqa.yml)

Icohedron:
Factored out the workflow_run_info step from reportStatus and moved it into the parseComment job.
Uses github actions' built-in success(), failure(), and cancelled() conditionals instead of a custom github action for status reporting, resulting in better reliability.
As a result, reportStatus has been broken up into three jobs, one for each status.

Fixes two issues mentioned in #2222 
> Depending on when the workflow was cancelled, the status report comment may say "Build(s) successful!" instead of "Build(s) cancelled!" because the runBuild part of the workflow was skipped while the parseComment part of the workflow was successful.

> A "Build(s) failed" error message is displayed even when no builds were started. A minor issue but I see no obvious fix. (See example error below)

Also fixes another issue in which a build may fail yet the reportStatus job will still report a success due to a discrepency in the run order of the jobs.

I have checked to make sure this change does not reproduce issue #2289

### DIFF
--- a/.github/workflows/runAqa.yml
+++ b/.github/workflows/runAqa.yml
@@ -10,6 +10,7 @@ jobs:
       workflow_url: ${{ steps.workflow_run_info.outputs.url }}
       workflow_id: ${{ steps.workflow_run_info.outputs.id }}
       build_parameters: ${{ steps.argparse.outputs.build_parameters }}
+      failed: ${{ steps.failure_report.outputs.failed }} # For reportStatus to check if parseComment has already reported a failure (in parsing).
     steps:
     - name: Get workflow run info
       run: |
@@ -52,6 +53,10 @@ jobs:
             repo: context.repo.repo,
             body: comment_body
           })
+    - name: Failure report
+      if: failure()
+      run: echo ::set-output name=failed::true
+      id: failure_report
     - name: Create success comment
       uses: actions/github-script@v3
       with:
@@ -129,8 +134,8 @@ jobs:
 
   reportFailure:
     runs-on: ubuntu-latest
-    if: failure()
     needs: [parseComment, runBuild]
+    if: failure() && !needs.parseComment.outputs.failed
     steps:
     - name: Create comment
       uses: actions/github-script@v3
@@ -150,8 +155,8 @@ jobs:
 
   reportCancelled:
     runs-on: ubuntu-latest
-    if: cancelled()
     needs: [parseComment, runBuild]
+    if: cancelled()
     steps:
     - name: Create comment
       uses: actions/github-script@v3
@@ -171,8 +176,8 @@ jobs:
 
   reportSuccess:
     runs-on: ubuntu-latest
-    if: success()
     needs: [parseComment, runBuild]
+    if: success()
     steps:
     - name: Create comment
       uses: actions/github-script@v3
--- a/.github/workflows/runAqa.yml
+++ b/.github/workflows/runAqa.yml
@@ -7,8 +7,15 @@ jobs:
     runs-on: ubuntu-latest
     if: startsWith(github.event.comment.body, 'run aqa') && github.event.issue.pull_request
     outputs:
+      workflow_url: ${{ steps.workflow_run_info.outputs.url }}
+      workflow_id: ${{ steps.workflow_run_info.outputs.id }}
       build_parameters: ${{ steps.argparse.outputs.build_parameters }}
     steps:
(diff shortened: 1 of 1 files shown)


sophia-guo:
Any reason the status report be a separate job? Could this status reports be steps in the job of `runBuild` as it is to report the result of build?

In this way report status step would only depend on the status of the previous steps, which would be clearer and there will be no concern about #2289. We could  use ` uses: martialonline/workflow-status@v2` to combine the three different status to one or three status jobs to one. Something like:

```
    - uses: martialonline/workflow-status@v2
      id: workflow_status
    - name: Create report comment
      uses: actions/github-script@v3
      with:
        github-token: ${{secrets.GITHUB_TOKEN}}
        script: |
          comment_body = `
          @${{ github.actor }} Build(s) status is ${{ steps.workflow_status.outputs.status}}
          Workflow Run ID: [${{ steps.workflow_run_info.outputs.id }}](${{ steps.workflow_run_info.outputs.url }})
          `;
          github.issues.createComment({
            issue_number: context.issue.number,
            owner: context.repo.owner,
            repo: context.repo.repo,
            body: comment_body
          })
```

This way we don't need to repeat report steps or repeat report jobs three times.  If `steps.workflow_status.outputs.status` doesn't not work well in the comment. We could do

```
if: steps.workflow_status.outputs.status == 'failure' ( or 'success' or cancelled)
   echo ::set-output name=Buildstatus::failed. ( or successful, or cancelled)
...
...
comment_body = `
          @${{ github.actor }} Build(s) ${{ steps.***.outputs.Buildstatus}}
...
```

Icohedron:
@sophia-guo 
The issue with the current implementation of the reportStatus job is that if a bad platform were specified, github actions would run reportStatus *before* running and reporting the failure status on runBuild, therefore causing the status report to be "Build(s) successful!" due to there being no failures at the time the reportStatus job is ran. I do not know why github actions exhibits this odd behavior.

Example: See run [605808102](https://github.com/Icohedron/openjdk-tests/actions/runs/605808102) from my fork
image

The issue is that the `uses: martialonline/workflow-status@v2` is not able to correctly determine that the workflow will fail in this situation, but the built-in `failure()` function in github actions can.

Alternatively, I could incorporate the status reporting into the runBuild job, which would result in a status report for each build ran. I suppose this is a good alternative to having three separate reportStatus jobs. However it may result in a lot of comments being made if the user specified multiple builds.

Icohedron:
I'm thinking of perhaps restricting the workflow to only running one build. If the user wants to make multiple builds with different parameters, they will need to make multiple comments. 
Currently, each parameter accepts more than one argument, and passes them into the runBuild job with creates a matrix.

sophia-guo:
> The issue is that the uses: martialonline/workflow-status@v2 is not able to correctly determine that the workflow will fail in this situation

I see. Jobs failed at the right beginning due to no runner matching the specified platform. So job hasn't started on the runners, which is probably the reason that martialonline/workflow-status@v2 is not able to correctly detect it.
For this case probably reconsider my suggestion at https://github.com/AdoptOpenJDK/openjdk-tests/pull/2222#discussion_r578516229, which setup a map between supported platforms to github runners. For not supported ones the parseComment job could throw an error and runbuild job will and should be skipped.

> Alternatively, I could incorporate the status reporting into the runBuild job, which would result in a status report for each build ran

In this case I'm fine to keep it as three separate reportStatus jobs. Just to confirm the workflow behaviour `if: failure()
    needs: runBuild	    needs: [parseComment, runBuild]` .  In the case of parseComment fails ( which already reports in the step of `name: Create error comment`) if the step of `reportFailure:` of job `reprotStatus` comments the failure again?

Icohedron:
Yes, at the moment `reportStatus` will report a failure after the `parseComment` reports its own failure. I suppose I can fix this by having `parseComment` output a variable which `reportStatus` checks before running.

Icohedron:
I have pushed a change to prevent reportStatus from reporting "Build(s) failed" if parseComment encountered an error in parsing and created an error comment already.

Icohedron:
If there are no more comments or suggestions, I believe this PR is ready for review @sophia-guo @llxia

sophia-guo (approved):
LGTM

llxia (approved):
LGTM

smlambert:
Thanks for the reviews @llxia and @sophia-guo and for this work @Icohedron !

Issue #2298, opened 2021-02-23
Update Renaissance suite to latest release v0.11.0


smlambert:
We are currently using Renaissance v0.9.0 and should update to latest release (https://github.com/renaissance-benchmarks/renaissance/releases/tag/v0.11.0) - update can be made in perf/renaissance/[build.xml_L37](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/perf/renaissance/build.xml#L37) file.

We should then check if this addresses any of the renaissance benchmark failures reported under
- https://github.com/AdoptOpenJDK/openjdk-tests/issues/1932
- https://github.com/AdoptOpenJDK/openjdk-tests/issues/1994
- https://github.com/AdoptOpenJDK/openjdk-tests/issues/2159
- https://github.com/AdoptOpenJDK/openjdk-tests/issues/2183
- https://github.com/AdoptOpenJDK/openjdk-tests/issues/2253
- https://github.com/AdoptOpenJDK/openjdk-tests/issues/2254
- https://github.com/AdoptOpenJDK/openjdk-tests/issues/2255

sophia-guo:
Renaissance's releases provides binary file formats which means we can get the latest releases by auto script. 

`curl -s https://api.github.com/repos/renaissance-benchmarks/renaissance/releases/latest`
`curl -s https://api.github.com/repos/renaissance-benchmarks/renaissance/releases/latest | grep "browser_download_url"`

smlambert:
Ah, sorry @sophia-guo - I did not see your comment before merging the PR related to this issue.

sophia-guo:
That's ok. I've related this with #1971 we can close this.

PR #2314, opened 2021-03-01
Change Renaissance 0.9.0 to 0.11.0 in build.xml

xfxcwynlc:
Change the download address to v0.11.0 at line 37.
Closes: #2298

### DIFF
--- a/perf/renaissance/build.xml
+++ b/perf/renaissance/build.xml
@@ -34,7 +34,7 @@
 		        <echo message="${FILE_NAME} exists. Hence, not downloading it." />
 		    </then>
 		    <else>
-				<var name="curl_options" value="-Lks https://github.com/renaissance-benchmarks/renaissance/releases/download/v0.9.0/renaissance-mit-0.9.0.jar -o ${FILE_NAME}"/>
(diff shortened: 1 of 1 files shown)


sophia-guo (approved):

sophia-guo:
LGTM. Just wondered what's the difference between GPL version and MIT version? Looks like former releases have two versions and we downloaded MIT version. Now the latest release only have the GPL version. Is it good using GPL version? @smlambert

smlambert:
re: https://github.com/AdoptOpenJDK/openjdk-tests/pull/2314#issuecomment-787911322 - GPL is fine in this case, we are neither changing or distributing it, but merely running this benchmark (and given MIT licensed one is no longer offered we will proceed with this one).

smlambert:
Grinder (jdk8/j9/sanity.perf): https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Perf/18
Grinder (jdk16/hs/sanity.perf): https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Perf/19

smlambert (approved):
Running Grinders to see if this update fixes some issues with jdk16 runs (so that the related issues may also be closed, but either way, we want this update, so will merge).

PR #2318, opened 2021-03-01
Ignore leading and trailing whitespace in parsing of comments (runAqa.yml)

Icohedron:
Fixes the issue in which extra whitespace caused the parser to read them as extra arguments, leading to odd error messages. For example, ending the comment with a mix of spaces and newlines results in an error such as this:
image

Also simplifies the mapping of comment/command arguments to outputs in the code.


Icohedron:
In effect, extra whitespace is completely ignored, so one could make a comment like this and it would correctly be parsed.
image

sophia-guo (approved):

smlambert (approved):
LGTM

Issue #2296, opened 2021-02-22
Comment triggered PR builds enhancements


llxia:
Recently, we enabled comment-triggered PR builds https://github.com/AdoptOpenJDK/openjdk-tests/issues/2199 (Thanks @Icohedron). There are a couple of enhancements that we would like to add:

- `--sdk_resource` can be a URL. It is the same as `CUSTOMIZED_SDK_URL ` in Grinder that users can provide a URL for downloading SDK. We should use this SDK for testing. We can ignore credentials for now.
```
--sdk_resource=customized --customized_sdk_url=https://xxx
```
- build SDK using github action at https://github.com/AdoptOpenJDK/build-jdk or https://github.com/eclipse/build-openj9. Users should provide jdk/openj9 repo and branch for now.
```
--sdk_resource=build
```

Icohedron:
The `AdoptOpenJDK/install-jdk` action requires an additional parameter `archiveExtension` when fetching a customized SDK from a URL. Should the workflow try to predict the archive extension from the URL or should the archive extension be provided as an additional argument?

Icohedron:
> Users should provide jdk/openj9 repo and branch for now.

How should the repo and branch to be provided for building the SDK?

sophia-guo:
Both build-jdk and build-openj9 currently auto supports PRs in related repos, i.e. AdoptOpenjdk/openjdk-builds, Eclipse/openj9, Eclispe/openj9-omr, ibmruntimes/openj9-openjdk-jdk${version}. Need to update if we'd like to support explicitly passing in repo/branch info.

sophia-guo:
> The AdoptOpenJDK/install-jdk action requires an additional parameter archiveExtension when fetching a customized SDK from a URL. Should the workflow try to predict the archive extension from the URL or should the archive extension be provided as an additional argument?

I think if we could predict the archive extension the prediction logic should be in AdoptOpenJDK/install-jdk action. That is the action could be enhanced to enable it.

Icohedron:
> Both build-jdk and build-openj9 currently auto supports PRs in related repos, i.e. AdoptOpenjdk/openjdk-builds, Eclipse/openj9, Eclispe/openj9-omr, ibmruntimes/openj9-openjdk-jdk${version}. Need to update if we'd like to support explicitly passing in repo/branch info.

So if I'm understanding this correctly, then `--sdk_resource=build` should only work for PRs in those repos?

sophia-guo:
[...]

PR #2322, opened 2021-03-01
Allow the use of a customized SDK in comment triggered PR builds (runAqa.yml)

Icohedron:
Allows the use of a customized SDK in comment triggered PR builds (runAqa.yml) as per #2296 
Introduces two new parameters and one new option for `jdk_resource`.
- The `customized_sdk_url` argument accepts a URL that points to a JDK, expected to be in a zip, tar, or 7z format.
- The `archive_extension` argument must be supplied alongside the `customized_sdk_url`  and can be either `.zip`, `.tar`, or `.7z`. This should match the file from `customized_sdk_url`. For `.tar.gz` file formats, use the `.tar` option.
- The `customized` option for `jdk_resource` will download and install the JDK from `customized_sdk_url`, extracting the archive  based on the provided `archive_extension`.
- 
Note: The `jdk_impl` argument has no effect if `jdk_resource` is set to `customized`

Known issues:
- The `customized_sdk_url` can not be one that causes a redirect, otherwise it will fail. See issue AdoptOpenJDK/install-jdk#12

Potential improvements:
- Determine the `archive_extension` based on the URL provided in `customized_sdk_url` instead of asking the user to supply it. This would be better implemented in the `AdoptOpenJDK/install-jdk` action itself rather than in this workflow. See issue AdoptOpenJDK/install-jdk#13
- Add flags for TEST_IMAGES_REQUIRED and DEBUG_IMAGES_REQUIRED. See issue AdoptOpenJDK/install-jdk#14

Example usage/run:
image

### DIFF
--- a/.github/workflows/runAqa.yml
+++ b/.github/workflows/runAqa.yml
@@ -65,6 +65,8 @@ jobs:
           comment_body = `
           @${{ github.actor }} Build(s) started with the following parameters:
           - sdk_resource: ${{ steps.argparse.outputs.sdk_resource }}
+          - customized_sdk_url: ${{ steps.argparse.outputs.customized_sdk_url }}
+          - archive_extension: ${{ steps.argparse.outputs.archive_extension }}
           - build_list: ${{ steps.argparse.outputs.build_list }}
           - target: ${{ steps.argparse.outputs.target }}
           - platform: ${{ steps.argparse.outputs.platform }}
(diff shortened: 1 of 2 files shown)


.github/workflows/runAqa.yml
```diff
@@ -97,9 +101,19 @@ jobs:
       matrix: ${{ fromJson(needs.parseComment.outputs.build_parameters) }}
     steps:
     - uses: AdoptOpenJDK/install-jdk@v1
+      if: matrix.sdk_resource != 'customized'
       with:
         version: ${{ matrix.jdk_version }}
         source: ${{ matrix.sdk_resource }}
+        sourceType: 'buildType'
```
sophia-guo, line 108: sourceType can be automatically infer. Though it doesn't hurt to explicitly set it.

.github/workflows/runAqa.yml
```diff
@@ -97,9 +101,19 @@ jobs:
...
+        impl: ${{ matrix.jdk_impl }}
+    - uses: AdoptOpenJDK/install-jdk@v1
+      if: matrix.sdk_resource == 'customized'
+      with:
+        version: ${{ matrix.jdk_version }}
+        source: ${{ matrix.customized_sdk_url }}
+        archiveExtension: ${{ matrix.archive_extension }}
+        sourceType: 'url'
```
sophia-guo, line 116: Same here.

sophia-guo (approved):
LGTM

llxia (approved):
LGTM

PR #2328, opened 2021-03-02
Excluding ResetPeakMemoryUsage on openj9

psoujany:
This PR is to exclude hotspot related jdk_management test on openj9.
java/lang/management/MemoryMXBean/ResetPeakMemoryUsage.java - exclusion on openj9 for jdk11

### DIFF
--- a/openjdk/ProblemList_openjdk11-openj9.txt
+++ b/openjdk/ProblemList_openjdk11-openj9.txt
@@ -133,6 +133,7 @@ sun/management/HotspotRuntimeMBean/GetSafepointCount.java	https://github.com/Ado
 sun/management/HotspotRuntimeMBean/GetSafepointSyncTime.java	https://github.com/AdoptOpenJDK/openjdk-tests/issues/932 generic-all
 sun/management/HotspotRuntimeMBean/GetTotalSafepointTime.java	https://github.com/AdoptOpenJDK/openjdk-tests/issues/932 generic-all
 sun/management/HotspotThreadMBean/GetInternalThreads.java	https://github.com/AdoptOpenJDK/openjdk-tests/issues/932 generic-all
+java/lang/management/MemoryMXBean/ResetPeakMemoryUsage.java https://github.com/AdoptOpenJDK/openjdk-tests/issues/1297 generic-all
 
 ############################################################################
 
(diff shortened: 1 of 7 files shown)


smlambert:
Thanks @psoujany, some initial questions:
- is this specific to jdk11 (if applicable also to jdk16+, should also update those files)?
- is there an open issue that can be referenced (as in future we will have automation to verify if issue is closed, and can re-include test.  If the URL is unreachable/internal, automation will not work)?

llxia:
Is ResetPeakMemoryUsage a part of `jdk_management ` test? This test target is excluded for openj9 https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/openjdk/playlist.xml#L753 as part of https://github.com/eclipse/openj9/issues/10757
 
Do we have a plan to re-enable `jdk_management ` after this PR? If not, maybe we do not need the problem list exclusion.

psoujany:
Hi @ShelleyLambert 
1. This test exclusion is specific to openj9. So, yes this change would be applicable to other releases as well. I'll make changes to releases. 
2. Yes, there's an open issue which is raised in internal repo i.e, openj9-openjdk-jdk11-zos/issues/454.

Thank you.

psoujany:
Hi @llxia 
Yes, ResetPeakMemoryUsage is a part of jdk_management test. We don't have any plan to re-enable jdk_management after this PR. There's an open issue for this testcase in openj9-openjdk-jdk11-zos/issues/454, as part of fixing the issue we need this test exclusion. 

This testcase is also not supported on openj9, so even in future enabling jdk_management this testcase would fail. Hence exclusion would be valid fix.

smlambert:
> This testcase is also not supported on openj9, so even in future enabling jdk_management this testcase would fail. Hence exclusion would be valid fix.

So it sounds like in this case that this is a permanent exclude (the test case is not valid for openj9).  For permanent excludes, then please change the URL from the internal one (runtimes/openj9-openjdk-jdk11-zos/issues/454) that we have no visibility of to https://github.com/AdoptOpenJDK/openjdk-tests/issues/1297.

I see the confusion, when I referred to 'open issue' I meant an issue in an open-source repo (not an internal repo) that is visible to everyone (and can be queried by automation we will run from the open-source projects).  :)

openjdk/ProblemList_openjdk11-openj9.txt
```diff
@@ -133,7 +133,7 @@ sun/management/HotspotRuntimeMBean/GetSafepointCount.java	https://github.com/Ado
 sun/management/HotspotRuntimeMBean/GetSafepointSyncTime.java	https://github.com/AdoptOpenJDK/openjdk-tests/issues/932 generic-all
 sun/management/HotspotRuntimeMBean/GetTotalSafepointTime.java	https://github.com/AdoptOpenJDK/openjdk-tests/issues/932 generic-all
 sun/management/HotspotThreadMBean/GetInternalThreads.java	https://github.com/AdoptOpenJDK/openjdk-tests/issues/932 generic-all
-
+java/lang/management/MemoryMXBean/ResetPeakMemoryUsage.java https://github.ibm.com/runtimes/openj9-openjdk-jdk11-zos/issues/454 generic-all
```
smlambert, line 136: @psoujany - please use https://github.com/AdoptOpenJDK/openjdk-tests/issues/1297 instead of the https://github.ibm.com/runtimes/openj9-openjdk-jdk11-zos/issues/454 internal issue

smlambert (changes requested):

psoujany:
Hi, @ShelleyLambert. I applied change as you mentioned to all releases i.e, jdk11+. Please review the changes. Thank you.

smlambert (approved):
thanks @psoujany

smlambert:
@llxia - anything further updates you would like to see on this PR, or are we good ?

llxia (approved):
LGTM

PR #2334, opened 2021-03-03
Only run foreign native test on x86_64 or aarch64 platforms

adamfarley:
This test requires CLinker, and the upstream community has no
current intention to port it to new platforms. So we need to exclude
the test on platforms that aren't x86-64 or aarch64.


adamfarley:
Linked to https://github.com/AdoptOpenJDK/openjdk-tests/issues/2265

Requesting reviews from Shelley and Sophia, as they're familiar with that issue.

sophia-guo (approved):

smlambert (approved):

PR #2335, opened 2021-03-03
Update timeout logic

llxia:
We can use Jenkins timeout to simplify the logic. The test build will be aborted if the timeout is reached.
By default, if there is no node available, abort the test build immediately. If ACTIVE_NODE_TIMEOUT is set,
we will set timeout to ACTIVE_NODE_TIMEOUT. Also, for dynamic vm agent, we need to use node(LABEL) directly
to spin up the agent. nodesByLabel(LABEL).size() will not work.

### DIFF
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -223,30 +223,14 @@ if (JDK_VERSIONS.size() > 1 || JDK_IMPLS.size() > 1 || PLATFORMS.size() >1 || PL
         println "LABEL: ${LABEL}"
 
         stage('Queue') {
-            def nodes = nodesByLabel(LABEL).size()
-            if (nodes < 1) {
-                // If no active node matches the label, we see if there's a timeout value set.
-                // If there is, we wait and check again periodically. If not, we fail immediately.
+            if (nodesByLabel(LABEL).size() < 1) {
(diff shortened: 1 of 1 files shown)


llxia:
- regular run: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/7426/
- use an invalid label to mimic the scenario that we do not have the machine and no ACTIVE_NODE_TIMEOUT is set: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/7425/console

smlambert (approved):
that's nice clean code to read, thanks @llxia !

sophia-guo:
What if there are available nodes though all nodes are busying? Are the jobs waiting until nodes available?

nodesByLabel: `list of nodes by Label, by default excludes offline nodes.`https://www.jenkins.io/doc/pipeline/steps/pipeline-utility-steps/#nodesbylabel-list-of-nodes-by-label-by-default-excludes-offline-nodes`

llxia:
nodesByLabel returns all nodes that match with the label, including the nodes that are busy. That is nodesByLabel(LABEL).size() > 0 when all nodes are busy. The script will proceed to the next step.

sophia-guo:
That is if `nodesByLabel(LABEL).size() > 0` the job will wait until next available node, which is a separate issue from this PR. Will merge this one first.

sophia-guo (approved):

PR #2336, opened 2021-03-04
use sw.os.alpine-linux label instead of sw.os.musl for Alpine

sxa:
Ref: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/1929#issuecomment-790781251


smlambert (approved):
thanks @sxa !

sophia-guo (approved):

PR #2338, opened 2021-03-05
Removing unneeded trycatch

adamfarley:
We don't need this, as the inner exec task doesn't need to be
failonerror. Removing the failonerror also covers us if the trycatch
doesn't catch the failure, which we've seen in a couple of places.

### DIFF
--- a/openjdk/build.xml
+++ b/openjdk/build.xml
@@ -188,16 +188,12 @@
 								<property name="jdkName" value="openjdk-aarch64-jdk8u" />
 							</then>
 							<else>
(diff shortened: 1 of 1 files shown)


adamfarley:
Test: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/7489/console

sophia-guo (approved):

smlambert (approved):

adamfarley:
Thanks @smlambert & Sophia :)

PR #2339, opened 2021-03-05
Increase xmx for jdk_utils to prevent TimSortStackSize2 OOM

adamfarley:
When you run that test with the xmx set to 512m and the CompressedOops
disabled, we hit an oom on jdk8. This is not the case for jdk11+
because the test has code to increase the xms and xmx values
appropriately. We've only been hitting this oom recently because
we've only just started running jdk_utils with the CompressedOops
disabled.

However, putting the change into playlist.xml means we'll see the same
oom if you try to run a grinder for this testcase, as that runs it under
the jdk_custom target which uses the small xmx value.

So this will be a temporary change, kept until the jdk8 updates project
integrates the long-term change to the test itself. See JDK-8263099.


smlambert (approved):

sophia-guo (approved):

sophia-guo:
Same failure in run-aqa. LGTM.

PR #2341, opened 2021-03-08
Fix mapping from x86-64_mac to macos GitHub runner in Comment Triggered PR Build Workflow

Icohedron:
`x86-64_mac` was incorrectly mapped to `mac-latest` which is an invalid GitHub runner name. The correct GitHub runner name for MacOS is `macos-latest`


sophia-guo (approved):
LGTM

llxia (approved):
LGTM

PR #2342, opened 2021-03-08
Add JDK_NATIVE_OPTIONS in jdk_net

llxia:
Resolve: https://github.com/eclipse/openj9/issues/12081

### DIFF
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -485,7 +485,7 @@
 			<variation>Mode1000</variation>
 		</variations>
 		<command>$(JAVA_COMMAND) -Xmx512m -jar $(Q)$(TEST_RESROOT)$(D)jtreg$(D)lib$(D)jtreg.jar$(Q) \
-	$(JTREG_BASIC_OPTIONS) -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS)$(Q) \
+	$(JTREG_BASIC_OPTIONS) $(JDK_NATIVE_OPTIONS) -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS)$(Q) \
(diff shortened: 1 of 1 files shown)


llxia:
|JDK Version|impl|Grinder|
|--|--|--|
|17| openj9| Grinder/14075|
|11| openj9| https://ci.adoptopenjdk.net/job/Grinder/7521/|
|8| openj9| https://ci.adoptopenjdk.net/job/Grinder/7524/|
|11| hotspot| https://ci.adoptopenjdk.net/job/Grinder/7525/|

smlambert (approved):
thanks @llxia - as this is a fix for failure, it will get merged despite our code freeze ahead of next week's release.

sophia-guo:
Looks like more native tests are continually | gradually  being added for new jdk version. Instead of keeping adding JDK_NATIVE_OPTIONS when error happens I think we can add JDK_NATIVE_OPTIONS to all jdk_* testcases( exclude jdk_custom). And add $(JVM_NATIVE_OPTIONS) to all other testcases. 

That is JDK_NATIVE_OPTIONS is one of default options for jdk_* testcase and JVM_NATIVE_OPTIONS is one of default options for non jdk_* testcases. We have limited those two native options to jdk8+ and exist(TESTIMAGE_PATH) so it won't hurt if there is no testimage available.

sophia-guo:
As this is a fix we can merge this first and update other target later.

sophia-guo (approved):

smlambert:
Might be worth updating all as suggested in https://github.com/AdoptOpenJDK/openjdk-tests/pull/2342#issuecomment-796169726, since there are other targets that also have new native tests blended into their targets, see related: https://github.com/AdoptOpenJDK/openjdk-tests/issues/2380

llxia:
I close this PR as we would like to add the option as default.

PR #2345, opened 2021-03-09
Allow for cgroups in MEMORY_SIZE detection

sxa:
This is a bit unpleasant but if we're in a restricted docker container we need to restrict the number of CPUs and RAM quantity selected otherwise it'll mis-detect the physical machine's quantity and overwhelm the container and probably exhaust the container's RAM. While it won't affect other containers, but will likely cause lots of timeouts as the container tries to deal with all the process contention and result in potentially intermittent hard-to-diagnose test failures. I'm not sure how many Linux systems won't have /sys/fs/cgroup but I've put some detection in to cover the case where that doesn't exist (It should fall back to the same values as before)

This is an enhancement to https://github.com/AdoptOpenJDK/openjdk-tests/pull/1427 and will hopefully resolve https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/2002 (Also Ref: General container issue https://github.com/AdoptOpenJDK/openjdk-tests/issues/2138)

I'm going to run some more testing on this, but it seems to do the right thing in several of the checks I've done.

### DIFF
--- a/openjdk/openjdk.mk
+++ b/openjdk/openjdk.mk
@@ -18,7 +18,7 @@ MEMORY_SIZE:=1024
 OS:=$(shell uname -s)
 
 ifeq ($(OS),Linux)
-	NPROCS:=$(shell KPROCS=`grep -c ^processor /proc/cpuinfo`; CGPROCS=`if [ -r /sys/fs/cgroup/cpu,cpuacct/cgroup.procs ]; then wc -l < /sys/fs/cgroup/cpu,cpuacct/cgroup.procs; else echo "$$KPROCS"; fi`; if [ "$${KPROCS}" -lt "$${CGPROCS}" ]; then echo "$${KPROCS}"; else echo "$${CGPROCS}"; fi)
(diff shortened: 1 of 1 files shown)


sxa:
Based on running the two jobs at [7544](https://ci.adoptopenjdk.net/job/Grinder/7544) and [7545](https://ci.adoptopenjdk.net/job/Grinder/7545) on a host that was showing problems it is now doing a better job of restricting itself. The load on the host is around 15-17 when running on two 8-core docker images, and the RAM usage of each is about 2Gb out of the 8Gb they have allocated. This contrasts with a load of over 100 and all 8Gb in use before this patch was applied.

jerboaa:
@sxa Sorry, but what is this `cpu.cfs_quota_us` fiddling supposed to do? Are you inferring CPU settings from `cpu.cfs_quota_us` only? This doesn't seem right. I'd think you'd need some form of fraction between `cpu.cfs_quota_us` **and** `cpu.cfs_period_us` at least. But then there are cpu.shares how quota could be enforced and what not else. Not sure that's worth it. See also:
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/resource_management_guide/sec-cpu

Are you sure you'll need to do anything in terms of cgroups CPU quotas? As an initial shot I'd only factor in the cgroup (v1!) memory limit and tests will likely run better with it in place. It's probably what you are observing now anyway. If you really need CPU too, this is a bit of a can of worms as it might get cpu limited in a variety of ways. This will get even more fun once you add in cgroups v2 hosts into the mix and docker containers.

sxa:
I'm not sure of anything at the moment :-) I was just trying to find something that would stop the jobs failing due to being overloaded by detecting more resources than the containers had and this was the first option that seemed to work. This close to the JDK16 GA I want to get something in that will help even if we change it later

jerboaa (approved):
This patch seems OK.

sxa:
Looks to be working - the container I'm testing with is sitting at less than it's capacity with just the memory check :-)

smlambert (approved):
Thanks @sxa !

PR #2364, opened 2021-03-10
Define JDK_REPO, JDK_BRANCH as variables of test job template

### LINKED ISSUE #2362: Enable JDK_REPO, JDK_BRANCH as variables of test job template
Current test job template doesn't take JDK_REPO, JDK_BRANCH as variables, which means test jobs can only be created with JDK_REPO, JDK_BRANCH as default empty string. Enabling them as variables gives user the flexibility of creating jobs with user defined value as default value. Hence jobs can be triggered by inputting parameters as less as possible.

sophia-guo:
Close #2362 


llxia (approved):
LGTM

llxia (approved):
LGTM

smlambert (approved):

PR #2372, opened 2021-03-10
Revert "auto exclude test jdk_imageio plat=*mac*"

smlambert:
Reverts AdoptOpenJDK/openjdk-tests#2371


smlambert:
Reverting due to issue with regex

PR #2374, opened 2021-03-12
Allow VENDOR params to be passed to Test Autogen

AdamBrousseau:
- Repos, Branches, Dirs. It is unlikely that
  a user would need to default hardcode a
  sha.
- This allows test jobs to be generated with
  default vendor values. Default to empty if
  nothing is passed.

@llxia for review
Internal sandbox autogen job run 93


llxia:
dup of https://github.com/AdoptOpenJDK/openjdk-tests/pull/2364

AdamBrousseau:
Sorry, I saw that PR but didn't look at it. Thanks.

PR #2377, opened 2021-03-12
Set environment variable only it is not empty

sophia-guo:
Related with https://github.com/AdoptOpenJDK/openjdk-tests/issues/2181#issuecomment-765468523

### DIFF
--- a/openjdk/build.xml
+++ b/openjdk/build.xml
@@ -90,10 +90,13 @@
 				</exec>
 			</then>
 		</if>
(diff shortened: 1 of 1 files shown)


sophia-guo:
Grinder with no vmagent and set variable as empty: https://ci.adoptopenjdk.net/view/work-in-progress/job/grinder_sandbox/324/

Grinder with vmagent and set variable as empty: https://ci.adoptopenjdk.net/view/work-in-progress/job/grinder_sandbox/323/

smlambert (approved):

llxia (approved):
LGTM

Issue #799, opened 2019-01-11
Add solaris jdk8 testing (openjdk + system)


smlambert:
Now that there are [solaris builds for jdk8](https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk8u/job/jdk8u-solaris-sparcv9-hotspot), we should enable testing

jdk8 openjdk: 

- [ ] [openjdk8_hs_openjdktest_sparcv9_solaris](https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk8_hs_openjdktest_sparcv9_solaris/)

jdk8 system:

- [ ] openjdk8_hs_systemtest_sparc_solaris

Do not yet see a live jdk11 build for solaris, so will wait for it before enabling testing in that version.

smlambert:
smlambert:solarisTest has the updates required, now just need a solaris machine with the test prereqs installed on it:

Test job fails due to "Can't locate XML/Parser.pm" which is one of the test prereqs:
https://ci.adoptopenjdk.net/view/Test_openjdk/job/openjdk8_hs_openjdktest_sparcv9_solaris/7/console

smlambert:
Now we have a dedicated solaris machine for test, will revisit this item.  https://ci.adoptopenjdk.net/computer/test-inspira-solaris10u11-sparcv9-1/

smlambert:
related: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/846

smlambert:
Once https://github.com/eclipse/openj9/pull/6737 is merged, and we create the test jobs for versions required, this can be closed.

smlambert:
Grinder: 
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/706

smlambert:
Closing this as test portion of this is completed, we still require a solaris machine with all of the prereqs installed for use by testing, but there is no further test work to do.  

Related: https://github.com/AdoptOpenJDK/openjdk-infrastructure/issues/631

sxa:
Let's reopen this and try to complete it :-) We've put in a couple of PRs recently and at the time of writing the jobs still aren't quite able to complete:
- https://github.com/AdoptOpenJDK/TKG/pull/110
- https://github.com/AdoptOpenJDK/TKG/pull/157
[...]

PR #2392, opened 2021-03-18
Use env for solaris

sophia-guo:
related #799 


smlambert (approved):

sxa (approved):
I've been running with this change already to make Solaris/x64 work, so I very much approve!

PR #2404, opened 2021-03-22
Excluding a 6hr test that causes timeout

adamfarley:
TestAESMain takes 6+ hours when run without the JIT, which jdk8 on
Hotspot (for linux_s390) doesn't have.

Excluding it until the issue is resolved, if ever, to enable the
execution and triage of the other tests in this bucket.

### DIFF
--- a/openjdk/ProblemList_openjdk8.txt
+++ b/openjdk/ProblemList_openjdk8.txt
@@ -12,6 +12,12 @@
 # limitations under the License.
 #############################################################################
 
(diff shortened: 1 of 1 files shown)


smlambert (approved):

sxa (approved):
Based on [the analysis I did last night](https://github.com/AdoptOpenJDK/openjdk-tests/issues/1051#issuecomment-804359944) excluding this seems the prudent thing to do so I approve

PR #2405, opened 2021-03-22
Update jtreg location

sophia-guo:

### DIFF
--- a/openjdk/build.xml
+++ b/openjdk/build.xml
@@ -52,9 +52,9 @@
 							<copy file="custom_jtreg.tar.gz" tofile="${jtregTar}.gz"/>
 					</then>
 				<else>
-					<echo message="Getting Jtreg from https://ci.adoptopenjdk.net/job/jtreg/lastSuccessfulBuild/artifact/${jtregTar}.gz"/>
+					<echo message="Getting Jtreg from https://ci.adoptopenjdk.net/job/dependency_pipeline/lastSuccessfulBuild/artifact/jtreg/${jtregTar}.gz"/>
 					<exec executable="curl" failonerror="true">
-						<arg line="--retry 5 -OLJSks https://ci.adoptopenjdk.net/job/jtreg/lastSuccessfulBuild/artifact/${jtregTar}.gz" />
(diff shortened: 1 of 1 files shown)


sophia-guo:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/7970/console
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/7969/console

smlambert (approved):

openjdk/build.xml
```diff
@@ -52,9 +52,9 @@
 							<copy file="custom_jtreg.tar.gz" tofile="${jtregTar}.gz"/>
 					</then>
 				<else>
-					<echo message="Getting Jtreg from https://ci.adoptopenjdk.net/job/jtreg/lastSuccessfulBuild/artifact/${jtregTar}.gz"/>
+					<echo message="Getting Jtreg from https://ci.adoptopenjdk.net/view/Dependencies/job/dependency_pipeline/lastSuccessfulBuild/artifact/jtreg/${jtregTar}.gz"/>
 					<exec executable="curl" failonerror="true">
-						<arg line="--retry 5 -OLJSks https://ci.adoptopenjdk.net/job/jtreg/lastSuccessfulBuild/artifact/${jtregTar}.gz" />
+						<arg line="--retry 5 -OLJSks https://ci.adoptopenjdk.net/view/Dependencies/job/dependency_pipeline/lastSuccessfulBuild/artifact/jtreg/${jtregTar}.gz" />
```
llxia, line 57: If Jenkins changes the view tab name, this link will be invalid. Instead of using view in the link, could we use the job directly? For example:
```
https://ci.adoptopenjdk.net/job/dependency_pipeline/lastSuccessfulBuild/artifact/jtreg/${jtregTar}.gz
```

sophia-guo: Updated. Though personally I think if anything changed we should be notified and update accordingly. Using the job directly feels like we are doing some hacking to make it work anyway.

llxia: IMO, the direct link is not hacking. A view can be customized/changed/renamed or a job can be move from one view to another. Using a direct link saves the trouble of updating the link in case these things happens.

llxia (approved):
LGTM

PR #2409, opened 2021-03-25
Update condition

sophia-guo:
To use CLOUD_PROVIDER test jobs need to be regenerated with updated template. This fix allows the current jobs without regeneration work. To use dynamic agent we need to regenerate test jobs.


sophia-guo:
Test:
https://ci.adoptopenjdk.net/job/grinder_sandbox/410/console

This is the case no available machine and no parameter ClOUD_PROVIDER. Job falls back to local machine waiting.

sophia-guo:
For case 3 I think current logic can't cover the case if user define the LABEL by a machine name directly.

sophia-guo:
https://github.com/AdoptOpenJDK/openjdk-tests/pull/2409#issuecomment-806976880 should be covered by commit check if label are set specifically

sophia-guo:
Close this one, changes will go to #2417

PR #2419, opened 2021-03-26
Exclude long-running tests on s390 JDK8 HS extended.openjdk tests

adamfarley:
On s390, JDK8 with Hotspot has no JIT. This results in a number of
tests running for over an hour, eventually causing an
extended.openjdk test run on Jenkins to be aborted due to the usual
10hr timeout.

While it seems likely we will not be running extended.openjdk on s390
(for JDK8 on Hotspot, anyway) for much longer, I'm excluding these
tests for now so that the test bucket is viable, should we choose to
run it again in the future.

### DIFF
--- a/openjdk/excludes/ProblemList_openjdk8.txt
+++ b/openjdk/excludes/ProblemList_openjdk8.txt
@@ -14,7 +14,8 @@
 
 # hotspot_all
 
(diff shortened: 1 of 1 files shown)


sophia-guo (approved):

smlambert:
Please resolve conflicts @adamfarley

smlambert (approved):

adamfarley:
Conflicts resolved @smlambert :)

adamfarley:
Thx all :)

Issue #2317, opened 2021-03-01
Update external tests to allow for autogen of Docker images for jdk16+


smlambert:
**👾 Description of the issue**

jdk 16 sanity.external jobs are being launched but the build fails with this error:

11:26:31  build_image:
11:26:31       [echo] Executing external.sh --build --dir camel --tag nightly --version 16 --impl openj9 
11:26:31       [exec] ERROR: Invalid Version: 16
11:26:31       [exec] Usage: /home/jenkins/workspace/Test_openjdk16_j9_sanity.external_ppc64le_linux/openjdk-tests/TKG/../../jvmtest/external/external.sh [8 9 10 11 12 13 14 15]

As a quick fix, this will require an update to [common_functions.sh:L17](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/external/common_functions.sh#L17) to add 16 and 17 to the list.  

As a future fix, perhaps we can come up with a mechanism to check for the existence of jdk git repos to verify which versions are released (and then add one to include the main/tip of development).  This future fix does not have to happen as part of this issue.  It is fine to focus on the quick fix defined above.

In order to test the fix locally, you can [download a jdk16 binary](https://adoptopenjdk.net/?variant=openjdk16&jvmVariant=hotspot) from the API or website then follow these instructions for running a test locally: https://github.com/AdoptOpenJDK/openjdk-tests/tree/master/external#running-external-tests-locally 

If the fix was good, then you will not hit `ERROR: Invalid Version: 16` in the output during compilation.

**📋 Step by Step**

To solve this issue and contribute a fix you should check the following step-by-step list. A more detailed documentation of the workflow can be found [here](https://github.com/firstcontributions/first-contributions/blob/master/README.md).

- [ ] Claim this issue: Comment below.
- [ ]  Fork the repository in github by simply clicking the 'fork' button.
- [ ]  Check out the forked repository
- [ ]  Create a feature branch for the issue. We do not have any naming definition for branches.
- [ ]  Commit your changes.
- [ ]  Start a Pull Request.
- [ ]  Done 👍 Ask in comments for a review :)
- [ ]  If the reviewer find some missing peaces or a problem he will start a discussion with you and describe the next steps how the problem can be solved.
- [ ]  You did it 🎉 We will merge the fix in the master branch.
- [ ]  Thanks, thanks, thanks for being part of this project as an open source contributor ❤️

**🤔❓ Questions**

[...]

PR #2456, opened 2021-04-02
Update external tests to allow for autogen of Docker images for jdk16+

brngylni:
Issue:#2317  

I have updated the common-functions.sh file and I created a file named "future-update.sh". This can be useful for checking available releases in future updates.

Test results : 

Before : 
Screenshot from 2021-04-02 14-28-43
After : 

Screenshot from 2021-04-02 14-34-37


sophia-guo:
We could probably move `function function get_versions()` to common_functions.sh and update https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/external/common_functions.sh#L17 to be supported_versions="'$(get_versions)'". However external tests will be running on different platforms and may run locally, which means the platform may not be `linux`. So `dpkg-query` , `sudo apt-get` and `jq` may not work.

sophia-guo:
Probably we can first only include https://github.com/AdoptOpenJDK/openjdk-tests/pull/2456/commits/3ca00128913c548a1009765331189b69a47e6785 in this PR, and works on https://github.com/AdoptOpenJDK/openjdk-tests/pull/2456/commits/c7fea3509170cfc0de809f5d5ee2e4a2aa94fd04 in a separate PR.

sophia-guo:
Please also sign off your commit and update your comment https://github.com/AdoptOpenJDK/openjdk-tests/pull/2456#issue-608064152 with Close #2317  , which will automatically close the issue #2317  when this PR is merged.

brngylni:
> Please also sign off your commit and update your comment [#2456 (comment)](https://github.com/AdoptOpenJDK/openjdk-tests/pull/2456#issue-608064152) with Close #2317 , which will automatically close the issue #2317 when this PR is merged.

I rebased my commits but they are not belongs to any branch now. Is that a problem?

external/future_update
```diff
@@ -0,0 +1,15 @@
+function get_versions(){
```
smlambert, line 1: Please add a comment to describe the purpose of this function.  Since currently it is not called, there is no way for us to test it in automation.  I believe @sophia-guo was suggesting this function get moved to common_functions.sh and the supported versions line get updated to call it.  

Once this is done, we would have to test it on all platforms where these tests are expected to run.  I can give instructions on how you will do that once this PR is updated.

smlambert (changes requested):

external/future_update
brngylni: I hope it fulfills your request.

sophia-guo:
Yes @smlambert I would suggest to remove the `function get_versions()` be included separately under a different PR, which will also move the function to the common_functions.sh and get called. The `function get_vesions()` are expected to be updated when running on all platforms we support the external testing on. By separating that we can get the fixes for #2317 in. And then enhance the story in the separate PR.

brngylni:
@smlambert @sophia-guo 

So I am removing it if both of you are agreed.

smlambert:
Yes, let's separate it out so we can bring the one liner change in quickly.  Thanks @brngylni !

brngylni:
> Yes, let's separate it out into a separate PR so we can bring the one liner change in quickly. Thanks @brngylni !

OK I removed it.

smlambert (approved):

sophia-guo (approved):
LGTM

PR #2424, opened 2021-03-30
Add documentation for the Run AQA (Comment-Triggered PR Builds) workflow

Icohedron:
Anything else I should add to the documentation?

Amendments will be made to the documentation when SDK builds #2399 is implemented.

PR #2457, opened 2021-04-05
Add help command to Run AQA (Comment-Triggered PR Build) Workflow 

Icohedron:
Adds a `--help` / `-h` argument to the Run AQA workflow #2199 which links to the documentation #2424 
image


sophia-guo (approved):
LGTM

llxia (approved):
LGTM

Icohedron:
An issue was overlooked in which the help flag itself is being passed to runBuild's job matrix.
But since the help flag is a boolean and not an array of values, it prevents the matrix from being constructed and hence no builds are ran.
#2465 Should be merged to revert this PR, and adoptium/TKG#173 contains a fix for this issue.

PR #2462, opened 2021-04-06
If DOCKERIMAGE_TAG is not set set nightly as default

sophia-guo:
Pass in DOCKERIMAGE_TAG as string so external.sh get check if it is empty just set as default nightly.

Related https://github.com/adoptium/run-aqa/issues/80

### DIFF
--- a/external/camel/playlist.xml
+++ b/external/camel/playlist.xml
@@ -28,9 +28,9 @@
 	</test>
 	<test>
 		<testCaseName>camel_test</testCaseName>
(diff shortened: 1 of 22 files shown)


sophia-guo:
Tested with target _camel_test.  https://github.com/sophia-guo/runaqaTest/runs/2280862940?check_suite_focus=true

Failure is related with camel_test itself.

llxia (approved):
LGTM

smlambert (approved):
Thanks for the find and the quick turnaround on this one @sophia-guo !

Issue #2399, opened 2021-03-20
Port comment triggered PR builds into Openj9 and TKG repo


llxia:
Comment-triggered PR builds are enabled at Adopt test repo with nightly/releases. And we are working on [PR builds enhancement](https://github.com/AdoptOpenJDK/openjdk-tests/issues/2296). In order for more people to benefit from the current comment triggered PR builds feature, I think we should port it into openj9 and TKG repo. The enhancement can be ported at a later stage.

For [openj9 repo](https://github.com/eclipse/openj9):
PR builds workflow should take openj9 repo and branch from the openj9 PR. The rest repo and branch should use the default values.

For [TKG repo](https://github.com/adoptium/TKG):
PR builds workflow should take TKG repo and branch from the TKG PR. The rest repo and branch should use the default values.

llxia:
FYI @Icohedron

Icohedron:
> The rest repo and branch should use the default values.

Could you clarify what you mean by this?

> For openj9 repo:
PR builds workflow should take openj9 repo and branch from the openj9 PR.

The defaults for the openj9 repo are to build the SDK from the head repo/ref of an openj9 PR, correct?

> For TKG repo:
PR builds workflow should take TKG repo and branch from the TKG PR.

I'm not sure how this would work. Could you give an example of the default parameters you'd expect for the TKG repo?

Icohedron:
> > For TKG repo:
> > PR builds workflow should take TKG repo and branch from the TKG PR.
> 
> I'm not sure how this would work. Could you give an example of the default parameters you'd expect for the TKG repo?

Nevermind, I figure that the PR Comment Triggered Build action needs a way for the user to specify the TKG repository to use when running AdoptOpenJDK/run-aqa

llxia:
The workflow needs to be added into openj9 and TKG repo respectively. Similar to port auto test exclusion in openj9: https://github.com/eclipse/openj9/tree/master/.github/workflows

Use Case for openj9 repo:
- users open a PR in openj9 repo (not Adopt test repo)
- users add a comment to trigger PR builds. The user does NOT need to specify openj9 repo and branch in the comment. The workflow should be able to grep the information.
- PR builds are launched using openj9 user's repo and branch from the original PR (openj9 user's repo and branch should be used when running run-aqa). 

The use case for TKG repo is very similar. PR builds are launched using TKG user's repo and branch from the original PR.

smlambert:
[...]

PR #2465, opened 2021-04-07
Run AQA Workflow: Use runAqaArgParse.py script from TKG to remove duplicate code.

Icohedron:
In accordance with adoptium/TKG#172 the `runAqaArgParse.py` script has been moved to `scripts/testBot` within the [adoptium/TKG](https://github.com/adoptium/TKG) repo to avoid duplicating code when porting the Run AQA workflow to other repositories (i.e. #2399)

For repo-specific settings, a configuration file (`runAqaConfig.json`) is used instead.

Changes the AdoptOpenJDK/run-aqa@v1 action to adoptium/run-aqa@v1.

Note that this change will revert #2457, but the functionality will be reintroduced when adoptium/TKG#173 is merged.


llxia (approved):
LGTM

sophia-guo:
The workflows are still duplicate( it probably has to be). runAqaArgParse.py has been moved to TKG repo to reduce the duplication in one way. However there will be extra configure duplicate files (configure.json) in openjdk-tests and TKG repo. And it also requires the workflow to add  extra two git checkout steps (checkout both openjdk-tests and TKG repo). Those two steps are not necessary if runAqaArgParse.py also stays as run-aqa does checkout both repos. So now actually openjdk-tests and TKG repo are checkout twice. Yes, it will help if there are other repos ( e.g. eclipse/openj9) do the same thing. 

> ("Second rule of fight club, once you knock down the first opponent, a new one steps up" - Shelley Lambert)

Probably we can also convert runAqaArgParse.py to a github action.

sophia-guo (approved):
LGTM. Though probably we can think more about it.

PR #2468, opened 2021-04-07
Renaissance als failure

patkarns:
More tests are passing, but we're still seeing some failures: https://github.com/AdoptOpenJDK/openjdk-tests/issues/2253#issuecomment-801519480


patkarns:
@ShelleyLambert I've reverted to the old version of renaissance last week ([following this thread](https://adoptium.slack.com/archives/C5219G28G/p1617096126107500 )), I just want to clarify if I should be using version 9 or 11?

Thanks! :)

sophia-guo:
Yes, if the tests still fail. We'd like to exclude it for 16+.

smlambert:
Yes, I guess we will fall back to 9 for now, as if we use 11, then all renaissance on windows platforms fails. (we should likely report this upstream to the renaissance project too, but will have to gather details around the failure before doing that).

Can you please add the ADD_OPENS_CMD to the 9 other renaissance targets that fail with similar issue (as reported in: https://github.com/AdoptOpenJDK/openjdk-tests/issues/2253#issuecomment-799483599)

renaissance-als_0
renaissance-chi-square_0
renaissance-db-shootout_0
renaissance-finagle-http_0
renaissance-gauss-mix_0
renaissance-log-regression_0
renaissance-mnemonics_0
renaissance-movie-lens_0
renaissance-par-mnemonics_0
renaissance-philosophers_0

thanks @patkarns !

sophia-guo (approved):
LGTM

smlambert:
Grinders:
- jdk16 https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/8164/tapResults/ - some fails for new reasons 👍 
- jdk11 https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/8166 - pass
- jdk8 https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/8167 - pass

smlambert (approved):
thanks, LGTM

PR #2474, opened 2021-04-07
added jck logic to jck.mk

ellschi:
added a copy of [#L154-L165 from Jenkins testJobTemplate](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/testJobTemplate#L154-L165) to jck.mk #L49-L62

### DIFF
--- a/jck.mk
+++ b/jck.mk
@@ -46,18 +46,10 @@ endif
 
 
 ifndef JCK_VERSION
-  	// for jck builds, we need to pass in EXTRA_OPTIONS
-	def EXTRA_OPTIONS = "";
-	if (GROUP == "jck" && JDK_IMPL == "openj9") {
-				EXTRA_OPTIONS = "-Xfuture"
-				if (JDK_VERSION != "8") {
-					  	EXTRA_OPTIONS += " --enable-preview"
-				}
-				if (ARCH_OS == "s390x_zos") {
-						// zOS has Attach API disabled by default
-						EXTRA_OPTIONS += " -Dcom.ibm.tools.attach.enable=yes"
-				}
-	}
+  ifeq (8, $(JDK_VERSION))
+    export JCK_VERSION=jck8c
+  else
+    export JCK_VERSION=jck$(JDK_VERSION)
   endif
 endif
 
@@ -68,7 +60,10 @@ endif
 OTHER_OPTS=
 # if JDK_IMPL is openj9 or ibm
 ifneq ($(filter openj9 ibm, $(JDK_IMPL)),)
- OTHER_OPTS=-Xtrace:maximal=all{level2}
+ OTHER_OPTS +=-Xtrace:maximal=all{level2} -Xfuture --enable-preview
+ ifeq ($(OS),OS/390)
+    OTHER_OPTS += -Dcom.ibm.tools.attach.enable=yes
+  endif
 endif
 
 SYSTEMTEST_RESROOT=$(TEST_RESROOT)/../../system
--- a/external/README.md
+++ b/external/README.md
@@ -5,7 +5,7 @@ Third Party container tests help verify that the AdoptOpenJDK binaries are *good
 ## Running External tests locally
 To run any AQA tests locally, you follow the same pattern:
 
-0. Ensure your test machine is set up with [test prereqs](https://github.com/eclipse/openj9/blob/master/test/docs/Prerequisites.md).  For external tests, you do need Docker installed.
+0. Ensure your test machine is set up with [test prereqs](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/doc/Prerequisites.md).  For external tests, you do need Docker installed.
 
 1. Download/unpack the SDK you want to your test machine
 1. `export TEST_JDK_HOME=</pathToWhereYouInstalledSDK>` 


llxia:
Thanks for working on this issue @ellschi 
- We need to move the code to the existing file https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/jck/jck.mk, not creating a new jck.mk.
- We cannot simply copy and paste the code form testJobTemplate to jck.mk. The syntax and logic are different.
- I think you can add options into `OTHER_OPTS` https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/jck/jck.mk#L63

ellschi:
Hi @llxia, thank you for for the review! 

- _"We need to move the code to the existing file https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/jck/jck.mk, not creating a new jck.mk."_ That sounds as if I made it too complicated. If I get you right I should just commit the changes and create a PR in the file you linked to.
- _"We cannot simply copy and paste the code form testJobTemplate to jck.mk. The syntax and logic are different.
I think you can add options into OTHER_OPTS https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/jck/jck.mk#L63"_
Thanks for pointing out where the changes should go and what I have to look out for. However I'm a bit confused: In the issue itself the example fist given was [this](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/testJobTemplate#L154-L165)

```
ARCH_OS_LIST.each { ARCH_OS ->
	JDK_VERSIONS.each { JDK_VERSION ->
		LEVELS.each { LEVEL ->
			GROUPS.each { GROUP ->
				def ACTUAL_TEST_JOB_NAME = TEST_JOB_NAME ? TEST_JOB_NAME : "Test_openjdk${JDK_VERSION}_${JDK_IMPL_SN}_${LEVEL}.${GROUP}_${ARCH_OS}${SUFFIX}"
				def ACTUAL_TARGET = TARGET ? TARGET : "${LEVEL}.${GROUP}"
				def ACTUAL_BUILD_LIST = BUILD_LIST ? BUILD_LIST : GROUP


				// for jck builds, we need to pass in EXTRA_OPTIONS
				def EXTRA_OPTIONS = "";
				if (GROUP == "jck" && JDK_IMPL == "openj9") {
					EXTRA_OPTIONS = "-Xfuture"
```

And then a comment of yours was pointed out, containing this:

As mentioned above, move the logic to mk, **we should check for JDK_IMPL and SPEC, then append the jvm option in EXTRA_OPTIONS in openjdk.mk and jck.mk**. The rest will be take care of by TKG. There is no playlist.xml change needed.

Similar to the following:

openjdk-tests/openjdk/openjdk.mk

Lines 120 to 121 in ed05e61

```
 else ifeq ($(JDK_IMPL), openj9) 
 	JVM_NATIVE_OPTIONS := -nativepath:"$(TESTIMAGE_PATH)$(D)openj9" 
```
So the location of the addition is clear, it is in "EXTRA-OPTIONS" or "OTHER-OPTS" as you pointed out. But you wrote about appending the JVM option which I don't see in the first example. I see that L120 in your example looks a lot like L164 in the other example and I see that where you pointed to (L63 in jck.mk)
```
OTHER_OPTS=
# if JDK_IMPL is openj9 or ibm
ifneq ($(filter openj9 ibm, $(JDK_IMPL)),)
 OTHER_OPTS=-Xtrace:maximal=all{level2}
endif
```
is the "same" place within the code as in the examples but I don't know where I can look up what I should change in the code so that it fits into jck.mk.

Thank you for reading through all this!
Looking forward to your feedback!

Mesbah-Alam:
> If I get you right I should just commit the changes and create a PR in the file you linked to. 

hi @ellschi , 

What Lan had meant is that, the change should be made in [jck.mk](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/jck/jck.mk) file itself. No other file needs to be touched or created anew.

llxia:
Since we have `OTHER_OPTS` in jck.mk and it is used in `JCK_CMD_TEMPLATE ` for all tests, we can just add the logic in there instead of using `EXTRA_OPTIONS `.

```
OTHER_OPTS=
# if JDK_IMPL is openj9 or ibm
ifneq ($(filter openj9 ibm, $(JDK_IMPL)),)
 OTHER_OPTS="-Xtrace:maximal=all{level2} -Xfuture"
 ifeq ($(OS),OS/390)
     OTHER_OPTS +="  -Dcom.ibm.tools.attach.enable=yes"
  endif
endif
```

Mesbah-Alam:
I believe this would work even better: 

```
OTHER_OPTS=$(JVM_OPTIONS)
# if JDK_IMPL is openj9 or ibm
ifneq ($(filter openj9 ibm, $(JDK_IMPL)),)
 OTHER_OPTS +=-Xtrace:maximal=all{level2} -Xfuture --enable-preview
 ifeq ($(OS),OS/390)
     OTHER_OPTS += -Dcom.ibm.tools.attach.enable=yes
  endif
endif
```

Tested it in internal/Grinder/14657. 

Since we are adding quotes ($(Q)) around the value of `OTHER_OPTS` when it's being used in `JCK_CMD_TEMPLATE`, we do not need to add any quotes around the OTHER_OPTS value while constructing it.

ellschi:
Hi @llxia and @Mesbah-Alam thanks again for your input and patience! I added the code that @Mesbah-Alam suggested and uploaded it. Which seems to be showing in this chat ... so I hope this is solved now.

llxia:
Re https://github.com/AdoptOpenJDK/openjdk-tests/pull/2474#issuecomment-815853332, we do not need line `OTHER_OPTS=$(JVM_OPTIONS)`. 

The JVM_OPTIONS is already added later in the code https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/jck/jck.mk#L72

llxia:
Also, the PR still shows a new jck.mk file is added. @ellschi could you fix it? Thanks

jck.mk
```diff
@@ -0,0 +1,84 @@
...
+endif
+
+ifndef JCK_ROOT
+  export JCK_ROOT=$(TEST_ROOT)/../../../jck_root/JCK$(JDK_VERSION)-unzipped
+endif
+
+OTHER_OPTS=
+# if JDK_IMPL is openj9 or ibm
+ifneq ($(filter openj9 ibm, $(JDK_IMPL)),)
+ OTHER_OPTS +=-Xtrace:maximal=all{level2} -Xfuture --enable-preview
```
llxia, line 63: `--enable-preview` should only be added if JDK_VERSION !=8. Please check the original logic
https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/testJobTemplate#L167

llxia (changes requested):
Please address the above concerns. Thanks

Mesbah-Alam:
I had tried the change https://github.com/AdoptOpenJDK/openjdk-tests/pull/2474#issuecomment-815813170 and it failed with malformed quotes, as follows: 

""-Xtrace:maximal=all{level2} -Xfuture --enable-preview"  "

Please see internal/Grinder/14655 output for details. 

> we do not need line OTHER_OPTS=$(JVM_OPTIONS).

The JVM_OPTIONS is already added later in the code https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/jck/jck.mk#L72 

Ah, I forgot to mention, In the logic that worked for me, I had moved $(JVM_OPTIONS) out of JCK_CMD_TEMPLATE definition and placed it as an initial value for OTHER_OPTS. That looked like the simpler thing to do. 

And yes, a check for JDK_VERSION !=8 needs to be added for `--enable-preview`.

jck.mk
ellschi: Hi! I created #2486 - please review.

ellschi:
Hi @llxia and @Mesbah-Alam - thank you for reviewing! As you might have noticed I am not able to really understand what I'm doing here. I don't find the resources informative enough to independently create the code needed. And the fact that I created new files was due to trying to follow the guidelines (the way that I understood them). So my impression is that I lack some knowledge that I would need to adress this issue. And I don't know where to research it. So that's a little frustrating - I'm guessing for you, too. Hope we can close this issue soon! :)

Mesbah-Alam:
@ellschi , We appreciate your effort so far! 

Just wondering, did you start off this work by creating a fork and clone of https://github.com/AdoptOpenJDK/openjdk-tests? 

I've just updated description of https://github.com/AdoptOpenJDK/openjdk-tests/issues/1754 and included a list of "step by step" steps there that you might want to follow those steps. I hope then our comments above would make more sense to you.

ellschi:
Hi @Mesbah-Alam! 

> @ellschi , We appreciate your effort so far!

Thank you! I wasn't doubting that but I felt that I didn't get to the point where I understood what I was doing to solve this issue. So I was second guessing if I was the right fit for it. :) With other issues that I worked on I didn't have that problem. I just looked up some things here and there and then found out how to solve them.

> Just wondering, did you start off this work by creating a fork and clone of https://github.com/AdoptOpenJDK/openjdk-tests?

I did - that's what led to creating a new jck.mk every time. But as that is not what was intended I seem to have some misgivings about the steps. I'm going to look into them again over the weekend.

So have a good weekend and - again and also to @llxia - thank you for the support!

llxia:
>I had moved $(JVM_OPTIONS) out of JCK_CMD_TEMPLATE definition and placed it as an initial value for OTHER_OPTS. 

$(JVM_OPTIONS) has to be added at last as it contains `EXTRA_OPTIONS`. `EXTRA_OPTIONS ` needs to be at the end so the user can override jvm options if needed. If we want to move $(JVM_OPTIONS) out of JCK_CMD_TEMPLATE, please make sure it is added at the end of all other jvm opts.

PR #2476, opened 2021-04-07
Remove spec

jwnmoon:
Hello, 
I removed the spec references from the tests and functions. 
Please let me know if I should fix anything!

### DIFF
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -221,7 +221,7 @@
                       <impl>openj9</impl>
               </disabled>
 		<variations>
-              <variation>Mode150</variation>
+                      <variation>Mode150</variation>
 			<variation>Mode650</variation>
 			<variation>Mode1000</variation>
 		</variations>
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -65,11 +65,11 @@
 		</impls>
 	</test>
 	<test>
-                <testCaseName>jvm_native_sanity</testCaseName>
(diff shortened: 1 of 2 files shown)


llxia:
Please resolve the conflicts. Thanks

openjdk/playlist.xml
```diff
@@ -1092,6 +1092,7 @@
                       <comment>https://github.com/AdoptOpenJDK/openjdk-tests/issues/1071</comment>
                       <impl>openj9</impl>
               </disabled>
+=======
```
llxia, line 1095: Please remove this change.

llxia (changes requested):

llxia:
All the SPEC reference in Jenkinsfilebase needs to be removed. For example, https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L42. I think can use PLATFORM instead of SPEC for checks.

jwnmoon:
@llxia When there is a SPEC reference within a function like here https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L582, is it correct to delete the entire if and else if statements?

llxia:
Re https://github.com/AdoptOpenJDK/openjdk-tests/pull/2476#issuecomment-815301571, No, we cannot delete any if blocks. We need to use PLATFORM instead of SPEC.

PR #2479, opened 2021-04-08
commit

benny-coy:
external/Payara-mp-tck;
added  README.md and test.properties.


benny-coy:
@ShelleyLambert 
@karianna

PR #2489, opened 2021-04-10
Have one disable tag for each version

Mesbah-Alam:
1) TKG requires that we have a single version tag under each disable tag - that was fixed 
2) The awt test actually should not be disabled completely on 8. Specific disabling of sub-tests were delivered for awt suite in jck8.jtx in separate pr. 


smlambert (approved):

smlambert:
merging as Admin to unblock a pipeline

Mesbah-Alam:
Thanks a lot for the prompt review, @ShelleyLambert

PR #2526, opened 2021-04-20
Tweetest prep work - Add external_custom target

codeupall:


smlambert (commented):
hi @codeupall - please align with those changes in https://github.com/smlambert/openjdk-tests/tree/external_custom branch.  Where 1) the directory and test target being added is called external_custom 2) There is no need for a Dockerfile 3) the test script is called simply test.sh 4) updating common_functions.sh like https://github.com/smlambert/openjdk-tests/blob/external_custom/external/common_functions.sh#L222-L235

Once those changes are done, we will make some additional changes to create variables for test repo and command to run as described by https://github.com/AdoptOpenJDK/openjdk-tests/issues/2076.

codeupall:
I have made the update.

external/common_functions.sh
```diff
@@ -14,7 +14,7 @@
 #
 
 # Supported versions
-supported_versions="8 9 10 11 12 13 14 15 16 17"
+supported_versions="8 9 10 11 12 13 14 15"
```
smlambert, line 17: You can change this line back to include the 16 and 17 (my branch was older and we want to keep this change).

external/common_functions.sh
```diff
@@ -219,6 +219,21 @@ function set_test_info() {
         ubi_packages="git wget unzip"
         ubi_minimal_packages="${ubi_packages}"
         ;;
+    external_custom)
+        github_url="https://github.com/apache/flink.git"
```
smlambert, line 223: Instead of "https://github.com/apache/flink.git" we will eventually use a variable so that we can pass in various URLs to application repositories

external/external_custom/test.sh
```diff
@@ -0,0 +1,53 @@
...
+	export JAVA_BIN="$java_root"
+	$JAVA_BIN/java -version
+	export JAVA_HOME="${java_root%/bin}"
+fi
+
+echo_setup
+
+# export MAVEN_OPTS="-Xmx1g"
+
+cd /flink
```
smlambert, line 43: This will also be a variable (based on the github_url passed in elsewhere

external/external_custom/test.sh
```diff
@@ -0,0 +1,53 @@
...
+pwd
+echo "Compile and run external_custom tests"
+mvn clean install
```
smlambert, line 46: Not sure if we need to, but we may also make the command that needs to be run a variable (though for now we will leave as mvn clean install).

smlambert (commented):
hi @codeupall - you can remove the first-test directory and contents as we will not need it.  

Before you address the other changes, we can run the external_custom test locally and see that it works as expected.

external/common_functions.sh
codeupall: Done

external/common_functions.sh
codeupall: Am waiting for the variable on this.

codeupall: where should I save the vaeriable?
I need more light on this 
Thanks

smlambert: If you have tried running this locally, you will have seen the 'flow' that is used.  

`make compile` invokes the build.xml ant script 
`make _external_custom` calls external.sh (which subsequently invokes the other shell scripts including common_functions.sh)

as a trial, you could see if you could set a new environment variable called CUSTOM_GITHUB_URL set with the value of "https://github.com/apache/flink.git" and use that environment variable in the common_functions.sh script.  

Alternatively, you can try passing in a variable to the script by the same name, but for that, you will have to understand the flow of the commands and scripts and understand what pieces call what other pieces.

Nonye55: The update is done.
Waiting for review.

SAY-droid427:
@smlambert This is the PR related to external_test

smlambert:
Will close this in favour of more progressed work in https://github.com/adoptium/aqa-tests/pull/2660 (thanks for this initial effort though!).

Issue #2511, opened 2021-04-14
Prototype a 'rerun' script 


smlambert:
User story:
As a developer, I want to recreate a test failure that was reported 2 months ago.  I know which JDK binary was being tested, and I have downloaded and installed that binary locally on my machine.  I want to set up the same set of test material that was used 2 months ago.  I have the SHA.txt file from that test run from 2 months ago.  How do I setup the correct test material so that I can rerun the test that failed locally?

-----------------

Related to the [reproducible tests epic](https://github.com/AdoptOpenJDK/openjdk-tests/issues/2347), it would be good to write a script that can read and parse a 'rerun in Grinder link' and/or the contents of a SHA.txt file, set the appropriate env vars based on those 2 types of inputs, and then 'rerun' the test target (defined in the 'rerun in Grinder link').

A good starting point for this work, is to first try to run a test target locally (instructions for how to do that can be found [here](https://github.com/eclipse/openj9/wiki/Reproducing-Test-Failures-Locally#general-steps)).  Once can run locally, try changing some variables and see what happens when you next compile and run.

Example: 
- install your test JDK and export TEST_JDK_HOME and try compiling and running different test material and running different TARGETs
- export BUILD_LIST=openjdk, make compile and then make _jdk_math
- then change and export BUILD_LIST=system, make compile and make _MathLoadTest_all_5m

rerun.sh --help

rerun.sh --link someRerunURL [--shas someSHAtxtFile]

example:
rerun.sh --link `https://ci.adoptopenjdk.net/job/Grinder/parambuild/?SDK_RESOURCE=nightly&TARGET=jdk_imageio_0&TEST_FLAG=&UPSTREAM_TEST_JOB_NAME=&DOCKER_REQUIRED=false&ACTIVE_NODE_TIMEOUT=&VENDOR_TEST_DIRS=&EXTRA_DOCKER_ARGS=-v+%24%7BTEST_JDK_HOME%7D%3A%2Fopt%2Fjava%2Fopenjdk&TKG_OWNER_BRANCH=AdoptOpenJDK%3Amaster&TEST_PARALLELIZATION_PARAMS=&PLATFORM=x86-64_linux` --shas /tmp/SHA.txt (edited) 

where a SHA.txt file is an optional input to the rerun.sh script but if present, would set some environment variables to specify which specific test repos we want to use

smlambert:
Some additional reference:
[...]

PR #2531, opened 2021-04-21
Added rerun.sh prorotype

brngylni:
I have made a script to be a prototype and give an idea what can be done. It is relevant with #2511 
We can discuss the problems of this script and how to solve them under here.   

Signed-off-by: Mehmet Geylani <barankara5@gmail.com>

### DIFF
--- a/rerun.sh
+++ b/rerun.sh
@@ -1,0 +1,139 @@
+#!/bin/bash
+
+if [[ ! $# -gt 0 ]] || [[ "$1" != "--link"  ]] || [[ "$3" != "--file" ]]
+then
+	echo "Usage: "
+	echo "rerun.sh --link rerunLink --file SHA_file"
+	exit 1
+fi
+
+link=$2
+sha=$4
+declare -A values
+state=false
+# This loop reads the file line by line and stores the content in an array.
+# Structure of array = url,sha=repo dir
+
+while read line;
+do
+	if [[ $line == *"repo dir:"* ]]
+	then
+		read -ra value <<< "$line"
+		dir="${value[2]}"
+		value=""
+	fi
+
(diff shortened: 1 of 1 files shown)


rerun.sh
```diff
@@ -0,0 +1,139 @@
...
+	elif [[ "${j[0]}" == *"TKG.git"* ]] 
+	then
+		echo "Fetching TKG..."
+		cd openjdk-tests	
+		echo "git clone ${j[0]} "
+		git clone "${j[0]}" 
+		cd "$(basename "${j[0]}" .git)" 
+		git checkout "${j[1]}"
+		cd ..
+		cd ..
```
llxia, line 121: We should call [get.sh](https://github.com/adoptium/aqa-tests/blob/master/get.sh) for git clone openj9 and TKG repo and move folders.

brngylni: But don't we need to checkout before moving folders?

llxia: The logic for git clone openj9 and TKG repo and moving folders are in get.sh. There is no need to duplicate the logic. We can just call get.sh and passing the required parameters.
https://github.com/adoptium/aqa-tests/blob/master/get.sh#L426
https://github.com/adoptium/aqa-tests/blob/master/get.sh#L465

The only exception is git clone https://github.com/adoptium/aqa-tests repo, which we do need to have.

brngylni: Check my last commit please.

llxia (approved):
A couple of enhancements are needed., but I am ok to deliver it as a prototype.

smlambert:
@brngylni - can you please switch this from Draft to PR now?  thanks!

smlambert (approved):
thanks @brngylni !

PR #2561, opened 2021-05-04
Exclude openjdk test java/io/File/GetXSpace.java

lumpfish:

### DIFF
--- a/openjdk/excludes/ProblemList_openjdk16-openj9.txt
+++ b/openjdk/excludes/ProblemList_openjdk16-openj9.txt
@@ -228,6 +228,7 @@ java/net/httpclient/websocket/PendingTextPongClose.java https://github.com/eclip
 # jdk_io
 
 java/io/CharArrayReader/OverflowInRead.java	https://github.com/AdoptOpenJDK/openjdk-tests/issues/1500	generic-all
+java/io/File/GetXSpace.java	https://github.com/AdoptOpenJDK/openjdk-tests/issues/2461 windows-all
 java/io/FileInputStream/UnreferencedFISClosesFd.java  https://github.com/eclipse/openj9/issues/3543 generic-all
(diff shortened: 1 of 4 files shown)


lumpfish:
Ref: https://github.com/AdoptOpenJDK/openjdk-tests/issues/2461

sophia-guo:
Might update the issue link to https://bugs.openjdk.java.net/browse/JDK-8251466

lumpfish:
I thought if I pointed to https://github.com/adoptium/aqa-tests/pull/2561 that would also include the investigation we had done at AdoptOpenJDK (as well as the reference to https://bugs.openjdk.java.net/browse/JDK-8251466).

sophia-guo:
https://github.com/adoptium/aqa-tests/pull/2561#issuecomment-832053009 yes, that also makes sense. Though would prefer the bug system one so we know at first sight that it's the upstream issue instead of adoptium.

smlambert:
Ultimately the link that is used in the comment of the exclude should be the core issue that needs to be fixed in order to reinclude the test and we should not keep other peripheral issues open or else we experience issue bloat where we live in a world where we have 2 open issues for 1 problem.  2561 served its purpose as a placeholder for excellent additional triage and now can be closed (if not being referenced by an exclude). 

On a related note: we need to go through this repo and 'fix' the bloat before we can turn on the automation to check if its safe to re-include tests based on whether the core issue is resolved or not.  I especially do not like when we hold open a test issue when it has become clear that it is an infra issue.  When we do this, it makes it hard to see overall trends and make program plans and decisions.

lumpfish:
Updated the link to https://bugs.openjdk.java.net/browse/JDK-8251466

smlambert (approved):

llxia (approved):
LGTM

PR #2568, opened 2021-05-06
Preparing SVTTestRepo to ebcdic

psoujany:
Converting SVTTestRepo to ebcdic to execute SVT internal tests on z/OS.

Added gitattributes to SVTTestRepo and modified system/common.xml which does conversion.

Verification Links
Linux :- runtimes-team-svt-jenkins/Test_grinder/job/Grinder/1246
z/OS:- runtimes-jenkins/Test_grinder/job/Grinder/15293

Fixes: runtimes/backlog/issues/517

### DIFF
--- a/system/common.xml
+++ b/system/common.xml
@@ -37,6 +37,25 @@
 		<isset property="env.STF_BRANCH"/>
 	</condition>
 	
+	<target name="checksvttestrepo">
+		<condition property="check_svttest_repo">
+			<contains
+				string="${env.VENDOR_TEST_REPOS}"
+				substring="SVTTestRepo" />
+		</condition>
+		<if>
+			<isset property="check_svttest_repo" />
+			<then>
+				<condition property="svttest_repo" value="${env.VENDOR_TEST_REPOS}" else="Please mention SVTTestRepo in VENDOR_TEST_REPOS field">
+					<isset property="env.VENDOR_TEST_REPOS"/>
+				</condition>
+				<condition property="svttest_repo_branch" value="${env.VENDOR_TEST_BRANCHES}" else="master">
+					<isset property="env.VENDOR_TEST_BRANCHES"/>
+				</condition>
+			</then>
+		</if>
+	</target>
+	
 	<target name="common_init">
 		<if>
 			<not>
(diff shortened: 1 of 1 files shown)


system/common.xml
```diff
@@ -37,6 +37,13 @@
 		<isset property="env.STF_BRANCH"/>
 	</condition>
 	
+	<condition property="svttest_repo" value="${env.VENDOR_TEST_REPOS}" else="Please mention SVTTestRepo in VENDOR_TEST_REPOS field">
+		<isset property="env.VENDOR_TEST_REPOS"/>
+	</condition>
+	<condition property="svttest_repo_branch" value="${env.VENDOR_TEST_BRANCHES}" else="master">
+		<isset property="env.VENDOR_TEST_BRANCHES"/>
+	</condition>
```
llxia, line 45: `VENDOR_TEST_REPOS` and `VENDOR_TEST_BRANCHES` could be multiple repos and branches. And they may not be system test repo. I do not think cannot simply set it this way.

system/common.xml
```diff
@@ -151,6 +158,47 @@
...
+			<mkdir dir="${SYSTEMTEST_ROOT}/SVTTestRepo/.git/info" />
+			<move file="${SYSTEMTEST_ROOT}/SVTTestRepo/.gitattributes.zos" tofile="${SYSTEMTEST_ROOT}/SVTTestRepo/.git/info/attributes" /> 
+			<exec executable="git" dir="${SYSTEMTEST_ROOT}/SVTTestRepo" failonerror="false">
+				<arg value="rm" />
+				<arg value="--cached" />
+				<arg value="-r" />
+				<arg value="-q" />
+				<arg value="." />
+			</exec>
+			<exec executable="git" dir="${SYSTEMTEST_ROOT}/SVTTestRepo" failonerror="false">
```
Mesbah-Alam, line 191: I see that you have basically done here what we did for the [JCK compilation story on z/OS](https://github.com/adoptium/aqa-tests/blob/master/jck/build.xml#L138).

The question is why? Do you see some of the SVT tests failing on z/OS due to conversion issues? Please note that we do not need to do these extra steps (remove and hard reset) on aqa system tests. 

Also, how long of a delay does this extra rm and reset steps introduce on the SVT repo compilation?

psoujany: Yes, SVT tests building is failing on z/OS as files were not converted. Hence I'm trying to convert SVTTestRepo on z/OS.

system/common.xml
psoujany: Agree, I'll find a way to get property set for only SVTTestRepo. Thank you.

system/common.xml
psoujany: Actually I tried to understand how openjdk-systemtest is getting converted and tried the same way for SVTTestRepo.

system/common.xml
```diff
@@ -177,6 +237,13 @@
 				<echo message="${SYSTEMTEST_ROOT}/openjdk-systemtest exists, skip cloning" />
 			</else>
 		</if>
+		<if>
+			<isset property="isZOS" />
+			<then>
+				<echo message="Clone SVTTestRepo freshly on z/OS" />
+				<antcall target="clone_svttestrepo" inheritall="true" />
```
Mesbah-Alam, line 244: This change would cause system/common.xml to clone and compile SVTTestRepo every time we are on z/OS, regardless of SDK version. But SVTTestRepo contains test materials that do not compile on OpenJ9, whereas system/common.xml is used commonly across all SDK versions and types.

system/common.xml
psoujany: We're running SVT tests on z/OS on Jenkins for the first time, so we are not sure what is the regular conversion. Could you please point us where regular conversion happens.

Mesbah-Alam: By regular conversion, I referred to the conversion performed by Git itself (https://git-scm.com/docs/gitattributes/2.19.2).

smlambert:
@Mesbah-Alam @psoujany - if I heard correctly you have arrived at a solution that is contained within the vendor repo?  if that is the case, can we close this PR?

psoujany:
> 
> 
> @Mesbah-Alam @psoujany - if I heard correctly you have arrived at a solution that is contained within the vendor repo? if that is the case, can we close this PR?

Yes, Shelley this can be closed. Thank you.

PR #2552, opened 2021-04-29
Exclude openjdk jdk/dynalink tests on openj9

lumpfish:
Signed-off-by: Simon Rushton <srushton@redhat.com>

PR #2573, opened 2021-05-10
Exclude dynalink subtests

lumpfish:


lumpfish:
PR https://github.com/adoptium/aqa-tests/pull/2552 excluded dynalink tests from running against openj9 because they use command line arguments not recognised by openj9.  Unfortunately it didn't actually exclude them.

The tests are defined with the `@test id=xxx` annotation, and it appears that the `xxx` needs to be included after the test source file for each individual subtest for the exclusion to take effect.

A jdk17 test runs with this amended PR: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/570/console

llxia (approved):
LGTM

PR #2589, opened 2021-05-25
Update bisheng JDK_IMPL to hotspot

eapenzhan:
Bisheng JDK is hotspot implementation.
This change can passed JDK_IMPL as "hotspot" to jenkins job when bisheng is tested, I think.

### DIFF
--- a/buildenv/jenkins/testJobTemplate
+++ b/buildenv/jenkins/testJobTemplate
@@ -19,7 +19,7 @@
  */
 
 def getBaseJDKImpl(JDK_IMPL) {
-	if (JDK_IMPL == "hotspot" || JDK_IMPL == "sap" || JDK_IMPL == "corretto" || JDK_IMPL == "dragonwell") {
+	if (JDK_IMPL == "hotspot" || JDK_IMPL == "sap" || JDK_IMPL == "corretto" || JDK_IMPL == "dragonwell" || JDK_IMPL == "bisheng") {
 		return "hotspot"
 	}
 	return JDK_IMPL


adamfarley (approved):
LGTM

sxa (approved):
At some point we need to resolve the TODO in https://github.com/adoptium/temurin-build/blob/921ec1893c2611d5b7865d2b83be418ec3c5831e/test/functional/buildAndPackage/src/net/adoptium/test/VendorPropertiesTest.java#L55 in order to perform the smoke test checks properly but this should resolve the immediate problem :-)

smlambert:
Based on the comment above, is this PR addressing a particular smoke test failure?  If so, will be good to reference it somewhere. 

If this PR is to resolve an immediate problem of smoke tests that should not target Bisheng, then possibly use vendor tag, 
re: https://github.com/adoptium/aqa-tests/pull/2589#pullrequestreview-667653517 - that is now resolved via https://github.com/adoptium/TKG/pull/142 where vendor support is added to playlists, where we use System.getProperty('java.vendor') as the value in the playlist if its not 'known' by a case in the case statement.

smlambert:
Reference to the Slack discussion to give context to this PR: https://adoptium.slack.com/archives/C09NW3L2J/p1621579653024100

adamfarley:
TLDR: This PR was to fix a problem where a smoke test job was run against a bisheng build, but would fail immediately because the detected JDK implementation (hotspot) didn't match the JDK_IMPL parameter (bisheng) passed from the build job.

**Details**

Smoke test error message:
```
19:49:18  /home/jenkins/workspace/build-scripts/jobs/jdk11u/jdk11u-linux-x64-bisheng_SmokeTests/openjdk-tests/TKG/../TKG/envSettings.mk:39: *** DETECTED_JDK_IMPL value is hotspot, settled JDK_IMPL value is bisheng. JDK_IMPL value does not match. Please reset or unset JDK_IMPL.  Stop.
```

Underlying cause: The test job template incorrectly identified bisheng as a discrete JDK implementation, as opposed to a variant on the hotspot implementation. In this context, the latter would be more appropriate.

Solution: The proposed "correct" behaviour was to follow the trend laid down by corretto, dragonwell, etc, by having the bisheng smoke tests' default value for JDK_IMPL set to "hotspot".

sxa:
@smlambert Are you happy with this change now that the context has been clarified?

smlambert (approved):

PR #2610, opened 2021-06-01
Hide all non Temurin tests from public view

gdams:
For now, I've chosen to allow everyone in [adoptium-aqavit-committers](https://github.com/orgs/adoptium/teams/adoptium-aqavit-committers/members) to see all non Temurin tests. We can always change this with different team memberships if required.


smlambert:
I do not believe this PR is required.  The testJobTemplate is for generating test jobs (not build jobs), so there is no need to restrict the view of non-Temurin test jobs.

gdams:
Closing for now then, thanks

PR #2611, opened 2021-06-01
Fix label for dynamic agent

llxia:


llxia:
With the original code, the test pipeline removes `ci.role.test`, but leaves `&&`. As a result, there is no machine matc. For example, label is `&&hw.arch.s390x&&sw.os.linux&&ci.agent.dynamic`.

image

internal/Test_openjdk8_j9_special.system_s390x_linux/63/console

sophia-guo (approved):

llxia:
Tested using the[ modified code](https://github.com/adoptium/aqa-tests/compare/master...llxia:test6?expand=1) to mimic all nodes are busy:
Grinder/15808
Grinder/15809

smlambert (approved):
LGTM

PR #2625, opened 2021-06-02
Set machine limit for dynamic vm to 100

llxia:
In parallelization, nodesByLabel(LABEL) for dynamic only returns the number of active vm agents.
This number may be less than the required number. For example, 1 active vm agent is
available and the parallel build requires 5 machines for parallelization. As a result,
we only have 1 machine for parallelization.

This PR sets the machine limit to 100 if LABEL contains dynamic. This number (100) limit
is per parallel parent build and it can be modified as needed in the future.

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -313,6 +313,7 @@ def setupParallelEnv() {
 def getNumMachines(){
 	int num = params.NUM_MACHINES ? params.NUM_MACHINES.toInteger() : 1
 	int limit = getMachineLimit()
+	echo "machine limit is ${limit}"
 	if (num > limit) {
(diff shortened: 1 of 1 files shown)


smlambert:
Haha, I am imagining based on this PR's existence at what might have occurred during testing...  more than 100 agents created?

llxia:
Not yet. Just want to set a large number in case we get there. If dynamic agents work well, maybe we can remove the machine limit completely in the future. :)

llxia:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/721

Internal:
Grinder/15834 //using dynamic agent

renfeiw (approved):
LGTM

smlambert (approved):

PR #2627, opened 2021-06-02
Update name to jenkins-helper in parentsJobPipeline

### LINKED ISSUE #2626: Update parentsJobPipeline to reference jenkins-helper
https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/parentsJobPipeline.groovy#L52 should be updated to reference the renamed jenkins-helper at https://github.com/adoptium/jenkins-helper

otherwise hitting this issue:
```
ERROR: No library named openjdk-jenkins-helper found
Finished: FAILURE
```

smlambert:
Fixes #2626


external/common_functions.sh
```diff
@@ -219,6 +219,21 @@ function set_test_info() {
...
+        environment_variable="MODE=\"java\""
+        debian_packages="git maven"
+        debianslim_packages="${debian_packages}"
+        ubuntu_packages="${debian_packages}"
+        alpine_packages="git maven"
+        centos_packages="git maven"
+        clefos_packages="${centos_packages}"
+        ubi_packages="git maven"
+        ubi_minimal_packages="${ubi_packages}"
+        ;;
```
llxia, line 236: Are these external test changes intended for this PR?

smlambert: crikey, no they are not!

llxia:
I think this PR will affect other Jenkins. We need to update the openj9 and internal Jenkins configuration as well. We should update to use the git URL: https://github.com/adoptium/jenkins-helper and the updated libaray name.

@AdamBrousseau could you please update openj9 Jenkins https://openj9-jenkins.osuosl.org once this PR is merged? I do not have permission. Also, did we update the internal mirror to point new git URL?

smlambert:
This is not urgent to merge, so we can coordinate, as I can point to my branch for the short-term for the configuration at the private server where I am hitting the issue.

sophia-guo:
This PR shouldn't affect other jenkins. The file is only used by a jenkins job to manually trigger external or jck tests. The url can be referenced automatically. Though eventually updating the url is needed.

sophia-guo (approved):

sophia-guo:
This is a standalone jenkinsfile. Not the one in https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L254

llxia:
ah, I see. Should we keep the library name to be the same?

llxia (approved):
LGTM

smlambert:
Both groovy scripts should eventually get updated to use the real repo name (which is changed to jenkins-helper), but there is no rush to do so necessarily.

PR #2653, opened 2021-06-17
Switching OpenJDK Test src to use adoptium mirrors

adamfarley:
Mirrors for new versions of OpenJDK (such as jdk17) are not being
created at AdoptOpenJDK anymore, so it seems the right time to
update the test build.xml so that grinders with no jdk repo specified
look to adoptium for the test code they need.

Two other changes include a typo fix, and changing the aarch64 repo to
aarch32 (arm). The latter is because the aarch64 support has been
folded into the normal openjdk src mirror repos, but aarch32 still has
its own, separate src mirror repo at adoptium.

### DIFF
--- a/openjdk/build.xml
+++ b/openjdk/build.xml
@@ -197,40 +197,40 @@
 					<else>
 						<if>
 							<and>
-								<contains string="${env.SPEC}" substring="aarch64" />
+								<contains string="${env.SPEC}" substring="arm" />
 								<equals arg1="${JDK_VERSION}" arg2="8" />
 							</and>
 							<then>
-								<property name="jdkName" value="openjdk-aarch64-jdk8u" />
+								<property name="jdkName" value="aarch32-jdk8u" />
 							</then>
 							<else>
 								<echo message="Using a git ls-remote command to determine if the jdk source repo has an available 'u' version yet." />
-								<echo message="git ls-remote -h https://github.com/AdoptOpenJDK/openjdk-jdk${JDK_VERSION}u.git" />
+								<echo message="git ls-remote -h https://github.com/adoptium/jdk${JDK_VERSION}u.git" />
 								<exec executable="git" resultproperty="repo.exist" failonerror="false" timeout="30000">
 									<env key="GIT_TERMINAL_PROMPT" value="0"/>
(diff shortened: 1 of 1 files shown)


adamfarley:
Aarch32 jdk8 test: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/941/console

Windows jdk17 test: https://ci.adoptopenjdk.net/job/Grinder/942/console

karianna (approved):
LGTM

openjdk/build.xml
```diff
@@ -197,40 +197,40 @@
 					<else>
 						<if>
 							<and>
-								<contains string="${env.SPEC}" substring="aarch64" />
+								<contains string="${env.SPEC}" substring="arm" />
 								<equals arg1="${JDK_VERSION}" arg2="8" />
 							</and>
 							<then>
-								<property name="jdkName" value="openjdk-aarch64-jdk8u" />
+								<property name="jdkName" value="aarch32-jdk8u" />
```
sophia-guo, line 204: I think https://github.com/AdoptOpenJDK/openjdk-aarch64-jdk8u and https://github.com/AdoptOpenJDK/openjdk-aarch32-jdk8u are different repos? Should https://github.com/AdoptOpenJDK/openjdk-aarch64-jdk8u be updated to https://github.com/adoptium/aarch64-jdk8u ( though I checked https://github.com/adoptium/aarch64-jdk8u wasn't setup)? Is this expected?

sophia-guo: Never mind. Just saw the comment in the commit https://github.com/adoptium/aqa-tests/pull/2653#issue-672456426

sophia-guo (approved):
LGTM

openjdk/build.xml
adamfarley: To your first question: Yup, those are different repos. 

I was told that aarch64-specific code had been folded into the main openjdk src repos, so I didn't see the point of specifically redirecting to the aarch64 variant repo anymore.

aarch32, however, does have a repo at adoptium (https://github.com/adoptium/aarch32-jdk8u), and I'm told it's still needed for aarch32 support.

These reasons are why I swapped the aarch64 redirect for an aarch32 redirect.

I hope that makes sense?

adamfarley:
If folks are happy with this, please merge when ready. Thank you. :)

adamfarley:
Thanks!

PR #2656, opened 2021-06-17
Exclude suspected jdk_management hang tests for openj9

sophia-guo:
Those tests causes hang problem, which has been excluded for hotspot. Should also  be excluded for
openj9.

Excluded those tests jdk_management only takes 1 minute, though still other test failures. Will keep jdk_managment disabled. 

https://ci.adoptopenjdk.net/view/work-in-progress/job/grinder_sandbox_new/87/


llxia (approved):
LGTM

smlambert (approved):

PR #2660, opened 2021-06-17
Added external_custom_target

SAY-droid427:
The following tasks have been performed:
- external_custom_target folder added to external directory
- function added to the common_functions to read the test.properties file


external/external_custom/dockerfile/test.sh
```diff
@@ -0,0 +1,53 @@
...
+	export JAVA_BIN="$java_root"
+	$JAVA_BIN/java -version
+	export JAVA_HOME="${java_root%/bin}"
+fi
+
+echo_setup
+
+# export MAVEN_OPTS="-Xmx1g"
+
+cd /flink
```
smlambert, line 43: this is specific to repo that was cloned, so should not hardcode to flink

external/external_custom/dockerfile/test.sh
```diff
@@ -0,0 +1,53 @@
...
+pwd
+echo "Compile and run external_custom tests"
+mvn clean install
```
smlambert, line 46: This is the EXTERNAL_TEST_CMD variable which should be passed in for cases where its not mvn clean install (which we can use as the default command, as its used most often by applications testing).

external/external_custom/dockerfile/test.sh
```diff
@@ -0,0 +1,53 @@
...
+test_exit_code=$?
+echo "Build external_custom completed"
+
+find ./ -type d -name 'surefire-reports' -exec cp -r "{}" /testResults \;
```
smlambert, line 50: we likely can not assume everything will go into 'surefire-reports' dir, but we can address this in a later PR.

external/common_functions.sh
```diff
@@ -219,6 +226,23 @@ function set_test_info() {
         ubi_packages="git wget unzip"
         ubi_minimal_packages="${ubi_packages}"
         ;;
+    external_custom)
```
smlambert, line 229: Does it complicate things to put this content into a test.properties file in the external_custom dir and read from there?  for now we can leave as is, but ultimately the set_test_info() function would be removed completely, and we would read_test_info from each test.properties file found test subdir.

external/common_functions.sh
```diff
@@ -219,6 +226,23 @@ function set_test_info() {
...
+        PROPERTY_FILE=external_custom/test.properties
+        GITHUB_URL=$(getProperty "github_url")
+        github_url="${EXTERNAL_CUSTOM_TARGET}"
+        script="test.sh"
+        test_results="testResults"
+        tag_version="master"
```
smlambert, line 235: tag_version="${EXT_REPO_BRANCH}"

external/common_functions.sh
```diff
@@ -219,6 +226,23 @@ function set_test_info() {
         ubi_packages="git wget unzip"
         ubi_minimal_packages="${ubi_packages}"
         ;;
+    external_custom)
+        PROPERTY_FILE=external_custom/test.properties
+        GITHUB_URL=$(getProperty "github_url")
+        github_url="${EXTERNAL_CUSTOM_TARGET}"
+        script="test.sh"
```
smlambert, line 233: test_command="${EXTERNAL_TEST_CMD}"

smlambert (commented):
Some notes about how to handle branch name and test command.

external/common_functions.sh
SAY-droid427: I have made the required changes for the EXT_REPO_BRANCH and successfully run the spring tests too which couldn't be run earlier due to branch name issues.
spring

external/external_custom/dockerfile/test.sh
SAY-droid427: For this  we have to add the properties in the test.properties of the individual tests and get the values from the test.properties of the test being run and use it in the common_functions.sh, according to my understanding. Is this correct?

external/external_custom/dockerfile/test.sh
SAY-droid427: We should use another variable for this and for getting the value for this can we set up a variable DIR in the test.properties file and read from it?

external/common_functions.sh
SAY-droid427: I will add a test.properties file in the external_custom dir and add all the contents there for now. After all the variables have been set up and working, the set_test_info() would be removed. I think this should work.

external/external_custom/dockerfile/test.sh
smlambert: this will always be the name of the external test repo that we cloned, so we can figure it out based on the value of that variable (which I think you are currently calling EXTERNAL_CUSTOM_TARGET but really that is a repo_url... probably be clearer to call that EXTERNAL_CUSTOM_REPO

so, for https://github.com/apache/camel-quarkus.git, cd /[camel-quarkus](https://github.com/adoptium/aqa-tests/blob/master/external/camel/dockerfile/camel-test.sh#L45)  and so forth

llxia: I think we can just reuse the cmd to get folder name from git URL
https://github.com/adoptium/aqa-tests/blob/14f5a8348e22e5616dea2ea001ab8d693cddfeee/external/dockerfile_functions.sh#L450

external/external_custom/dockerfile/test.sh
smlambert: yes, that is what I was thinking we do, for tests in category 1 (unsuitable for external_custom), you do not define test_command in test.properties (because there are many commands), for those, we need to keep their Dockerfile/xx-test.sh file which contains all the special set up, for tests in category 2 (using mvn clean install) or category 3 (some other simple one line command) those can be added to test_command in test.properties file

for the contents of external_custom/test.properties, then test_command="${EXT_TEST_CMD}" if EXT_TEST_CMD is defined, if not, we use the default value 'mvn clean install'

external/external_custom/dockerfile/test.sh
SAY-droid427: @smlambert Yes, I will change the name of the variable @lxia Yes we should get the command from the git URL. I will make the changes. Thanks!

SAY-droid427:
@smlambert I have added the test_command property to the  test.properties file of eligible tests. I have also changed the name of the EXTERNAL_CUSTOM_TARGET to EXTERNAL_CUSTOM_REPO.
As discussed in the meet I have to add an external directory where the repo has to be cloned and then we can cd into that EXTERNAL_DIR(or any name) directory. But I couldn't get in which file I have to make the change. If you could please point that out a bit. Thanks!

SAY-droid427:
@smlambert @llxia I have made all the changes as discussed in the meeting. Please review the changes and let me know of any other changes needed.

external/dockerfile_functions.sh
```diff
@@ -451,7 +451,7 @@ print_clone_project() {
 
     echo -e "# Clone ${test} source" \
             "\nENV ${test_tag}=\$${test_tag}" \
-            "\nRUN git clone ${github_url}" \
+            "\nRUN git clone ${github_url} external_test_repo" \
```
llxia, line 454: I believe this is a common script. Changing the folder name here means we change the behavior for all external tests - all external tests check out into `external_test_repo` folder. Therefore, all external tests need to cd into `external_test_repo` in their shell script, not just external_custom test. 

On the second thought, it would be easier to figure out the repo dir for external_custom. Can you get EXTERNAL_CUSTOM_REPO in test.sh? If so, just use cmd to figure out the folder (similar to 450).

SAY-droid427: Since we have set the EXTERNAL_CUSTOM_TARGET as an environment variable, I think we can get the EXTERNAL_CUSTOM_REPO from the EXTERNAL_CUSTOM_TARGET using a function in test.sh itself.

SAY-droid427: @lxia I have changed the logic of the external_test_repo as you told but one thing I noticed is that while moving into the jacoco folder, we have to cd into the following:[ https://github.com/adoptium/aqa-tests/blob/b766eb2c81252e8beefa3dcdf7e8c99c99210b68/external/jacoco/dockerfile/jacoco-test.sh#L41](https://github.com/adoptium/aqa-tests/blob/b766eb2c81252e8beefa3dcdf7e8c99c99210b68/external/jacoco/dockerfile/jacoco-test.sh#L41). So is it necessary to cd into the org.jacoco.build subfolder?

smlambert: interesting, so jacoco might be an exception case (not as suitable to use this external_custom approach as others, because it doesn't conform to the ideal scenario... where ideal is clone repository, cd into repo dir, run a single command to execute tests)

you can check if its necessary by seeing what happens if you cd into repo name (and not subfolder), are you still able to run tests or does it fail to work.

SAY-droid427: @smlambert So I run the tests and it appears that we don't really need to cd into the org.jacoco.build folder again as the tests run successfully.
sfd

smlambert: very good, thanks for confirming!

SAY-droid427:
@smlambert @lxia Could you please let me know what should I do next now that the changes have been made? Thanks!

smlambert (approved):
Looks like we are at the point where we can and should merge this.  Then we can use the other branch from https://github.com/adoptium/aqa-tests/pull/2687 to test with in Grinder_Tweetest job.  Thanks @SAY-droid427 !

external/dockerfile_functions.sh
SAY-droid427: Sorry for the misinformation i provided earlier! Although the make compile target can be run successfully, but the make _jacoco_test cannot be run without cd ing into the org.jacoco folder. So I think this cannot be included in the external_custom target tests.

smlambert:
@llxia - anything else you'd like to see here before we merge?  (we expect other incremental improvements to follow also, but merging this enables Grinder_Tweetest testing of https://github.com/adoptium/aqa-tests/pull/2687.

external/external_custom/dockerfile/test.sh
```diff
@@ -0,0 +1,50 @@
...
+	export JAVA_BIN="$java_root"
+	$JAVA_BIN/java -version
+	export JAVA_HOME="${java_root%/bin}"
+fi
+
+echo_setup
+
+external_test_repo="$(echo ${EXTERNAL_CUSTOM_REPO} | awk -F'/' '{print $NF}' | sed 's/.git//g')"
+
+cd /external_test_repo
```
llxia, line 43: Does this work without `$`?

SAY-droid427: Yes the `make compile` works fine but `make _external_custom` does not work both with and without $. The whole output:
output1
output2
output3
output4
output7
output8
output9

llxia: The `test.sh` will run in the actual test execution (make _external_custom), not `make compile`. 
I will approve this PR, so we can run this in Grinder. Instead of debugging in your local, I think it will be easier to debug in Grinder.

llxia (approved):

external/external_custom/dockerfile/test.sh
SAY-droid427: Thanks @llxia ! Could you please guide me a bit on how to run this in Grinder?

llxia: By Grinder, I mean https://ci.adoptopenjdk.net/job/Grinder_Tweetest/. Once #2687 is merged, we can run tests on Jenkins.

Issue #2686, opened 2021-06-29
Test new Tweetest Parameters with a fresh set of application repositories


SAY-droid427:

smlambert:
As a starting point at #172 (we should go through others in the list to see if they are suitable for external_custom, but starting at zookeeper)
- doc: https://cwiki.apache.org/confluence/display/ZOOKEEPER/HowToContribute
  - repo: https://github.com/apache/zookeeper 
  - branch: git checkout -b ZOOKEEPER-1234
  - command: ant test

SAY-droid427:
@smlambert Thank you for pointing out the test! Should I check whether the tests can be run using the changes in https://github.com/adoptium/aqa-tests/pull/2660?

smlambert:
Yes, and in fact what I was hoping is that we can get #2660 merged, so then we can run testing in Grinder_Tweetest, to save you from having to run locally.  However, given many committers are out until Monday (so likely won't be merged until then), if you feel inspired please do so.

We should also review the remaining tests listed in #172 to see if others are appropriate candidates for external_custom and could be added to this issues list for 'testing a fresh set'.

SAY-droid427:
> Yes, and in fact what I was hoping is that we can get #2660 merged, so then we can run testing in Grinder_Tweetest, to save you from having to run locally. However, given many committers are out until Monday (so likely won't be merged until then), if you feel inspired please do so.
> 
> We should also review the remaining tests listed in #172 to see if others (in the list that are 'unchecked') are appropriate candidates for external_custom and could be added to this issues list for 'testing a fresh set'.

Thank you so much @smlambert ! I will try to run the test above and all those listed in the related issue list to check if they can be added to our eligible test lists :)

SAY-droid427:
TEST RESULTS FOR ZOOKEEPER with the following parameters set:
EXTERNAL_CUSTOM_REPO:  https://github.com/apache/zookeeper
EXTERNAL_REPO_BRANCH: master
EXTERNAL_TEST_CMD: ant test

Results for make compile:
![make_compile_zookeeper](https://user-images.githubusercontent.com/59281636/124414514-8a390380-dd70-11eb-9010-dd36081fa255.png)

Results for make external_custom:
![make_external_zoo](https://user-images.githubusercontent.com/59281636/124414564-9c1aa680-dd70-11eb-97bc-5dc554398190.png)

SAY-droid427:
[...]

PR #2687, opened 2021-06-29
Update Jenkinsfilebase with new parameters for Tweetest

SAY-droid427:
Jenkinsfilebase has been updated to read the EXTERNAL_CUSTOM_REPO,  EXT_TEST_CMD, EXT_REPO_BRANCH and set environment variables.

Fixes https://github.com/adoptium/aqa-tests/issues/2685


SAY-droid427:
@smlambert According to my understanding of the issue, I have created the PR. If any other changes need to be done I will change it immediately. Thanks!

buildenv/jenkins/JenkinsfileBase
```diff
@@ -79,6 +79,9 @@ def setupEnv() {
 	env.KEEP_REPORTDIR = params.KEEP_REPORTDIR ? params.KEEP_REPORTDIR : false
 	SDK_RESOURCE = params.SDK_RESOURCE ? params.SDK_RESOURCE : "upstream"
 	env.AUTO_DETECT = params.AUTO_DETECT
+	env.EXTERNAL_CUSTOM_REPO=params.EXTERNAL_CUSTOM_REPO ? params.EXTERNAL_CUSTOM_REPO : ""
+	env.EXT_REPO_BRANCH=params.EXT_REPO_BRANCH ? params.EXT_REPO_BRANCH : "master"
+	env.EXT_TEST_CMD=params.EXT_TEST_CMD ? params.EXT_TEST_CMD : "mvn clean install"
```
llxia, line 84: Could we have a common naming convention? 
EXTERNAL_TEST_REPO
EXTERNAL_TEST_BRNACH
EXTERNAL_TEST_CMD

SAY-droid427: @lxia Sure. I will change all the variable names to the above both in this file and the external_custom target.

SAY-droid427: @smlambert I think we need to  change the variable names in [https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Tweetest/](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Tweetest/)?

smlambert: Updated Grinder_Tweetest with new names now

llxia: Thanks @smlambert

llxia:
Thanks @SAY-droid427 
Please make sure https://github.com/adoptium/aqa-tests/pull/2660 is updated accordingly. :)

llxia (approved):
LGTM

SAY-droid427:
> Thanks @SAY-droid427
> Please make sure #2660 is updated accordingly. :)

I have made the updates.

smlambert:
FYI: 
https://ci.adoptopenjdk.net/job/Grinder_Tweetest/5/ - the Grinder_Tweetest job config is updated to point at this PR branch, so we can run some testing.

llxia:
@SAY-droid427 the Grinder_Tweetest is pointing to the repo and branch in this PR. In order for it to work properly, please rebase this PR to include the changes in #2660. Otherwise, there is no external_custom to run.

SAY-droid427:
@lxia I have fetched and merged the changes from this repo to my branch. The changes in https://github.com/adoptium/aqa-tests/pull/2660 are included now I think.

SAY-droid427:
@lxia I have run the Grinder just now and got this output:
[https://ci.adoptopenjdk.net/job/Grinder_Tweetest/6/](https://ci.adoptopenjdk.net/job/Grinder_Tweetest/6/)

smlambert:
```
11:18:39  =JAVA VERSION OUTPUT END=
11:18:39  /test.sh: line 43: cd: /external_test_repo: No such file or directory
```
That external_test_repo should be a variable that gets interpreted as zookeeper in your run.

SAY-droid427:
Corrected the errors

SAY-droid427:
Latest run after usng set -e like in kafka-tests:
https://ci.adoptopenjdk.net/job/Grinder_Tweetest/8/console

smlambert:
Based on the output of that test, the ${external_test_repo} variable from [test.sh_L43](https://github.com/SAY-droid427/openjdk-tests/blob/new-param-jenkins/external/external_custom/dockerfile/test.sh#L43-L49) is empty.

It may be useful to add echo statements for all 3 variables in test.sh script to see what they are set to so we can look at the [console output](https://ci.adoptopenjdk.net/job/Grinder_Tweetest/8/console) of Grinder_Tweetest job and learn if things are being passed through correctly.

```
22:16:22  /
22:16:22  Compile and run external tests
22:16:22  Build external_custom completed
22:16:23  
22:16:23  external_custom_0_PASSED
```

SAY-droid427:
The output:
```
08:13:20  =JAVA VERSION OUTPUT END=
08:13:20  
08:13:20  
08:13:20  
08:13:20  /
08:13:20  Compile and run external tests
```
The variables are not getting printed. So I think the variables are not getting passed correctly.

SAY-droid427:
The current output:
[https://ci.adoptopenjdk.net/job/Grinder_Tweetest/11/console](https://ci.adoptopenjdk.net/job/Grinder_Tweetest/11/console)

SAY-droid427:
I think the derby which is hard-coded and we were unable to run the make _external_custom is in this line:
https://github.com/adoptium/aqa-tests/blob/c35812bbdf0a1b59c662c618910dcc0b7b75bfb6/external/external.sh#L26

smlambert:
re: https://github.com/adoptium/aqa-tests/pull/2687#issuecomment-877238522, yes, my guess is that derby is set to the default in case something is left blank.  And since external_custom remains blank, its getting set to derby.

SAY-droid427:
To make sure that the variables get passed to external.sh should I create a function separately in common_functions.sh(from where I believe it gets called) to pass the values to it?

llxia:
> The variables are not printed in the test.sh.

What output are you expecting? In [test.sh](https://github.com/adoptium/aqa-tests/blob/master/external/external_custom/dockerfile/test.sh), I do not see echo of EXTERNAL_CUSTOM_REPO, EXTERNAL_REPO_BRANCH or EXTERNAL_TEST_CMD

SAY-droid427:
@llxia I was trying to investigate till where the variables  are getting passed correctly and so in this PR I have tried to echo the variables at different points. I found that the variables are not getting passed to the external.sh and that is why it is using derby as default and make _external_custom is failing.

SAY-droid427 (commented):
Here are the changes I had made for debugging

smlambert:
Yes, that looks like a correct assessment, and based on https://ci.adoptopenjdk.net/job/Grinder_Tweetest/11/console during the external.sh --run phase, the APPLICATION_NAME says external_custom, but I would expect it to be zookeeper in that run.  Also for the purpose of testing, useful to set the EXTERNAL_CUSTOM_BRANCH to ZOOKEEPER-1234 (as noted in https://github.com/adoptium/aqa-tests/issues/2686#issuecomment-872676219) for extra clarity since master is the default setting so it will be harder to differentiate.

```
09:45:46  APPLICATION_NAME=external_custom
09:45:46  APPLICATION_TAG=master
```

We need the APPLICATION_NAME=zookeeper (it gets printed out [here](https://github.com/SAY-droid427/openjdk-tests/blob/new-param-jenkins/external/dockerfile_functions.sh#L431-L439) as part of[ generate_dockerfile()](https://github.com/SAY-droid427/openjdk-tests/blob/new-param-jenkins/external/dockerfile_functions.sh#L493)).  This implies that we may need to do some extra handling for the Dockerfile generation for external_custom inside of generate_dockerfile.  

Pseudo-code:
if ${test} == external_custom
    ${test} = the repo name stripped off the end of the  EXTERNAL_CUSTOM_REPO variable)
    ${tag_version} = EXTERNAL_CUSTOM_BRANCH

--------------

Adding extra printouts/echos is a great start to understanding what is happening each step of the way.  Also helpful to summarize the steps, to review what happens when building and running this test target (called external_custom), here is a summary:

**Build the right docker image:**
params set via Jenkinsfilebase :arrow_right: 
TKG / make compile :arrow_right: 
build.xml :arrow_right: 
external.sh --build :arrow_right: 
Dockerfile is generated :arrow_right:  
Docker image is built using generated Dockerfile 

(**TODO**: review generate_dockerfile() and see what updates are needed there, look at a generated Dockerfile on local runs to see if it looks right)

**Run the image with the right variables:**
TKG / make _external_custom :arrow_right: 
external.sh --run :arrow_right: 
docker run :arrow_right: 
test.sh inside Docker image is invoked

SAY-droid427:
@smlambert On making the changes in the generate_dockerfile() in dockerfile_functions.sh, I get the error :
error

coming from [here](https://github.com/adoptium/aqa-tests/blob/d6e362aa5c43bdcff4866d69eec82d7b54284e77/external/common_functions.sh#L539) in the common_functions.sh because it is trying to match the test with zookeeper which it does not get and fails.

smlambert:
Is https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Tweetest/12/console a test of the latest changes?

```
02:33:30       [exec] Step 12/22 : COPY zookeeper/dockerfile/test.sh /test.sh
02:33:30       [exec] COPY failed: file not found in build context or excluded by .dockerignore: stat zookeeper/dockerfile/test.sh: file does not exist
02:33:30  
02:33:30  BUILD FAILED
```

sophia-guo:
I'm confused and looks like the changes https://github.com/adoptium/aqa-tests/pull/2687/commits/25f04c3f554fa1a6574c5bb88b4d661237772699  won't fix the issue

https://ci.adoptopenjdk.net/job/Grinder_Tweetest/14/console

```
16:57:51       [exec]  ---> b69b7919ef4a
16:57:51       [exec] Step 12/22 : COPY zookeeper/dockerfile/test.sh /test.sh
16:57:51       [exec] COPY failed: file not found in build context or excluded by .dockerignore: stat zookeeper/dockerfile/test.sh: file does not exist
```

sophia-guo:
The current framework suppose each test has its own specific directory and the directory is also the test name, which won't work for external_custom any more.  If this makes the code confuse and have some extra hacks we may probably just add one more variable for directory name. Just a quick look.

smlambert:
Directory name can just be the application name, but the generic (good for all external_custom applications) test.sh file will need to be copied from the external_custom folder to somewhere where it can be found and run.

SAY-droid427:
@smlambert @sophia-guo I haven't pushed the latest changes. I was waiting for approval to include the changes
```
for external_custom_test in ${supported_tests}
    do
        if [[ "${test}" == "${external_custom_test}" ]]; then
             echo -e "# This is the main script to run ${test} tests" \
            "\nCOPY external_custom/dockerfile/${script} /${script}" \
            "\nCOPY test_base_functions.sh test_base_functions.sh\n" >> ${file}
        fi
    done
```
Adding these changes makes the make compile run successfully. The changes I have pushed till now to make APPLICATION_NAME=zookeeper leads to BUILD failure.

SAY-droid427:
The result after the latest changes:
https://ci.adoptopenjdk.net/job/Grinder_Tweetest/16/console

 The error in the make _external_custom is:
```
11:50:22  Unable to find image 'adoptopenjdk-external_custom-test:8-jdk-ubuntu-hotspot-full' locally
11:50:24  docker: Error response from daemon: pull access denied for adoptopenjdk-external_custom-test, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
11:50:24  See 'docker run --help'.
11:50:24  
11:50:24  external_custom_0_FAILED
```
Before these changes it was getting hardcoded to adoptopenjdk-derby-test:8-jdk-ubuntu-hotspot-full. I am trying to debug this bit too.

SAY-droid427:
The Dockerfile generated in jvmtest in my local machine looks fine to me:
zookeeper

SAY-droid427:
UPDATE: The external.sh now has access to a value 'external_custom' in the test variable. Earlier, nothing was getting passed to it. But now the problem is that the dockerimage created is adoptopenjdk-zookeeper-test:8-jdk-ubuntu-hotspot-full but during make _external_custom it is trying to access the adoptopenjdk-external_custom-test:8-jdk-ubuntu-hotspot-full. So I think we have to try and make the test passed to external.sh to be zookeeper instead of external_custom.
```
14:39:48  Unable to find image 'adoptopenjdk-external_custom-test:8-jdk-ubuntu-hotspot-full' locally
14:39:50  docker: Error response from daemon: pull access denied for adoptopenjdk-external_custom-test, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
14:39:50  See 'docker run --help'.
14:39:50  
14:39:50  external_custom_0_FAILED
14:39:50  
14:39:50  test in external.sh points to external_custom 
14:39:50  Error: No such image: adoptopenjdk-external_custom-test:8-jdk-ubuntu-hotspot-full
```

smlambert:
'docker build' step creates an image called
```
02:20:08       [exec] Successfully tagged adoptopenjdk-zookeeper-test:8-jdk-ubuntu-hotspot-full
02:20:08  
```
'docker run' step attempts to run an image called adoptopenjdk-external_custom-test:8-jdk-ubuntu-hotspot-full

SAY-droid427:
> 'docker build' step creates an image called
> 
> ```
> 02:20:08       [exec] Successfully tagged adoptopenjdk-zookeeper-test:8-jdk-ubuntu-hotspot-full
> 02:20:08  
> ```
> 
> 'docker run' step attempts to run an image called adoptopenjdk-external_custom-test:8-jdk-ubuntu-hotspot-full
> 
> so, being careful to either create an image called adoptopenjdk-external_custom-test:8-jdk-ubuntu-hotspot-full or run an image called adoptopenjdk-zookeeper-test:8-jdk-ubuntu-hotspot-full, but you can't create one and attempt to run another, because it will not be found.
> 
> (notice in the code where/how the name of the image is set for both external --build and external --run).

Yes I was trying to make sure that the same image is created by changing the line 43 in external.sh but it gave an error. I will try to investigate a bit more. Thanks!

sophia-guo:
I think you got the point @SAY-droid427 https://github.com/adoptium/aqa-tests/pull/2687#issuecomment-879237315.

There are three times external.sh is called and each time `test` was reset. Two them are set by ${TEST},  property of build.xml:

for example:
<property name="TEST" value="external_custom"/>

https://github.com/adoptium/aqa-tests/blob/master/external/build.xml#L61
https://github.com/adoptium/aqa-tests/blob/master/external/build.xml#L86

And another one is set hardcoded in playlist.xml
https://github.com/adoptium/aqa-tests/blob/master/external/external_custom/playlist.xml#L31

SAY-droid427:
> I think you got the point @SAY-droid427 [#2687 (comment)](https://github.com/adoptium/aqa-tests/pull/2687#issuecomment-879237315).
> 
> There are three times external.sh is called and each time `test` was reset. Two them are set by ${TEST}, property of build.xml:
> 
> for example:
> 
> https://github.com/adoptium/aqa-tests/blob/master/external/build.xml#L61
> https://github.com/adoptium/aqa-tests/blob/master/external/builges d.xml#L86
> 
> And another one is set hardcoded in playlist.xml
> https://github.com/adoptium/aqa-tests/blob/master/external/external_custom/playlist.xml#L31

Thank you so much @sophia-guo ! So instead of making the changes to external.sh, docker_functions.sh separately shouldn't I make the changes to playlist.xml and build.xml directly?

SAY-droid427:
The latest change to the file external.sh in the lines does not seem to work. I think I should change the [lines](https://github.com/adoptium/aqa-tests/blob/c0c016e8a49ddbecee68c9241323dc33a923851a/external/external.sh#L146) to use
``` ${echo ${EXTERNAL_CUSTOM_REPO} | awk -F'/' '{print $NF}' | sed 's/.git//g'}
```
in place of ${test}

SAY-droid427:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Tweetest/24/console
Based on this output, I think that no parameters are getting passed to it. I fail to understand what could be the possible reason. Could I please receive some help as to what can be the possible reason?

sophia-guo:
@SAY-droid427 to make the parameters are visible to test.sh you can either pass them to test.sh or export them as environment variables. 

That is in the generated Dockerfile you will need something like:

```
ENTRYPOINT ["/bin/bash", "/test.sh"] 
CMD ["parameters"] 
```

or 

```
ENV EXTERNAL_CUSTOM_REPO yourvalue.  

ENTRYPOINT ["/bin/bash", "/test.sh"] // EXTERNAL_CUSTOM_REPO is visible to test.sh
```
The environment variable solution is easier as test.sh won't need to do extra parsing. 

To pass in those parameters to Dockerfile you will need ARG command in Dockerfile. Example:

https://github.com/adoptium/aqa-tests/blob/master/external/dockerfile_functions.sh#L89

external/common_functions.sh
```diff
@@ -226,14 +226,18 @@ function set_test_info() {
         ubi_packages="git wget unzip"
         ubi_minimal_packages="${ubi_packages}"
         ;;
-    external_custom)
+    ${test})
```
sophia-guo, line 229: I know it's draft but would be better to point it out earlier rather than later.  This is not a good solution.  This is basically saying  if ${test} == ${test}.

SAY-droid427: Yes, I will update this soon by maybe creating an array of supported_external_custom_tests and then when it does not match any of the tests then it will check if any of the tests match the ${test} and if not then use the default error case. I will update it soon.

SAY-droid427:
> @SAY-droid427 to make the parameters are visible to test.sh you can either pass them to test.sh or export them as environment variables.
> 
> That is in the generated Dockerfile you will need something like:
> 
> ```
> ENTRYPOINT ["/bin/bash", "/test.sh"] 
> CMD ["parameters"] 
> ```
> 
> or
> 
> ```
> ENV EXTERNAL_CUSTOM_REPO yourvalue.  
> 
> ENTRYPOINT ["/bin/bash", "/test.sh"] // EXTERNAL_CUSTOM_REPO is visible to test.sh
> ```
> 
> The environment variable solution is easier as test.sh won't need to do extra parsing.
> 
> To pass in those parameters to Dockerfile you will need ARG command in Dockerfile. Example:
> 
> https://github.com/adoptium/aqa-tests/blob/master/external/dockerfile_functions.sh#L89

Thank you so much @sophia-guo for pointing this out. I had no idea about this. I will work on this problem this week.

SAY-droid427:
@sophia-guo I have implemented the print_external_custom_parameters() in dockerfile.sh to pass the parameters using ARG as you said. This is the output: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Tweetest/28/console
The parameters are passed to the test.sh (hopefully) as can be seen from the ant test command which is printed out in the console output, but the print statements and the cd command is not giving the desired output. Could you please check once if I have implemented the print_external_custom_parameters() function correctly? I had run the tests locally on my machine too and here is the screenshot of the changes reflected in the Dockerfile. Thank you!
Dockerfile_output

SAY-droid427:
@sophia-guo @smlambert I tried running the netty tests using this method and while the variables are getting passed correctly, the cd /{external_test_repo} shows cd / as below:
```
22:13:29  mvn clean package
22:13:29  /
22:13:29  Compile and run external tests
```
```
22:13:33  [[1;31mERROR[m] The goal you specified requires a project to execute but there is no POM in this directory (/). Please verify you invoked Maven from the correct directory. -> [1m[Help 1][m
```

SAY-droid427:
As I have separated the two functions in the common_functions.sh into two: set_test_info and set_external_custom_test_info, I think we can now completely remove the switch case inside the set_test_info() and use the getProperty() function I had already written in the to read the data directly from the test.properties file. I have tested it in my local machine and the tests run smoothly using it too. Looking forward to your suggestions on this.

SAY-droid427:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Tweetest/39/consoleFull
The tests are running successfully. Looking forward to your review. Thanks!

smlambert (approved):

external/dockerfile_functions.sh
```diff
@@ -500,7 +518,22 @@ generate_dockerfile() {
...
+    if [ ${test} == 'external_custom' ]; then
+        check_external_custom_test=1
+        echo "EXTERNAL_CUSTOM_REPO points to ${EXTERNAL_CUSTOM_REPO} in dockerfile_functions.sh"
+        echo "EXTERNAL_CUSTOM_BRANCH points to ${EXTERNAL_REPO_BRANCH} in dockerfile_functions.sh"
+        test="$(echo ${EXTERNAL_CUSTOM_REPO} | awk -F'/' '{print $NF}' | sed 's/.git//g')"
+        echo ${test}
+        tag_version=${EXTERNAL_REPO_BRANCH}
+    fi
+
+    if [[ ${check_external_custom_test} -eq 1 ]]; then
```
sophia-guo, line 532: We may  just move condition logic between Line 524 to Line 529  to line 532 using the same condition ` if [ ${test} == 'external_custom' ];` no need for the extra flag check_external_custom_test.

external/common_functions.sh
```diff
@@ -178,6 +181,36 @@ function getProperty() {
     echo $PROP_VALUE
 }
 
+function set_external_custom_test_info(){
+    test=$1
+    for current_test in ${supported_external_custom_tests}
```
sophia-guo, line 186: This for loop requires for each new third party app we will need to update the source code, which means it can't be used directly by tweeter api. For external_custom target we actually don't need to check the test target. This can be simplified as :
 
```
function set_external_custom_test_info(){
                PROPERTY_FILE=external_custom/test.properties
                GITHUB_URL=$(getProperty "github_url")
                github_url="${EXTERNAL_CUSTOM_REPO}"
                echo "The value of the EXTERNAL_CUSTOM_REPO in common_functions.sh is ${EXTERNAL_CUSTOM_REPO}"
                echo "The test name is ${test}"
                script="test.sh"
                test_command="${EXTERNAL_TEST_CMD}"
                echo "The value of the EXTERNAL_TEST_CMD in common_functions.sh is ${EXTERNAL_TEST_CMD}"
                test_results="testResults"
                tag_version="${EXTERNAL_REPO_BRANCH}"
                echo "The value of the EXTERNAL_REPO_BRANCH in common_functions.sh is ${EXTERNAL_REPO_BRANCH}"
                environment_variable="MODE=\"java\""
                debian_packages="git maven"
                debianslim_packages="${debian_packages}"
                ubuntu_packages="${debian_packages}"
                alpine_packages="git maven"
                centos_packages="git maven"
                clefos_packages="${centos_packages}"
                ubi_packages="git maven"
}

This can be updated when you do the cleanup to remove unnecessary echoes in the following PRs.

external/dockerfile_functions.sh
SAY-droid427: I used the flag check_external_custom_test because in the line 527 we are changing the test to point to individual tests like netty, zookeeper instead of external_custom, so now the test in all the subsequent function calls, test no longer refers to external_custom.  So if we are to pass the external_custom_parameters to Dockerfile we need a check whether it comes from an external_custom_target eligibe test or not.

external/common_functions.sh
SAY-droid427: Yes this can be updated. I will do so in the next PR. Thank you!

Issue #2263, opened 2021-02-11
jdk_util_2 FAILED in Test_openjdk8_hs_sanity.openjdk_arm_linux


smlambert:
**Test Info**
Test Name: jdk_util_2
Test Duration: 1 hr 2 min 5 sec
Machine: test-sxamb-armv7l-raspbian-pi4
TRSS link for the test output: https://trss.adoptopenjdk.net/output/test?id=6023fc3d5730424dbc6f337c

**Build Info**
Build Name: Test_openjdk8_hs_sanity.openjdk_arm_linux
Jenkins Build start time: Feb 10 2021, 08:41 am
Jenkins Build URL: https://ci.adoptopenjdk.net/job/Test_openjdk8_hs_sanity.openjdk_arm_linux/165/
TRSS link for the build: https://trss.adoptopenjdk.net/allTestsInfo?buildId=6023fb7c5730424dbc6f2d13

**Java Version**
openjdk version "1.8.0_282"
OpenJDK Runtime Environment (build 1.8.0_282-b08)
OpenJDK Client VM (build 25.282-b08, mixed mode)

**This test has been failed 10 times since Jan 27 2021, 05:48 am**
**Java Version when the issue first seen**
openjdk version "1.8.0_282-internal"
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_282-internal-202101270913-b01)
OpenJDK Client VM (AdoptOpenJDK)(build 25.282-b01, mixed mode)
Jenkins Build URL: https://ci.adoptopenjdk.net/job/Test_openjdk8_hs_sanity.openjdk_arm_linux/156/

The test failed on machine test-sxamb-armv7l-raspbian-pi4 2 times 
The test failed on machine build-scaleway-ubuntu1604-armv7-2 2 times 
The test failed on machine build-scaleway-ubuntu1604-armv7-1 2 times 
The test failed on machine test-aahlenst-armv7-raspbian-pi4-2 3 times 
The test failed on machine test-aahlenst-armv7-raspbian-pi4-1 1 times 

[...]

PR #2705, opened 2021-07-01
Exclude failing linux arm java.util.stream testcases due to issue #2263

andrew-m-leonard:
Excluded for: https://github.com/adoptium/aqa-tests/issues/2263


smlambert (approved):

sxa (approved):
As mentioned I'm tentatively ok with this going in, so I'll approve but we should definitely investigate further as a relatively high priority since it appears that it may be a regression compared to our previous GAs. In particular we shoudl determine if it fails with a similar frequency on the non-docker systems such as the scaleway and aahlenst boxes.

Issue #2655, opened 2021-06-17
extended.openjdk jdk_tools demo/jvmti tests fail because native libs cannot be found


lumpfish:
Some demo/jvmti tests fail with message similar to:
```
Could not find agent library /home/jenkins/workspace/Test_openjdk8_hs_extended.openjdk_aarch64_linux/openjdkbinary/j2sdk-image/jre/../demo/jvmti/gctest/lib/libgctest.so in absolute path, with error: /home/jenkins/workspace/Test_openjdk8_hs_extended.openjdk_aarch64_linux/openjdkbinary/j2sdk-image/jre/../demo/jvmti/gctest/lib/libgctest.so: cannot open shared object file: No such file or directory
```
The tests only exist in jdk8.  All platforms are affected.

These are failing tests and the their associated missing native library:
```
TEST: demo/jvmti/gctest/Gctest.java
j2sdk-image/jre/../demo/jvmti/gctest/lib/libgctest.so
TEST: demo/jvmti/heapTracker/HeapTrackerTest.java
j2sdk-image/jre/../demo/jvmti/heapTracker/lib/libheapTracker.so
TEST: demo/jvmti/heapViewer/HeapViewerTest.java
j2sdk-image/jre/../demo/jvmti/heapViewer/lib/libheapViewer.so
TEST: demo/jvmti/minst/MinstTest.java
j2sdk-image/jre/../demo/jvmti/minst/lib/libminst.so
TEST: demo/jvmti/mtrace/TraceJFrame.java
j2sdk-image/jre/../demo/jvmti/mtrace/lib/libmtrace.so
TEST: demo/jvmti/versionCheck/FailsWhenJvmtiVersionDiffers.java
j2sdk-image/jre/../demo/jvmti/versionCheck/lib/libversionCheck.so
TEST: demo/jvmti/waiters/WaitersTest.java
j2sdk-image/jre/../demo/jvmti/waiters/lib/libwaiters.so
```
Full output of a failing test:
```
19:12:32  TEST: demo/jvmti/gctest/Gctest.java
19:12:32  TEST JDK: /home/jenkins/workspace/Test_openjdk8_hs_extended.openjdk_aarch64_linux/openjdkbinary/j2sdk-image
19:12:32  
19:12:32  ACTION: compile -- Passed. Compilation successful
19:12:32  REASON: User specified action: run compile ../DemoRun.java 
19:12:32  TIME:   0.091 seconds
19:12:32  messages:
[...]

PR #2707, opened 2021-07-05
Exclude demo/jvmti tests jdk8

lumpfish:


lumpfish:
Ref: https://github.com/adoptium/aqa-tests/issues/2655
Test run with this PR: https://ci.adoptopenjdk.net/job/Grinder/1019/console

smlambert (approved):
Ultimately this will be a decision as to whether we build the native libs for these demo tests (which would get raised as a temurin-build repo issue) or not, if not, these will remain permanently excluded.

sophia-guo (approved):

PR #2709, opened 2021-07-05
Add impl ibm and openj9 for jck disabled elements

andrew-m-leonard:
A number of openj9 or ibm disabled jck tests are missing \<impl\> tags.

### DIFF
--- a/jck/compiler.api/playlist.xml
+++ b/jck/compiler.api/playlist.xml
@@ -85,6 +85,13 @@
 			<comment>Temporarily disabled on aix on JDK8 for test setup issue: backlog/issues/506</comment>
 			<platform>ppc64_aix(_mixed)?</platform>
 			<version>8</version>
+			<impl>ibm</impl>
+		</disabled>
+		<disabled>
+			<comment>Temporarily disabled on aix on JDK8 for test setup issue: backlog/issues/506</comment>
+			<platform>ppc64_aix(_mixed)?</platform>
+			<version>8</version>
+			<impl>openj9</impl>
 		</disabled>
 		<variations>
 			<variation>NoOptions</variation>
--- a/jck/compiler.lang/playlist.xml
+++ b/jck/compiler.lang/playlist.xml
@@ -34,6 +34,13 @@
 			<comment>Disabled on aix due to RTC: 144784</comment>
 			<platform>ppc64_aix(_mixed)?</platform>
 			<version>11</version>
+			<impl>ibm</impl>
+		</disabled>
+		<disabled>
+			<comment>Disabled on aix due to RTC: 144784</comment>
+			<platform>ppc64_aix(_mixed)?</platform>
+			<version>11</version>
+			<impl>openj9</impl>
 		</disabled>
 		<variations>
 			<variation>NoOptions</variation>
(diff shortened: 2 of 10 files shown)


jck/compiler.api/playlist.xml
```diff
@@ -85,7 +85,14 @@
 			<comment>Temporarily disabled on aix on JDK8 for test setup issue: backlog/issues/506</comment>
 			<platform>ppc64_aix(_mixed)?</platform>
 			<version>8</version>
+                        <impl>ibm</impl>
```
llxia, line 88: Could you please fix the indentation?

jck/compiler.api/playlist.xml
```diff
@@ -85,7 +85,14 @@
...
 		</disabled>
+                <disabled>
+                        <comment>Temporarily disabled on aix on JDK8 for test setup issue: backlog/issues/506</comment>
+                        <platform>ppc64_aix(_mixed)?</platform>
```
llxia, line 92: All platforms default to mixed. We can remove `(_mixed)?`

andrew-m-leonard: ah that's because the original code uses tabs, which when copy and pasted turns into spaces

andrew-m-leonard: i'll remove the tabs

jck/compiler.api/playlist.xml
andrew-m-leonard: original code uses tabs, which is not a great idea, as a lot of editors when copy&paste turn them into spaces, like it is here. In an editor it aligns for me.

jck/compiler.api/playlist.xml
andrew-m-leonard: I'll keep Mesbah's change asis

jck/compiler.api/playlist.xml
andrew-m-leonard: fixed indents

jck/compiler.api/playlist.xml
llxia: When Mesbah added the disabled, we were using mixed. Now we do not explicitly use mixed anymore. All platforms default to mixed (except zos).

andrew-m-leonard: Yes, so i'm thinking Mesbah should change those under #2713, that's not the prupose of this PR

Mesbah-Alam: I am okay with the idea of cleaning up existing `(_mixed)` in a separate PR!

Mesbah-Alam:
Thanks @andrew-m-leonard  for taking care of this. the PR looks fine to me. However, could you please run a grinder to ensure the playlists are working fine?

smlambert (approved):
Tested via temurin-compliance/job/Test_openjdk8_hs_sanity.jck_x86-64_mac/19 job.  For the platform update to remove _mixed, we can create a "good first issue" if it doesn't already exist.

andrew-m-leonard:
Thanks @Mesbah-Alam , as Shelley pointed to the job temurin-compliance/job/Test_openjdk8_hs_sanity.jck_x86-64_mac/19  verifies the change, I don't think you can see it, but it is successful.
cheers

Mesbah-Alam (approved):
LGTM.

Issue #2708, opened 2021-07-05
extended.openjdk hotspot_jre jdk8 gc/metaspace/TestMetaspaceMemoryPool.java fails java.lang.RuntimeException: Expected to find a memory pool with name Compressed Class Space


lumpfish:
extended.openjdk hotspot_jre jdk8 gc/metaspace/TestMetaspaceMemoryPool.java fails on 64 bit platforms with
```
java.lang.RuntimeException: Expected to find a memory pool with name Compressed Class Space
```

The test case only executes the check which gives rise to the exception on 64 bit platforms (so the test "passes" on 32 bit).

Example failing job: https://ci.adoptopenjdk.net/job/Test_openjdk8_hs_extended.openjdk_aarch64_linux/28/consoleFull

To rerun: https://ci.adoptopenjdk.net/job/Grinder/parambuild/?JDK_VERSION=8&JDK_IMPL=hotspot&JDK_VENDOR=Temurin&BUILD_LIST=openjdk&PLATFORM=aarch64_linux&TARGET=hotspot_jre_0

Full output of the failing test:
```
[2021-07-03T13:28:38.261Z] TEST: gc/metaspace/TestMetaspaceMemoryPool.java
[2021-07-03T13:28:38.261Z] TEST JDK: C:\Users\jenkins\workspace\Test_openjdk8_hs_extended.openjdk_x86-64_windows_testList_2\openjdkbinary\j2sdk-image
[2021-07-03T13:28:38.261Z] 
[2021-07-03T13:28:38.261Z] ACTION: build -- Passed. Build successful
[2021-07-03T13:28:38.261Z] REASON: Named class compiled on demand
[2021-07-03T13:28:38.261Z] TIME:   0.24 seconds
[2021-07-03T13:28:38.261Z] messages:
[2021-07-03T13:28:38.261Z] command: build TestMetaspaceMemoryPool
[2021-07-03T13:28:38.261Z] reason: Named class compiled on demand
[2021-07-03T13:28:38.261Z] Test directory:
[2021-07-03T13:28:38.261Z]   compile: TestMetaspaceMemoryPool
[2021-07-03T13:28:38.261Z] elapsed time (seconds): 0.24
[2021-07-03T13:28:38.261Z] 
[2021-07-03T13:28:38.261Z] ACTION: compile -- Passed. Compilation successful
[2021-07-03T13:28:38.261Z] REASON: .class file out of date or does not exist
[...]

PR #2710, opened 2021-07-05
Exclude TestMetaspaceMemoryPool jdk8

lumpfish:


lumpfish:
Ref: https://github.com/adoptium/aqa-tests/issues/2708
Test run with this PR: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1020/console.  There are failures in the run but they are not from the test excluded by this PR.

sophia-guo (approved):

smlambert (approved):

PR #2718, opened 2021-07-06
Fixing nativepath location issue on z/OS

psoujany:
This PR fixes -nativepath location on z/OS. 

Verification :- hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/16791

This PR fixes runtimes/openj9-openjdk-jdk11-zos/issues/594

### DIFF
--- a/openjdk/openjdk.mk
+++ b/openjdk/openjdk.mk
@@ -121,9 +121,7 @@ CUSTOM_NATIVE_OPTIONS :=
 
 ifneq ($(JDK_VERSION),8)
 	ifdef TESTIMAGE_PATH
-		ifneq ($(OS),OS/390)
-			JDK_NATIVE_OPTIONS := -nativepath:"$(TESTIMAGE_PATH)$(D)jdk$(D)jtreg$(D)native"
-		endif
+		JDK_NATIVE_OPTIONS := -nativepath:"$(TESTIMAGE_PATH)$(D)jdk$(D)jtreg$(D)native"
 		ifeq ($(JDK_IMPL), hotspot)
 			JVM_NATIVE_OPTIONS := -nativepath:"$(TESTIMAGE_PATH)$(D)hotspot$(D)jtreg$(D)native"
 		# else if JDK_IMPL is openj9 or ibm


openjdk/openjdk.mk
```diff
@@ -121,7 +121,7 @@ CUSTOM_NATIVE_OPTIONS :=
 
 ifneq ($(JDK_VERSION),8)
 	ifdef TESTIMAGE_PATH
-		ifneq ($(OS),OS/390)
+		ifeq ($(OS),OS/390)
```
sophia-guo, line 124: I believe this will make jdk native tests for other platforms fail.

sophia-guo (changes requested):

openjdk/openjdk.mk
llxia: I agree with Sophia, the current change will break other platforms. If we have zos native now, we can simply set `JDK_NATIVE_OPTIONS` without if check.

psoujany: Got it. Thank you. I made JDK_NATIVE_OPTIONS common across platforms. could you please review now.

openjdk/openjdk.mk
```diff
@@ -121,9 +121,7 @@ CUSTOM_NATIVE_OPTIONS :=
 
 ifneq ($(JDK_VERSION),8)
 	ifdef TESTIMAGE_PATH
-		ifneq ($(OS),OS/390)
-			JDK_NATIVE_OPTIONS := -nativepath:"$(TESTIMAGE_PATH)$(D)jdk$(D)jtreg$(D)native"
-		endif
+		JDK_NATIVE_OPTIONS := -nativepath:"$(TESTIMAGE_PATH)$(D)jdk$(D)jtreg$(D)native"
```
sophia-guo, line 124: This is the initial setup before the https://github.com/adoptium/aqa-tests/pull/1707/files. Does this mean zos jdk has testimage available now?

psoujany: Yes sophia, we've jdk test image for z/OS now. [This is the PR](https://github.com/eclipse-openj9/openj9/pull/13073) which had enabled jdk natives on z/OS

sophia-guo (approved):
LGTM

llxia (approved):
LGTM

PR #2721, opened 2021-07-09
Use KEEP_REPORTDIR for archiving test results on successful runs

smlambert:


andrew-m-leonard (approved):
lgtm

smlambert:
Talked with @llxia - and we will use the ARCHIVE_TEST_RESULTS param instead (implication is that for nightlies of openjdk and jck runs where we want to KEEP_REPORTDIR but not archive results in order to save space in Jenkins master or artifactory, we can do so).  Closing this in favour of https://github.com/adoptium/aqa-tests/pull/2723

PR #2728, opened 2021-07-12
Exclude automated jck tests requiring a real UI to be run manually

andrew-m-leonard:
Exclude those UI jck tests requiring manual running https://github.com/temurin-compliance/temurin-compliance/issues/29

### DIFF
--- a/jck/runtime.api/playlist.xml
+++ b/jck/runtime.api/playlist.xml
@@ -349,6 +349,12 @@
 			<platform>x86-64_mac(_mixed)?</platform>
 			<impl>openj9</impl>
 		</disabled>
+                <disabled>
+                        <comment>Requires UI https://github.com/temurin-compliance/temurin-compliance/issues/29</comment>
+                        <version>8</version>
+                        <platform>x86-64_mac</platform>
+                        <impl>hotspot</impl>
+                </disabled>
 		<variations>
 			<variation>NoOptions</variation>
 		</variations>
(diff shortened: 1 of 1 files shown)


sxa (approved):
LGTM

gdams (approved):

sophia-guo:
The disable feature supports regular expression. Probably use regular expression for some similar disable would be better. https://github.com/adoptium/aqa-tests/blob/master/doc/userGuide.md#exclude-test-against-multiple-criteria

To disable for all implementations we can simply do it by not adding  the  <impl> element.

Mesbah-Alam:
> To disable for all implementations we can simply do it by not adding the element. 

The down side of this would be that we'd miss documentation on why we disabled a target on which impl - something we may not want. 

For example, `jck-runtime-api-javax_naming-spi` is being disabled on hotspot for https://github.com/temurin-compliance/temurin-compliance/issues/29. It was disabled previously on openj9 for infrastructure/issues/5254.  So, when infrastructure/issues/5254 is fixed, we'd want to enable it on just openj9, but still keep it disabled on hotspot. Having separate `disabled` tags helps in this case.

andrew-m-leonard:
> > To disable for all implementations we can simply do it by not adding the element.
> 
> The down side of this would be that we'd miss documentation on why we disabled a target on which impl - something we may not want.
> 
> For example, `jck-runtime-api-javax_naming-spi` is being disabled on hotspot for [temurin-compliance/temurin-compliance#29](https://github.com/temurin-compliance/temurin-compliance/issues/29). It was disabled previously on openj9 for infrastructure/issues/5254. So, when infrastructure/issues/5254 is fixed, we'd want to enable it on just openj9, but still keep it disabled on hotspot. Having separate `disabled` tags helps in this case.

Thanks @Mesbah-Alam yes agree makes good sense.

PR #2745, opened 2021-07-16
Exclude jck tests requiring manual for hotspot on all versions

andrew-m-leonard:


sxa (approved):
To clarify, this is excluding tests from the automated runs because we've identified them as having to be run manually.

Assuming that's the intent of this, I approve (The title of the PR looks like we're just excluding some manual ones)

smlambert (approved):

PR #2752, opened 2021-07-16
Updated directoriesFilesChangePR workflow file to test with jdk16 instead of 15

### LINKED ISSUE #2715: Update directoriesFilesChangePR workflow file to test with jdk16 instead of 15
With the release cadence of Java marching forward, jdk15 is already done, we have moved to jdk16 and the PR testing needs to reflect that.  This is a good first issue to check the workflow files in the .github directory to ensure they are using jdk16 instead of jdk15 to test against.

For example: 
https://github.com/adoptium/aqa-tests/blob/master/.github/workflows/directoriesFilesChangePR.yml#L37 should get updated.

ManasviGoyal:
Fixed #2715

### DIFF
--- a/.github/workflows/directoriesFilesChangePR.yml
+++ b/.github/workflows/directoriesFilesChangePR.yml
@@ -34,7 +34,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


sophia-guo (approved):
LGTM

smlambert (approved):

PR #2753, opened 2021-07-17
#1728 Remove _xl and _mixed from map (save zos, riscv)

### LINKED ISSUE #1728: Jenkins script cleanup
- [x] there is no need to [iterationCount](https://github.com/AdoptOpenJDK/openjdk-tests/blob/481fb72c1030a2693af9d7b078c252f3d457a450/buildenv/jenkins/JenkinsfileBase#L393), we can just use variable `i`

- [x] we do not have `native-test-lib` anymore, following code can be removed
https://github.com/AdoptOpenJDK/openjdk-tests/blob/481fb72c1030a2693af9d7b078c252f3d457a450/get.sh#L349

- [ ] due to [Merge compressedrefs and non-compressedrefs tests builds into one](https://github.com/eclipse/openj9/issues/9231) work, `*_xl` should be removed (pending on https://github.com/eclipse/openj9/issues/9257). For example, 
https://github.com/AdoptOpenJDK/openjdk-tests/blob/481fb72c1030a2693af9d7b078c252f3d457a450/buildenv/jenkins/openjdk_tests#L12

jeffalder:
Fixes #1728 

> the last checkbox still needs doing, which is to update the PLATFORM_MAP in the openjdk_tests file to remove entries that end in _xl and _mixed, with the exception of zos and riscv platforms.

### DIFF
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -117,9 +117,6 @@ if (JDK_VERSIONS.size() > 1 || JDK_IMPLS.size() > 1 || PLATFORMS.size() >1 || PL
                     def ACTUAL_PLATFORM = PLATFORM
                     if (PLATFORM == "all") {
                         ACTUAL_PLATFORM = "ppc64_aix,ppc64le_linux,s390x_linux,x86-64_linux,x86-64_mac,x86-64_windows"
-                        if (JDK_IMPL == "openj9" || JDK_IMPL == "ibm"){
-                            ACTUAL_PLATFORM = ACTUAL_PLATFORM + ",ppc64_aix_xl,ppc64le_linux_xl,s390x_linux_xl,x86-64_linux_xl,x86-64_mac_xl,x86-64_windows_xl"
-                        }
                         if (JDK_VERSION == "8"){
                             ACTUAL_PLATFORM = ACTUAL_PLATFORM + ",x86-32_windows"
                         } else {
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -34,7 +34,7 @@ def PLATFORM_MAP = [
         'LABEL' : 'ci.role.test&&hw.arch.ppc64le&&sw.os.linux',
     ],
     'riscv64_linux' : [
-        'SPEC' : 'linux_riscv64_cmprssptrs',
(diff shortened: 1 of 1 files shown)


jeffalder:
@smlambert Sorry for the re-do on this; I closed and re-opened, trying to get something to fix itself. It needs workflow approval again.

buildenv/jenkins/openjdk_tests
```diff
@@ -13,14 +13,6 @@ def PLATFORM_MAP = [
         'SPEC' : 'linux_aarch64_cmprssptrs',
```
renfeiw, line 13: Thanks for your fix! We also need to remove the `_cmprssptrs` from all the `SPEC`, as the default will be mixed now. There are a few of them for different platforms.

jeffalder: Thanks for the feedback! I'll take a look when I'm back in the office.

jeffalder:
```suggestion
        'SPEC' : 'linux_aarch64',
```

I'm guessing that this is the correct change. I'm going to move forward by making this assumption throughout this file. If this is not correct, please let me know the right one.

jeffalder: OK. I _think_ I've done what you've asked; if there was xl or mixed and I removed at least one, then I removed `_cmprssptrs` from the remaining SPEC. The remaining `_cmprssptrs` are on RISC and zOS (because they still distinguish), and Solaris (because I didn't remove any xl or mixed definitions).

renfeiw: Thanks for the update! For Solaris, riscv and zos, I believe you can also remove the _cmprssptrs. @llxia What do you think?

llxia: Yes, please remove _cmprssptrs for solaris, riscv and zos. Thanks @jeffalder

jeffalder: Awesome, thanks, this is done for all as of 1ab18b3.

renfeiw (approved):
Thanks, LGTM.

llxia (approved):
Thanks @jeffalder . LGTM
 We are in the code freeze due to the release. This PR will be merged after the code freeze lifts.

smlambert:
Code freeze lifted, thanks for this one @jeffalder !

Issue #2754, opened 2021-07-18
Pre-release triage of weekly test runs


smlambert:
This report represents this view:

<img width="207" alt="Screen Shot 2021-07-18 at 11 11 43 AM" src="https://user-images.githubusercontent.com/2836948/126072423-8c3a393f-36c4-48c6-90e6-8e562898038e.png">

Notable: 
- all failures on mac related to https://github.com/adoptium/infrastructure/issues/2281 (3 machines in bad state)
- most failures on aix related to https://github.com/adoptium/infrastructure/issues/2282 (test-osuosl-aix72-ppc64-2 in bad state)

--------------

TRSS [link](https://trss.adoptium.net/buildDetail?parentId=60f2ca8e73f1507c79e52612&testSummaryResult=failed&buildNameRegex=%5ETest) 
Build URL https://ci.adoptopenjdk.net/job/build-scripts/job/openjdk8-pipeline/1709/ 
Started by upstream project "build-scripts/weekly-openjdk8-pipeline" build number 30 

[**Test_openjdk8_hs_extended.openjdk_ppc64le_linux_testList_0**](https://ci.adoptopenjdk.net/job/Test_openjdk8_hs_extended.openjdk_ppc64le_linux_testList_0/10/) 
hotspot_jre_0 => [deep history 0/2 passed](https://trss.adoptium.net/deepHistory?testId=60f3096773f1507c79e565c5) | [possible issues](https://trss.adoptium.net/possibleIssues?buildId=60f3088573f1507c79e563aa&buildName=Test_openjdk8_hs_extended.openjdk_ppc64le_linux_testList_0&testId=60f3096773f1507c79e565c5&testName=hotspot_jre_0) 
jdk_jmx_0 => [deep history 0/3 passed](https://trss.adoptium.net/deepHistory?testId=60f3096773f1507c79e565c1) | [possible issues](https://trss.adoptium.net/possibleIssues?buildId=60f3088573f1507c79e563aa&buildName=Test_openjdk8_hs_extended.openjdk_ppc64le_linux_testList_0&testId=60f3096773f1507c79e565c1&testName=jdk_jmx_0) 
[...]

PR #2757, opened 2021-07-19
Exclude the test as the fix isn't backported to upstream

sophia-guo:
Related with #2754 


smlambert (approved):

gdams (approved):

smlambert:
Noting that I believe this was causing both the jdk_jfr and jdk_jmx test targets to fail (for all platforms).

PR #2759, opened 2021-07-19
Updated tag_version

### LINKED ISSUE #2703: Please use a released version of tomcat for the test.
https://github.com/adoptium/aqa-tests/blob/97830a294fb593edb3486b82fce39c99f18858c3/external/tomcat/test.properties#L4
10.0.0-M5 is old 10.0.7 is the actual one.

It would be nice to test the 9.0.x branch too as that is the most used version.

ManasviGoyal:
fixed #2703

### DIFF
--- a/external/common_functions.sh
+++ b/external/common_functions.sh
@@ -462,7 +462,7 @@ function set_test_info() {
         github_url="https://github.com/apache/tomcat.git"
         script="tomcat-test.sh"
         home_path=""
-        tag_version="10.0.0-M5"
+        tag_version="10.0.7"
         ant_version="1.10.6"
         openssl_version="1.1.1d"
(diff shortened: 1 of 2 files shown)


sophia-guo:
I believe switching to use test.properties file hasn't been done yet. Probably need to update the line https://github.com/adoptium/aqa-tests/blob/master/external/common_functions.sh#L465 too.

ManasviGoyal:
> I believe switching to use test.properties file hasn't been done yet. Probably need to update the line https://github.com/adoptium/aqa-tests/blob/master/external/common_functions.sh#L465 too.

Updated

smlambert (approved):

sophia-guo (approved):

PR #2762, opened 2021-07-20
Un-exclude devtools schema tests

Mesbah-Alam:
This PR is to un-exclude the devtools schema test targets as the issue preventing these tests from running is now fixed via https://github.com/adoptium/aqa-systemtest/pull/439

FYI @JasonFengJ9 
 
### DIFF
--- a/jck/devtools.java2schema/playlist.xml
+++ b/jck/devtools.java2schema/playlist.xml
@@ -16,20 +16,6 @@
 	<include>../jck.mk</include>
 	<test>
 		<testCaseName>jck-devtools-java2schema-CustomizedMapping</testCaseName>
-		<disables>
-			<disable>
-				<comment>Disabled on x64 Linux + aarch64 + ppc64le on jdk8 due to backlog/issues/302. Awaiting potential test / setup issue resolution</comment>
-				<platform>x86-64_linux(_mixed)?|aarch64_linux(_mixed)?|ppc64le_linux(_mixed)?</platform>
-				<version>8</version>
-				<impl>ibm</impl>
-			</disable>
-			<disable>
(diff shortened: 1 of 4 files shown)


Mesbah-Alam:
Test : Grinder/17046 

(Note: The two targets that didn't get executed (reason for the grinder being marked 'failed') were due to a space in front of them in the `testList TESTLIST` value I had supplied for TARGET. They can be ignored - as the 10 that did get executed (and passed) were all exercising the same set of *.sh scripts as well - which had initially caused the failure fixed by adoptium/aqa-systemtest#439)

llxia:
Since we are in the release week and this PR re-includes 20+ test targets, maybe we should merge after the release? But if you (@JasonFengJ9 @Mesbah-Alam ) think the risk is low, I am ok to merge it as well.

Mesbah-Alam:
Not sure if this info can be used to determine level of risk, but the un-excluded targets here would only run in extended.jck, i.e., in weekend builds.

llxia:
release builds will run all tests (including extended.jck).

JasonFengJ9:
@llxia It is good to get more test coverage, the risk is the extra noise for triage if the test still fails. I think it can be merged after release cycle.

Mesbah-Alam:
Since we will not release this PR until after the release week, moving this PR to draft. 

Also, please note that, this PR needs to be delivered after https://github.com/adoptium/aqa-tests/pull/2769 - otherwise there will be a merge conflict.

smlambert (approved):

Mesbah-Alam:
Test for the un-excluded targets works fine (Grinder/17506), thanks to https://github.com/adoptium/aqa-systemtest/pull/439. We can deliver this PR now - which will enable these devtools schema tests in the next weekend tck builds.

llxia (approved):
LGTM

Issue #1971, opened 2020-09-18
Update external tests to latest stable releases


smlambert:
We have been doing this manually, changing the tag_version values in set_test_info in [common_functions.sh](https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/external/common_functions.sh#L176), but should also look at what it would take to automate this.

Related:
https://developer.github.com/v3/repos/releases/#get-the-latest-release

sophia-guo:
Tried with https://developer.github.com/v3/repos/releases/#get-the-latest-release looks like somerepos supports releases, some not. 
For exampley: apache/camel-quarkus, quarkusio/quarkus-quickstarts: not
https://github.com/sophia-guo/runaqaTest/runs/1311298479?check_suite_focus=true

AdoptOpenJDK/run-aqa: yes
https://github.com/sophia-guo/runaqaTest/runs/1311290502?check_suite_focus=true

Looks like it depends on how the release is created. https://gist.github.com/steinwaywhw/a4cd19cda655b8249d908261a62687f8#gistcomment-2602381

sophia-guo:
Maybe use tags instead though it can't be assumed that the tags are in chronological order

sophia-guo:
[renaissance ](https://github.com/renaissance-benchmarks/renaissanc) is a repo which we can use  `https://developer.github.com/v3/repos/releases/#get-the-latest-release` to get latest info.

https://github.com/AdoptOpenJDK/openjdk-tests/issues/2298#issuecomment-787915946

sophia-guo:
Github api https://developer.github.com/v3/repos/releases/#get-the-latest-release is not always reliable. It appears that it depends on how the Dev creates their "releases". https://gist.github.com/steinwaywhw/a4cd19cda655b8249d908261a62687f8#gistcomment-2602381

For example we can't get apache/camel-quarkus/ releases info. Same for quarkusio/quarkus-quickstarts.

```
sophiags-MacBook-Pro:bin sophiag$ curl -H "Accept: application/vnd.github.v3+json" "https://api.github.com/repos/apache/camel-quarkus/releases"
[

]
```

Probably go through all external repos and first work with the repos which we can get releases info.

sophia-guo:
It turns out if you don't explicitly create a release help.github.com/articles/creating-releases the tags are not actually being listed as releases. They just look like the release.

[...]

PR #2784, opened 2021-07-31
Update tags to latest release

SAY-droid427:
Added code to update the tags  to the latest release. 

Related to https://github.com/adoptium/aqa-tests/issues/1971


llxia:
Could you provide sample input and output of the running this script? Thanks

llxia:
Could you provide sample input and output of the running this script? Thanks

SAY-droid427:
@llxia  This is the video of the working of this script.
https://drive.google.com/file/d/1Nhprw-5GFi013vwKcsBNBUuTV1c3zA3H/view?usp=sharing
Thanks!

.github/workflows/updateToLatestRelease.py
```diff
@@ -0,0 +1,56 @@
+import requests
+import json
+
+filepath=input("Enter the file path: ")
```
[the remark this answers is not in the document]
SAY-droid427: OK I will make the changes.

.github/workflows/updateToLatestRelease.py
```diff
@@ -0,0 +1,56 @@
...
+modified_repos = []
+no_releases = []
+path = "https://api.github.com/repos/"
+updateflag = False
+
+with open(filepath) as fhand:
+  line = fhand.readline()
+  cnt = 0
```
sophia-guo, line 12: what's the `cnt` for

.github/workflows/updateToLatestRelease.py
```diff
@@ -0,0 +1,56 @@
...
+        elem = request_url+"\n"
+        no_releases.append(elem)
+
+    modified_repos.append(line)
+    line = fhand.readline()
+    cnt += 1
+fhand.close()
+
+# printing the urls for which latest/release donot exist
+print("No latest/release for url:")
```
sophia-guo, line 46: May be more clearer put those no latest one under the condition of ` if no_releases in not empty`

sophia-guo:
What's the input? One test.properties file or all test.properties files under specific directory?

.github/workflows/updateToLatestRelease.py
SAY-droid427: This can be removed as it is not being used. I was using it while debugging. Will remove that.

SAY-droid427:
> What's the input? One test.properties file or all test.properties files under specific directory?
> 
> Suppose this is only for repo with releases?
The input is test.properties of a particular repo we want to check for latest release. Do we want to check it like that or any other way?

smlambert:
The way I read what @sophia-guo is asking is whether you want this script to take a single test.properties file as input (as in feed this script (https://github.com/adoptium/aqa-tests/blob/master/external/netty/test.properties), or whether you could input https://github.com/adoptium/aqa-tests/blob/master/external and look in all subdirectories for all test.properties files.

I think if we intend to run this regularly to check if there is anything that needs updating, we may want the latter case, but in order to work for all files, you need to make sure it works first for a single file, which is what I believe you are doing here, correct?

SAY-droid427:
@sophia-guo @smlambert @llxia  I have updated the logic to take command-line input of https://github.com/adoptium/aqa-tests/blob/master/external and search for all the test.properties under that and update them simultaneously in one run. I will attach a working demo in a while because the API rate limits has exceeded for my machine. Looking forward to your review. Thank you!

.github/workflows/updateToLatestRelease.py
SAY-droid427: Done

sophia-guo (approved):
This is the first step of automatically updating latest releases. LGTM.

llxia (approved):
LGTM. We can enhance this in future PR.

PR #2791, opened 2021-08-06
Exclude testing in Graal JIT mode

sophia-guo:


smlambert (approved):

sophia-guo:
related with https://github.com/adoptium/adoptium/issues/63, which will not fix the issue of jvm_compiler. However according to the console output, it is the right way.

tellison:
I can't leave a review (no write access to repo?), but looks good to me :-)

smlambert:
re: https://github.com/adoptium/aqa-tests/pull/2791#issuecomment-894825898 that is odd, all committers should have this access and @tellison you are a committer (https://projects.eclipse.org/projects/adoptium.aqavit/who).  We will have to follow-up with Eclipse for why you are missing permissions.

karianna (approved):
LGTM

PR #2793, opened 2021-08-08
Remove spec completely

### LINKED ISSUE #2363: remove SPEC in openjdk_tests
With TKG auto-detecting SPEC, I think we can remove all SPEC in openjdk_tests. For example, https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/buildenv/jenkins/openjdk_tests#L5

This may affect special builds like Valhalla. 

FYI @renfeiw

jeffalder:
Fixes #2363 

Goal: Use checks like `==` instead of `startsWith` to avoid string formatting requirements

Method: Replace `SPEC`, which is overly-general, with `BASE_OS` which is targeted towards all the `if` conditions I could find (except one, noted in comments)

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -27,8 +27,8 @@ def setupEnv() {
 	env.JOBSTARTTIME = sh(script: 'LC_TIME=C date +"%a, %d %b %Y %T %z"', returnStdout: true).trim()
 
 	// Check what existing java processes are running
-		echo "PROCESSCATCH: Java processes which are currently on the machine:"
-	if (PLATFORM.contains("windows")) {
+	echo "PROCESSCATCH: Java processes which are currently on the machine:"
+	if (BASE_OS == 'windows') {
 		sh(script:'powershell -c "Get-Process java | select id,processname,starttime"', returnStatus:true)
 	} else {
 		sh(script:'ps -fu '+"${env.USER}"+' | grep java | egrep -v "remoting.jar|agent.jar|grep"', returnStatus: true)
@@ -39,7 +39,7 @@ def setupEnv() {
 		env.JDK_VERSION = params.JDK_VERSION.equalsIgnoreCase("next") ? "" : params.JDK_VERSION
 	}
 
-	env.SPEC = "${SPEC}"
+	env.BASE_OS = "${BASE_OS}"
 
 	ADOPTOPENJDK_REPO = params.ADOPTOPENJDK_REPO ? params.ADOPTOPENJDK_REPO : "https://github.com/adoptium/aqa-tests.git"
 	OPENJ9_REPO = params.OPENJ9_REPO ? params.OPENJ9_REPO : "https://github.com/eclipse-openj9/openj9.git"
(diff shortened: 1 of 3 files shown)


buildenv/jenkins/JenkinsfileBase
```diff
@@ -27,8 +27,8 @@ def setupEnv() {
 	env.JOBSTARTTIME = sh(script: 'LC_TIME=C date +"%a, %d %b %Y %T %z"', returnStdout: true).trim()
 
 	// Check what existing java processes are running
-		echo "PROCESSCATCH: Java processes which are currently on the machine:"
```
jeffalder, line 30: misformatted

buildenv/jenkins/JenkinsfileBase
```diff
@@ -445,12 +445,12 @@ def setup() {
 			RESOLVED_MAKE = "if [ `uname` = AIX ] || [ `uname` = SunOS ] || [ `uname` = *BSD ]; then MAKE=gmake; else MAKE=make; fi"
 			
 			dir( WORKSPACE) {
-				// use sshagent with Jenkins credentials ID for all platforms except zOS
 				// on zOS use the user's ssh key
-				if (!env.SPEC.startsWith('zos')) {
```
jeffalder, line 450: This and other `if` conditions were inverted so that we're always comparing positively. 
* It's easier for me to understand "if OS is zos then get_sources otherwise get_sources_with_authentication" than the other way around.
* In most of the cases I changed, there was a specific `if` and a specific` else` case; no fall-throughs, optional branches, or anything else.

buildenv/jenkins/openjdk_tests
```diff
@@ -150,136 +150,135 @@ if (JDK_VERSIONS.size() > 1 || JDK_IMPLS.size() > 1 || PLATFORMS.size() >1 || PL
         parallel testJobs
     }
 } else {
-    if (PLATFORM_MAP.containsKey(params.PLATFORM)) {
-        SPEC = PLATFORM_MAP[params.PLATFORM]["SPEC"]
-        LABEL = params.LABEL ? params.LABEL : PLATFORM_MAP[params.PLATFORM]["LABEL"]
+    if (!PLATFORM_MAP.containsKey(params.PLATFORM)) {
```
jeffalder, line 153: This is one example of an `if` that I negated. It is much easier for me to read "if not this then exit" compared to "if this .... 130 lines later ... else exit". I know the plain diff on GH is tough to read, so if you look at only the first commit, and select "Hide whitespace changes", you'll see that's how the change works.

openjdk/build.xml
```diff
@@ -77,7 +77,7 @@
 			<arg line="-d ${jtregTar}.tar.gz" />
 		</exec>
 		<if>
-			<contains string="${SPEC}" substring="zos" />
+			<equals arg="${BASE_OS}" arg2="zos" />
```
jeffalder, line 80: I didn't see `SPEC` set anywhere in the file, so I'm assuming this is launched from either `JenkinsfileBase` or `openjdk_tests`.

openjdk/build.xml
```diff
@@ -197,7 +198,7 @@
 					<else>
 						<if>
 							<and>
-								<contains string="${env.SPEC}" substring="arm" />
+								<equals arg1="${PLATFORM}" arg2="arm_linux" />
```
jeffalder, line 201: This is the only example I could find where the hardware component of SPEC was being used. That said, I'm pretty sure this branch never worked properly, as I wouldn't expect `env.SPEC` to resolve properly. The only `arm` SPEC was for the `arm_linux` PLATFORM so I figured this was an equivalent substitution.

llxia:
Thanks @jeffalder. Please see https://github.com/adoptium/aqa-tests/issues/2363#issuecomment-895306689 for details.

jeffalder:
@llxia Thanks for looking at this, and I apologize for letting this rot. I'm actually going to close this out. I don't feel comfortable being able to make this change, and I don't know how to test it.

I'm going to take another look at how I approach the project and see if I can find a better fit.

PR #2243, opened 2021-02-08
OpenJ9 AArch64: excluding stream/CountLargeTest

knn-k:
The following test is known to timeout with OpenJ9 AArch64.
This commit excludes it.

- java/util/stream/test/org/openjdk/tests/java/util/stream/CountLargeTest.java eclipse/openj9#9040

Signed-off-by: KONNO Kazuhiro <konno@jp.ibm.com>

PR #1947, opened 2020-09-03
OpenJ9 AArch64: excluding stream/CountLargeTest

knn-k:
The following test is known to timeout with OpenJ9 AArch64.
This commit excludes it.

- java/util/stream/test/org/openjdk/tests/java/util/stream/CountLargeTest.java eclipse/openj9#9040

Signed-off-by: KONNO Kazuhiro <konno@jp.ibm.com>

PR #1710, opened 2020-04-02
OpenJ9 AArch64: excluding tests known to timeout

Akira1Saitoh:
Excluding following tests known to timeout for OpenJ9 AArch64:

- java/lang/String/UnicodeCasingTest.java https://github.com/eclipse/openj9/issues/9081
- java/math/BigInteger/LargeValueExceptions.java https://github.com/eclipse/openj9/issues/9082
- java/net/httpclient/http2/HpackHuffmanDriver.java https://github.com/eclipse/openj9/issues/9041
- java/net/httpclient/websocket/ReaderDriver.java https://github.com/eclipse/openj9/issues/9083
- java/util/stream/test/org/openjdk/tests/java/util/stream/CountLargeTest.java https://github.com/eclipse/openj9/issues/9040

Signed-off-by: Akira Saitoh <saiaki@jp.ibm.com>

PR #2817, opened 2021-08-12
OpenJ9 AArch64: Stop excluding java/util/stream/stream/CountLargeTest

knn-k:
The following test used to timeout with OpenJ9 for AArch64, and it was
excluded by #1710, #1947, and #2243.

- java/util/stream/test/org/openjdk/tests/java/util/stream/CountLargeTest.java

This commit removes its entries from the exclude lists.


knn-k:
It seems I need to change the name of my fork.

smlambert (approved):

llxia (approved):

PR #2831, opened 2021-08-26
AIX: create display number based on date +"%u%M" and not use port 6000

aixtools:
A security scan complains about port 6000 being active. This is being opened by, among other tests, java_beans_X.

This change prevents, on AIX, the port 6000 being opening on the IP interfaces.


smlambert:
Based on a conversation with @sxa - we have not been able to reproduce the failure that this PR addresses, I would like to leave this open in draft form while we observe and work to reproduce.  If after a period of a few weeks we do not see the issue again, we will close this PR as not needed.

aixtools:
> 
> 
> Based on a conversation with @sxa - we have not been able to reproduce the failure that this PR addresses, I would like to leave this open in draft form while we observe and work to reproduce. If after a period of a few weeks we do not see the issue again, we will close this PR as not needed.

The `failure` is that a security scan tool sees port 6000 open. This is not something we can generate on our own. Not unless we were `lucky` and a process hangs and leaves the port open.

sxa:
> The `failure` is that a security scan tool sees port 6000 open. This is not something we can generate on our own. Not unless we were `lucky` and a process hangs and leaves the port open.

I thought `-secIP 000` was the solution to that problem - the port number switching was a separate issue, and just moving it around doesn't fix the fact that if it's an exposure on 6000 it'd be an exposure even if on another port. I believe we only looked at adding the port moving code to try and work around the case when there was an existing X server left over from a previous run.

aixtools:
> 
> 
> > The `failure` is that a security scan tool sees port 6000 open. This is not something we can generate on our own. Not unless we were `lucky` and a process hangs and leaves the port open.
> 
> I thought `-secIP 000` was the solution to that problem - the port number switching was a separate issue, and just moving it around doesn't fix the fact that if it's an exposure on 6000 it'd be an exposure even if on another port. I believe we only looked at adding the port moving code to try and work around the case when there was an existing X server left over from a previous run.

re: -secIP 000 - have to verify that resolves the security scan issues. If so, I'll eat some words (I forgot in any case).
re: it's an exposure: correct, could be an exposure - BUT - scanners being generally dumb, this issue seems to have been merely that that (standard) port was available. They are not scanning all ports, and checking protocols on all open ports.

iirc: I was trying to be more generic rather than AIX only.

aixtools:
> 
> 
> This problem is unlikely to be AIX specific so I would personally not support doing this for AIX only. Also I feel we should resolve the underlying problem (X servers being left over) instead of working around it like this which is likely to just grow more and more X servers over time. We already have some process cleanup code at the start of the test runs and adding an X server clean up to that would be a better approach, so we can ensure it's stopped before trying to start a new one.

I made a change - as a start to make it less AIX specific - but that was never accepted.

I have merged 'master' and dropped the generalization.

Considering that there were - already - platform specific conditional tests - I see no reason to 'improve' the AIX _then_ block.

smlambert (approved):

PR #2842, opened 2021-09-02
Exclude HelpFlagsTest.java on Windows 32bit

adamfarley:
An anticipated upstream fix should resolve the failure affecting
this test. Basically we're verifying the help output of a pair of
.exe files that (it appears) should be ignored by this test.

Excluding this test until the issue is resolved.


smlambert (approved):

llxia (approved):

adamfarley:
Thank you @smlambert and @llxia :)

PR #2880, opened 2021-09-18
Change concurrency setting for java_net test target

smlambert:
Not all platforms behave well with concurrency=cpus, so set to concurrency=1 for now to ensure we do not introduce contention for network resources that could result is undesirable failures.


llxia (approved):

Mesbah-Alam (approved):
LGTM

PR #2883, opened 2021-09-18
Moved unzip jtreg.jar command into if statement

### LINKED ISSUE #2879: Cannot run `make compile` twice with openjdk tests
In the local run, if we follow [the doc](https://github.com/eclipse-openj9/openj9/wiki/Reproducing-Test-Failures-Locally#run-specific-test-class-from-openjdk-tests-on-jdk11-linux-s390x-openj9-sdk) to run openjdk test. We found that we cannot run `make compile` twice with openjdk tests.

```
getJtreg:
    [mkdir] Created dir: /root/TEST/jvmtest/openjdk
     [exec] gzip: jtreg_6_1.tar already exists; not overwritten
BUILD FAILED
/root/TEST/aqa-tests/TKG/scripts/build_test.xml:76: The following error occurred while executing this line:
/root/TEST/aqa-tests/openjdk/build.xml:278: The following error occurred while executing this line:
/root/TEST/aqa-tests/openjdk/build.xml:76: exec returned: 2
```

I suspect when we run it the second time, the jtreg.jar is download and it fails to unzip again https://github.com/adoptium/aqa-tests/blob/master/openjdk/build.xml#L76

amandanguyeen:
Was previously trying to unzip this file even if it already existed
from a previous `make compile` command

Signed-off-by: Amanda Nguyen <amanda6@ualberta.ca>

Fixes #2879

### DIFF
--- a/openjdk/build.xml
+++ b/openjdk/build.xml
@@ -68,14 +68,14 @@
 							<copy file="custom_jtreg.tar.gz" tofile="${jtregTar}.tar.gz"/>
 					</then>
 				</if>
+				<exec executable="gzip" failonerror="true">
+					<arg line="-d ${jtregTar}.tar.gz" />
+				</exec>
 			</then>
 			<else>
    				<copy file="${LIB_DIR}/${jtregTar}.tar.gz" tofile="${jtregTar}.tar.gz"/>
 			</else>
 		</if>
-		<exec executable="gzip" failonerror="true">
-			<arg line="-d ${jtregTar}.tar.gz" />
-		</exec>
 		<if>
 			<contains string="${SPEC}" substring="zos" />
 			<then>


amandanguyeen:
I noticed that there were two other files that were changed: `openjdk/jtreg_6_1.tar` and `openjdk/openjdk-jdk/` Is this something that I need to include or is it something that happens when I run tests locally?

amandanguyeen:
Link to grinder job: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1442/

llxia (approved):
LGTM

sophia-guo:
I'm not sure the PR would work. @amandanguyeen did you try the scenario mentioned in #2879 to verify the change? Please see my comment https://github.com/adoptium/aqa-tests/issues/2879#issuecomment-925440158

PR #2890, opened 2021-09-21
Enable jtreg test selection by keywords

sophia-guo:
Enable jtreg test selection by keywords so jtreg can select tests automatically

Add keyword '!headful' so tests need ci won't run on headless jdk.

### DIFF
--- a/openjdk/openjdk.mk
+++ b/openjdk/openjdk.mk
@@ -90,6 +90,13 @@ JTREG_XML_OPTION = -xml:verify
 JTREG_BASIC_OPTIONS += $(JTREG_XML_OPTION)
 # Add any extra options
 JTREG_BASIC_OPTIONS += $(EXTRA_JTREG_OPTIONS)
+JTREG_KEY_OPTIONS :=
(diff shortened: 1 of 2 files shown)


sophia-guo:
Run jdk_bean, all passed. 4 tests failed for other issues: https://ci.adoptopenjdk.net/view/work-in-progress/job/grinder_sandbox_new/330/

sophia-guo:
Should we limit the selection to Temurin build? Do IBM or others want to test a headful alpine binary?

smlambert:
re: https://github.com/adoptium/aqa-tests/pull/2890#issuecomment-925133039 - IMHO, if someone wants to update this later if they are producing a full build for alpine-linux then it can come in as a separate PR.  In PMC discussions, it was noted that in known cases, alpine-linux applications are headless, which lead us to the decision to produce headless (let's assume for now other distributors would be same, until we hear differently).

smlambert (approved):
thanks @sophia-guo !

llxia (approved):
LGTM

Issue #2680, opened 2021-06-25
Accept STF sha in test pipeline


llxia:
In order to support [Reproducible Tests ](https://github.com/adoptium/aqa-tests/issues/2347), our test pipeline should accept STF sha from [STF repo](https://github.com/adoptium/STF). The user/upstream should be able to set STF sha. The test pipeline will use STF sha if it is provided. If not, the branch will be used.

In order to avoid parameter explosion, we should reuse `STF_OWNER_BRANCH`, so it can be 
`<owner>:<branch>`, `<owner>:<tag>`, or `<owner>:<sha>`. SImilar to PR https://github.com/adoptium/aqa-tests/pull/2829 (get.sh change)

In this case, code needs to be updated in ant:
https://github.com/adoptium/aqa-tests/blob/master/system/common.xml#L51

Please make sure to test both in local and Grinder.

xius666:
I would like to take this issue thanks

xius666:
-Grinder STF_OWNER_BRANCH example and its output

1. 

adoptium:master
output from grinder:
<img width="1074" alt="Screen Shot 2021-09-23 at 1 20 56 AM" src="https://user-images.githubusercontent.com/55036290/134604034-1c9a501c-9cbd-41d5-b231-b09e3f7f0b13.png">

2.

adoptium:2129b191b1111b665758fb755851025e35888a1b
output from grinder:
<img width="1015" alt="Screen Shot 2021-09-23 at 7 23 49 PM" src="https://user-images.githubusercontent.com/55036290/134604210-f8355817-d595-4f4a-9746-f26c85a2bd78.png">

3.
adoptium:invalid
output from grinder:
<img width="828" alt="Screen Shot 2021-09-23 at 7 13 04 PM" src="https://user-images.githubusercontent.com/55036290/134604326-ba3d4881-78e9-4b00-bbe9-b19a8b4bc225.png">

llxia:
This issue is resolved.

PR #2882, opened 2021-09-18
 Query git issue url to get issue status and add GIT_ISSUE_STATUS to the json

xius666:
solve issue #2701

PR #2895, opened 2021-09-24
accept STF sha in test pipeline 

xius666:
related to issue #2680


xius666:
-Grinder STF_OWNER_BRANCH example and its output

1. 

adoptium:master
output from grinder:
Screen Shot 2021-09-23 at 1 20 56 AM

2.

adoptium:2129b191b1111b665758fb755851025e35888a1b
output from grinder:
Screen Shot 2021-09-23 at 7 23 49 PM

3.
adoptium:invalid
output from grinder:
Screen Shot 2021-09-23 at 7 13 04 PM

xius666:
this code also has the change related to the PR #2882

system/common.xml
```diff
@@ -49,17 +49,49 @@
...
+				</exec>
+			</then>
+		</if>
+		
+		<if>
+ 			<equals arg1="${code2}" arg2="128"/>
+			<then>
+				<echo message="This branch does not exist!"/>
+			</then>
+		</if>
```
llxia, line 80: Could we move the if block inside the above if (after line 71)? We only need to check code2 if we execute line 68.

llxia:
Thanks @xius666 . The PR looks good except for the minor issue above. Also, Please remove the change that is related to PR #2882

system/common.xml
xius666: changed it pl review

llxia:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1485/

llxia (approved):
Thanks @xius666

Mesbah-Alam (approved):
LGTM

PR #1516, opened 2019-12-18
Set default concurrency to 1 for openjdk test

llxia:
fixes: #1513

Signed-off-by: lanxia <lan_xia@ca.ibm.com>

PR #1515, opened 2019-12-17
Update test lib location

renfeiw:
- update JenkinsfileBase for test lib location
- update MBCS datetime build.xml

Signed-off-by: Renfei Wang <renfeiw@ca.ibm.com>

Issue #1514, opened 2019-12-17
Intermittent build failures list


M-Davies:
Me, @adam-thorpe and @andrew-m-leonard have been tracking sporadic build failures at ci.adoptopenjdk for triage purposes. 
This list should now be made public so it can be contributed to by anyone observing the nightly builds.

Intermittent Nightly test list: https://github.com/AdoptOpenJDK/openjdk-tests/issues/1511

M-Davies:
jdk11-openj9/hotspot-windowsXL 13-12-2019
~jdk11-openj9/hotspot-mac 09-12-2019~ FIXED
~jdk11-openj9/hotspot-macXL 12-12-2019~ FIXED 
```
Scheduling project: build-scripts » release » sign_build
Starting building: build-scripts » release » sign_build #6387
build-scripts » release » sign_build #6387 completed with status FAILURE (propagate: false to ignore)
```
~https://github.com/AdoptOpenJDK/openjdk-build/issues/1428~
~https://github.com/AdoptOpenJDK/openjdk-build/issues/1429~

M-Davies:
jdk11-openj9/hotspot-windows 09-12-2019
```
21:14:15  + ./openjdk-tests/get.sh -s /cygdrive/c/Users/jenkins/workspace/Test_openjdk11_hs_sanity.openjdk_x86-64_windows -t /cygdrive/c/Users/jenkins/workspace/Test_openjdk11_hs_sanity.openjdk_x86-64_windows/openjdk-tests -p x64_windows -r upstream -j 11 -i hotspot --openj9_repo https://github.com/eclipse/openj9.git --openj9_branch master --tkg_repo https://github.com/AdoptOpenJDK/TKG.git --tkg_branch master
21:14:15  ./openjdk-tests/get.sh: line 14: $'\r': command not found
21:14:15  ./openjdk-tests/get.sh: line 15: set: -
: invalid option
21:14:15  set: usage: set [-abefhkmnptuvxBCHP] [-o option-name] [--] [arg ...]
21:14:15  ./openjdk-tests/get.sh: line 16: $'\r': command not found
21:14:15  ./openjdk-tests/get.sh: line 39: $'\r': command not found
21:14:15  ./openjdk-tests/get.sh: line 40: $'\r': command not found
21:14:15  ./openjdk-tests/get.sh: line 41: syntax error near unexpected token `$'\r''
[...]

PR #2913, opened 2021-09-29
Fix stf sha issue 

xius666:


system/common.xml
```diff
@@ -75,21 +75,37 @@
 						<echo message="This branch does not exist!"/>
 					</then>
 				</if>
+				<if>
```
llxia, line 78: Instead of if, should this be `else block`? Also, I would like to have `git checkout` logic once.
pseudocode:
```
run git rev-parse ${stf_branch} // failonerror="false" and get stf_sha1
If code=128 {
    git rev-parse origin/${stf_branch} // set failonerror="true" and set stf_sha2
     if code2 =128 {
           // error. fail the build. if set failonerror="true" works as expected, we do not need this if block
      } else {
          // success. Set stf_sha
         <property stf_sha value="${stf_sha2}"/>
      }
}
if  stf_sha is not set {
      // set it to stf_sha1
      <property stf_sha value="${stf_sha1}"/>
}
run git checkout -q -f ${stf_sha}
```

xius666: I changed the logic according to the pseudocode, it is indeed a cleaner approach, and I also rerun on the grinder(#1514,#1515,#1516).

llxia: Thanks a lot, @xius666 !

llxia:
Grinder for the default case: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1518/

xius666:
@llxia thanks for the suggestion, changed it pl review

llxia (approved):
LGTM

LongyuZhang (approved):
LGTM

PR #2938, opened 2021-10-01
Update langtools_custom target

### LINKED ISSUE #2692: Update langtools_custom target
Currently we run specific langtools testscase using jdk_custom target, which is a little bit confusing ( especially the description is wrong https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/testJobTemplate#L313, for the target both langtools and jdk custom should use the same target jdk_custom). 

It would be nice to add a separate langtools_custom for langtools tests especially langtools may need different options.

totorororo:
Fixes #2692 
Update langtools_custom target

### DIFF
--- a/openjdk/openjdk.mk
+++ b/openjdk/openjdk.mk
@@ -112,7 +112,6 @@ endif
 
 JDK_CUSTOM_TARGET ?= java/math/BigInteger/BigIntegerTest.java
 LANGTOOLS_CUSTOM_TARGET ?= tools/javac/4241573/T4241573.java
-
 ifneq (,$(findstring $(JDK_VERSION),8-9))
 	JTREG_JDK_TEST_DIR := $(OPENJDK_DIR)$(D)jdk$(D)test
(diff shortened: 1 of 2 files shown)


llxia:
Thanks @totorororo 
Grinder is launched: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1540/

llxia:
There is no default LANGTOOLS_CUSTOM_TARGET set. @sophia-guo any suggestion what should be good default value?

sophia-guo:
This should be a good one tools/javac/4241573/T4241573.java

totorororo:
added default for LANGTOOLS_CUSTOM_TARGET

sophia-guo (approved):
LGTM.

smlambert:
Tested via https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1541/

smlambert (approved):

PR #2946, opened 2021-10-04
Set BUILD_LIST default to empty in jenkins file

renfeiw:
- remove unused getBuildList function
- set BUILD_LIST default to empty
- when in dynmaic compilation, user doesn't need to set BUILD_LIST


smlambert (approved):
Love to see this simplicity!  thanks @renfeiw !

Issue #2910, opened 2021-09-28
Dynamically map platforms from ProblemList to PLATFORM


llxia:
Currently, we have a [hardcoded map ](https://github.com/adoptium/aqa-tests/blob/master/disabledTestParser/generateDisabledTestListJson.py#L7) to map [platform value in problem list](https://github.com/adoptium/aqa-tests/blob/master/openjdk/excludes/ProblemList_openjdk11-openj9.txt#L30) to [PLATFORM](https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/openjdk_tests#L3). For enhancement, I think we can map this dynamically. We can split the platform value in the problem list by -. The matching PLATFORM should contain the tokens with a few exceptions.

Related: https://github.com/adoptium/aqa-tests/issues/2698

prahasR:
Hi @llxia can you please elaborate a little more over this. What exactly do you mean by dynamical mapping? I am getting it as that the "platform_map" is a dictionary that stores string as keys and list as values. So you want to split the keys by - and then map the splitted keys to the list?

llxia:
For example, if we get `linux-ppc32`, we can set `PLATFORM` to be `ppc32_linux` by split the value using `-` and reverse the first and the second token. We do not really need a hardcoded map.

There are a few mistakes in the map:
https://github.com/adoptium/aqa-tests/blob/master/disabledTestParser/generateDisabledTestListJson.py#L8
https://github.com/adoptium/aqa-tests/blob/master/disabledTestParser/generateDisabledTestListJson.py#L17

You can ignore them for now. They should be fixed in https://github.com/adoptium/aqa-tests/issues/2908

prahasR:
I think that one of the possible solution is to directly add the result we get by doing above mentioned operation over `unresolved_platform_name` in the `revolved_platform_list` instead of doing it by `if-elif` conditions here https://github.com/adoptium/aqa-tests/blob/master/disabledTestParser/generateDisabledTestListJson.py#L49. I am working toward this @llxia, please assign this issue to me.

prahasR:
Is the solution seems correct to you? Should I start working toward this @llxia ?

gervaisj:
I'd like to work on this.

[...]

PR #2955, opened 2021-10-06
Dynamical mapping of platform from problemlist to PLATFORM added

prahasR:
Solves #2910 please review @llxia @smlambert


smlambert:
@llxia - I will defer to your feedback on this PR.

llxia:
Thanks @prahasR for working on this. Sorry, this fell off my radar. 

The expected solution is to remove `platform_map` completely, so we do not need to update the map for future platforms.
For example, `linux-aarch64` can be matched to `aarch64_linux` by switching the order and changing the `-` to `_`.

@prahasR could you update this PR? Thanks

llxia:
I will close this PR as it is obsolete. Please feel free to open a new PR with the update. Thanks.

Issue #2918, opened 2021-09-30
consume testenv.properties in system tests


llxia:
details in: https://github.com/adoptium/aqa-tests/issues/2889

testenv.properties is in testenv folder. See https://github.com/adoptium/aqa-tests/pull/2921

llxia:
The feature is implemented. Close this issue.

PR #2967, opened 2021-10-07
consume property file in system test

xius666:
issue #2918


xius666:
https://ci.adoptopenjdk.net/job/Grinder/1648/

psaunderualberta:
Did you mean for it to be 'complie'? Shouldn't it be "compile"

Issue #2701, opened 2021-06-30
Query git issue url to get issue status


llxia:
As part of [Create a Disabled tests pipeline](https://github.com/adoptium/aqa-tests/issues/1853), we need to query `GIT_ISSUE` url to get the issue status (open/closed). 

- input: 

```
[{
"JDK_VERSION" : 17,
"JDK_IMPL" : "hotspot",
"TARGET" : "jdk_custom",
"CUSTOM_TARGET" : "java/lang/ClassLoader/nativeLibrary/NativeLibraryTest.java",
"GIT_ISSUE" : "https://github.com/adoptium/temurin-build/issues/248",
"PLATFORM" : "all"
},{
"JDK_VERSION" : 16,
"JDK_IMPL" : "openj9",
"TARGET" : "jdk_custom",
"CUSTOM_TARGET" : "java/lang/ClassLoader/Assert.java",
"GIT_ISSUE" : "https://github.com/eclipse-openj9/openj9/issues/6668",
"PLATFORM" : "x86-64_mac"
}]
```

- query git to get the issue status and add GIT_ISSUE_STATUS in json. Expected output:
```
[{
"JDK_VERSION" : 17,
"JDK_IMPL" : "hotspot",
"TARGET" : "jdk_custom",
"CUSTOM_TARGET" : "java/lang/ClassLoader/nativeLibrary/NativeLibraryTest.java",
"GIT_ISSUE" : "https://github.com/adoptium/temurin-build/issues/248",
"PLATFORM" : "all",
"GIT_ISSUE_STATUS": "closed"
},{
"JDK_VERSION" : 16,
"JDK_IMPL" : "openj9",
"TARGET" : "jdk_custom",
"CUSTOM_TARGET" : "java/lang/ClassLoader/Assert.java",
"GIT_ISSUE" : "https://github.com/eclipse-openj9/openj9/issues/6668",
"PLATFORM" : "x86-64_mac",
"GIT_ISSUE_STATUS": "open"
}]
```

llxia:
example: https://github.com/adoptium/aqa-test-tools/blob/master/MachineLearningPrototype/dataCollection/collect_data.py

xius666:
I finished and please review it thanks

PR #2974, opened 2021-10-10
Update ProblemList.json

King-kay:
updated .json file with Git issue status


llxia:
Close this PR. The original issue https://github.com/adoptium/aqa-tests/issues/2701 is fixed via #2882.

PR #2995, opened 2021-10-14
enhance the function getProperty() to deal with quotations

norbline:


norbline:
Hello sir @andrew-m-leonard, i have even configure my email adress both locally and globally nothing seems to change

PR #3007, opened 2021-10-15
Exclude 2 jfr testcases on aarch64 platform

sainamercy:
Related to issue #2985


sainamercy:
@smlambert @sophia-guo I am not sure If am doing this right. I am still learning to use git, git commands and github

PR #3008, opened 2021-10-15
enhance the function getProperty() to deal with quotations

norbline:


andrew-m-leonard:
@norbline Azah, I think your changes look good. Now there's two things to check:
1) Check with @sophia-guo if the intention was also to fix the test.properties? It's debateable whether that's needed, at least in this PR anyway.
2) You need to test your changes. The easiest way to do that is to submit what we call a Grinder using your branch of aqa-tests. So launch this job: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/build?delay=0sec
specify your fork and branch for ADOPTOPENJDK_REPO and ADOPTOPENJDK_BRANCH, then specify the test build list that will invoke the code you changed, I believe that's probably any external test suite, so BUILD_LIST=external, TARGET=sanity

norbline:
@andrew-m-leonard i recieved this message when trying to test my changes " Access Denied
norbline is missing the Job/Build permission"

smlambert:
@andrew-m-leonard - The reviewers are responsible for testing Outreachy PRs.  Outreachy interns during their term will be given access, but for contribution period, it is still restricted access.

Testing via: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1719

norbline:
@ShelleyLambert Hello mam, is like the test failed "Failed>console output" what does that mean and what next step am i supposed to take

Issue #2949, opened 2021-10-05
update sytem stf project name


sophia-guo:
System stf git repo name/project has changed to `STF`,  to make it work on case-sensitive platform we need to update the name accordingly. Otherwise we will get the following error 

```/home/runner/work/run-aqa/run-aqa/aqa-tests/system/common.xml:194: The following error occurred while executing this line:
/home/runner/work/run-aqa/run-aqa/aqa-tests/system/common.xml:60: The directory /home/runner/work/run-aqa/run-aqa/aqa-tests/system/stf does not exist
```

For example it happened run-aqa github action on linux
https://github.com/adoptium/run-aqa/runs/3796240585?check_suite_focus=true

To fix it, the `stf` in https://github.com/adoptium/aqa-tests/blob/master/system/common.xml should be replaced by `STF`

sophia-guo:
Looks like https://github.com/adoptium/STF may need to update. I will take a look and if the change should be there I will transfer this PR to STF repo.

abigailsleek:
I will like to assign this to myself.

Lovelynchristian:
I would love this to be assigned to me.

keithc-ca:
It might be better to go the other way: update scripts to use a working directory based on the actual repo name (`STF` instead of `stf`).

llxia:
@Lovelynchristian do you have any questions regarding this issue?

llxia:
Thanks @xius666

sophia-guo:
Agreed https://github.com/adoptium/aqa-tests/issues/2949#issuecomment-945911127. There are multiple places in aqa-tests and aqa-systemtest are using as `directoryVariable/stf/....`. Setting a working directory based on the actual repo name (STF instead of stf) should make the job easier. For now we'd like this first be addressed.

Probably open a separate issue to improve this(Setting a working directory based on the actual repo name for STF. @Mesbah-Alam thoughts?

PR #3038, opened 2021-10-19
fix stf to STF 

xius666:
issue #2949


sophia-guo:
According to the failure, we'll need the related changes in https://github.com/adoptium/aqa-systemtest/blob/master/openjdk.build/build.xml#L32. Probably a few other places. I will take a look. This one is ready for merge.

llxia:
@xius666 Do we test this PR in Grinder?

llxia:
Running internal zos Grinder/18915/

PR #3148, opened 2021-11-24
Exclude java/nio/channels/FileChannel/Transfer2GPlus.java from jdk17 linux arm32

Haroon-Khel:
Known upstream issue https://bugs.openjdk.java.net/browse/JDK-8272095

ref https://github.com/adoptium/infrastructure/issues/2374

The same upstream exclusion is required for arm32 as it fails on all of our arm32 linux boxes


sxa (approved):
LGTM but I'll let @smlambert confirm she's happy too :-)

smlambert (approved):
thanks @Haroon-Khel !

PR #3158, opened 2021-12-02
Update to test target names

### LINKED ISSUE #3157: Standardize testCaseName for each of the external tests
**Describe the bug**
current test name does not flow with the workflow

SL/Dec2: Since we are improving the automation of the external tests and continuing to reduce the duplication, it is important that we establish a testCaseName naming pattern that is easy to predict (and avoids needing to look into the playlist file to read the testCaseName when we want to know the TARGET to use).

| directory name | app repo name |  testCaseName(s) in playlist | notes & allowable exceptions | 
| ------ | -------- | ---------- | ----------- | 
| camel |[camel-quarkus](https://github.com/apache/camel-quarkus.git)|   [camel_test](https://github.com/adoptium/aqa-tests/blob/master/external/camel/playlist.xml#L30) |  allowable to shorten dir name and testCaseName to camel|
|  derby |[derby](https://github.com/apache/derby.git)|  [derby_test_junit_all](https://github.com/adoptium/aqa-tests/blob/master/external/derby/playlist.xml#L30) | change testCaseName to derby_test | 
|  elasticsearch |[elasticsearch](https://github.com/elastic/elasticsearch.git) | multiple entries mostly following the existing naming rule stated below | ignore this one for now, we may remove this test as we want to pull test material from repos with open-source licenses | 
| example-test |  example-test | ... | ignore for now, we may remove this folder as well, '-' should be replaced with '_' if we keep it | 
| external_custom |  $EXTERNAL_CUSTOM_REPO |  external_custom | ignore for now, this is a special target that can take parameters to set repo and target name | 
|  functional-test | functional-test is not an app, uses aqa-tests repo | multiple targets | ignore for now, this is a special case, should have '-' replaced with '_' | 
| ... |  ... | ... | ... | 
| ... |  ... | ... | ... | 

**Existing naming rules** already applied to many cases:
- we may use a pattern that is repeated, repo name = dir name = 'testCaseName+_test' suffix.  
- lowercase is used.  
- because we are using shell scripts to process, when names contained '-' those have typically been replaced with '_', to make processing easier.
- For cases where there are multiple `<test>` entries in the playlist.xml file, even those variations should follow a pattern.  For tests that encompass some version number and up, like 11+, we would name that using the same pattern as described.  For other entries that are limited to different JDK_IMPL or a single JDK_VERSION, we add that information to the name.  An example would be, a particular test is for `<impl>hotspot</impl>` and `<version>8</version>`, we would append _hotspot_jdk8 to the end of the testCaseName.

LizyBbethy:
update to external tests' TestNameCase in the playlist.xml
From analysis ,the few  passing tests seem to have this naming convention
and some have not been
when run with the current upstream master ,
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2694/console this grinder came up with error cannot find the following tests [wycheproof_test] in TESTLIST in the playlist.xml .

this PR corrects the errors in documentation and the playlist.xml for some external tests.
grinder after troubleshooting the issue finds the wycheproof in the TESTLIST under the common_functions.sh and playlist.xml
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2707/console

Fixes #3157

### DIFF
--- a/external/elasticsearch/playlist.xml
+++ b/external/elasticsearch/playlist.xml
@@ -27,7 +27,7 @@
 		</groups>
 	</test>
 	<test>
-		<testCaseName>elasticsearch_test_openj9_jdk8</testCaseName>
+		<testCaseName>elasticsearch_test</testCaseName>
 		<disables>
 			<disable>
 				<comment>https://github.com/adoptium/aqa-tests/issues/1720</comment>
--- a/external/example-test/playlist.xml
+++ b/external/example-test/playlist.xml
@@ -17,7 +17,7 @@
 		TestKitGen can produce the equivalent make command lines to be executed in the build 
 	-->
 	<test>
-		<testCaseName>example-test</testCaseName>
+		<testCaseName>example_test</testCaseName>
 		<command>docker run --rm $(EXTRA_DOCKER_ARGS) adoptopenjdk-example-test:latest ; \
 			$(TEST_STATUS); \
 			docker rmi -f adoptopenjdk-example-test
--- a/external/lucene-solr/playlist.xml
+++ b/external/lucene-solr/playlist.xml
@@ -27,7 +27,7 @@
 		</groups>
 	</test>
 	<test>
-		<testCaseName>lucene_solr_nightly_smoketest</testCaseName>
+		<testCaseName>lucene-solr_test</testCaseName>
 		<command> $(TEST_ROOT)$(D)external$(D)external.sh --run --tag "${DOCKERIMAGE_TAG}" --version ${JDK_VERSION} --impl ${JDK_IMPL} --dir lucene-solr --reportdst $(REPORTDIR) --docker_args "$(EXTRA_DOCKER_ARGS)"; \
 		$(TEST_STATUS); \
 		$(TEST_ROOT)$(D)external$(D)external.sh --clean --tag "${DOCKERIMAGE_TAG}" --version ${JDK_VERSION} --impl ${JDK_IMPL} --dir lucene-solr
--- a/external/openliberty-mp-tck/playlist.xml
+++ b/external/openliberty-mp-tck/playlist.xml
@@ -27,7 +27,7 @@
 		</groups>
 	</test>
 	<test>
-		<testCaseName>openliberty_microprofile_tck_jdk8_j9</testCaseName>
+		<testCaseName>openliberty-mp-tck_test</testCaseName>
 		<command> $(TEST_ROOT)$(D)external$(D)external.sh --run --tag "${DOCKERIMAGE_TAG}" --version ${JDK_VERSION} --impl ${JDK_IMPL} --dir openliberty-mp-tck --reportsrc /testResults/surefire-reports --reportdst $(REPORTDIR) --docker_args "$(EXTRA_DOCKER_ARGS)" ; \
 		$(TEST_STATUS); \
 		$(TEST_ROOT)$(D)external$(D)external.sh --clean --tag "${DOCKERIMAGE_TAG}" --version ${JDK_VERSION} --impl ${JDK_IMPL} --dir openliberty-mp-tck
--- a/external/payara-mp-tck/playlist.xml
+++ b/external/payara-mp-tck/playlist.xml
@@ -27,7 +27,7 @@
 		</groups>
 	</test>
 	<test>
-		<testCaseName>payara_microprofile_tck</testCaseName>
+		<testCaseName>payara-mp-tck_test</testCaseName>
 		<disables>
 			<disable>
 				<comment>https://github.com/adoptium/aqa-tests/issues/1160</comment>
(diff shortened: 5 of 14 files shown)


LizyBbethy:
cc @sophia-guo @smlambert @llxia

joeyleeeeeee97 (approved):
LGTM.

external/elasticsearch/playlist.xml
```diff
@@ -27,7 +27,7 @@
 		</groups>
 	</test>
 	<test>
-		<testCaseName>elasticsearch_test_openj9_jdk8</testCaseName>
+		<testCaseName>elasticsearch_test</testCaseName>
```
smlambert, line 30: This can stay as `elasticsearch_test_openj9_jdk8`, we will be changing the elasticsearch targets in a separate PR.

external/lucene-solr/playlist.xml
```diff
@@ -27,7 +27,7 @@
 		</groups>
 	</test>
 	<test>
-		<testCaseName>lucene_solr_nightly_smoketest</testCaseName>
+		<testCaseName>lucene-solr_test</testCaseName>
```
smlambert, line 30: '-' dashes need [special handling](https://github.com/adoptium/aqa-tests/blob/master/external/dockerfile_functions.sh#L62-L74) in the shell scripts, so our naming standard should be standardizing to use '_' underscores, so ideally we rename both the directory names and target names to use underscores.

external/openliberty-mp-tck/playlist.xml
```diff
@@ -27,7 +27,7 @@
 		</groups>
 	</test>
 	<test>
-		<testCaseName>openliberty_microprofile_tck_jdk8_j9</testCaseName>
+		<testCaseName>openliberty-mp-tck_test</testCaseName>
```
smlambert, line 30: See the 'For cases where there are multiple <test> entries...' section in the description of #3157 for guidance on this, so this would get changed to openliberty_mp_tck_openj9_jdk8.  As with lucene_solr, we may choose to change the directory name to replace the '-' with '_'.

external/payara-mp-tck/playlist.xml
```diff
@@ -27,7 +27,7 @@
 		</groups>
 	</test>
 	<test>
-		<testCaseName>payara_microprofile_tck</testCaseName>
+		<testCaseName>payara-mp-tck_test</testCaseName>
```
smlambert, line 30: payara_mp_tck_test and updating the folder to payara_mp_tck ...

external/quarkus/playlist.xml
```diff
@@ -27,7 +27,7 @@
 		</groups>
 	</test>
 	<test>
-		<testCaseName>quarkus_java_test</testCaseName>
+		<testCaseName>quarkus_test</testCaseName>
```
smlambert, line 30: Let's leave this one alone for now, this was called quarkus_java_test because it runs in 'java mode', versus the test below it in the playlist called quarkus_native_test, which is to be run in 'native mode' (if it were enabled).

external/thorntail-mp-tck/playlist.xml
```diff
@@ -27,7 +27,7 @@
 		</groups>
 	</test>
 	<test>
-		<testCaseName>thorntail_microprofile_tck</testCaseName>
+		<testCaseName>thorntail-mp-tck_test</testCaseName>
```
smlambert, line 30: thorntail test will be removed https://github.com/adoptium/aqa-tests/issues/3182 (so it won't matter that this is using '-')

external/example-test/playlist.xml
```diff
@@ -17,7 +17,7 @@
 		TestKitGen can produce the equivalent make command lines to be executed in the build 
 	-->
 	<test>
-		<testCaseName>example-test</testCaseName>
+		<testCaseName>example_test</testCaseName>
```
smlambert, line 20: example-test directory will be removed https://github.com/adoptium/aqa-tests/issues/3181

external/tomee/playlist.xml
```diff
@@ -27,7 +27,7 @@
 		</groups>
 	</test>
 	<test>
-		<testCaseName>tomee_test_j9</testCaseName>
+		<testCaseName>tomee_test</testCaseName>
```
smlambert, line 30: tomee is a microprofile tck test, so we should really name the directory and target as we do the others tomee_mp_tck.  In this case, since there are multiple targets in the playlist, this one would then get changed to tomee_mp_tck_openj9 and the target below it would be tomee_mp_tck_hotspot.

smlambert (commented):
Since we broadened the issue that this PR is fixing, I've got a lot of comments @LizyBbethy :)
It may be worth meeting in a video call and stepping through the updates if you have questions about some of my comments.

smlambert:
hi @LizyBbethy, based on the virtual deskside review that you and @sophia-guo and I just had - if you want to remove the changes from the openliberty-mp-tck, payara-mp-tck and tomee files for now, we can return to those in a separate PR.  The other changes can be merged and we can follow with separate PRs for removal of example-test and thorntail-mp-tck.

PR #3162, opened 2021-12-06
jck: hardcode jck.env.runtime.url.ftpURL value

gdams:
without this we get: `Configuration file can not be imported due to question with specified tag "jck.env.runtime.url.httpURL" was not found on path. The path is:jck.intro`

### DIFF
--- a/jck/jck.mk
+++ b/jck/jck.mk
@@ -62,7 +62,7 @@ ifndef JCK_ROOT
 endif
 
 ifndef CONFIG_ALT_PATH
-  export CONFIG_ALT_PATH:=$(TEST_ROOT)$(D)jck$(D)jtrunner/config/default
+  export CONFIG_ALT_PATH:=$(TEST_ROOT)$(D)jck$(D)jtrunner$(D)config$(D)default
 endif
 
 OTHER_OPTS=
--- a/jck/jck.mk
+++ b/jck/jck.mk
@@ -62,7 +62,7 @@ ifndef JCK_ROOT
 endif
 
 ifndef CONFIG_ALT_PATH
-  export CONFIG_ALT_PATH:=$(TEST_ROOT)$(D)jck$(D)/jtrunner/config/default
+  export CONFIG_ALT_PATH:=$(TEST_ROOT)$(D)jck$(D)jtrunner/config/default
 endif
 
 OTHER_OPTS=
--- a/jck/jtrunner/JavaTestRunner.java
+++ b/jck/jtrunner/JavaTestRunner.java
@@ -224,13 +224,13 @@ public class JavaTestRunner {
 
 		testSuiteFolder = "JCK-" + testSuite.toString().toLowerCase() + "-" + jckVersionNo;
 		jckBase = jckRoot + File.separator + testSuiteFolder; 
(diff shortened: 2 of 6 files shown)


Mesbah-Alam:
@gdams , could you share the name of the test target that this fixes?

gdams:
@Mesbah-Alam it fixes tests in `jck-runtime-api-java_net_0`. The fix is actually carried over from the old systemtest repo: https://github.com/adoptium/aqa-systemtest/blob/master/openjdk.test.jck/config/jck8c/runtime.jti#L199

jck/jck.mk
```diff
@@ -62,7 +62,7 @@ ifndef JCK_ROOT
 endif
 
 ifndef CONFIG_ALT_PATH
-  export CONFIG_ALT_PATH:=$(TEST_ROOT)$(D)jck$(D)/jtrunner/config/default
+  export CONFIG_ALT_PATH:=$(TEST_ROOT)$(D)jck$(D)jtrunner/config/default
```
Mesbah-Alam, line 65: Thanks for this update. Since it's being fixed here, could we get all the `/` replaced with `$(D)` in `CONFIG_ALT_PATH` ?

Mesbah-Alam (approved):
LGTM

jck/jtrunner/JavaTestRunner.java
```diff
@@ -224,13 +224,13 @@ public static boolean generateJTB() throws Exception {
...
-		jtliteJarFullPath = jckRoot + File.separator + testSuiteFolder + File.separator + "lib" + File.separator + "jtlite.jar"; 
-		classesFullPath = jckRoot + File.separator + testSuiteFolder + File.separator + "classes";
+		jckPolicyFileFullPath = jckBase + File.separator + "lib" + File.separator + "jck.policy";
+		javatestJarFullPath = jckBase + File.separator + "lib" + File.separator + "javatest.jar";
+		jtliteJarFullPath = jckBase + File.separator + "lib" + File.separator + "jtlite.jar"; 
+		classesFullPath = jckBase + File.separator + "classes";
 		nativesLoc = jckRoot + File.separator + "natives" + File.separator + platform;
 		jtiFile = testRoot + File.separator + "jck" + File.separator + "jtrunner" + File.separator + CONFIG + File.separator + jckVersion + File.separator + testSuite.toLowerCase() + ".jti"; 
-		fileUrl = "file:///" + testSuiteFolder + "/testsuite.jtt";
+		fileUrl = "file:///" + jckBase + "/testsuite.jtt";
```
sophia-guo, line 233: Is this change on purpose? As @gdams mentioned this PR will hit `Unexpected exception:java.io.FileNotFoundException: /JCK-runtime-8c/testsuite.jtt (No such file or directory)` would suggest to change back to `testSuiteFolder`.

gdams: This change is on purpose, currently the directory cannot be found and fails. This uses the correct test directory

Mesbah-Alam: Thanks for spotting this @gdams!

smlambert (approved):

Issue #2401, opened 2021-03-20
Use of ci.role.perf label


sxa:
**Describe the bug**
Performance tests should run on machines with the `ci.role.perf` label

**To Reproduce**
Run any sanity.perf test

**Expected behavior**
Machines will only schedule on machines with `ci.role.perf` labels in jenkins. 

**Additional context**
I thought we had been using the `ci.role.test&&ci.role.perf` label for performance test suites but from what I can tell we only have [one machine](https://ci.adoptopenjdk.net/label/ci.role.perf/) with that label now and I can't see any indication in the logs that we're targetting any of the `perf` jobs to machines with that label. I would propose setting up a suitable set of machines with a `ci.role.perf` label which would go on machines that can produce consistent results where possible, and ensure that the `perf` tests only go onto those. I was about to set up some labels before realising they're not currently used.

I think I've mentioned this before but I would also propose using this label INSTEAD of `ci.role.test` and not require `ci.role.test&&ci.role.perf` to be present, to cover any case where we might be using a completely separate machien that we don't want to be used for other general testing.

smlambert:
We **HAD** configured certain perf test jobs to use ci.role.perf instead of ci.role.test back before when we had dedicated perf machines.  Since that time, we have:
- lost those dedicated machines and only recently added 2 mac minis back in that role
- regenerated the perf jobs overwriting the special label they were configured with
- made changes to now auto-generated test jobs from build pipelines in cases where test jobs do not exist

I had intended to create an issue similar to this one, but as you might be able to guess, am overwhelmed and am quite raw when taking 'suggestions/criticisms' about things on the AQA roadmap, that we have yet to do, that then suddenly someone takes an interest in.  Be gentle.

Indeed we need to add support back in, and should update our test pipelines for perf testing to check first if there is an option to send jobs to ci.role.perf&&hw.arch.xxx&&sw.os.yyyy machines, and if no label matches then select ci.role.test&&hw.arch.xxx&&sw.os.yyyy machines.

sxa:
> I'm quite raw when taking 'suggestions/criticisms' about things on the AQA roadmap

I know exactly what you mean - I feel the same way ref infra :-(

Thanks for the info. Was mostly trying to get up to speed on where we are at the moment since I didn't match what I thought I'd seen before and figure out what needed to happen on the infra side to make things happen.

smlambert:
[...]

PR #3168, opened 2021-12-08
Updated openjdk_test file - use of ci.role.perf label in Jenkins 

norbline:


buildenv/jenkins/openjdk_tests
```diff
@@ -152,7 +152,7 @@ if (JDK_VERSIONS.size() > 1 || JDK_IMPLS.size() > 1 || PLATFORMS.size() >1 || PL
 } else {
     if (PLATFORM_MAP.containsKey(params.PLATFORM)) {
         SPEC = PLATFORM_MAP[params.PLATFORM]["SPEC"]
-        LABEL = params.LABEL ? params.LABEL : PLATFORM_MAP[params.PLATFORM]["LABEL"]
+        LABEL = BUILD_LIST.contains("perf") ? params.LABEL : PLATFORM_MAP[params.PLATFORM]["LABEL"]
```
llxia, line 155: This does not set `ci.role.perf` in LABEL

llxia:
The change of adding `ci.role.perf ` needs to wait until both internal and external Jenkins have the machine labeled properly.

I think we should start with documenting the Perf machine requirements and `ci.role.perf ` machines at Adopt Jenkins. Ideally, we need at least one perf machine on each platform.

norbline:
> The change of adding `ci.role.perf ` needs to wait until both internal and external Jenkins have the machine labeled properly.
> 
> I think we should start with documenting the Perf machine requirements and `ci.role.perf ` machines at Adopt Jenkins. Ideally, we need at least one perf machine on each platform.

Hi @llxia looking at @sxa comment https://github.com/adoptium/aqa-tests/issues/2401#issuecomment-817853626.. is like he workes on it or something else

buildenv/jenkins/openjdk_tests
```diff
@@ -152,8 +152,15 @@ if (JDK_VERSIONS.size() > 1 || JDK_IMPLS.size() > 1 || PLATFORMS.size() >1 || PL
...
-
+      if (params.LABEL) {
+          LABEL = params.LABEL
+} else {
+  LABEL = PLATFORM_MAP[params.PLATFORM]["LABEL"]
+  if (params.BUILD_LIST.contains("perf")) {
+     LABEL = LABEL.minus("ci.role.test&&")
+     LABEL += '&&ci.role.perf'
+  }
+}
```
smlambert, line 163: Thanks @norbline - can you format/indent the block so its formatted like the other code?

smlambert (commented):
We could merge this as is, in which case if there are any platforms where we do not have machines, the test jobs would fail to find a machine that is online with a matching LABEL, https://github.com/adoptium/aqa-tests/blob/7ebdbe3d1408a93747a2b89cbb6181dddfd42cbe/buildenv/jenkins/openjdk_tests#L181-L182.  

An improvement to this (which we can do as part of this PR, or in a follow-up PR), is to move L181-182 to its own method, areNodesWithLabelOnline() that returns a boolean and call it in your else block to check we have nodes with your newly built LABEL, if not, then print out a warning and drop back to using ci.role.test for the time being.

sophia-guo (approved):

smlambert:
hi @norbline - based on my chat with @llxia and since we are trying not to break anyone who is currently using our scripts, the prudent thing to do as part of this PR is to create a method that checks if nodes of a particular label are online, something like:
```
def areNodesWithLabelOnline(labelToCheck) {
            String[] onlineNodes = nodesByLabel(labelToCheck)
            if (onlineNodes.size() < 1) {
                        return false
            } 
            return true
}        
```
before changing label, in your L159 check 
```
if (params.BUILD_LIST.contains("perf") && areNodesWithLabelOnline(LABEL.minus("ci.role.test&&").concat("&&ci.role.perf")) {
```

and since we have a method, also replace [L181-L182](https://github.com/adoptium/aqa-tests/blob/7ebdbe3d1408a93747a2b89cbb6181dddfd42cbe/buildenv/jenkins/openjdk_tests#L181-L182) with 
```
areNodesWithLabelOnline(LABEL) {
```

Can you make these changes to this PR please?  then we can proceed to test and merge.  Thanks!

norbline:
> hi @norbline - based on my chat with @llxia and since we are trying not to break anyone who is currently using our scripts, the prudent thing to do as part of this PR is to create a method that checks if nodes of a particular label are online, something like:
> 
> ```
> def areNodesWithLabelOnline(labelToCheck) {
>             String[] onlineNodes = nodesByLabel(labelToCheck)
>             if (onlineNodes.size() < 1) {
>                         return false
>             } 
>             return true
> }        
> ```
> 
> before changing label, in your L159 check
> 
> ```
> if (params.BUILD_LIST.contains("perf") && areNodesWithLabelOnline(LABEL.minus("ci.role.test&&").concat("&&ci.role.perf")) {
> ```
> 
> and since we have a method, also replace [L181-L182](https://github.com/adoptium/aqa-tests/blob/7ebdbe3d1408a93747a2b89cbb6181dddfd42cbe/buildenv/jenkins/openjdk_tests#L181-L182) with
> 
> ```
> areNodesWithLabelOnline(LABEL) {
> ```
> 
> Can you make these changes to this PR please? then we can proceed to test and merge. Thanks!

wow, thanks so much for the help @smlambert...

smlambert:
Note: I did not check yet if my code suggestions will work, so once you make the changes, you will need to run some Grinders on the changes @norbline :)

norbline:
> Note: I did not check yet if my code suggestions will work, so once you make the changes, you will need to run some Grinders on the changes @norbline :)
is this okay @smlambert ..failed on grinder https://ci.adoptopenjdk.net/job/Grinder/2778/
stage('Queue') {
            def areNodesWithLabelOnline(labelToCheck) {
            String[] onlineNodes = nodesByLabel(labelToCheck)
            if (onlineNodes.size() < 1) {
                        return false
            }
            return true
} {
                int ACTIVE_NODE_TIMEOUT = params.ACTIVE_NODE_TIMEOUT ? params.ACTIVE_NODE_TIMEOUT : 0
                timeout(ACTIVE_NODE_TIMEOUT) {
                    // If there is available node before timeout
                    node(LABEL) {
                        echo "find the node with label as ${env.NODE_NAME}"
                    }
                }
            }

smlambert:
hi @norbline - I may have caused you to go in the wrong direction :)

The new method you added should not go where you have put it, but rather it can go near the end of the file just before the other method, checkErrors(errorList) is defined at Line303.

Also, the reformatting you have applied is not desirable, as it has gone and changed all of the " double quotes to ' single quotes, which is unneeded.  Perhaps you can undo your last commit and just adjust the indentation of the small section of code by hand (without invoking formatting in your IDE).

Message me if you want to connect and walk through the next code review over a video call where we share a screen.  :)

PR #3160, opened 2021-12-03
Add temporary tck 11 excludes for z/OS

Mesbah-Alam:
- Add temporary tck 11 excludes for jdk 11 z/OS
- Related to backlog/issues/583

Signed-off-by: Mesbah-Alam <Mesbah_Alam@ca.ibm.com>

PR #3170, opened 2021-12-10
Exclude more failing tests on z/OS tck 11 + Clean up dups in previous excludes

Mesbah-Alam:
- Additional excludes for /backlog/issues/669 for failures observed in the latest extended tck 11 run on z/OS + cleanup of duplications in excludes introduced in https://github.com/adoptium/aqa-tests/pull/3160.
- Custom target updated to only run `api/java_math/BigInteger`. 

### DIFF
--- a/jck/runtime.api/playlist.xml
+++ b/jck/runtime.api/playlist.xml
@@ -357,11 +357,6 @@
 				<impl>openj9</impl>
 			</disable>
 			<disable>
-				<comment>Disabled on osx for infrastructure/issues/5254. Awaiting some automation improvements, in the interim these targets will be run manually</comment>
-				<platform>x86-64_mac</platform>
-				<impl>openj9</impl>
-			</disable>
-			<disable>
 				<comment>Temporarily disabled on z/OS for backlog/issues/679</comment>
 				<platform>.*zos.*</platform>
 				<impl>ibm</impl>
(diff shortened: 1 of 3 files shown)


smlambert (approved):
LGTM

jck/runtime.api/playlist.xml
```diff
@@ -16,6 +16,14 @@
 	<include>../jck.mk</include>
 	<test>
 		<testCaseName>jck_custom</testCaseName>
+		<disables>
```
llxia, line 19: If we exclude `jck_custom` on zos, it means we cannot use it for triaging or debugging user-specified tests. Not sure if there is a better way to exclude the subtest. For example, change default `JCK_CUSTOM_TARGET` from `api/java_math` `api/java_math/BigInteger`?

Mesbah-Alam: Yes. not a fan of running the whole `api/java_math` suite under `jck_custom` either - as that means we are running a target twice. 

Updated the PR to only run `api/java_math/BigInteger` - as suggested. This is faster + also works normally on z/OS , so no exclude necessary for jck_custom i: (results: Grinder_JCK/342).

llxia:
Just for the record, `jck_custom` is a special target for triaging or debugging purposes. The testing for `api/java_math` is covered in  `jck-runtime-api-java_math`

llxia (approved):
LGTM

PR #3179, opened 2021-12-14
use of ci.role.perf label in Jenkins fixes #2401

### LINKED ISSUE #2401: Use of ci.role.perf label

norbline:
Fixes #2401

### DIFF
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -155,7 +155,7 @@ if (JDK_VERSIONS.size() > 1 || JDK_IMPLS.size() > 1 || PLATFORMS.size() >1 || PL
                   LABEL = params.LABEL
                 } else {
                       LABEL = PLATFORM_MAP[params.PLATFORM]["LABEL"]
-                       if (params.BUILD_LIST.contains("perf")) && areNodesWithLabelOnline(LABEL.minus("ci.role.test&&").concat("&&ci.role.perf")){
+                       if (params.BUILD_LIST.contains("perf") && areNodesWithLabelOnline(LABEL.minus("ci.role.test&&").concat("&&ci.role.perf"))){
                          LABEL = LABEL.minus("ci.role.test&&")
                           LABEL += '&&ci.role.perf'
                         }
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -191,6 +191,7 @@ if (JDK_VERSIONS.size() > 1 || JDK_IMPLS.size() > 1 || PLATFORMS.size() >1 || PL
                 if (params.CLOUD_PROVIDER != null && params.CLOUD_PROVIDER in dynamicAgents && LABEL == PLATFORM_MAP[params.PLATFORM]["LABEL"]) {
                     boolean isNodeIdle = false
                     node {
+                        String[] onlineNodes = nodesByLabel(LABEL)
                         for (String onlineNode : onlineNodes) {
                             def nodeStatus = sh(script: "curl -s ${env.JENKINS_URL}computer/${onlineNode}/api/xml?xpath=/*/idle",
                                                 returnStdout: true,
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -157,8 +157,9 @@ if (JDK_VERSIONS.size() > 1 || JDK_IMPLS.size() > 1 || PLATFORMS.size() >1 || PL
              LABEL = PLATFORM_MAP[params.PLATFORM]["LABEL"]
                 if (params.BUILD_LIST.contains("perf") {
                     def perfLabel = LABEL.minus("ci.role.test&&").concat("&&ci.role.perf")
-                    if (areNodesWithLabelOnline(perfLabel))
-                    {  LABEL = perfLabel }
+                    if (areNodesWithLabelOnline(perfLabel)){
+                       ( LABEL = perfLabel )
+                    }
                 }
             }
 
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -151,7 +151,15 @@ if (JDK_VERSIONS.size() > 1 || JDK_IMPLS.size() > 1 || PLATFORMS.size() >1 || PL
 } else {
     if (PLATFORM_MAP.containsKey(params.PLATFORM)) {
         SPEC = PLATFORM_MAP[params.PLATFORM]["SPEC"]
-        LABEL = params.LABEL ? params.LABEL : PLATFORM_MAP[params.PLATFORM]["LABEL"]
+        if (params.LABEL) {
+                  LABEL = params.LABEL
+                } else {
+                      LABEL = PLATFORM_MAP[params.PLATFORM]["LABEL"]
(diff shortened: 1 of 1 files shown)


smlambert (commented):
This change is looking good @norbline - please add the Grinder links in the comments once they've been run.

norbline:
link to grinder https://ci.adoptopenjdk.net/job/Grinder/2787/

buildenv/jenkins/openjdk_tests
```diff
@@ -165,8 +173,7 @@ if (JDK_VERSIONS.size() > 1 || JDK_IMPLS.size() > 1 || PLATFORMS.size() >1 || PL
         println "LABEL: ${LABEL}"
 
         stage('Queue') {
-            String[] onlineNodes = nodesByLabel(LABEL)
```
llxia, line 168: `onlineNodes` is used in line 194. By removing it here, I think it will break line 194.

buildenv/jenkins/openjdk_tests
```diff
@@ -151,7 +151,15 @@ if (JDK_VERSIONS.size() > 1 || JDK_IMPLS.size() > 1 || PLATFORMS.size() >1 || PL
...
-        LABEL = params.LABEL ? params.LABEL : PLATFORM_MAP[params.PLATFORM]["LABEL"]
+        if (params.LABEL) {
+                  LABEL = params.LABEL
+                } else {
+                      LABEL = PLATFORM_MAP[params.PLATFORM]["LABEL"]
+                       if (params.BUILD_LIST.contains("perf") && areNodesWithLabelOnline(LABEL.minus("ci.role.test&&").concat("&&ci.role.perf"))){
+                         LABEL = LABEL.minus("ci.role.test&&")
+                          LABEL += '&&ci.role.perf'
+                        }
+                    }
```
llxia, line 162: Could we make the logic simpler in the `else` block: 
```
LABEL = PLATFORM_MAP[params.PLATFORM]["LABEL"]
def perfLabel = LABEL.minus("ci.role.test&&").concat("&&ci.role.perf")
if (params.BUILD_LIST.contains("perf") && areNodesWithLabelOnline(perfLabel)) {
     LABEL = perfLabel
}
```

buildenv/jenkins/openjdk_tests
norbline: how do mean @llxia it was moved to L297

llxia: By moving to line 297, `onlineNodes` becomes a local variable in the function. However, we need to use `onlineNodes` in https://github.com/adoptium/aqa-tests/blob/ad669e153b314745f8a68aabecd640a1be4a264d/buildenv/jenkins/openjdk_tests#L194

smlambert: Yes, that is my bad, I did not see that it was being used below in this case for determining dynamic agents... so either we skip calling the new method here or change the new method to just return the array and do the check for less than zero in the 2 places we need to know the answer.

smlambert: hi @norbline - to resolve this, for now you can put 
`String[] onlineNodes = nodesByLabel(LABEL)` as a line between L193 and L194
https://github.com/adoptium/aqa-tests/blob/ad669e153b314745f8a68aabecd640a1be4a264d/buildenv/jenkins/openjdk_tests#L193-L194

norbline: link to grinder https://ci.adoptopenjdk.net/job/Grinder/2795/

llxia:
https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/466/ (testing for no matching perf label)

norbline:
> https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/466/ (testing for no matching perf label)

test failed

llxia:
rebuild: https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/467

norbline:
> rebuild: https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/467

build was successful

llxia (approved):
Thanks @norbline

buildenv/jenkins/openjdk_tests
```diff
@@ -151,7 +151,15 @@ if (JDK_VERSIONS.size() > 1 || JDK_IMPLS.size() > 1 || PLATFORMS.size() >1 || PL
 } else {
     if (PLATFORM_MAP.containsKey(params.PLATFORM)) {
         SPEC = PLATFORM_MAP[params.PLATFORM]["SPEC"]
-        LABEL = params.LABEL ? params.LABEL : PLATFORM_MAP[params.PLATFORM]["LABEL"]
+        if (params.LABEL) {
+                  LABEL = params.LABEL
+                } else {
+                      LABEL = PLATFORM_MAP[params.PLATFORM]["LABEL"]
+                        def perfLabel = LABEL.minus("ci.role.test&&").concat("&&ci.role.perf")
```
sophia-guo, line 158: would prefer to move this `def perfLabel ` inside the if condition.

```
LABEL = PLATFORM_MAP[params.PLATFORM]["LABEL"]
if (params.BUILD_LIST.contains("perf") && areNodesWithLabelOnline(perfLabel)) { 
   def perfLabel = LABEL.minus("ci.role.test&&").concat("&&ci.role.perf")
   LABEL = perfLabel 
}
```

llxia: @sophia-guo the above will not work. `areNodesWithLabelOnline() ` uses `perfLabel `

sophia-guo: yes, didn't notice it. Probably the following code? The main point is perfLabel is only related with perf so moving under 'perf' condition may be better. 

```
LABEL = PLATFORM_MAP[params.PLATFORM]["LABEL"] 
if (params.BUILD_LIST.contains("perf") {
    def perfLabel = LABEL.minus("ci.role.test&&").concat("&&ci.role.perf")
    if (areNodesWithLabelOnline(perfLabel))
         {  LABEL = perfLabel }
}
```

norbline: @sophia-guo changes made ...grinder link https://ci.adoptopenjdk.net/job/Grinder/2815/

sophia-guo (approved):

smlambert (approved):
thanks @norbline!

smlambert:
Rerunning failed Grinder [466](https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/467) at (failed due to wrong credential):
https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/476

smlambert:
Grinder 476 passes 👍

PR #3184, opened 2021-12-14
Support openj9 sha without branch

llxia:
Currently, we take both OPENJ9_BRANCH and OPENJ9_SHA. However, in testenv.properties, we only output one value - OPENJ9_SHA (adoptium/TKG#255). We will not have branch information. The logic in `get.sh` was highly optimized to have minimal checkout for openj9 repo (i.e., we do not fetch other branches). This causes a problem as it is not able to git clone with sha only (if the sha is from other branches).

This PR adds the support to git clone personal openj9 sha without branch info, so we can consume the OPENJ9_SHA without OPENJ9_BRNACH info.

Related: adoptium/TKG#255
Related: adoptium#2889

Signed-off-by: lanxia lan_xia@ca.ibm.com

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -484,7 +484,13 @@ getFunctionalTestMaterial()
 			echo "git fetch -q --tags $OPENJ9_REPO +refs/pull/*:refs/remotes/origin/pr/*"
 			git fetch -q --tags $OPENJ9_REPO +refs/pull/*:refs/remotes/origin/pr/*
 			echo "git checkout -q $OPENJ9_SHA"
-			git checkout -q $OPENJ9_SHA
+			if ! git checkout $OPENJ9_SHA; then
+				echo "SHA not yet found. Continue fetching all the branches on origin..."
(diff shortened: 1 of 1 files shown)


llxia:
|USE_TESTENV_PROPERTIES| Grinder|Comments|
|--|--|--|
| false| https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2788||
|true| https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/2791|used [customized testenv.properties](https://github.com/adoptium/aqa-tests/commit/ef6dbb8e3a7d809969a3f502c6a8343849a4671d). It contians OPENJ9_SHA|
|false| internal/Grinder/20000| testing with both OPENJ9_BRANCH and OPENJ9_SHA|

renfeiw (approved):
LGTM thanks!

smlambert (approved):

PR #3194, opened 2021-12-31
Use xlc16 for running JCK tests on AIX

sxa:
Required for our new AIX machines. We should verify whether this will cause a problem for IBM before merging this. Worst case we can put in a switch so it can use either one that's available on the system.

### DIFF
--- a/jck/jtrunner/JavaTestRunner.java
+++ b/jck/jtrunner/JavaTestRunner.java
@@ -774,7 +774,7 @@ public class JavaTestRunner {
 			}
 
 			fileContent += "concurrency " + concurrencyString + ";\n";
-			fileContent += "timeoutfactor 4" + ";\n";							// All Devtools tests take less than 1h to finish.
+			fileContent += "timeoutfactor 40" + ";\n";							// All Devtools tests take less than 1h to finish.
 
 			if (platform.equals("win")) {
 				// On Windows set the testplatform.os to Windows and set systemRoot, but do not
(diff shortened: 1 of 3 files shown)


Mesbah-Alam:
This PR changes compiler to use for AIX from `/opt/IBM/xlC/13.1.3/bin/xlc` to `/opt/IBM/xlC/16.1.0/bin/xlc`. 

@AdamBrousseau , do we have `/opt/IBM/xlC/16.1.0/bin/xlc` on all internal AIX machines?

sxa:
Paging @AdamBrousseau

karianna:
@sxa Is this required for the release?

Mesbah-Alam (approved):
LGTM

smlambert:
It's in draft and from @sxa - We can merge it when he is ready (or run from his branch for AIX if we need).

sxa:
> It's in draft and from @sxa - We can merge it when he is ready (or run from his branch for AIX if we need).

I'm fairly happy this should be safe. Basically on the new machines we'll be using xlc16 is the only one installed, but that's not on the older machines (that we hopefully won't use!) so yep let's get this in.

smlambert (approved):
Adding a note for others, the timeout and timeoutfactor adjustments do not affect the overall execution time of healthy good / passing test runs.  What they do affect is in the case of a failing run, how long to let the tests try before giving up, marking as failed and moving on.  Changing it gives a longer window to try and run (good for some systems we may be running on), but be aware that if there are failures in nightlies/weeklies, due to dev work, these may end up causing what appears to be 'hangs', but would eventually time out, possibly after the next nightly is kicked off.  (I am not concerned with this, it can be adjusted to suit as needed).

Issue #3189, opened 2021-12-17
Assess the health of the perf test group


smlambert:
- Look at weekly perf runs and check that all perf targets that are enabled are passing (on all versions/impls/platforms), if not then annotate the failing targets (first in this issue - then we can assess and if needed find existing issues or raise new ones as required.  
- Likely easiest to start by looking at the grid view of the latest weekly run to assess what is running and its status, example: [latest weekly hotspot run for jdk8](https://trss.adoptium.net/resultSummary?parentId=61b494c5a7df1a2924c74087) shows which sanity.perf and extended.perf runs are green (passing), yellow (some targets failing) or red (job completely failed to run)
- Tracking status through the shared spreadsheet at this [link](https://docs.google.com/spreadsheets/d/1jcRO5fsrqXOBT7DxqNhaCCZa9Emupg3rw-iIdbbNo3k/edit?usp=sharing)

smlambert:
Activity update:
- many of the failures in the perf group are due to certain benchmarks in the renaissance suite that check whether the environment where they are running is sufficient.  These benchmarks use Apache Spark and this is related to the [Apache Spark configuration](https://books.japila.pl/apache-spark-internals/spark-properties/).  We currently are using the default values found in  spark-defaults.conf (the calculation that occurs is found in the [UnifiedMemoryManager.scala](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/memory/UnifiedMemoryManager.scala#L194-L208) file and can inform us on what we may need to change in our environment or configuration in order to be able to run the affected renaissance benchmarks on our machines at the project).  

FYI @sxa

norbline:
As Mentioned by @smlambert  These two platforms below seem to be affected after [running tests](https://ci.adoptopenjdk.net/job/Grinder/2903/console) check on the failed Benchmarks.

**Affected Platforms**
- arm_linux
- x86-32_windows

**Failed Benchmarks**
- renaissance-als
- renaissance-chi-square
- renaissance-dec-tree
- renaissance-gauss-mix	
- renaissance-log-regression 	
- renaissance-movie-lens
`17:36:36  Exception occurred in org.renaissance.apache.spark.GaussMix@143529a: System memory 259522560 must be at least 471859200. Please increase heap size using the --driver-memory option or spark.driver.memory in Spark configuration.
[...]

PR #3201, opened 2022-01-06
exclude some renaissance tests from arm_linux $ x86-32_windows

norbline:
Related: https://github.com/adoptium/aqa-tests/issues/3189


sophia-guo:
To make it simple `<platform>x86-32_windows|arm_linux</platform>` should be enough. No need to use wildcard character.

sophia-guo (approved):

smlambert (approved):
thanks @norbline !

PR #3231, opened 2022-01-18
Exclude tests with known issues

### LINKED ISSUE #3230: MiniMix_aot_5m_0 failed on windows32 with jdk11, jdk17
MiniMix_aot_5m_0 failed on windows32 with with jdk11, jdk17. Might happen to other version? Deep history shows it has been failed since Dec26 (maybe earlier than that, as earlier history is deleted)

```
12:00:07  STF 11:00:16.343 - Monitoring processes: LT 
12:00:07  LT  Error occurred during initialization of VM
12:00:07  LT  Could not reserve enough space for 2097152KB object heap
12:00:07  STF 11:00:16.380 - **FAILED** Process LT  ended with exit code (1) and not the expected exit code/s (0)
12:00:07  STF 11:00:16.380 - Monitoring Report Summary:
12:00:07  STF 11:00:16.380 -   o Process LT  ended with exit code (1) and not the expected exit code/s (0)
12:00:07  STF 11:00:16.381 - Killing processes: LT 
12:00:07  STF 11:00:16.381 -   o Process LT  pid 44304 is not running
12:00:07  **FAILED** at step 1 (Run mixed unit tests). Expected return value=0 Actual=1 at E:/workspace/Test_openjdk17_hs_extended.system_x86-32_windows/aqa-tests/\TKG\output_16423517415225\MiniMix_aot_5m_0/20220116-110014-MixedLoadTest/execute.pl line 98.
12:00:07  STF 11:00:16.397 - **FAILED** execute script failed. Expected return value=0 Actual=1

```

https://ci.adoptopenjdk.net/job/Test_openjdk17_hs_extended.system_x86-32_windows/134/consoleFull
https://ci.adoptopenjdk.net/job/Test_openjdk11_hs_extended.system_ppc64le_linux/587/

sophia-guo:

### DIFF
--- a/system/security/playlist.xml
+++ b/system/security/playlist.xml
@@ -36,6 +36,10 @@
 				<comment>https://github.com/eclipse-openj9/openj9/issues/13892</comment>
 				<impl>openj9</impl>
 			</disable>
+			<disable>
+				<comment>https://github.com/adoptium/aqa-systemtest/issues/474</comment>
+				<version>17+</version>
+			</disable>
 		</disables>
 		<command>$(SYSTEMTEST_CMD_TEMPLATE) -test=OAuthTest; $(TEST_STATUS)</command>
 		<levels>
--- a/openjdk/excludes/ProblemList_openjdk17.txt
+++ b/openjdk/excludes/ProblemList_openjdk17.txt
@@ -363,4 +363,8 @@ compiler/codegen/ClearArrayTest.java https://github.com/adoptium/aqa-tests/issue
 compiler/c2/TestJumpTable.java https://github.com/adoptium/aqa-tests/issues/2382 windows-x86
 compiler/loopstripmining/TestPinnedUseInOuterLSMUnusedBySfpt.java https://github.com/adoptium/aqa-tests/issues/2382 windows-x86
 compiler/vectorapi/TestNoInline.java https://github.com/adoptium/aqa-tests/issues/2382 windows-x86
+############################################################################
 
+#runtime_nestmate
+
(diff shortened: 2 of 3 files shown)


smlambert (approved):
thanks @sophia-guo !

llxia:
> close #3230

If this PR is going to close #3230, then we should not reference #3230 in the disable comment. https://github.com/adoptium/aqa-tests/blob/8849710aad46a7aea3416b3e337df8ef6b136a74/system/otherLoadTest/playlist.xml#L54

This creates a loop where a disabled test is referencing a closed issue. It will create problem for disabled test pipeline in the future.

system/security/playlist.xml
```diff
@@ -41,6 +41,10 @@
 		<levels>
 			<level>extended</level>
 		</levels>
+		<versions>
+			<version>8</version>
+			<version>11</version>
+		</versions>
```
llxia, line 47: Is this a temporary exclusion? If so, please use disable feature and reference the open issue. Thanks.

smlambert:
and I guess this is really only going to exclude in nightlies running off master...  after we review the other triaged test, we could consider pushing exclusions to the v0.8.0-release branch

system/security/playlist.xml
sophia-guo: Quick answer is we don't know.  @joeyleeeeeee97 could probably confirm that https://github.com/adoptium/aqa-systemtest/issues/474

llxia: @sophia-guo could we use disable feature and reference adoptium/aqa-systemtest#474 for now?

smlambert: that is a good approach as this is a temporary exclusion based on joeyleeeeeee97's reply

llxia (approved):
Thanks @sophia-guo

Issue #3156, opened 2021-12-02
Update github workflows to use LTS jdk versions


smlambert:
We should check what is currently being used in all of the various workflow yaml files and update them if they are found to be using versions that are no longer interesting:

For example, 16 can be updated to 17 (which is an LTS version):
https://github.com/adoptium/aqa-tests/blob/master/.github/workflows/directoriesFilesChangePR.yml#L37

nadeensami:
I'll take this one

PR #3257, opened 2022-01-21
Update git workflows to use LTS JDK versions

nadeensami:
This patch updates the directoriesFilesChangePR.yml file to use the LTS version 17 rather than non-LTS version 16.

Fixes: #3156

Signed-off-by: Nadeen Mohamed <nadeen@ualberta.ca>


smlambert (approved):
thanks @nadeensami !

llxia (approved):
LGTM

PR #3261, opened 2022-01-21
Setting customized sdk url and sdk resource as non customized ignores the sdk resource value #2097

YeeSkywalker:
Check out: Yee Lin
Date: Jan. 21st, 2022
Issue: https://github.com/adoptium/aqa-tests/issues/2097


YeeSkywalker:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3173/console

PR #3276, opened 2022-01-24
re-add Solaris support

gdams:
Solaris support seemed to disappear as part of the refactor


sxa (approved):
Looks like it should be ok

smlambert (approved):

PR #3308, opened 2022-02-02
Add sw.os.mac label for xmac and amac

AdamBrousseau:
sw.os.osx is not as correct now so we should
migrate to sw.os.mac. A future change can remove
the orignal label once all farms are updated.


AdamBrousseau:
@llxia and @ShelleyLambert looking for feedback on this, thanks.

llxia:
Do we have an issue tracking this? Will Adopitum change as well?

llxia (approved):

smlambert (approved):
I hope I am not committing to technical debt by merging this (and having not everyone evolve to the new semantically correct naming).

Issue #2841, opened 2021-09-01
Support server and client testing


llxia:
In some cases, two machines are needed for testing. For example, some performance tests will need to communicate between server and client. We need to update our test pipeline to support this case.

Proposed solution:
- create a Jenkins job `Machine-Reservation` - this is a simple job to grep user-specified machine and sleep for the user-specified time. The main idea is to reserve the machine.
- in the test pipeline, a new parameter `RELATED_NODES` should be added. If the user provides a node name in `RELATED_NODES`, the test pipeline will trigger `Machine-Reservation` to book the specific machine before starting any testing. Otherwise, proceed with testing normally. For this case, we just need to pass in a single node. This can be further extended later for multiple nodes (i.e., comma separated).
- `RELATED_NODES` is set as an env variable to pass into the test scripts
- individual job will connect to `RELATED_NODES` for testing 
- once the testing is completed, abort `Machine-Reservation` if it is still running

awsafsakif:
We also need a check to see if `CLIENT_NODE` machine is idle before triggering the reservation job. Otherwise we will run into an issue where if the client machine is already busy running a different benchmark, the reservation job will be scheduled but not triggered.

Extra variables required: TIMEOUT: How long server machine will be active in jenkins till job is cancelled, INTERVAL: How often to check if CLIENT_NODE is idle before triggering machine reservation and continuing with the benchmark run.

llxia:
Yes, for checking if a node is available https://stackoverflow.com/questions/37227562/how-to-know-whether-the-jenkins-build-executors-are-idle-or-not-in-jenkins-using/37238045

Also, we have similar code to use the API: https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/openjdk_tests#L300

smlambert:
Created https://ci.adoptopenjdk.net/job/Machine-Reservation job as a placeholder for a pipeline described in this issue.

@norbline - if you have not already done so, please rebase your branch (described in [Contributing.md](https://github.com/adoptium/aqa-tests/blob/master/Contributing.md)) in preparation to begin work on this issue.  I believe @llxia will share some code with you relating to Machine-Reservation shortly.

llxia:
TODO:

- [ ] create Jenkins job Machine-Reservation https://ci.adoptopenjdk.net/view/Test_grinder/job/Machine-Reservation/

- [ ]  checkin initial Machine-Reservation script in aqa-tests repo (Lan)

[...]

PR #3312, opened 2022-02-02
trigger Machine-Reservation job with RELATED_NODES and TIME_LIMIT

norbline:
related to:  #2841

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -601,6 +601,22 @@ def runTest( ) {
 					sh "iconv -f iso8859-1 -t ibm-1047 ${parallelList} > ${parallelList}.ebcdic; rm ${parallelList}; mv ${parallelList}.ebcdic ${parallelList}"
 				}
 			}
+
+			echo "about to check if TARGET equals IdleMicrobenchmark_HS and if params.RELATED_NODES exist"
+			if (params.TARGET == 'IdleMicrobenchmark_HS' && params.RELATED_NODES ) {
+				// assume for now params.RELATED_NODES contains a single hostname 
+				// or label (not comma-separated list for now)
+				echo "About to trigger the Machine-Reservation job"
+			 timeout(time: 45, unit: "MINUTES"){
+					build job: 'Machine-Reservation', parameters: [
+					string(name: 'NODE', value: "${params.RELATED_NODES}"),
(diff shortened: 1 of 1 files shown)


norbline:
Test plan:
like what happens if you try to
1) set RELATED_NODES to hostname of a machine that is already busy running a job. 
2) set RELATED_NODES to a hostname of a machine that is offline?  https://ci.adoptopenjdk.net/job/Machine-Reservation/9/console
3) set RELATED_NODES to a hostname that does not exist? https://ci.adoptopenjdk.net/job/Machine-Reservation/10/console
4) what happens if you launch Machine-Reservation successfully, but then manually abort the Machine-Reservation job?
5) temporarily insert an error into the machineRes.groovy script to cause the script to fail, to see how your code changes will handle a case when launching Machine-Reservation fails. https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Perforama/14/console

buildenv/jenkins/JenkinsfileBase
```diff
@@ -601,6 +601,22 @@ def runTest( ) {
...
+				echo "About to trigger the Machine-Reservation job"
+			 timeout(time: 30, unit: 'MINUTES'){
+					build job: 'Machine-Reservation', parameters: [
+					string(name: 'NODE', value: "${params.RELATED_NODES}"),
+					string(name: 'RESERVATION_LENGTH', value: "${params.TIME_LIMIT}")] 
+					propagate: true
+					wait: true	
+				}
+			}
+			echo "after launching M-R job"
```
llxia, line 618: This echo should be inside the if statement

buildenv/jenkins/JenkinsfileBase
```diff
@@ -753,6 +770,18 @@ def post(output_name) {
 	}
 }
 
+def getjobStatus(){
+	steps {
+		script {
+			def job = jenkins.model.Jenkins.instance.getItemByFullName("Machine-Reservation")
+			def result = job.getLastBuild().getResult().toString()
```
llxia, line 777: `getLastBuild` may not be the build that you triggered. The correct way is to get job info when you are triggering the job (line 611). For this PR, let's just remove this getjobStatus() for now.

llxia:
Also, we need to investigate what API to use to get the job stage.

llxia (changes requested):

llxia:
@norbline any update on this? Thanks

PR #3216, opened 2022-01-13
work around bazel tool

LizyBbethy:
draft testing how new changes run

PR #3321, opened 2022-02-05
wycheproof and scala dockerfile generation fix

### LINKED ISSUE #3320: Some external projects (wycheproof, scala) Dockefile can't be generated correctly.
Some external projects ( wycheproof) Dockefile can't be generated correctly, some contents are missing.

For example: wycheproof need bazel installed however there is no installation step when build the image by dockerfile.
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Tether/73/console

Believe some information are missing in [set_test_info() ](https://github.com/adoptium/aqa-tests/blob/master/external/common_functions.sh#L195-L213)

A quick look shows that we need to add following test info, similar way as ant_version [ant_version=$(getProperty "ant_version")](https://github.com/adoptium/aqa-tests/blob/master/external/common_functions.sh#L205)

gradle_version
sbt_version
bazel_version
openssl_version

LizyBbethy:
a few updates to the [set_test_info()](https://github.com/adoptium/aqa-tests/blob/master/external/common_functions.sh#L195-L213) function
grinder https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3496/console
Fixes #3320


external/wycheproof/test.properties
```diff
@@ -3,5 +3,5 @@ script="wycheproof.sh"
 home_path="\${WORKDIR}"
 test_results="testResults"
 tag_version="master"
-bazel_version="1.2.1"
+bazel_version="4.2.2"
```
sophia-guo, line 6: I think this change should be part of #3216 to update the bazel version, which should not be part of this PR. This PR is a general fix of parsing and setting set information.

sophia-guo:
Grinder job ` https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3496/console` failed, which is  bazel specific .

From console output I can see 

```
13:40:15       [exec] Step 9/26 : ARG BAZEL_VERSION=4.2.2
```

which means test information is correctly parsed and set and Dockerfile can be generated correctly. Please remove the changes in test.properties. All others LGTM.

sophia-guo (approved):

smlambert (approved):

PR #3330, opened 2022-02-09
Undo excludes for openjdk-build#893

sxa:
Removal of the excludes made via the issue https://github.com/adoptium/temurin-build/issues/893.

This had excluded the test case changed in 11.0.14.1 on Linux platforms, but a fix appears to have gone into 11.0.7 so we can re-enable. I've verified the individual SpecialHeaders test case on xLinux and it passes - I'm running the whole of java_net with this exclusion and expect them to pass too, so I'm submitting this PR.

I'll look at a separate PR for 17+ later, but I haven't verified those yet.


sophia-guo (approved):

sophia-guo:
Those tests are also excluded for jdk17 and jdk18 for same reason. I would suggest to re-enable them in jdk17 and jdk18

sxa:
> Never mind. I saw that you will do it in a separate PR.

Yep I was going to include those too but felt it was best to verify that it really was safe to enable them first, so intend do it separately post-release.

llxia (approved):
LGTM

Issue #3325, opened 2022-02-07
Enhance the archive***File() with filename parameter


sophia-guo:
With the requirements expand we have three functions to archive files.
archiveSHAFile()
archiveTestenvFile()
archiveAQACertFile()

The only difference of those function is the specific file we need to archive, which can be done by adding one function parameter. Like:

archiveFile(fileName) {
}

This way to reduce the redundant and be easy to maintain and expand.

Stergios13:
Hello. I can work on this if it is possible, please. Happy to help on this issue

smlambert:
Definitely @Stergios13 - note that before we can merge any PR from you, you will need to create an Eclipse profile and sign the ECA (details [here](https://github.com/adoptium/documentation/blob/main/documentation-documentation/eca-sign-off.adoc)).

Stergios13:
Sure i will. Can you please navigate me to the file or files containing those three functions? i have trouble finding them. Thank you for assigning me!

smlambert:
https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L688-L690
https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L881-L912

sophia-guo:
Can be closed by #3333 and #3394

PR #3333, opened 2022-02-09
A solution in the issue #3325

Stergios13:
To begin with, i created a single function called archiveFile(filename, filePath) that takes 2 arguments. The first one declares the name of the file that has to be saved in jerkins and the second one declares the whole path of the file. In the post function (especially in line 690) i created a hash map that each key declares the filename and each value of the key declares the whole path of it. The function is called in a for-each loop after the hash map declaration.

PS1. i am not 100% sure if that works, but if it doesn't i'll be happy to contribute again
PS2: thanks for assigning me for this issue, i really wanted to contribute as it is my first open source contribution!

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -698,7 +698,6 @@ def post(output_name) {
 				archiveFile("$entry.key", "$entry.value")
 			}
 
-
 			if (env.BUILD_LIST.startsWith('jck')) {
 				xunit (
 						tools: [Custom(customXSL: "$WORKSPACE/aqa-tests/jck/xUnit.xsl",
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -687,16 +687,15 @@ def post(output_name) {
 			junit allowEmptyResults: true, keepLongStdio: true, testResults: '**/work/**/*.jtr.xml, **/junitreports/**/*.xml, **/external_test_reports/**/*.xml'
 
 
-			//defining a hashMap with filenames->keys and filePaths->values
-			def map = [
-					[filename: "testenv.properties", filePath: "aqa-tests/testenv/testenv.properties"],
-					[filename: "SHA", filePath: "aqa-tests/TKG/SHA.txt"],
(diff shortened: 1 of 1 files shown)


llxia:
Thanks @Stergios13 for working on this issue. Please see https://github.com/adoptium/aqa-tests/blob/master/Contributing.md for commit guidelines.

Stergios13:
I changed the file again listening to your tips. 
Additionally, i do not format the whole file, i just change the lines that need to be changed and yet it seems like i format the whole file.

sophia-guo:
Running good https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3586/

buildenv/jenkins/JenkinsfileBase
```diff
@@ -685,18 +685,26 @@ def post(output_name) {
...
+
+            //Defining the files to be archived
+            def testenvFile = "aqa-tests/testenv/testenv.properties"
+            def shaFile = "aqa-tests/TKG/SHA.txt"
+            def AQACertFile = "aqa-tests/TKG/AQACert.log"
+
+            //call the archive function for each file
+            archiveFile(testenvFile)
+	    archiveFile(shaFile)
+            archiveFile(AQACertFile)
```
sophia-guo, line 698: As those filename is only used once may not need to define the variable. 

```
            archiveFile("aqa-tests/testenv/testenv.properties")
	    archiveFile("aqa-tests/TKG/SHA.txt")
            archiveFile("aqa-tests/TKG/AQACert.log")
```

smlambert: @Stergios13 - can you update your PR based on @sophia-guo's suggestion (pass the filenames directly, as we do not need to separately define variables for our purpose.

Stergios13: Sorry for being inactive i was busy due to university purposes. Of course i will during the day!

smlambert:
FYI @llxia @sophia-guo - if @Stergios13 can update this quickly we can wait, otherwise suggest merging and having a followup PR to clean up.

#3394 is somewhat related.

sophia-guo (approved):
Minor issue could be updated later.

smlambert (approved):

PR #3344, opened 2022-02-11
jck: remove !headful tag on Alpine

gdams:
It seems that `set jck.env.testPlatform.headless Yes;` is all that is required


karianna (approved):
LGTM

calderast:
Is `set jck.env.testPlatform.headless Yes;` included elsewhere as an alternative to the `!headful` keyword, or are we not indicating headless on jdk11 and jdk8? This change looks good, but graphical tests will still fail on Alpine if we don't indicate headless somewhere

gdams:
Yeah we set that as well!

smlambert:
How did this get merged with only 1 reviewer?  

I had been holding off my review to watch the runs done off the branch over the weekend.

PR #3352, opened 2022-02-14
Use JAVA_TOOL_OPTIONS to specify overwritable JVM options

babsingh:
Setting JVM_OPTIONS overwrites the modes.

EXTRA_OPTIONS will always append to the JVM options for the modes. If the
modes already specify -Djava.security.manager=*, then it will be overwritten
by EXTRA_OPTIONS.

JAVA_TOOL_OPTIONS works with both OpenJ9 and Hotspot. It prepends JVM options
at startup so it will not overwrite the JVM options for the modes.

Leading tabs have the potential to cause elements of makefiles to be
considered part of a recipe instead of part of the global structure: use
spaces instead. For consistency, 2 spaces are used in this file for a
tab.

Fixes: https://github.com/eclipse-openj9/openj9/issues/14495
Related: https://github.com/eclipse-openj9/openj9/pull/14449


babsingh:
`Hide whitespace` will make it easier to review changes.

babsingh:
re https://github.com/adoptium/aqa-tests/pull/3352#issuecomment-1039264121:

The above builds is not picking my changes from this PR even after `ADOPTOPENJDK_REPO` was set to pick up my changes and `LIGHT_WEIGHT_CHECKOUT` was set to false.

babsingh:
Also tried the grinder:
- JDK8: https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/20983/
- JDK18: https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/20984/

The JDK8 grinder correctly picks up changes from my fork. But, the JDK18 grinder still does not get changes from my fork.

babsingh:
@Mesbah-Alam @llxia Can you please review this PR?

Mesbah-Alam (approved):
LGTM

pshipton:
@llxia is it possible to get this merged before weekend testing?

llxia:
@babsingh Could we echo JAVA_TOOL_OPTIONS value in makefile?

llxia:
I think you can use `@echo` 
https://www.gnu.org/software/make/manual/html_node/Echoing.html
https://github.com/adoptium/TKG/blob/master/settings.mk#L230

Anyway, I think `warning ` is ok in this case.

llxia (approved):

babsingh:
Thanks for merging. re https://github.com/adoptium/aqa-tests/pull/3352#issuecomment-1045037369:

In https://github.com/adoptium/aqa-tests/pull/3352#issuecomment-1044988870, I had tried `@echo`. It resulted in the below error:
```
@echo "Environment variable JAVA_TOOL_OPTIONS is set to $(JAVA_TOOL_OPTIONS)"
  
# Error
../systemtest.mk:25: CYGWIN is 0
../systemtest.mk:60: *** missing separator.  Stop.
```
As per the GNU ref (https://www.gnu.org/software/make/manual/html_node/Echoing.html), `@echo` only works within the scope of a recipe. Example:
```
recipe:
    @echo "Message"
```
In the suggested example (https://github.com/adoptium/TKG/blob/master/settings.mk#L230),
```
# Only the variable is assigned a value. The @echo command is not invoked.
TEST_TEARDOWN=@echo "Nothing to be done for teardown."

# autoGen.mk has a recipe which invokes $(TEST_TEARDOWN). Example:
DaaLoadTest_daa2_special_5m_19:
    ...
    ...
    $(TEST_TEARDOWN); # @echo gets invoked from within the recipe.
```

PR #3369, opened 2022-02-23
Update compile.sh permission

llxia:
compile.sh does not have correct permission to run locally. Change the permission to 755.


smlambert (approved):

llxia:
We do not have issue with Jenkins. Permission change is for the local case.

keithc-ca:
My suggestion is that the documentation should advise use of `bash compile.sh ...` instead of modifying permissions.

llxia:
Currently, users can directly run `get.sh` (without bash...). IMO, I think we should keep the user experience to be the same.

keithc-ca:
I agree there should be consistency, but we aren't even close right now. There are 39 *.sh files with execute permission and 44 without:
```
$ find aqa-tests -type f -name \*.sh -perm -100 | wc -l
39
$ find aqa-tests -type f -name \*.sh ! -perm -100 | wc -l
44
```
Having execute permission isn't always enough (on some platforms `#!/bin/bash` or even `#!/usr/bin/env bash` won't work). In my opinion, we should be advising people to be explicit using `bash ...`. Adding execute permission is a step in the wrong direction.

smlambert:
The opinion in https://github.com/adoptium/aqa-tests/pull/3369#issuecomment-1049925095 is noted however I agree with @llxia that we have 2 scripts that we expect users to execute directly (for local execution).

I am happy to have compile.sh and get.sh have execute permissions to have a consistent user experience (and to not have to change a diverse set of documentation that already exists across a variety of repositories).

keithc-ca:
I'm happy to be heard, but, of course, committers will make the final decision.

For my own education: where is the use of `compile.sh` documented?

smlambert:
re: https://github.com/adoptium/aqa-tests/pull/3369#issuecomment-1049985073 In instructions & presentations provided to vendors who wish to list their binaries in the Adoptium marketplace.

keithc-ca:
Any links to those instructions & presentations?

smlambert:
Some of the places I can think of (not an exhaustive list & some that are not easily revokable/updatable):

PDFs shared with Adoptium Working Group members (shared in the Adoptium WG slack channel, not publicly reachable), presentations to different companies who will be listing in the marketplace (done internally and sent via email to various member companies), [AQAvit verification doc](https://github.com/adoptium/website-v2/blob/main/src/asciidoc-pages/docs/aqavit-verification.adoc), recorded webinars talking about AQAvit verification basic steps (found from AQAvit bi-weekly meeting recordings and linked from the [supplemental info](https://www.youtube.com/watch?v=1EUi3iTZSzg) of the verification document.

Is there a point where this enquiry gets dropped?   it seems like it's become an obsession.

keithc-ca:
I'm satisfied that the guidance in aqavit-verification.adoc supports this change. I still think it's bad practice, but I'll drop it.

sophia-guo (approved):

PR #3371, opened 2022-02-23
Moving Java detection logic to common function

### LINKED ISSUE #3370: Duplicate java detection logic is in each third party application test script file
For each third party application test we'd like to detect and set java environment. Currently the logic is duplicate in all test script files.

```
	if [ -d /java/jre/bin ];then
		echo "Using mounted Java8"
		export JAVA_BIN=/java/jre/bin
		...
		echo "Using default openjdk image Java"
		java_path=$(type -p java)
		suffix="/java"
		java_root=${java_path%$suffix}
		export JAVA_BIN="$java_root"
		export JAVA_HOME="${java_root%/bin}"
	fi
```

Better to move the logic to existing common functions.

sophia-guo:
Fix #3370 

### DIFF
--- a/external/camel/dockerfile/camel-test.sh
+++ b/external/camel/dockerfile/camel-test.sh
@@ -13,29 +13,7 @@
 # limitations under the License.
 
 source $(dirname "$0")/test_base_functions.sh
-
-# Set up Java to be used by the the camel-test
-
-if [ -d /java/jre/bin ];then
-	echo "Using mounted Java8"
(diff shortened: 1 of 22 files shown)


sophia-guo:
Grinder jacoco https://ci.adoptopenjdk.net/view/work-in-progress/job/grinder_sandbox_new/434/

llxia (approved):
LGTM

smlambert (approved):
thanks @sophia-guo !

PR #3375, opened 2022-02-24
AArch64 macOS: Exclude some MathLoadTest tests

knn-k:
This commit excludes the following tests on AArch64 macOS for the time
being:

- MathLoadTest_all_special_5m
- MathLoadTest_bigdecimal_special_5m

### DIFF
--- a/system/mathLoadTest/playlist.xml
+++ b/system/mathLoadTest/playlist.xml
@@ -129,6 +129,10 @@
 				<comment>rtc 139897</comment>
 				<variation>Mode554</variation>
 			</disable>
+			<disable>
+				<comment>https://github.com/eclipse-openj9/openj9/issues/14590</comment>
+				<platform>aarch64_mac.*</platform>
+			</disable>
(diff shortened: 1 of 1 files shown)


system/mathLoadTest/playlist.xml
```diff
@@ -129,6 +129,10 @@
 				<comment>rtc 139897</comment>
 				<variation>Mode554</variation>
 			</disable>
+			<disable>
+				<comment>https://github.com/eclipse-openj9/openj9/issues/14590</comment>
+				<platform>aarch64_mac.*</platform>
```
llxia, line 134: We can use `<platform>aarch64_mac</platform>` without `.*`

knn-k: No, I need `.*` here because we only have `aarch64_mac_xl` build at the moment.

llxia (approved):
LGTM

smlambert (approved):

PR #3386, opened 2022-02-28
Exclude javax_net and javax_rmi from z/OS JDK 11

Mesbah-Alam:
- Exclude javax_net and javax_rmi from z/OS JDK 11
- backlog/issues/744

### DIFF
--- a/jck/runtime.api/playlist.xml
+++ b/jck/runtime.api/playlist.xml
@@ -931,6 +931,12 @@
 				<impl>ibm</impl>
 				<platform>.*zos.*</platform>
 			</disable>
+			<disable>
+				<comment>Disabled on z/OS due to backlog/issues/744</comment>
+				<version>11</version>
+				<impl>ibm</impl>
+				<platform>s390x_zos</platform>
+			</disable>
 		</disables>
 		<variations>
 			<variation>NoOptions</variation>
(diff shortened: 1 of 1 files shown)


jck/runtime.api/playlist.xml
```diff
@@ -931,6 +931,12 @@
 				<impl>ibm</impl>
 				<platform>.*zos.*</platform>
 			</disable>
+			<disable>
+				<comment>Disabled on z/OS due to backlog/issues/744</comment>
+				<version>11</version>
+				<impl>ibm</impl>
+				<platform>.*zos.*</platform>
```
sophia-guo, line 938: Maybe specify s390x_zos and s390_zos?

Mesbah-Alam: Could you please state exactly what string value you are suggesting for `<platform>` here?  I thought `.*zos.*` was clear enough - it includes all z/OS platforms.

Mesbah-Alam: Actually, since the only z/OS platform on which we run tck 11 is s390x, having `s390x_zos` for `<platform>` should suffice. I've updated the PR.

llxia (approved):
LGTM

smlambert (approved):
LGTM

PR #1706, opened 2020-03-31
Handle both debug-image and debugimage

llxia:
Due to inconsistent naming, handling both debug-image and debugimage
until the issue is fixed.

Issue: https://github.com/eclipse/openj9/issues/9028 and
https://github.com/AdoptOpenJDK/openjdk-tests/issues/1696

Signed-off-by: lanxia <lan_xia@ca.ibm.com>

PR #1692, opened 2020-03-24
Auto download test-images and debug-images

llxia:
- add auto download test-images and debug-images
- to simplify the story, auto download will only be triggered if users
provide only one CUSTOMIZED_SDK_URL and URL filename contains
OpenJ9-JDK. In this way, if users want to provide multiple separate URLs
or multiple separate URLs from separate servers, they can still do so
and auto-download will not be triggered.
- Once auto-download is triggered, by default, the flag
test_images_required and debug_images_required are set to true in
get.sh. If the flag is set to true, it means the script will retry if
the download fails and the program will fail if all retry attempts fail
(the same behavior as downloading SDK). If the flag is set to false, it
means the script will not try to auto download.
- In JenkinsfileBase, pass in test_images_required and
debug_images_required to get.sh if users want to overwrite the flag(s)
- During unzipping process, if debug-images jar is found, overlay
debug-images on top of sdk j2sdk-image/jre dir if it exists. Otherwise,
extracts into j2sdk-image dir

Issue: eclipse/openj9#8874

Signed-off-by: lanxia <lan_xia@ca.ibm.com>

PR #3388, opened 2022-03-01
Always move debug image to end of array

AdamBrousseau:
Also
Check for other spelling of debugimage
Related https://github.com/adoptium/aqa-tests/pull/1706

Related backlog#745

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -316,23 +316,21 @@ getBinaryOpenjdk()
 	# if $jar_file_array contains debug-image, move debug-image element to the end of the array
 	# debug image jar needs to be extracted after jdk as debug image jar extraction location depends on jdk structure
 	# debug image jar extracts into j2sdk-image/jre dir if it exists. Otherwise, extracts into j2sdk-image dir
-	if [ $DEBUG_IMAGES_REQUIRED = true ]; then
-		last_index=$(( ${#jar_file_array[@]} - 1 ))
-		for i in "${!jar_file_array[@]}"; do
-			if [[ "${jar_file_array[$i]}" =~ "debug-image" ]] || [[ "${jar_file_array[$i]}" =~ "debugimage" ]]; then
-				if [ "$i" -ne "$last_index" ]; then
-					debug_image_jar="${jar_file_array[$i]}"
+	last_index=$(( ${#jar_file_array[@]} - 1 ))
+	for i in "${!jar_file_array[@]}"; do
+		if [[ "${jar_file_array[$i]}" =~ "debug-image" ]] || [[ "${jar_file_array[$i]}" =~ "debugimage" ]]; then
+			if [ "$i" -ne "$last_index" ]; then
+				debug_image_jar="${jar_file_array[$i]}"
 
-					# remove the element
-					unset jar_file_array[$i]
+				# remove the element
+				unset jar_file_array[$i]
 
-					# add $debug_image_jar to the end of the array
-					jar_file_array=( "${jar_file_array[@]}" "${debug_image_jar}" )
-					break
-				fi
+				# add $debug_image_jar to the end of the array
+				jar_file_array=( "${jar_file_array[@]}" "${debug_image_jar}" )
+				break
 			fi
-		done
-	fi
+		fi
+	done
 
 	for jar_name in "${jar_file_array[@]}"
 		do
@@ -400,10 +398,10 @@ getBinaryOpenjdk()
 
 checkURL() {
 	local filename="$1"
-	if [[ $filename =~ "test-image" ]] && [ $TEST_IMAGES_REQUIRED = false ]; then
-		required=false
-	elif [[ $filename =~ "debug-image" ]] || [[ "$jar_name" =~ "debugimage" ]] && [ $DEBUG_IMAGES_REQUIRED = false ]; then
-		required=false
+	if [[ $filename =~ "test-image" ]]; then
+		required=$TEST_IMAGES_REQUIRED
+	elif [[ $filename =~ "debug-image" ]] || [[ "$jar_name" =~ "debugimage" ]]; then
+		required=$DEBUG_IMAGES_REQUIRED
 	fi
 }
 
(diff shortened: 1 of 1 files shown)


AdamBrousseau:
@llxia I'm not sure if the other spelling is actually required in that if condition. But the fix is to remove the if condition on the rearrange logic.

get.sh
```diff
@@ -402,7 +400,7 @@ checkURL() {
 	local filename="$1"
 	if [[ $filename =~ "test-image" ]] && [ $TEST_IMAGES_REQUIRED = false ]; then
 		required=false
-	elif [[ $filename =~ "debug-image" ]] && [ $DEBUG_IMAGES_REQUIRED = false ]; then
+	elif [[ $filename =~ "debug-image" ]] || [[ "$jar_name" =~ "debugimage" ]] && [ $DEBUG_IMAGES_REQUIRED = false ]; then
```
AdamBrousseau, line 403: @keithc-ca is this ok or do I need another set of brackets around the first 2 conditions? It seems to work either way I tried it.

keithc-ca: I'm not a fan of using the global variable for communicating the answer; using `return` would be a cleaner solution.
Even without going there, this function could be improved:
```
checkURL() {
	local filename="$1"
	if [[ $filename =~ "test-image" ]]; then
		required=$TEST_IMAGES_REQUIRED
	elif [[ $filename =~ "debug-image" ]] || [[ "$jar_name" =~ "debugimage" ]]; then
		required=$DEBUG_IMAGES_REQUIRED
	else
		required=true
	fi
}
```
This might also be a good place to use `case`.

llxia:
This is by design that DEBUG_IMAGES_REQUIRED will not take effect if CUSTOMIZED_SDK_URL contains debug-image.

```
to simplify the story, auto download will only be triggered if users
provide only one CUSTOMIZED_SDK_URL and URL filename contains
OpenJ9-JDK. In this way, if users want to provide multiple separate URLs
or multiple separate URLs from separate servers, they can still do so
and auto-download will not be triggered.
```
Please see https://github.com/adoptium/aqa-tests/pull/1692

May I know the purpose here? If we just want to skip downloading debug-image, we can just simply not set debug-image link in CUSTOMIZED_SDK_URL.

smlambert:
I have no visibility of backlog#745 but would be actively pushing for standardizing how build pipelines at various servers are doing instead of making more compromises in this repo / script.

AdamBrousseau:
Let me summarize the failure and the fix.
CUSTOMIZED_SDK_URL includes a debug image and DEBUG_IMAGES_REQUIRED is set to false.
All urls in the param are downloaded. Debug image is untared first because we do not rearrange the debug-image to the end of `jar_file_array`. Untar fails because debug-image untar needs to happen after jdk untar. The fix is to always try to rearrange `jar_file_array` to avoid the failure. Another option would be do not download the debug image since DEBUG_IMAGES_REQUIRED is set to false but I think that bool is more about auto-download. Since the user has passed the url, I imagine we still want to download it.

```
00:00:17.970  get jdk binary...
00:00:17.970  _ENCODE_FILE_NEW=UNTAGGED curl -OLJSks --user ****:**** https://na.artifactory.swg-devops.com/artifactory/sys-rt-generic-local/hyc-runtimes-jenkins.swg-devops.com/build-scripts/jobs/jdk8u/jdk8u-linux-x64-openj9/327/ibm-semeru-open-debugimage_x64_linux_JDK8U_2022-02-25-03-30.tar.gz
00:00:35.197  _ENCODE_FILE_NEW=UNTAGGED curl -OLJSks --user ****:**** https://na.artifactory.swg-devops.com/artifactory/sys-rt-generic-local/hyc-runtimes-jenkins.swg-devops.com/build-scripts/jobs/jdk8u/jdk8u-linux-x64-openj9/327/ibm-semeru-open-jre_x64_linux_JDK8U_2022-02-25-03-30.tar.gz
00:00:36.072  _ENCODE_FILE_NEW=UNTAGGED curl -OLJSks --user ****:**** https://na.artifactory.swg-devops.com/artifactory/sys-rt-generic-local/hyc-runtimes-jenkins.swg-devops.com/build-scripts/jobs/jdk8u/jdk8u-linux-x64-openj9/327/ibm-semeru-open-jdk_x64_linux_JDK8U_2022-02-25-03-30.tar.gz
00:00:42.666  _ENCODE_FILE_NEW=UNTAGGED curl -OLJSks --user ****:**** https://na.artifactory.swg-devops.com/artifactory/sys-rt-generic-local/hyc-runtimes-jenkins.swg-devops.com/build-scripts/jobs/jdk8u/jdk8u-linux-x64-openj9/327/ibm-semeru-open-testimage_x64_linux_JDK8U_2022-02-25-03-30.tar.gz
00:00:43.070  Uncompressing ibm-semeru-open-debugimage_x64_linux_JDK8U_2022-02-25-03-30.tar.gz over ./j2sdk-image...
00:00:43.070  tar (GNU tar) 1.30
00:00:43.070  License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
00:00:43.070  tar: j2sdk-image: Cannot open: No such file or directory
00:00:43.070  tar: Error is not recoverable: exiting now
```

AdamBrousseau:
Perhaps the user shouldn't be supplying the debug-image in the url. In this case it's the grinder rerun on a test job that was initiated by an Adopt (semeru) pipeline. Is the debug image needed by test builds? Perhaps we shouldn't be passing it. Looks like Only the Semeru side passes CUSTOMIZED_SDK_URL. Temurin just uses upstream build.

https://github.com/ibmruntimes/ci-jenkins-pipelines/blob/fc3c87de3c489600c9994cd8243c95c55d9b59b0/pipelines/build/common/openjdk_build_pipeline.groovy#L386

smlambert:
Test builds do not use debug images directly as far as I am aware.  Some tests do need the native test images, but that does not sound like it's causing an issue for you.

AdamBrousseau:
OK. I think i will make a change on the build side to not pass the debug image url. I still think this change makes the test script more robust though.

llxia (approved):
LGTM

smlambert (approved):
👍

Issue #3382, opened 2022-02-25
quarkus_quickstarts_test_0 failed with jdk17


sophia-guo:
https://ci.adoptopenjdk.net/view/work-in-progress/job/grinder_sandbox_new/442/console
quarkus_quickstarts_test_0 failed with jdk17.

```
20:00:35  /quarkus-quickstarts
20:00:35  Compile and run quarkus_quickstarts tests
20:00:35  Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8 -Djava.security.egd=file:/dev/./urandom
20:00:38  [[1;31mERROR[m] Error executing Maven.
20:00:38  [[1;31mERROR[m] java.lang.IllegalStateException: Unable to load cache item
20:00:38  [[1;31mERROR[m] Caused by: Unable to load cache item
20:00:38  [[1;31mERROR[m] Caused by: Could not initialize class com.google.inject.internal.cglib.core.$MethodWrapper
20:00:38  Build quarkus_quickstarts completed
```

smlambert:
Suggest an update of quarkus version first, since we take a very old release at present (we are pulling [1.3.2.Final](https://github.com/adoptium/aqa-tests/blob/master/external/quarkus_quickstarts/test.properties#L11), and the latest tag looks to be [2.7.2.Final](https://github.com/quarkusio/quarkus-quickstarts/releases/tag/2.7.2.Final)).

amir-waheed:
I can work on this issue.

sophia-guo:
related with https://github.com/adoptium/aqa-tests/issues/3450

sophia-guo:
Same failure as https://github.com/adoptium/aqa-tests/issues/3390Looks like this mvn version doesn't work with jdk17.

PR #3457, opened 2022-03-13
Changing tag version for quarkus tests

amir-waheed:
This patch updates quarkus version from 1.3.2 to 2.7.2.

Fixes: #3382 

### DIFF
--- a/external/camel/test.properties
+++ b/external/camel/test.properties
@@ -2,6 +2,6 @@
  script="camel-test.sh"
  test_command="mvn clean install"
  test_results="testResults"
- tag_version="main"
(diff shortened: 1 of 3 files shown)


sophia-guo (approved):

sophia-guo:
I've updated the description to be _related_ as the updated won't fix the #3382. Though the update is still necessary as we can get the latest test material.

https://ci.adoptopenjdk.net/view/work-in-progress/job/grinder_sandbox_new/516/

smlambert (approved):
thanks @amir-waheed !

PR #3459, opened 2022-03-13
Updating triggerGroovy to trigger the scripts

### LINKED ISSUE #3445: Update AutoReenableJob
- Update AutoReenableJob to use the latest python script in https://github.com/adoptium/aqa-tests/tree/master/scripts/disabled_tests
- Consume git token. For example:
https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L485
https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L494

https://ci.adoptopenjdk.net/view/Test_grinder/job/AutoReenableJob/
https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/triggerGrinder.groovy

amir-waheed:
Updating triggerGroovy to trigger the python scripts and consuming git token

Fixes #3445 


amir-waheed:
I have added the functionality to consume git token however to be able to test this, credentials need to be added in Jenkins and we would also need to add a parameter in AutoReeanbleJob. (I have assumed the name of the parameter to be AQA_ISSUE_TRACKER_CREDENTIAL_ID)

smlambert:
I've updated the AutoReenableJob to have the credentials (and set the default one to be the credentials for our [eclipse-aqavit-bot](https://github.com/eclipse-aqavit-bot) functional ID).  @amir-waheed - you should be able to test this PR now (see [AutoReenableJob/87](https://ci.adoptopenjdk.net/view/Test_grinder/job/AutoReenableJob/87) for the latest run, shows some updates to script are needed).

amir-waheed:
@smlambert I was successfully able to trigger the first python script 'exclude_parser.py' however the script runs into a 'ModuleNotFound' [error](https://ci.adoptopenjdk.net/view/Test_grinder/job/AutoReenableJob/94/console) for module dataclasses.

smlambert:
Hmm, is that because we need to use a certain version of python?  We might want to pin this job to run on a certain set of machines if so.

amir-waheed:
Yes, the python scripts need a version of 3.8+ to run.

smlambert:
I've added a LABEL parameter to the [job](https://ci.adoptopenjdk.net/view/Test_grinder/job/AutoReenableJob/), so that we can run this job on a machine with a specific label (you can update the [groovy script](https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/triggerGrinder.groovy#L5) to run on a certain node(LABEL) like is done [elsewhere](https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/openjdk_tests#L186) in other groovy scripts).

Now we just need to find some machines with that version of python (or create an infra request for one to be updated).

amir-waheed:
@smlambert Thank you for the suggestions. 
One of the python scripts was able to run after adding the label parameter. I was also able to successfully grab the "eclipse_aqavit_bot_token", however I wasn't sure what the git user id would be so for now I just went with 'eclipse_aqavit'.
There is still an issue as seen in [here](https://ci.adoptopenjdk.net/view/Test_grinder/job/AutoReenableJob/110/console), the second script that was supposed to be run ran into 'ModuleNotFoundError' error. I think this is also due to python version requirement, I initially tried different labels and 'ci.role.test&&hw.arch.x86&&sw.os.linux' was the one that was able to successfully run the first script.

smlambert:
@amir-waheed - https://github.com/eclipse-aqavit-bot is the git id of the bot

@gervaisj - are the python scripts using modules that are not available in a standard python installation?  and is this the list, https://github.com/adoptium/aqa-tests/blob/master/scripts/disabled_tests/requirements.txt ?  

If we are depending on things that are not in the standard install, the scripts are less portable and it is suboptimal.  We will need to make infrastructure request to get dependencies on certain machines and carry that overhead/maintenance.

gervaisj:
@smlambert sorry, just saw the comment now

replying to https://github.com/adoptium/aqa-tests/pull/3459#issuecomment-1073834433

> are the python scripts using modules that are not available in a standard python installation? and is this the list, https://github.com/adoptium/aqa-tests/blob/master/scripts/disabled_tests/requirements.txt ?

That's right, `issue_status.py` needs `requests` and `playlist_parser.py` needs `lxml`

> If we are depending on things that are not in the standard install, the scripts are less portable and it is suboptimal. We will need to make infrastructure request to get dependencies on certain machines and carry that overhead/maintenance.

I can open a separate issue for that, but could we briefly discuss it first? Either on wednesday's meeting or (my preference) before that or later, whichever is more convenient! (I know there's a release going on, so no pressure)

smlambert:
@gervaisj - @llxia and I discussed several options to address this... instead of changing the scripts, we can decouple the creation of the json file, move it and run it in an environment that we can control and install the necessary prereqs on, either  in a container or on a github runner via a workflow.  That produces the JSON file to a known location which could then be referenced by the Jenkins script (which can trigger Grinder jobs on all platforms).

gervaisj:
@smlambert

replying to https://github.com/adoptium/aqa-tests/pull/3459#issuecomment-1080066434

If we go the container route, I would add a Dockerfile to the scripts folder, then we would run the image in a Jenkins node that has docker, correct? (the image can either be pre-built or built on demand in the groovy script). I guess the advantage of this method is if we need more dependencies in the future, or we want to update python version, everything is in the same place.

If we go the github action route, we don't need a Dockerfile really (although we could still have one if we wanted). That would work, the only thing is, I can't see where the "known location" would be. Did you have something in mind?

smlambert:
Dockerfile may be the best option, out of the 2 options, but to answer your question, one can archive artifacts on an action run, that could then be accessed later.  

So, one can query to [get latest action run](https://github.community/t/get-the-latest-workflow-run-from-github-actions-using-the-octokit-api/137908) and then query to get the zip file of artifacts (see an artifact example at the bottom of this [action summary page](https://github.com/smlambert/aqafer/actions/runs/1930458434)).

amir-waheed:
@smlambert I was able to run the pipeline using a label that @gervaisj used for testing, and it seems to be working ok. Here is the [link](https://ci.adoptopenjdk.net/view/Test_grinder/job/AutoReenableJob/150/console) to the test I ran on Jenkins. I will create an issue regarding the DockerFile and start working on it, if that's ok.

gervaisj:
Knowing you can archive files from a github action, I really like that solution now. It would provide more convenient access to log files, do "dry runs" on the repo to detect errors, and we could also trigger the script to generate the JSON only on certain repo events, like a PR being merged in.

And we can keep the benefit of a dockerfile (build instructions, dependencies, python version) by including one in the repo anyway, and using it in the github action.

smlambert:
Thanks @amir-waheed @gervaisj - we can have a design discussion on Wednesday to discuss both options and how it looks in the overall flow of the feature, but in either case, let's presume we'll create a Dockerfile with all of the dependencies/prereqs we need (as this also serves as living documentation).  

Even in a github workflow, we may deploy in a Docker container (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#docker-container-filesystem).

Issue #3345, opened 2021-06-30
Parse playlist.xml and generate json file


llxia:
As part of [Create a Disabled tests pipeline](https://github.com/adoptium/aqa-tests/issues/1853), we need to parse playlist.xml in multiple repos and generate json file:

- repo:
https://github.com/adoptium/aqa-tests
https://github.com/eclipse-openj9/openj9
- parse playlist and generate json file. For example,
   - input: https://github.com/adoptium/aqa-tests/blob/c81d573e3ee749b457ec10e7e4aaec69f513e34c/openjdk/playlist.xml#L153-L160
   - output:
 

 ``` 
[{
"JDK_VERSION" : all,
"JDK_IMPL" : "openj9",
"TARGET" : "jvm_native_sanity",
"GIT_ISSUE" : "https://github.com/eclipse-openj9/openj9/issues/10757",
"PLATFORM" : "all"
},{
"JDK_VERSION" : all,
"JDK_IMPL" : "ibm",
"TARGET" : "jvm_native_sanity",
"GIT_ISSUE" : "https://github.com/eclipse-openj9/openj9/issues/10757",
"PLATFORM" : "all"
}]
```

PR #3460, opened 2022-03-13
Parse playlist.xml and generate json file

gervaisj:
Parse playlist.xml and generate json file

- Move common code
- New dependency is needed: `lxml` (enables getting the source line number)
- Takes stdin as the input and stdout as the default output stream to enable command-line usage shown in the [README](https://github.com/gervaisj/aqa-tests/blob/issue-3345/scripts/disabled_tests/README.md)
- Info/debug logging modes

Fixes https://github.com/adoptium/aqa-tests/issues/3345

Signed-off-by: Jocelyn Gervais [gervaisj14@outlook.com](mailto:gervaisj14@outlook.com)

### Output samples

https://github.com/gervaisj/aqa-tests/blob/issue-3345/scripts/disabled_tests/tests/data/playlist_out.json#L1-L34

- On missing data:
```
ERROR - tests/data/playlist_input.xml:3 : test node has no 'testCaseName' child; skipping node
ERROR - tests/data/playlist_input.xml:15 : disable node has no 'comment' child; skipping node
ERROR - tests/data/playlist_input.xml:19 : could not find 'misspelled-variation' in defined variations; skipping node
```
- Info (`-v`) mode
```
...
INFO - Processed './jck/devtools.xml_schema/playlist.xml' : n_disables=0
INFO - Processed './jck/runtime.api/playlist.xml' : n_disables=65
INFO - Processed './jck/runtime.lang/playlist.xml' : n_disables=0
INFO - Processed './jck/runtime.vm/playlist.xml' : n_disables=8
INFO - Processed './jck/runtime.xml_schema/playlist.xml' : n_disables=8
INFO - Processed './openjdk/playlist.xml' : n_disables=52
INFO - Processed './perf/bumbleBench/playlist.xml' : n_disables=3
...
```
- Debug (`-vv`) mode
```
DEBUG - Taking file list from <stdin>
DEBUG - Processing 'tests/data/playlist.xml'...
ERROR - tests/data/playlist.xml:12 : could not find 'misspelled-variation' in defined variations, skipping
INFO - Processed 'tests/data/playlist.xml' : n_disables=4
DEBUG - Outputting JSON to <stdout>
```


gervaisj:
@smlambert @llxia ready for review!

Just making sure, is this [the intended output](https://github.com/gervaisj/aqa-tests/blob/issue-3345/scripts/disabled_tests/tests/data/playlist_out.json) when receiving [this input](https://github.com/gervaisj/aqa-tests/blob/issue-3345/scripts/disabled_tests/tests/data/playlist_input.xml)?
Some nodes are misspelled or misplaced to show that we query for child nodes at any depth and the parser still works.
Some nodes are also ignored because required info is missing:
```
ERROR - tests/data/playlist_input.xml:3 : test node has no 'testCaseName' child; skipping node
ERROR - tests/data/playlist_input.xml:15 : disable node has no 'comment' child; skipping node
ERROR - tests/data/playlist_input.xml:19 : could not find 'misspelled-variation' in defined variations; skipping node
```

renfeiw:
> ERROR - tests/data/playlist_input.xml:3 : test node has no 'testCaseName' child; skipping node
> ERROR - tests/data/playlist_input.xml:15 : disable node has no 'comment' child; skipping node

For the above two error checks, I am not sure if we need them as TKG already does the validation on them. All the playlists in the code base should not have these errors. Any suggestions? @ShelleyLambert @llxia

renfeiw:
> ERROR - tests/data/playlist_input.xml:3 : test node has no 'testCaseName' child; skipping node
> ERROR - tests/data/playlist_input.xml:15 : disable node has no 'comment' child; skipping node

For the above two error checks, I am not sure if we need them as TKG already does the validation on them. All the playlists in the code base should not have these errors. Any suggestions? @ShelleyLambert @llxia

scripts/disabled_tests/tests/test_playlist_parser.py
```diff
@@ -0,0 +1,18 @@
+import json
+import os
+from unittest import TestCase
+
+from playlist_parser import parse_all_files
```
renfeiw, line 5: Not sure if I missed anything, when I run python scripts/disabled_tests/tests/test_playlist_parser.py, it gave error: 
```
Traceback (most recent call last):
  File "scripts/disabled_tests/tests/test_playlist_parser.py", line 5, in <module>
    from playlist_parser import parse_all_files
ModuleNotFoundError: No module named 'playlist_parser'

File: scripts/disabled_tests/tests/data/playlist_input.xml [hunk 2]
```diff
@@ -0,0 +1,90 @@
...
+		<NewTestCaseName>ParallelStreamsLoadTest_special_J9</NewTestCaseName>
+	</test>
+	<test>
+		<a>
+			<b>
+				<c>
+					<testCaseName>ParallelStreamsLoadTest_special_J9</testCaseName>
+				</c>
+			</b>
+		</a>
```
renfeiw, line 13: Could you please explain more on why do we need ```<a> <b> <c>``` here?

scripts/disabled_tests/tests/data/playlist_input.xml
```diff
@@ -0,0 +1,90 @@
...
+		<disables_new_name>
```
renfeiw, line 14: I think it should be `<disables>` instead of `<disables_new_name>`

scripts/disabled_tests/tests/data/playlist_out.json
```diff
@@ -0,0 +1,34 @@
...
+  },
+  {
+    "JDK_VERSION": "11",
+    "PLATFORM": "all",
+    "ISSUE_TRACKER": "some non functional comment",
+    "TARGET": "jdk_custom",
+    "CUSTOM_TARGET": "ParallelStreamsLoadTest_special_J9_0",
+    "JDK_IMPL": "openj9"
+  }
+]
```
renfeiw, line 34: I saw you have a bunch of `<platform>.*windows.*</platform>` in playlist_input.xml, but I  couldn't find matching `"PLATFORM": ".*windows.*"`. Is there anything I missed?

scripts/disabled_tests/tests/test_playlist_parser.py
gervaisj: Hmm true the `README` is lacking in terms of how to run the unit tests.

See instructions added to README in 8ff24151ee40335c94c8dc0441509013b2420228

scripts/disabled_tests/tests/data/playlist_input.xml
gervaisj: They are simply there to show by example how the parser searches all (grand*)children of a `test` node to find a `testCaseName`.

Added a explanatory comment in acc49cb338dcd7dd67cc06feb4dcce9752cc6f7b

scripts/disabled_tests/tests/data/playlist_input.xml
gervaisj: The name is intentionally misspelled to show how the parser ignores the container name, and only searches for "disable" nodes.

Added a explanatory comment in fc4c5554c3371ed55a3b296079c4cb9a687ec007

smlambert: I think Renfei was looking at this from the TKG perspective, where `<disables_new_name>` is not allowed by the .[xsd](https://github.com/adoptium/TKG/blob/master/resources/playlist.xsd) file which is enforced by TKG, so while this parser would ignore parent container, the .xml files checked into the repo, it would only encounter them if someone manually changed the file (hopefully not a common case).  Same goes for the `<a><b><c>` examples... though I understand they are to describe generally how the script works.  (One could just use examples that are actually allowed by the schema).

scripts/disabled_tests/tests/data/playlist_out.json
gervaisj: Oops, I'm sorry for the late reply, I didn't see the comment.

Unfortunately, this is a somewhat undocumented decision that was made in a weekly meeting. The main issue stems from the fact that
a) the parsers don't carry information about all concrete platforms which we support (thus, we can't directly apply the regex and keep only applicable platforms); and
b) the `PLATFORM` field is intended to map to a runnable platform in Jenkins (which the regex expressions aren't).

Of course, both of those points can be changed. The current decision was to always output "ALL", launch it on Jenkins, and then let each TKG instance determine whether or not it should run the test.

scripts/disabled_tests/tests/data/playlist_input.xml
gervaisj: >  it would only encounter [the invalid .xml files] if someone manually changed the file

To me, the intent was more to show how various changes to the _official_ format (enforced by the .xsd) do or do not effect the parser (changes to the format that require an update to the parser or not). This intent to show how changes 
to the official format may require a change in the parser was my interpretation of Lan's advice:

> The idea is that we do not want this python parser to be duplicating the logic in TKG. Since the parser only has info about disables, it does not need to be maintained/updated for every playlist format change.
https://github.com/adoptium/aqa-tests/issues/3345#issuecomment-1036268620

---

> One could just use examples that are actually allowed by the schema

I can add more test files and include some as well if you would like. Perhaps:
- one formatted correctly (valid according to the .xsd)
- one formatted "fuzzy" (**not** valid according to the .xsd, but still accepted by the parser)
- and one formatted incorrectly (**not** valid according to the .xsd, and **not** accepted by the parser)

smlambert: You could increase the examples if you want... the main point is that if we give only the incorrect example, we risk 'confusing some people' who may think thats a big thumbs up to mis-format those files.  It may be sufficient to use a properly formatted one for the example and state that this script does not enforce the schema as TKG does (and reference the .xsd), so that even if you fed it incorrectly formatted files, it could still do its job.

scripts/disabled_tests/tests/data/playlist_out.json
smlambert: For the record, during code reviews of playlists, the reviewers are not advising people to put the full platform name in most cases.  There will still be always be regex though, and yes, thanks for the reminder that @llxia advised in one of our weekly meetings to launch 'all' when in doubt.  :)

scripts/disabled_tests/tests/data/playlist_input.xml
gervaisj: I see, yes it may be confusing, especially if we don't even have one "correct" xml test file.

I reverted the xml test file to a correct format and stated, in the readme, that the script does not rely on the xsd format to work. Please see 2157d9a3a0bc4d2a6692086882591e536fb0433e

scripts/disabled_tests/README.md
```diff
@@ -26,7 +26,31 @@ Python 3.8+
...
+# in scripts/disabled_tests
+python -m unittest discover tests
+```
+
+#### Individual tests
+```shell
+# in scripts/disabled_tests
+python -m unittest tests\test_playlist_parser.py
+```
+
```
renfeiw, line 42: Just tried the test, got the following error:
```
E
======================================================================
ERROR: teststest_playlist_parser (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: teststest_playlist_parser
Traceback (most recent call last):
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/unittest/loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
ModuleNotFoundError: No module named 'teststest_playlist_parser'


----------------------------------------------------------------------
Ran 1 test in 0.000s

FAILED (errors=1)

Comment (gervaisj) [reply scripts/disabled_tests/README.md:42]: Good thing Windows accept both forward and back slashes ;)

Running this should work fine now:
```shell
# in scripts/disabled_tests
python -m unittest tests/test_playlist_parser.py
```

Tested on Ubuntu 20.04 and Windows. Updated the README in 541e1637141cf34b936013ed662aa6188a2360e8

scripts/disabled_tests/tests/data/playlist_out.json
gervaisj: > to launch 'all' **when in doubt** (_emphasis mine_)

Just to clarify: at the moment, that the parser is set to always output "ALL"; we don't try to parse the platform by any means. Should I?

As reference, below are all the values from the `<platform>` nodes and their occurrences.
```
'aarch64_mac.*', 30
'x86-64_windows', 16
'x86-64_mac', 14
'.*windows.*', 14
'.*zos.*', 12
'x86-64_mac|aarch64_mac', 9
'x86-64_mac|ppc64_aix', 6
'x86-64_windows|x86-32_windows', 6
'.*mac.*', 6
'x86-32_windows|arm_linux', 6
'x86-32_windows', 5
'.*aix.*', 4
'ppc64_aix', 2
'x86-64_windows|x86-64_mac', 2
's390x_zos', 2
'aarch64.*', 2
'.*aarch64.*', 1
'x86-64_alpine-linux', 1
'.*x86-64_solaris.*', 1
'^((?!(x86-64_linux|x86-64_mac|x86-64_windows)).)*$', 1
'x86-64_solaris|sparcv9_solaris', 1
'ppc64le.*', 1
'aarch64.*mac.*', 1
'.*aarch.*', 1
```

scripts/disabled_tests/README.md
renfeiw: Thanks!

renfeiw (approved):
Thanks @gervaisj for addressing all the comments! Talked with @llxia, there might be a few enhancements. They can be addressed in the other PRs later.

smlambert (approved):
thanks @gervaisj !

PR #3467, opened 2022-03-15
Disable deferred wipeout before the build and cleanws after build even the build failed

sophia-guo:
Deferred wipeout means that deletion takes places asynchronously to
reduce the build time. If the disk space is low there is no reason to use
this feature as it actually asks for more disk space.

Cleanws at post stage for all build status. Only keep workspace if the parameter KEEP_WORKSPACE is set


llxia (approved):
LGTM

Issue #3466, opened 2022-03-15
Cleanup external tests dockerfile directory


sophia-guo:
Each external test has the dockerfile directory, which originally set as the location of docker related files( dockefile, shell script). For example https://github.com/adoptium/aqa-tests/tree/master/external/elasticsearch/dockerfile . External tests story has been updated to dynamically generate dockefile there is no need to keep the directory.

We'd like to remove the dockefile directory in external tests and move the shell script file to be the same level of all other files.

sophia-guo:
[print_test_files()](https://github.com/adoptium/aqa-tests/blob/master/external/dockerfile_functions.sh#L415-L434) will be affected. The line https://github.com/adoptium/aqa-tests/blob/master/external/dockerfile_functions.sh#L423 and https://github.com/adoptium/aqa-tests/blob/master/external/dockerfile_functions.sh#L427 will need to update.

abhijeetjejurkar:
Hey, Can I work on this?

abhijeetjejurkar:
@sophia-guo Except test.sh Can we delete other related files in dockerfile directory (eg: dockerfile) or we have to move those to  external test directories

smlambert:
For the subdirectories that still have static dockerfiles in that directory (system-test and functional-test), we have to check that they will work if we switch them over to use the autogenerated dockerfiles.  If they do, we can remove the static dockerfiles.  Handling these 2 cases can be done under a separate issue / PR.

abhijeetjejurkar:
Thankyou @smlambert , So I will keep those 2 dockerfile directory(inside system-test and functional-test) as it is. And just move test.sh script in both subdirectories to external test directories

smlambert:
Yes, thanks @abhijeetjejurkar

abhijeetjejurkar:
> Yes, thanks @abhijeetjejurkar

Welcome😊

PR #3483, opened 2022-03-22
Dockerfile folder has been removed.

sophia-guo:
Related #3478 

### DIFF
--- a/external/build.xml
+++ b/external/build.xml
@@ -84,9 +84,6 @@
 	</target>
 
 	<target name="move_scripts" description="move shell scripts to working directory">
-		<copy todir="${DEST_EXTERNAL}/${TEST}/dockerfile">
-			<fileset dir="${src}/dockerfile" includes="*.sh"/>
-		</copy>
 		<copy todir="${DEST_EXTERNAL}/${TEST}">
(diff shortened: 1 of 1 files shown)


smlambert:
Not sure its yet removed for system-test and functional-test, those still need to be dealt with... still contain a static Dockerfile (see recent comment https://github.com/adoptium/aqa-tests/issues/3466#issuecomment-1074264422 about handling those in another cleanup pass...).

sophia-guo:
This fails the job with target sanity or extended or specific third party targets (except for system and function, which have been set as 'special' level. That is the issue will be hidden there if we don't run special)
https://ci.adoptopenjdk.net/view/work-in-progress/job/grinder_sandbox_new/554/

smlambert:
Agree that we will not hit an immediate problem because we are not running system-test or functional-test atm, though @llxia and @LongyuZhang intend to use those targets to verify CRIU prototypes for OpenJ9 testing.

smlambert (approved):
LGTM (but noting that we will need to finish the removal of this directory from system-test and functional-test when we start using them).

llxia (approved):

Issue #1211, opened 2019-07-10
extended.external: scala_test failures


smlambert:
Placeholder issue as we triage and determine where to report issues.

hotspot: 
jdk8: 22 of 2818 failures [jdk8 hs Grinder link](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/21/)
jdk11: x of 2818 failures [jdk11 hs Grinder link](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/23/)

openj9: 
jdk8: 2 of 2818 failures [jdk8 j9 Grinder link](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/20/)
jdk11: 4 of 2818 failures [jdk11 j9 Grinder link](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_Advanced/22/)

- test group: external
- test target name: scala_test
- OS/architecture: x86-64_linux
- OpenJDK version: jdk8u222-b08

Optional info
-------------
- intermittent failure (yes|no): no
- regression or new test:  likely failing for a long while, tests are just now being triaged

Sample failure output:
jdk8 hotspot
19:10:51  Selected 2818 tests drawn from 3 named test categories
19:10:51  
19:10:51  # starting 1625 tests in pos
19:11:31  ........................................................................
19:12:55  .....................[[syntax trees at end of                     typer]] // t10344.scala
19:12:55  package <empty> {
19:12:55    object t10344 extends scala.AnyRef {
19:12:55      def <init>(): t10344.type = {
19:12:55        t10344.super.<init>();
19:12:55        ()
19:12:55      };
19:12:55      def unwrap[F[_] >: [_]Nothing <: [_]Any](f: F[Unit] => Unit): Unit = ();
19:12:55      private[this] val f: (=> Unit) => Unit = ((x$1: => Unit) => ());
19:12:55      <stable> <accessor> def f: (=> Unit) => Unit = t10344.this.f;
19:12:55      t10344.this.unwrap[<byname>](t10344.this.f)
19:12:55    }
19:12:55  }
19:12:55  
19:13:04  ...................................................
19:13:16  ......................................Name table size after compilation: 73597 chars
19:13:32  ..................................
19:17:18  .........................................
19:17:18  
19:17:18  # starting 1112 tests in neg
19:17:23  ........................................................................
19:17:33  ........................................................................
19:17:41  ........................................................................
[...]

PR #3484, opened 2022-03-22
Support default test.sh parameter by CMD

sophia-guo:
Support default test_options as the test suites or targets per
applications in docker image.
Keep --testtarget as user defined test suites or targets which override
the default one, which is helpful for debugging or setting different
suites for different jdk version or vendors.
Update $testtarget with double quote as docker run parameters to avoid
the string is cut by space to fix the issue mentioned https://github.com/adoptium/aqa-tests/issues/1211#issuecomment-1058532135


sophia-guo:
Grinder scala with default CMD setting https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3985/
Grinder with personalized testsuite https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3986/

external/external.sh
```diff
@@ -180,11 +180,19 @@ if [ $command_type == "run" ]; then
 	fi
 	if [ $reportsrc != "false" ]; then
 		echo "docker run $mountV --name $test-test adoptopenjdk-$test-test:${JDK_VERSION}-$package-$docker_os-${JDK_IMPL}-$build_type $testtarget"
-		docker run $mountV --name $test-test adoptopenjdk-$test-test:${JDK_VERSION}-$package-$docker_os-${JDK_IMPL}-$build_type $testtarget;
+		if [ -n "$testtarget" ]; then
+			docker run $mountV --name $test-test adoptopenjdk-$test-test:${JDK_VERSION}-$package-$docker_os-${JDK_IMPL}-$build_type "$testtarget";
+		else
+			docker run $mountV --name $test-test adoptopenjdk-$test-test:${JDK_VERSION}-$package-$docker_os-${JDK_IMPL}-$build_type;
+		fi
```
llxia, line 187: I am not very clear why we need if-else here

sophia-guo: As $testtarget may be the string include spaces we have to use "$testtarget" as parameter, which also means it could be an empty string. In that case the empty string may override the default setting, which is not what we expected. We could also try to test if $testtarget is null instead of an empty string, but I don't think that helps a lot.

external/payara-mp-tck/test.properties
```diff
@@ -1,6 +1,7 @@
 github_url="https://github.com/payara/MicroProfile-TCK-Runners.git"
 test_command="mvn --batch-mode clean verify"
 test_results="testResults"
+test_options="-pl !MicroProfile-OpenTracing/tck-runner,!MicroProfile-JWT-Auth,!MicroProfile-JWT-Auth/tck-arquillian-extension,!MicroProfile-JWT-Auth/tck-runner"
```
smlambert, line 4: At some point it may be interesting to try reincluding these, but definitely not needed as part of this PR

smlambert (approved):

llxia (approved):

PR #3519, opened 2022-03-31
Exclude the FFI specific test suites on the 64-bit platforms

ChengJin01:
The change is to exclude the FFI specific test suites that are
intended for the 32-bit platforms.

Signed-off-by: Cheng Jin <jincheng@ca.ibm.com>


llxia (approved):
LGTM

Mesbah-Alam (approved):
LGTM

PR #3520, opened 2022-04-01
Set excludes per jdk

sophia-guo:
Set excludes per jdk.
Rerun all failed projects, enable 11 projects which passed right now


llxia (approved):
LGTM

smlambert (approved):
thanks @sophia-guo !

PR #3530, opened 2022-04-04
Using gradle defined by spring project

sophia-guo:
Instead of manually setup gradle by test.properties we should use the gradle wrapper set up by project itself. Also update to v2.6.6.

### DIFF
--- a/external/spring/playlist.xml
+++ b/external/spring/playlist.xml
@@ -15,6 +15,12 @@
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../TKG/playlist.xsd">
 	<test>
 		<testCaseName>spring_test</testCaseName>
+		<disables>
+			<disable>
+				<comment>https://github.com/adoptium/aqa-tests/issues/3505#issuecomment-1086270612</comment>
+				<version>17</version>
+			</disable>
+		</disables>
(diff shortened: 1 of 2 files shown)


sophia-guo:
jdk11: https://ci.adoptopenjdk.net/view/work-in-progress/job/grinder_sandbox_new/613/
jdk8: https://ci.adoptopenjdk.net/view/work-in-progress/job/grinder_sandbox_new/612/

Failure `Compilation with Kotlin compile daemon was not successful` believe it's the compatibility issue with [Kotlin 1.6.10 ](https://github.com/spring-projects/spring-boot/blob/v2.6.6/gradle.properties#L7) 

Gradle is tested with Kotlin 1.3.72 through 1.6.0, not 1.6.10.
1.6.10 has fixes with 1.6.0 https://kotlinlang.org/docs/releases.html#release-details

Could manually update [Kotlin 1.6.10 ](https://github.com/spring-projects/spring-boot/blob/v2.6.6/gradle.properties#L7) to be 1.6.0 and see if the problem stays.

jdk 17: https://ci.adoptopenjdk.net/view/work-in-progress/job/grinder_sandbox_new/610/

Failure `General error during semantic analysis: Unsupported class file major version 61` shows the tag '2.6.6' setting gradle `6.9.2`, which doesn't work with jdk17.

smlambert (approved):
LGTM

sophia-guo:
The newest grinder link with v2.6.6 should be 

https://ci.adoptopenjdk.net/view/work-in-progress/job/grinder_sandbox_new/617/console
https://ci.adoptopenjdk.net/view/work-in-progress/job/grinder_sandbox_new/616/console

Issue #3510, opened 2022-03-29
Double check gradle version of external test projects as only 7.3 supports jdk17


sophia-guo:
There are a few external test projects are using gradle with specific version, which might not support jdk17 yet. https://docs.gradle.org/current/userguide/compatibility.html. Need to double check that and if that's the case we need to disable test with jdk17. 

Similar to https://github.com/adoptium/aqa-tests/issues/3381

amir-waheed:
Can I have this issue assigned to me?

sophia-guo:
Spring : https://github.com/adoptium/aqa-tests/issues/3505#issuecomment-1086270612

amir-waheed:
@sophia-guo 
These are all the test that require specific version of grade which doesn't support jdk17.
functional-test - gradle version 5.1
Kafta - gradle version 5.4
Spring - gradle version 5.1
system-test - gradle version 5.1
Should I go ahead and disable these tests for jdk17?

sophia-guo:
@amir-waheed we can leave functional-test and system-test as is for now.
Both Spring and Kafta actually use gradle wrapper, which means we don't need to manually setup gradle in test.properties file, please see #3530 as spring updates. You can try similar thing for Kafta.

sophia-guo:
@amir-waheed I also think we can just remove the function print_gradle_install in dockerfile_function.sh as all gradle related projects set up the gradle by gradle wrapper, which we will always use the same gradle version as the application community.

amir-waheed:
@sophia-guo I have a PR up with these changes.

PR #3534, opened 2022-04-06
Updating Kafta project to remove gradle 

amir-waheed:
This PR removes the print_gradle_install function in dockerfile_function.sh and removes the gradle tag from Kafta project. 

Part of #3510 


sophia-guo:
There should be some code which are calling the function print_gradle_install in the same file, which should be removed too.

Also please do a rebase to resolve the conflicts.

PR #3536, opened 2022-04-07
archive AQACert log in parent build

llxia:
1. Since all AQACert.log are the same for the parent build, only archive one AQACert.log from a child build
2. Update archiveFile() logic to handle the case where we want to archive on both Jenkins and Artifactory. *.tap files and AQACert.log have to be archived on Jenkins in order for the parent job to pick up.
3. Add cleanWs before and after archiving child build files to avoid copying the wrong files
4. Update the logic to archive a specific TAP file for each child (instead of **/*.tap). 3 and 4 are for resolving https://github.com/adoptium/aqa-tests/issues/3539
5. Add TODO for future enhancement

Resolve: https://github.com/adoptium/aqa-tests/issues/3539

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -685,18 +685,13 @@ def post(output_name) {
 
 			step([$class: "TapPublisher", testResults: "**/*.tap", outputTapToConsole: false, failIfNoResults: true])
 
-			// only archive children TAP result in parallel mode, the file will be copied into the parent job after parallel runs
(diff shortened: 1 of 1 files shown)


llxia:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4338/

image

llxia:
Parallel: internal/Grinder/22427/

smlambert (approved):
LGTM (thanks @llxia)...

llxia:
zos Parallel internal/Grinder/22445/

renfeiw (approved):
LGTM

PR #3607, opened 2022-04-25
Dragonwell update: get_source not needed, update test list

joeyleeeeeee97:


openjdk/playlist.xml
```diff
@@ -1757,4 +1757,215 @@
...
+		<versions>
+			<version>8</version>
+			<version>11</version>
+		</versions>
+		<levels>
+			<level>sanity</level>
+		</levels>
+		<groups>
+			<group>openjdk</group>
+		</groups>
```
smlambert, line 1781:
```suggestion
		</groups>
		<vendor>alibaba</vendor>
```

smlambert: Since each of these new `<test>` blocks are specific to alibaba dragonwell, please add `<vendor>` tags to all newly added `<test>` blocks so they will not be executed against temurin or others.

smlambert (changes requested):
thanks @joeyleeeeeee97 - request to add `<vendor>` tags as these tests are specific to alibaba (and will be filtered out and not run against other vendors products).

smlambert:
Ah, apologies, I gave poor instruction, I think it needs to be this...

```
<vendors>
<vendor>alibaba</vendor>
</vendors>
```

joeyleeeeeee97:
@smlambert Thanks for the tip. I have learned to use Grinder to verify my patch. :)
[Termurin no tests found](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4528/console)
[Dragonwell running](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4529/console)

joeyleeeeeee97:
ping.

smlambert:
The change looks good, the Grinder job fails because the configuration I used did not point to the Dragonwell OpenJDK fork for grabbing test material (and it won't look there for test material pulled for AQAvit verification, which means these new tests will fail when following the process for verification... which is to use the v0.9.1-release branch which pins to a particular OpenJDK tag, not vendor specific OpenJDK repos).  This is fine, in that vendor specific tests do not need to pass, but I was looking at what we may want to do so that you do to better support the case of vendor specific tests... (possibly a new level beyond sanity/extended/special that is 'dev').  Will approve this PR, as I believe we pull the 'right' Dragonwell OpenJDK repo in nightly testing.

smlambert (approved):

llxia (approved):
LGTM

PR #3638, opened 2022-05-10
Fix jck_custom target

### LINKED ISSUE #3598: jck_custom target appears to be 'broken'
At some point between January release to March release the ability to run jck_custom target appears to be broken.  Need to investigate what changes went into JavaTestRunner and related files that may have caused the breakage.

sophia-guo:
Fix #3598 

### DIFF
--- a/jck/runtime.api/playlist.xml
+++ b/jck/runtime.api/playlist.xml
@@ -19,7 +19,7 @@
 		<variations>
 			<variation>NoOptions</variation>
 		</variations>
-		<command>$(JCK_CMD_TEMPLATE) tests=$(JCK_CUSTOM_TARGET) jckRoot=$(JCK_ROOT) jckversion=$(JCK_VERSION); \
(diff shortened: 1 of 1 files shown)


sophia-guo:
Grinder with test which doesn't need the `configAltPath`:  util tests temurin-compliance/Grinder/1290 also works.

@Mesbah-Alam if there is any concern that we add `configAltPath` as a default setting for jdk_custom, which will make it work with testcases need that path.  As I mentioned trying with one util test which doesn't need this setting it works. Though I'm not sure if there is any side effects.

Mesbah-Alam:
@sophia-guo , The change you have in this PR atm should be fine. Tested with your changes: Grinder/23720.

smlambert (approved):
LGTM

smlambert:
Can you approve and merge it please @Mesbah-Alam ?  thanks!

llxia (approved):
LGTM

PR #3634, opened 2022-05-09
[arm32] Exclude problematic tests

sophia-guo:
1. linux-arm jdk8
https://github.com/adoptium/aqa-tests/issues/3603#issuecomment-1113123584
https://github.com/adoptium/aqa-tests/issues/3603#issuecomment-1121863541
2. Linux-arm jdk11 https://github.com/adoptium/aqa-tests/issues/3596#issuecomment-1104643594
3. Linux-arm jdk17 https://github.com/adoptium/aqa-tests/issues/3594#issuecomment-1104042051
4. Linux-arm jdk18 https://github.com/adoptium/aqa-tests/issues/3599#issuecomment-1105959116 
https://github.com/adoptium/aqa-tests/issues/3637
5. Fix the wrong os-linux information

Signed-off-by: Sophia Guo <sophia.gwf@gmail.com>

PR #3645, opened 2022-05-11
(v0.9.1-release) [arm32] Exclude problematic tests (#3634)

sophia-guo:
cherry-pick:  #3634  Exclude problematic tests


smlambert (approved):
thanks!

Issue #3546, opened 2022-04-08
Investigate and add jcstress in system test group


llxia:
We should investigate and add jcstress in the system test group:

https://github.com/openjdk/jcstress

Mesbah-Alam:
Download jcstress-latest.jar from https://builds.shipilev.net/jcstress locally on my Mac and was able to start the test by doing: `java -jar jcstress-latest.jar` - which is the simplest way of running these tests according to the [doc](https://github.com/openjdk/jcstress). 

However, when [I try to do the same](https://github.com/adoptium/aqa-tests/compare/master...Mesbah-Alam:add-jcstress?expand=1) on Jenkins, I get an error : 

```
Error: Invalid or corrupt jarfile /home/jenkins/workspace/Grinder/aqa-tests/TKG/../../jvmtest/system/jcstress/jcstress-latest.jar
```

(Grinder/23306). 

Checking the size of jcstress-latest.jar on Jenkins shows its only 8K : 

```
13:55:36  + du -h /home/jenkins/workspace/Grinder/aqa-tests/TKG/../../jvmtest/system/jcstress/jcstress-latest.jar
13:55:36  8.0K	/home/jenkins/workspace/Grinder/aqa-tests/TKG/../../jvmtest/system/jcstress/jcstress-latest.jar
```
(all-nodes-matching-labels/555). 

While the actual size of jcstress-latest.jar is 54.6 MiB (ref: https://builds.shipilev.net/jcstress). 

---- 

An alternative way to do this would be to checkout jcstress source and build it as part of systemtest.getDependency job - but that is a more complicated route. 

@llxia - would you be aware of any reasons why Jenkins may prevent download of jars from https://builds.shipilev.net/jcstress? (I had basically copied the `curl` command from [Dacapo](https://github.com/adoptium/aqa-tests/blob/9a25e28294e24e67db11bcaf83990dc6c48710f9/perf/dacapo/build.xml#L30).

llxia:
From what I can see, you are pointing https://builds.shipilev.net/jcstress, not a specific jar.
https://github.com/adoptium/aqa-tests/compare/master...Mesbah-Alam:add-jcstress?expand=1#diff-8a2f84603fb402a5feff98b12271e8a711baabf03d1efe2647ce1e3552bf629bR30

Mesbah-Alam:
[...]

PR #3666, opened 2022-05-17
Add jcstress test to system test

Mesbah-Alam:
Related : https://github.com/adoptium/aqa-tests/issues/3546

Should be delivered after: https://github.com/adoptium/TKG/pull/321

### DIFF
--- a/system/jcstress/build.xml
+++ b/system/jcstress/build.xml
@@ -1,0 +1,41 @@
+<?xml version="1.0"?>
+
+<!--
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+
+<project name="JCStressTest" default="build" basedir=".">
+	<taskdef resource="net/sf/antcontrib/antlib.xml" />
+	<description>JCStress Test</description>
+
+	<!-- Set global properties for this build -->
+	<property name="DEST" value="${BUILD_ROOT}/system/jcstress" />
+	<property name="SRC" location="." />
+
+	<property name="LIB" value="jcstress" />
+	<import file="${TEST_ROOT}/TKG/scripts/getDependencies.xml" />
+
+	<target name="init">
+		<mkdir dir="${DEST}" />
+	</target>
+	
+	<target name="dist" depends="init" description="generate the distribution">
+		<copy todir="${DEST}">
+			<fileset dir="${SRC}"/>
+		</copy>
+	</target>
+
+	<target name="build">
+		<antcall target="dist" inheritall="true" />
+	</target>
+</project>
--- a/system/jcstress/playlist.xml
+++ b/system/jcstress/playlist.xml
@@ -1,0 +1,28 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!--
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+<playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../TKG/playlist.xsd">
+	<test>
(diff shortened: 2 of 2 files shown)


Mesbah-Alam:
Test : Grinder/23850

system/jcstress/playlist.xml
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+<playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../TKG/playlist.xsd">
+	<test>
+		<testCaseName>jcstress_SampleTestBench</testCaseName>
+		<command>$(JAVA_COMMAND) $(JVM_OPTIONS) -jar $(Q)${TEST_ROOT}$(D)systemtest_prereqs$(D)jcstress$(D)jcstress-latest.jar$(Q) -jvmArgs $(Q)$(JVM_OPTIONS) -t SampleTestBench$(Q); \
+		$(TEST_STATUS)</command>
+		<versions>
+			<version>11</version>
```
llxia, line 21: Should this be 8+? If so, we do not need the version tag.

Mesbah-Alam: Good catch. Updated, works. Grinder/23855.

llxia:
Ideally, we should separate jcstress with STF as they are unrelated. However, this PR creates the dependency between jcstress and STF (i.e, we cannot run jcstress without STF). And without build.xml, I suspect dynamic compilation will not work in this case.

I am ok with the PR for now as this enables us to run jcstress. Please open an issue for decoupling jcstress and STF. I think the better way is to use getDependencies in TKG.

Mesbah-Alam:
> we cannot run jcstress without STF

We can run jcstress without STF. The only dependency is the top.xml changes (https://github.com/adoptium/STF/pull/127). However, since we want to run jckstress as part of systemtest, and system test will compile STF anyway, there is no extra cost of having the dependency on top.xml to fetch jcstress.jar.

smlambert:
Ideally for the dynamic compilation feature that TKG has, if I only wanted to run the jcstress target by itself, then it should not have to go and compile STF but agree any further decoupling can be in a 'phase 2'.

Mesbah-Alam:
Sounds good! Instead of doing it in phase 2 , let's move to using the getDependencies script in TKG now. 

I've added jcstress jar to the list of downloads in get Dependencies script in TKG : https://github.com/adoptium/TKG/pull/321. 

Manually ran test.getDependency job pointing to my branch for testing : https://hyc-runtimes-jenkins.swg-devops.com/job/test.getDependency/10261/. 

(Note, once https://github.com/adoptium/TKG/pull/321 is delivered, we need to run the [test.getDependency job at ci.adopt](https://ci.adoptopenjdk.net/job/test.getDependency/)). 

Updated the PR for playlist. New test : Grinder/23939.

llxia:
Thanks @Mesbah-Alam . 

Please also add build.xml file in this case. Otherwise, I think the local run will fail. :) 
The build.xml should contain the following. It will trigger getDependencies for downloading jcstress if needed.
```
	<property name="LIB" value="jcstress" />
	<import file="${TEST_ROOT}/TKG/scripts/getDependencies.xml" />
```

Mesbah-Alam:
Re: https://github.com/adoptium/aqa-tests/pull/3666#issuecomment-1130404099

Added. Please check if this will suffice!

system/jcstress/build.xml
```diff
@@ -0,0 +1,20 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+
+<project name="JCStressTest" basedir=".">
+	<property name="LIB" value="jcstress" />
+	<import file="${TEST_ROOT}/TKG/scripts/getDependencies.xml" />
+</project>
```
llxia, line 20: Sorry, the above is not sufficient. We need the default targets.

```
<project name="JCStressTest" default="build" basedir=".">
	<taskdef resource="net/sf/antcontrib/antlib.xml" />
	<description>JCStress Test</description>

	<!-- Set global properties for this build -->
	<property name="DEST" value="${BUILD_ROOT}/system/jcstress" />
	<property name="SRC" location="." />

	<property name="LIB" value="jcstress" />
	<import file="${TEST_ROOT}/TKG/scripts/getDependencies.xml" />

	<target name="init">
		<mkdir dir="${DEST}" />
	</target>
	
	<target name="dist" depends="init" description="generate the distribution">
		<copy todir="${DEST}">
			<fileset dir="${SRC}"/>
		</copy>
	</target>

	<target name="build">
		<antcall target="dist" inheritall="true" />
	</target>
</project>
```

llxia: This should be tested via `BUILD_LIST=system/jcstress`

Mesbah-Alam: Added (tested: Grinder/23946). Please check.

Mesbah-Alam: Here's the test with `BUILD_LIST=system/jcstress` : Grinder/24008.

llxia (approved):
LGTM

smlambert (approved):
Since the related TKG PR is merged, I will proceed to merge this one.  Thanks @Mesbah-Alam for this initial add, and as discussed, we can now refine what we run out of the jcstress repo quite easily in a follow-up PR. 👍

Issue #3696, opened 2022-05-22
./compile.sh will only download renaissance.jar at the first time, but will download dacapo.jar everytime


sendaoYan:
**Describe the bug**
./compile.sh will only download renaissance.jar at the first time, but will download dacapo.jar everytime.

**To Reproduce**
```bash
export JDK_VERSION=17
export JAVA_HOME=${HOME}/software/jdk/2022-05/dragonwell-17.0.3.0.3+7-GA
export TEST_JDK_HOME=$JAVA_HOME
export PATH=$JAVA_HOME/bin:$PATH
export USE_TESTENV_PROPERTIES=true
export JDK_IMPL=hotspot
export BUILD_LIST=perf
cd aqa-tests
./get.sh
./compile.sh

./compile.sh
```

**Expected behavior**
./compile.sh download the dacapo.jar at the first time. Same to renaissance, the ant build.xml check there is a dacapo.jar in aqa-tests/perf/dacapo/dacapo.jar or not, if there already a dacapo.jar file in aqa-tests/perf/dacapo directory, then skip the download step from internet, like renaissance ant build.xml do:

![image](https://user-images.githubusercontent.com/24123821/169690341-d37f2d8a-44b4-4ce5-b900-be0a9ff9f73a.png)

**Screenshots**

**Additional context**

This is patch:

```diff
git diff
diff --git a/perf/dacapo/build.xml b/perf/dacapo/build.xml
index a62bb47..71e04f1 100644
--- a/perf/dacapo/build.xml
+++ b/perf/dacapo/build.xml
@@ -27,15 +27,24 @@
        </target>
 
        <target name="getDacapoSuite" depends="init">
-               <var name="curl_options" value="-Lks https://sourceforge.net/projects/dacapobench/files/latest/download -o dacapo.jar"/>
-               <retry retrycount="3">
-                       <sequential>
-                               <echo message="curl ${curl_options}" />
-                               <exec executable="curl" failonerror="true">
-                                       <arg line="${curl_options}" />
-                               </exec>
-                       </sequential>
-               </retry>
+               <var name="FILE_NAME" value="dacapo.jar"/>
+               <if>
+                       <available file="${FILE_NAME}" type="file" />
+                       <then>
+                               <echo message="${FILE_NAME} exists. Hence, not downloading it." />
+                       </then>
+                       <else>
[...]

PR #3697, opened 2022-05-22
Download dacapo.jar form internet only at the first time

sendaoYan:
Similar to renaissance, check there is dacapo.jar in aqa-tests/perf/dacapo/ or not, if it does, then skip the download action from internet

Fixes: #3696


smlambert (approved):
thanks @sendaoYan !

karianna (approved):
LGTM

PR #3711, opened 2022-05-30
[AQA_Test_Pipeline] : Do not parallel for temurin functional and perf categories

sophia-guo:
For temurin the functional set is small so it's not necessary to run in parallel especially for platforms with limited resources. Machines with label `ci.role.perf` are also limited in Adoptium. Running in parallel may result in the waiting time is 90% of the whole job time. Disable parallel for funtional and perf tests. 

### DIFF
--- a/buildenv/jenkins/aqaTestPipeline.groovy
+++ b/buildenv/jenkins/aqaTestPipeline.groovy
@@ -64,6 +64,11 @@ JDK_VERSIONS.each { JDK_VERSION ->
             if (TARGET.contains("jck") || TARGET.contains("openjdk")) {
                 keep_reportdir = true
             }
+            if (TARGET.contains("functional") || TARGET.contains("perf") {
+                if (params.VARIANT == "temurin") {
+                    PARALLEL = "None"
+                }
+            }
             if (AUTO_AQA_GEN) {
                 String[] targetTokens = TARGET.split("\\.")
                 def level = targetTokens[0];


llxia:
The functional and perf do not run in parallel from what I can see. May I know where the request comes from?
https://ci.adoptopenjdk.net/job/Test_openjdk8_hs_sanity.functional_x86-64_mac
https://ci.adoptopenjdk.net/job/Test_openjdk8_hs_sanity.perf_x86-64_mac
https://ci.adoptopenjdk.net/job/Test_openjdk11_hs_extended.functional_ppc64_aix

sophia-guo:
This only affects [AQA_Test_Pipeline](https://ci.adoptopenjdk.net/view/Test_grinder/job/AQA_Test_Pipeline/aqatestpipeline) when trigger the job by multiple platforms and multiple test categories parallel is set to 3 by default https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/aqaTestPipeline.groovy#L9. If it's just for specific platform or specific category we can also add PARALLEL and NUM_MACHINES to specify. However for multiple platforms and test categories setting separate default would be preferred
https://ci.adoptopenjdk.net/job/Test_openjdk11_hs_extended.functional_x86-64_mac/245/

llxia (approved):
LGTM

smlambert (approved):

PR #3722, opened 2022-06-02
Temporarily exclude JLM remote class test on AIX

Mesbah-Alam:
As reported in https://github.com/eclipse-openj9/openj9/issues/11557, JLM remote class tests (all flavours) fail pretty consistently due to failure of STF's attempt to kill child subprocess. We need to investigate and improve the STF logic of   cleaning up child processes (especially on AIX). Until then, the targets which are consistently failing due to this are being excluded in this PR. 

### DIFF
--- a/system/jlm/playlist.xml
+++ b/system/jlm/playlist.xml
@@ -42,6 +42,13 @@
 	<!-- Temporarily excluding from z/OS due to : jdk11-zos/issues/567 -->
 	<test>
 		<testCaseName>TestJlmRemoteClassAuth</testCaseName>
+		<disables>
+			<disable>
+				<comment>https://github.com/eclipse-openj9/openj9/issues/11557</comment>
+				<platform>ppc64_aix</platform>
+				<impl>openj9</impl>
+			</disable>
+		</disables>
 		<variations>
 			<variation>Mode150</variation>
 			<variation>Mode650</variation>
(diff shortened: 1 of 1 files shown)


system/jlm/playlist.xml
```diff
@@ -42,6 +42,12 @@
 	<!-- Temporarily excluding from z/OS due to : jdk11-zos/issues/567 -->
 	<test>
 		<testCaseName>TestJlmRemoteClassAuth</testCaseName>
+		<disable>
+			<comment>Disabled on aix for https://github.com/eclipse-openj9/openj9/issues/11557</comment>
+			<platform>ppc64_aix</platform>
+			<impl>openj9</impl>
+			<impl>ibm</impl>
```
llxia, line 49: This format is not correct. 
- There is no `<disables>`. Please see https://github.com/eclipse-openj9/openj9/blob/master/test/docs/OpenJ9TestUserGuide.md#manually-exclude-a-test-target
- Cannot use two impl in one `<disable>`

llxia: We do not need `<impl>ibm</impl>` as there is no aix build uses impl=ibm.

Mesbah-Alam: Thanks! Fixed! Please review!

system/jlm/playlist.xml
```diff
@@ -42,6 +42,13 @@
 	<!-- Temporarily excluding from z/OS due to : jdk11-zos/issues/567 -->
 	<test>
 		<testCaseName>TestJlmRemoteClassAuth</testCaseName>
+		<disables>
+			<disable>
+				<comment>Disabled on aix for https://github.com/eclipse-openj9/openj9/issues/11557</comment>
```
llxia, line 47: Could you remove the text `Disabled on aix for` in the comment? We should only keep URL portion as this will be parsed via another script for running disabled tests.

Mesbah-Alam: Done!

llxia (approved):
LGTM

Mesbah-Alam:
Changing this PR to draft as there is a [fix](https://github.com/adoptium/STF/pull/128) that just got delivered. We want to monitor the builds for the next little while to check if it resolves the problem.

Mesbah-Alam:
The [related issue ](https://github.com/eclipse-openj9/openj9/issues/11557) is closed. We no longer need this exclude. Closing without merging.

PR #3742, opened 2022-06-08
Re-enable some disabled runtime tests as it's ready to run automatically

sophia-guo:


smlambert (approved):
thanks @sophia-guo !

llxia (approved):

PR #3744, opened 2022-06-08
No need to set JDK_VENDOR from jenkins

sophia-guo:
JDK_VENDOR has been set in TKG


sophia-guo:
As https://github.com/adoptium/TKG/pull/327 is merged we can just totally depends on TKG to export right JDK_VERSION.

Have been tested before https://github.com/adoptium/aqa-tests/pull/3723#issuecomment-1147611175

smlambert (approved):

llxia (approved):
LGTM

Issue #1112, opened 2019-05-03
Identify and prioritize open perf benchmarks 


smlambert:
We want to select various (open-source) performance benchmark suites/harnesses to run at AdoptOpenJDK (for validating JDK binaries and to make available for easy use by JDK developers).

We can first list any that could be useful for JDK measurement/verification, identify why they are interesting, and then prioritize accordingly.  This list can include both full suites and microbenchmarks with links to git repos (and pre-compiled jar/zip files when available).

Once we start adding test targets from different benches, we will also have to ensure that they are suitable for all implementations (not based on implementation specific classes, etc).  But we can leave that as a separate exercise.

**Larger suites**
HiBench:
https://github.com/Intel-bigdata/HiBench

AcmeAir:
https://github.com/blueperf/acmeair-monolithic-java
https://acmeair.github.io/acmeair/ 

DayTrader3:
https://github.com/WASdev/sample.daytrader3

DayTrader7:
https://github.com/WASdev/sample.daytrader7

Dacapo:
https://github.com/dacapobench/dacapobench
http://dacapobench.org/

**Micro/Macro Bench Harnesses**
Bumblebench:
https://github.com/AdoptOpenJDK/bumblebench

jmh:
https://openjdk.java.net/projects/code-tools/jmh/

piyush286:
Thanks Shelley for creating this list! Besides the ones that you've already mentioned, here are some more open benchmarks:

Startup time, Footprint and Throughput of OpenLiberty with different apps:
https://github.com/OpenLiberty/open-liberty + DT3, DT7, AcmeAir & More

Linkerd:
https://github.com/linkerd/linkerd-examples/tree/master/perf-baseline/linkerd1-perf

Databricks:
https://github.com/databricks/spark-perf

ES Rally:
https://github.com/elastic/rally

Spring Boot Startup Benchmark:
https://github.com/dsyer/spring-boot-startup-bench

FrameworkBenchmarks:
https://github.com/TechEmpower/FrameworkBenchmarks

**Perf Team's Microbenchmarks**
We've also written a few micro-benchmarks such as for crypto and GPU over the few years so we could think of adding those if you think they might be interesting and useful.

**SPEC**
[...]

PR #3746, opened 2022-06-09
Add FrameworkBenchmarks, starting from netty

joeyleeeeeee97:
FrameworkBenchmarks is mentioned in https://github.com/adoptium/aqa-tests/issues/1112 .

Personnaly I am using this test a lot, it's open, popular, and covers a wide range of frameworks.

https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Java

I am beginning with popular ones such as netty, we need to find a way to generlize different cases later.


joeyleeeeeee97:
Grinder not working, so I will rebuild this later. https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/4964/

smlambert:
Its unable to delete workspace because of the json file from this benchmark... could it be the case that something still has that file open, when we try to delete the workspace, which blocks the deletion?

llxia:
yes, I suspect the previous benchmark job failed in a bad state and left the process running (holding the file). The new job landed on the same machine cannot delete the workspace due to the file. 

At this point, the machine needs to be cleaned up. Also, we need to enhance the test framework to terminate the leftover processes and clean up the workspace. If this cannot be properly handled at the test level, it should be handled at TKG/Jenkins level.

joeyleeeeeee97:
image
These frameworks are more popular... I think.

These benchmarks are very sensitive and takes a long time to finish.

smlambert:
The issue seen with not being able to delete workspace seems to be related to the results directory being created as root (instead of Jenkins).  See https://github.com/adoptium/infrastructure/issues/2597#issuecomment-1153974613

joeyleeeeeee97:
@llxia Yes. Should I change the permission in `results` in running script to get removable results?

llxia:
Yes, we need to change to `jenkins`. Currently, the results folder is owned by `root` which cannot be removed by `jenkins` user.

joeyleeeeeee97:
@llxia 
I will change to current owner after each run, 
https://github.com/adoptium/aqa-tests/pull/3746/files#diff-f67c65a7fc8657a9fa5658270920dce990b1ee63a080ed9ed0d404dae3e0699fR21

Locally, I am getting right permissions for these files after each run, but this doesn't work if current jenkins job is forced to stop or killed during running.

joeyleeeeeee97:
@llxia  I added a extra step to make results removable.
docker run --entrypoint /bin/bash  --rm --network tfb -v /var/run/docker.sock:/var/run/docker.sock -v ${TEST_RESROOT}:/FrameworkBenchmarks techempower/tfb -c "chmod -R 777 /FrameworkBenchmarks/results" 


FrameworkBenchmarks is running with root, I raise an [issue](https://github.com/TechEmpower/FrameworkBenchmarks/issues/7413) for them.

joeyleeeeeee97:
http://ci.dragonwell-jdk.io/job/Test_openjdk11_dragonwell_special.perf_x86-64_linux/8/console
 testing in my local jenkins, hopefully get a green result.

joeyleeeeeee97:
@smlambert 
Update:
Removed changes in testenv.
This test is not pinned to any specific hardware, only needs to have docker.

To dispatch these to docker, maybe we could:
1. Install docker on all `ci.role.perf` machines if this is acceptable.
2. Add `DOCKER_REQUIRED` when running some test pipelines, but this also relies on that we have some machines that both labeled `ci.role.perf` and `sw.tool.docker`

I think adding `DOCKER_REQUIRED` is better because in long term we might have other tests rely on docker, for example some system tests and some other perf tests might need a docker.

smlambert:
I am glad you are asking these questions @joeyleeeeeee97 - I agree we need more machines that can support container based testing (ie. with Docker installed).  

If we pin special.perf to https://ci.adoptopenjdk.net/label/sw.tool.docker&&ci.role.perf/ that implies we can only run these benchmarks on 2 platforms at the project (x86-64_linux and s390x_linux), when the tests get sent to the other platforms (which based on the playlist, since there are no platform restrictions, we will try to do, they will fail with "no machines available with that tag").

I think we have to ensure that we have Docker installed on a broad base of platforms, and then add a platformCapabilities tag to the playlist in this PR to exclude the systems where we are not able to run Docker.

joeyleeeeeee97:
- ensure that we have Docker installed on a broad base of platforms
I think we could follow this in infrastructure issues.
- add a platformCapabilities tag to the playlist in this PR to exclude the systems
Currently only enable x86 arch
```
		<platformRequirements>os.linux,arch.x86,bits.64</platformRequirements>
```
Do we have a `or` semantic in platformRequirements? like
```
		<platformRequirements>os.linux,(arch.x86|arch.390),bits.64</platformRequirements>

```

smlambert:
Closing as stale

Issue #3751, opened 2022-06-10
VersionCheck.java run fail by dragonwell8


sendaoYan:
**Describe the bug**
jdk/test/tools/launcher/VersionCheck.java run fail by dragonwell8

**To Reproduce**
`jtreg jdk/test/tools/launcher/VersionCheck.java`

```log
Internal Strings Test: 0
Testing (-version): /home/yansendao/software/jdk/2022-05/dragonwell-8.11.12/bin/idlj
Testing (-version): /home/yansendao/software/jdk/2022-05/dragonwell-8.11.12/bin/jgroup
Testing (-version): /home/yansendao/software/jdk/2022-05/dragonwell-8.11.12/bin/jhat
Testing (-version): /home/yansendao/software/jdk/2022-05/dragonwell-8.11.12/bin/java
Testing (-version): /home/yansendao/software/jdk/2022-05/dragonwell-8.11.12/bin/javac
Testing (-version): /home/yansendao/software/jdk/2022-05/dragonwell-8.11.12/bin/javah
Testing (-version): /home/yansendao/software/jdk/2022-05/dragonwell-8.11.12/bin/javap
Testing (-version): /home/yansendao/software/jdk/2022-05/dragonwell-8.11.12/bin/jdb
STDERR:
java.lang.AssertionError: Some tests failed
        at VersionCheck.main(VersionCheck.java:227)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at com.sun.javatest.regtest.agent.MainWrapper$MainThread.run(MainWrapper.java:127)
        at java.lang.Thread.run(Thread.java:855)
```

**Expected behavior**
The resaon of this testcase run fail, is the command `jgroup -v` return non zero.

![image](https://user-images.githubusercontent.com/24123821/173037524-6ae43478-7ea7-4b89-a96b-bd71d439ac1f.png)

If we make some modify on this case, then this testcas will pass(add jgroup to BLACKLIST_VERSION String array):

![image](https://user-images.githubusercontent.com/24123821/173038308-d3a92c69-415b-44a2-8a75-d8c40af88429.png)

[...]

Issue #1297, opened 2019-08-22
Non applicable tests for openjdk11-openj9 implementation


M-Davies:
These tests have been permanently excluded since they currently unsuitable for OpenJ9:
- https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/965d0c0df359f3da224b865701bbcc044b7104c2/test/jdk/java/lang/StackWalker/LocalsAndOperands.java#L1 as per https://github.com/eclipse/openj9/issues/6704
- https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/965d0c0df359f3da224b865701bbcc044b7104c2/test/jdk/java/lang/StackWalker/ReflectionFrames.java#L1 as per https://github.com/eclipse/openj9/issues/6724
- https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/965d0c0df359f3da224b865701bbcc044b7104c2/test/jdk/java/lang/invoke/ClassSpecializerTest.java#L1 as per https://github.com/eclipse/openj9/issues/6779

adam-thorpe:
Can i suggest a change to this to help with readability; using the format:
java/lang/String/test1.java      [link to issue]

As it is quite hard to see which tests are in this list

M-Davies:
Done 👍

adam-thorpe:
jdk/internal/misc/TerminatingThreadLocal/TestTerminatingThreadLocal.java https://github.com/eclipse/openj9/issues/7167 11,13

M-Davies:
I noticed that the links to the files were all pointing to JDK11 files and not their own implementations. This has been fixed

adam-thorpe:
java/lang/Class/GetModuleTest.java on jdk11 and 13: https://github.com/eclipse/openj9/issues/3040

adam-thorpe:
java/util/concurrent/atomic/VMSupportsCS8.java on jdk8 and 11: https://github.com/AdoptOpenJDK/openjdk-tests/issues/1347

adam-thorpe:
java/lang/ClassLoader/LibraryPathProperty.java on jdk11+ https://github.com/eclipse/openj9/issues/8440

adam-thorpe:
java/util/logging/TestLoggerWeakRefLeak.java (all versions) https://github.com/AdoptOpenJDK/openjdk-tests/issues/1352

smlambert:
[...]

PR #3758, opened 2022-06-13
Exclude jdk_tools testcase VersionCheck.java

sendaoYan:
Exclude jdk_tools testcase tools/launcher/VersionCheck.java only vendor is alibaba

Fixes: #3751

### DIFF
--- a/openjdk/excludes/vendors/alibaba/ProblemList_openjdk8.txt
+++ b/openjdk/excludes/vendors/alibaba/ProblemList_openjdk8.txt
@@ -94,6 +94,7 @@
 ############################################################################
 
 # jdk_tools
+tools/launcher/VersionCheck.java https://github.com/adoptium/aqa-tests/issues/3751 generic-all
 
 ############################################################################
 


smlambert (commented):
This LGTM, but I will note that we will be adding auto-reenablement of tests in the future, where we check the status of the issue, and if its closed, as https://github.com/adoptium/aqa-tests/issues/3751 will be once this is merged, we will automatically try to run the test again to reinclude it.  For exclusions that will be "permanent", we may want to use a single issue that tracks all of the individual issues that we are covering as part of this effort to remove testcases that are not applicable to a Dragonwell implementation.  (Much like we did for OpenJ9, and then we can tell our automation to ignore re-enabling if issueX is referenced in the problemlist.

sendaoYan:
@smlambert 
Do you mean all the testcases which not applicable to Dragonwell implementation, should track in a single issue?
Bacause https://github.com/adoptium/aqa-tests/pull/3759 has been merged, we can track all the others in #3751.

smlambert:
I mean that we can create an issue similar to this one, https://github.com/adoptium/aqa-tests/issues/1297 but for Dragonwell, then it can be used as the reference issue in the problem list (and we would link all of the specific issues to it, so we can see the details and good explanation of each testcase failure that you've investigated).

I have created this issue, https://github.com/adoptium/aqa-tests/issues/3763 which we can use.

openjdk/excludes/vendors/alibaba/ProblemList_openjdk8.txt
```diff
@@ -94,6 +94,7 @@
 ############################################################################
 
 # jdk_tools
+tools/launcher/VersionCheck.java https://github.com/adoptium/aqa-tests/issues/3751 generic-all
```
smlambert, line 97:
```suggestion
tools/launcher/VersionCheck.java https://github.com/adoptium/aqa-tests/issues/3763 generic-all
```

smlambert (commented):
LGTM, but suggest a change to reference the umbrella issue.  It is still good and important to have the individual issues that you are raising to explain the details, and link those individual issues to the umbrella issue #3763

openjdk/excludes/vendors/alibaba/ProblemList_openjdk8.txt
sendaoYan: OK

smlambert (approved):
👍

PR #3780, opened 2022-06-21
Only use <impl>ibm</impl> disable blocks where it applies

Mesbah-Alam:
This PR simplifies existing disable blocks in tck playlists by removing unnecessary `<impl>ibm</impl>` disable blocks. The only impl=ibm is zos JDK11 atm. 


llxia (approved):
LGTM

PR #3849, opened 2022-07-13
Perm exclude java/lang/StringBuffer/HugeCapacity.java

mikezhang1234567890:
Similar to java/lang/StringBuilder/HugeCapacity.java (see https://github.com/adoptium/aqa-tests/pull/2141), this test is not
compatibile with OpenJ9.

The test assumes `new byte[Integer.MAX_VALUE` throws an OOM error, but
it does not on OpenJ9.

Excluding from jdk versions 17+ since it does not exist before this.

fyi: @pshipton


llxia (approved):

llxia:
@mikezhang1234567890 please also backport this PR into https://github.com/adoptium/aqa-tests/tree/v0.9.3-release. Thanks

renfeiw (approved):
LGTM

Issue #3861, opened 2022-07-15
run sanity.jck fail because of java.lang.NumberFormatException: For input string: "8b"


sendaoYan:
**Describe the bug**
run sanity.jck fail because of java.lang.NumberFormatException: For input string: "8b"

```log
mkdir -p "/home/yansendao/aqa-tests-jdk8/TKG/../TKG/output_16578765208911/jck-runtime-api-java_lang-instrument_0"; \
cd "/home/yansendao/aqa-tests-jdk8/TKG/../TKG/output_16578765208911/jck-runtime-api-java_lang-instrument_0"; \
"/home/yansendao/software/jdk/2022-05/dragonwell-8.11.12/bin/java" -Djvm.options=" " -Dother.opts="" -cp /home/yansendao/aqa-tests-jdk8/TKG/../jck/jtrunner/bin JavaTestRunner resultsRoot="/home/yansendao/aqa-tests-jdk8/TKG/../TKG/output_16578765208911/jck-runtime-api-java_lang-instrument_0" testRoot=/home/yansendao/aqa-tests-jdk8/TKG/.. tests=api/java_lang/instrument jckRoot=/home/yansendao/jck/jck8b jckversion=8b testsuite=RUNTIME; \
if [ $? -eq 0 ] ; then echo ""; echo "jck-runtime-api-java_lang-instrument_0""_PASSED"; echo ""; cd /home/yansendao/aqa-tests-jdk8/TKG/..;  else echo ""; echo "jck-runtime-api-java_lang-instrument_0""_FAILED"; echo ""; fi; } 2>&1 | tee -a "/home/yansendao/aqa-tests-jdk8/TKG/../TKG/output_16578765208911/TestTargetResult";
jckversion determined to be jck8b
test-args jckversion 8b does not match actual jckversion jck8b. Using actual jckversion jck8b
Using initial excludes list file /home/yansendao/jck/jck8b/JCK-runtime-8b/lib/jck8b.jtx
Unable to find additional excludes list file /home/yansendao/jck/jck8b/excludes/jck8b.jtx
Unable to find known failures list file /home/yansendao/jck/jck8b/excludes/jck8b.kfl
java.lang.NumberFormatException: For input string: "8b"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at java.lang.Integer.parseInt(Integer.java:615)
[...]

PR #3863, opened 2022-07-18
Jck test support jck8b

sendaoYan:
Some openjdk vendor can only obtain the jck test suite of the jck8b version, so the aqavit test framework needs to support both jck8b and jck8c

Fixes: #3861


Mesbah-Alam:
This change should be fine; checked with a grinder with IBM internal (8c) repo : Grinder/25815.

smlambert (approved):

Issue #3875, opened 2022-07-21
sanity.openjdk and extended.openjdk overwrite jtreg test report


sendaoYan:
**Describe the bug**
All the sanity.openjdk and extended.openjdk test target run with the same test report directory, so finally the report directory only contains the report of last test target.

**Expected behavior**
Every test target has a unique report directory. After all the test targets run finish, we can count which jtreg testcases has been run.

**Screenshots**

![image](https://user-images.githubusercontent.com/24123821/180167573-abc9b48a-e592-4068-8ad6-0c6d7df737e9.png)

**patch**

```diff
diff --git a/openjdk/playlist.xml b/openjdk/playlist.xml
index 5e1ff3cc..54aff297 100644
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -25,7 +25,7 @@
        $(JTREG_BASIC_OPTIONS) $(CUSTOM_NATIVE_OPTIONS) -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS) $(VMOPTION_HEADLESS)$(Q) \
        $(TIMEOUT_HANDLER) \
        -w $(Q)$(REPORTDIR)$(D)work$(Q) \
-       -r $(Q)$(TEST_RESROOT)$(D)report$(Q) \
+       -r $(Q)$(REPORTDIR)$(D)report$(Q) \
        -jdk:$(Q)$(TEST_JDK_HOME)$(Q) \
        $(Q)$(JTREG_JDK_TEST_DIR)$(D)$(JDK_CUSTOM_TARGET)$(Q); \
        $(TEST_STATUS)</command>
@@ -54,7 +54,7 @@
        $(JTREG_BASIC_OPTIONS) $(CUSTOM_NATIVE_OPTIONS) -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS) $(VMOPTION_HEADLESS)$(Q) \
        $(TIMEOUT_HANDLER) \
        -w $(Q)$(REPORTDIR)$(D)work$(Q) \
-       -r $(Q)$(TEST_RESROOT)$(D)report$(Q) \
+       -r $(Q)$(REPORTDIR)$(D)report$(Q) \
        -jdk:$(Q)$(TEST_JDK_HOME)$(Q) \
        $(Q)$(JTREG_LANGTOOLS_TEST_DIR)$(D)$(LANGTOOLS_CUSTOM_TARGET)$(Q); \
[...]

PR #3879, opened 2022-07-22
All the sanity.openjdk and extended.openjdk generate unique jtreg report directory

sendaoYan:
All the sanity.openjdk and extended.openjdk generate unique jtreg report directory, After all the test targets run finish, we can count which jtreg testcases has been run.

Fixes: #3875

### DIFF
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -25,7 +25,7 @@
 	$(JTREG_BASIC_OPTIONS) $(CUSTOM_NATIVE_OPTIONS) -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS) $(VMOPTION_HEADLESS)$(Q) \
 	$(TIMEOUT_HANDLER) \
 	-w $(Q)$(REPORTDIR)$(D)work$(Q) \
-	-r $(Q)$(TEST_RESROOT)$(D)report$(Q) \
+	-r $(Q)$(REPORTDIR)$(D)report$(Q) \
 	-jdk:$(Q)$(TEST_JDK_HOME)$(Q) \
 	$(Q)$(JTREG_JDK_TEST_DIR)$(D)$(JDK_CUSTOM_TARGET)$(Q); \
 	$(TEST_STATUS)</command>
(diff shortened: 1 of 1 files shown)


smlambert:
FYI to reviewers - this will increase what gets archived (though presumably not too much), unlikely to affect those pushing artifacts to artifactory, but may affect those keeping them on Jenkins.

sendaoYan:
> FYI to reviewers - this will increase what gets archived (though presumably not too much), unlikely to affect those pushing artifacts to artifactory, but may affect those keeping them on Jenkins.

Take test target jdk_util_0 for example, the jtreg report directory occupy 700KB disk space. The contents of the folder are relatively fixed.
On the other hand, the jtreg work directiry occupy 114MB disk space, which contains all the class files and all the intermediate files from running tests, such as hs_err_pid and core dump files.
So I think, compare the jtreg work directory, the jtreg report directory disk usage is acceptable.


image

openjdk/playlist.xml
```diff
@@ -25,7 +25,7 @@
 	$(JTREG_BASIC_OPTIONS) $(CUSTOM_NATIVE_OPTIONS) -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS) $(VMOPTION_HEADLESS)$(Q) \
 	$(TIMEOUT_HANDLER) \
 	-w $(Q)$(REPORTDIR)$(D)work$(Q) \
-	-r $(Q)$(TEST_RESROOT)$(D)report$(Q) \
+	-r $(Q)$(REPORTDIR)$(D)report$(Q)) \
```
llxia, line 28: I think there is an extra `)` at the end of the line. Same as the rest changes.

sendaoYan: All the extra `)` has been fixed. Please review it again, thanks.

llxia:
This PR will result in the artifact size to increase (for both Jenkins and Artifactory), but the increase seems to be small. Also, since we only archive `TKG/output_*` (a.k.a `REPORTDIR`) when there is a failure, I think it only has a limited impact.

llxia (approved):
LGTM

smlambert:
Grinder to check syntax is all good: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5303

smlambert (approved):
thanks @sendaoYan

Issue #3901, opened 2022-08-01
MissingResourceCauseTestRun.java/HashedPasswordFileTest.java/JImageExtractTest.java run fail by root user


sendaoYan:
**Describe the bug**
java/util/ResourceBundle/Control/MissingResourceCauseTestRun.java run fail by root user

```log
STDOUT:
Command line: [chmod 000 UnreadableRB.properties]
[2022-08-01T03:34:03.190952969Z] Gathering output for process 60223
[2022-08-01T03:34:03.211121049Z] Waiting for completion for process 60223
[2022-08-01T03:34:03.211226766Z] Waiting for completion finished for process 60223
Output and diagnostic info for process 60223 was saved into 'pid-60223-output.log'



[2022-08-01T03:34:03.213681223Z] Waiting for completion for process 60223
[2022-08-01T03:34:03.213758593Z] Waiting for completion finished for process 60223
Command line: [/mnt_array/yansendao.ysd/software/jdk/2022-07/corretto/x64/amazon-corretto-17.0.4.8.1-linux-x64/bin/java -ea -esa -cp /mnt_array/yansendao.ysd/git/dragonwell17/tmp/classes/java/util/ResourceBundle/Control/MissingResourceCauseTestRun.d:/mnt_array/yansendao.ysd/git/dragonwell17/test/jdk/java/util/ResourceBundle/Control:. MissingResourceCauseTest]
[2022-08-01T03:34:03.248908701Z] Gathering output for process 60228
[2022-08-01T03:34:03.344730939Z] Waiting for completion for process 60228
[2022-08-01T03:34:03.344829143Z] Waiting for completion finished for process 60228
Output and diagnostic info for process 60228 was saved into 'pid-60228-output.log'
Exception in thread "main" java.lang.RuntimeException: getBundle("UnreadableRB") doesn't throw class java.io.IOException
        at MissingResourceCauseTest.callGetBundle(MissingResourceCauseTest.java:51)
        at MissingResourceCauseTest.callGetBundle(MissingResourceCauseTest.java:43)
        at MissingResourceCauseTest.main(MissingResourceCauseTest.java:37)

[2022-08-01T03:34:03.345549329Z] Waiting for completion for process 60228
[2022-08-01T03:34:03.345626772Z] Waiting for completion finished for process 60228
Command line: [chmod 666 UnreadableRB.properties]
[2022-08-01T03:34:03.349959898Z] Gathering output for process 60251
[...]

PR #3902, opened 2022-08-01
Exclude MissingResourceCauseTestRun.java/HashedPasswordFileTest.java/JImageExtractTest.java because of run fail by root user

sendaoYan:
The testcases of dragonwell feature multi-tenant must run by root user, but MissingResourceCauseTestRun.java/HashedPasswordFileTest.java/JImageExtractTest.java run fail by root user. So we shoule exclude these testcases by problemlist.

Fixes: #3901

### DIFF
--- a/openjdk/excludes/vendors/alibaba/ProblemList_openjdk17.txt
+++ b/openjdk/excludes/vendors/alibaba/ProblemList_openjdk17.txt
@@ -42,6 +42,7 @@
 ############################################################################
 
 # jdk_management
+javax/management/security/HashedPasswordFileTest.java https://github.com/adoptium/aqa-tests/issues/3763 generic-all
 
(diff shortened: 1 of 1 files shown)


smlambert (commented):
This PR contains 3 testcases for exclusion, not just the testcase mentioned in the heading.  Is that intentional @sendaoYan ?

sendaoYan:
> This PR contains 3 testcases for exclusion, not just the testcase mentioned in the heading. Is that intentional @sendaoYan ?

I had combined two commit logs to one, so the heading message lose the other testcases. I have correct the description in heading.
In  https://github.com/adoptium/aqa-tests/issues/3901#issuecomment-1200808470, I also descript it.

smlambert (approved):

Issue #3911, opened 2022-08-07
dragonwell standard version run dragonwell8_feature_jdk0 test target error


sendaoYan:
**Describe the bug**
dragonwell standard version run dragonwell8_feature_jdk0 test target error, which complaint Error: Cannot find file: test/jdk/com/alibaba/wisp2

```log
===============================================
Running test dragonwell8_feature_jdk0_0 ...
===============================================
dragonwell8_feature_jdk0_0 Start Time: Sun Aug  7 20:16:24 2022 Epoch Time (ms): 1659874584004
Nothing to be done for setup.
variation: NoOptions
JVM_OPTIONS:  
{ itercnt=1; \
mkdir -p "/mnt_array/yansendao.ysd/202207/aqa-tests-std11/TKG/../TKG/output_16598745833020/dragonwell8_feature_jdk0_0"; \
cd "/mnt_array/yansendao.ysd/202207/aqa-tests-std11/TKG/../TKG/output_16598745833020/dragonwell8_feature_jdk0_0"; \
"/mnt_array/yansendao.ysd/software/jdk/2022-07/std/dragonwell-11.0.16.12+8-GA/bin/java" -Xmx512m -jar "/mnt_array/yansendao.ysd/202207/aqa-tests-std11/TKG/../../jvmtest/openjdk/jtreg/lib/jtreg.jar" \
                -agentvm -a -ea -esa -v:fail,error,time,nopass -retain:fail,error,*.dmp,javacore.*,heapdump.*,*.trc -ignore:quiet -timeoutFactor:8 -xml:verify -concurrency:24  -vmoptions:"-Xmx512m  " \
                -w ""/mnt_array/yansendao.ysd/202207/aqa-tests-std11/TKG/../TKG/output_16598745833020/dragonwell8_feature_jdk0_0"/work" \
                -r ""/mnt_array/yansendao.ysd/202207/aqa-tests-std11/TKG/../TKG/output_16598745833020/dragonwell8_feature_jdk0_0"/report" \
                -jdk:"/mnt_array/yansendao.ysd/software/jdk/2022-07/std/dragonwell-11.0.16.12+8-GA" \
                -exclude:"/mnt_array/yansendao.ysd/202207/aqa-tests-std11/TKG/../openjdk/openjdk-jdk/test/jdk/ProblemList.txt" \
[...]

PR #3912, opened 2022-08-07
Skip dragonwell feature test target when the test jdk is dragonwell standard version

sendaoYan:
Skip dragonwell feature test target when the test jdk is dragonwell standard version. When the test jdk is dragonwell standard version, export TEST_FLAG=Dragonwell_Standard

Fixes: #3911


llxia:
Just a heads up, as TEST_FLAG becomes more popular, we will set up a naming convention for it soon. (Suggestions are welcome!) The current TEST_FLAG value may or may not need to be adjusted by then.

llxia (approved):

sophia-guo (approved):

PR #3919, opened 2022-08-11
uplift: GHA version and use correct parameter

zdtsw:


smlambert (approved):

llxia:
Currently, we do not have a good way to test workflow before merging. I created an issue for looking into this: https://github.com/adoptium/aqa-tests/issues/3920

llxia (approved):

PR #3926, opened 2022-08-18
(v0.9.4.1-release) : Update jdk11 and 17 tags

Haroon-Khel:
Update the jdk11 and 17 tags to jdk-11.0.16.1+1 and jdk-17.0.4.1+1 respectively


smlambert (approved):
thanks @Haroon-Khel !

PR #3928, opened 2022-08-23
feat: add customized timeout for "Test" stage

zdtsw:
Job like smoketest (e.g https://ci.adoptopenjdk.net/job/build-scripts/job/jobs/job/jdk/job/jdk-alpine-linux-x64-temurin_SmokeTests/) should be finished within 5 mins but due to lack of timeout set on the stage,  by default , it timeout after 10hrs for the pipeline.
Even the job has parameter set TIME_LIMIT to 10, we cannot overwrite it per each run.

This PR is to add the same function which already in testBuild() but for runTest()

Cross ref https://github.com/adoptium/temurin-build/issues/3031

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -609,7 +609,7 @@ def buildTest() {
 
 def runTest( ) {
 	stage('Test') {
-		TIME_LIMIT =  params.TIME_LIMIT ? params.TIME_LIMIT.toInteger() : 10
+		TIME_LIMIT = params.TIME_LIMIT ? params.TIME_LIMIT.toInteger() : 10
 		timeout(time: TIME_LIMIT, unit: 'HOURS') {
 			timestamps{
 				echo 'Running tests...'
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -609,67 +609,70 @@ def buildTest() {
 
 def runTest( ) {
 	stage('Test') {
-		timestamps{
-			echo 'Running tests...'
-			def CUSTOM_OPTION = ''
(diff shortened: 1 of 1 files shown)


buildenv/jenkins/JenkinsfileBase
```diff
@@ -609,67 +609,70 @@ def buildTest() {
...
-			echo 'Running tests...'
-			def CUSTOM_OPTION = ''
-
-			TARGET = "${params.TARGET}";
-			if (TARGET.contains('custom') && CUSTOM_TARGET!='') {
-				if (TARGET == 'system_custom') {
-					env.SYSTEM_CUSTOM_TARGET=CUSTOM_TARGET
-				} else {
-					CUSTOM_OPTION = "${TARGET.toUpperCase()}_TARGET='${CUSTOM_TARGET}'"
+		TIME_LIMIT =  params.TIME_LIMIT ? params.TIME_LIMIT.toInteger() : 10
```
karianna, line 612:
```suggestion
		TIME_LIMIT = params.TIME_LIMIT ? params.TIME_LIMIT.toInteger() : 10
```

karianna (changes requested):

karianna:
A lot of whitespace was changed in this PR.  Do you have the correct space/tab stops set up?

zdtsw:
> A lot of whitespace was changed in this PR. Do you have the correct space/tab stops set up?

hum, could be I move the whole block with one tab forwards in vscode? but by default i have 4 spaces for indentation

smlambert:
This PR is needed, you would set the TIME_LIMIT parameter for the smoketest job upon launching it,  since https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L787

In future, to avoid extra work for you or reviewers, perhaps best to ask or discuss the need for a change before creating a PR.

zdtsw:
We can close this PR and do the change with a short timeout value in the jenkins side.

PR #3942, opened 2022-08-30
Consume vendor specific exclude list

Mesbah-Alam:
Consume vendor specific exclude list
Related : backlog/issues/885

Signed-off-by: Mesbah-Alam <Mesbah_Alam@ca.ibm.com>

PR #3952, opened 2022-09-07
If no custom JTX file is supplied, ensure it's not null

Mesbah-Alam:
This PR fixes an issue introduced by https://github.com/adoptium/aqa-tests/pull/3942 : If no custom JTX file is supplied, we need to ensure it's not null. Otherwise grinder jobs with tck targets will fail with a null value in generated command jtb. 


Mesbah-Alam:
Test : Grinder_JCK/655

llxia (approved):

LongyuZhang (approved):
LGTM

PR #3968, opened 2022-09-14
Define -jvmArgs in systemtest.mk for JCStress

Mesbah-Alam:
Define `APPLICATION_OPTIONS` in systemtest.mk and compose the JCStress option `-jvmArgs` therein. 

### DIFF
--- a/system/systemtest.mk
+++ b/system/systemtest.mk
@@ -64,6 +64,11 @@ ifeq (,$(findstring $(JDK_IMPL),hotspot))
   JAVA_ARGS += -Xdump:system:events=user
 endif
 
+APPLICATION_OPTIONS :=
+ifdef JVM_OPTIONS
+  APPLICATION_OPTIONS := -jvmArgs $(Q)$(JVM_OPTIONS)$(Q)
+endif
+
 define SYSTEMTEST_CMD_TEMPLATE
 perl $(SYSTEMTEST_RESROOT)$(D)STF$(D)stf.core$(D)scripts$(D)stf.pl \
   -test-root=$(Q)$(SYSTEMTEST_RESROOT)$(D)STF;$(SYSTEMTEST_RESROOT)$(D)aqa-systemtest$(OPENJ9_PRAM)$(Q) \
--- a/system/jcstress/playlist.xml
+++ b/system/jcstress/playlist.xml
@@ -13,9 +13,10 @@
 # limitations under the License.
 -->
 <playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../TKG/playlist.xsd">
+	<include>../systemtest.mk</include>
 	<test>
 		<testCaseName>jcstress_SampleTestBench</testCaseName>
-		<command>$(JAVA_COMMAND) $(JVM_OPTIONS) -jar $(Q)$(LIB_DIR)$(D)jcstress-latest.jar$(Q) -jvmArgs $(Q)$(JVM_OPTIONS) -t SampleTestBench$(Q); \
+		<command>$(JAVA_COMMAND) $(JVM_OPTIONS) -jar $(Q)$(LIB_DIR)$(D)jcstress-latest.jar$(Q) $(APPLICATION_OPTIONS) -t SampleTestBench; \
 		$(TEST_STATUS)</command>
 		<levels>
 			<level>extended</level>


Mesbah-Alam:
Test : https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5693/console

system/systemtest.mk
```diff
@@ -64,6 +64,11 @@ ifeq (,$(findstring $(JDK_IMPL),hotspot))
   JAVA_ARGS += -Xdump:system:events=user
 endif
 
+APPLICATION_OPTIONS :=
+ifdef JVM_OPTIONS
+  APPLICATION_OPTIONS += -jvmArgs $(JVM_OPTIONS)
```
llxia, line 69: - instead of `+=`, we just need `=`
- Please add `$(Q)` around `$(JVM_OPTIONS)`. `$(JVM_OPTIONS)` may contain multiple jvm options.

llxia:
We do not have `Unrecognized option: -Xt` anymore, but there is an error - `FATAL: No matching tests.` in Grinder.

system/systemtest.mk
```diff
@@ -64,6 +64,11 @@ ifeq (,$(findstring $(JDK_IMPL),hotspot))
   JAVA_ARGS += -Xdump:system:events=user
 endif
 
+APPLICATION_OPTIONS :=
+ifdef JVM_OPTIONS
+  APPLICATION_OPTIONS = $(Q)-jvmArgs $(JVM_OPTIONS)$(Q)
```
llxia, line 69: This should be `APPLICATION_OPTIONS = -jvmArgs $(Q)$(JVM_OPTIONS)$(Q)`

Mesbah-Alam:
The JVM_OPTIONS doesn't seem to be getting passed properly via systemtest.mk. Debugging. Putting this PR back to draft.

llxia:
> The JVM_OPTIONS doesn't seem to be getting passed properly via systemtest.mk. Debugging. Putting this PR back to draft.

This is because you did not set variations in the playlist.

Mesbah-Alam:
> This is because you did not set variations in the playlist.

I see.. I thought setting `JVM_OPTIONS` in the input parameter for the Grinder was going to cause JVM_OPTIONS in systemtest.mk to have that value.

smlambert:
> The JVM_OPTIONS doesn't seem to be getting passed properly via systemtest.mk. Debugging. Putting this PR back to draft.
> 
> This is because you did not set variations in the playlist.
> 
Or append or replace JVM_OPTIONS via the Grinder parameters:
Screen Shot 2022-09-14 at 3 28 56 PM

But that is what we are checking, if JVM_OPTIONS is not defined, do not pass in arguments to jcstress.  Is -jvmArgs a required parameter for jcstress?

---

other clean up notes (not related to this PR):
For some reason (prior to this PR), a new variable is introduced in the systemtest.mk file,  [JAVA_ARGS](https://github.com/adoptium/aqa-tests/blob/master/system/systemtest.mk#L62) ... do not think we need to rename the JVM_OPTIONS variable before using it in the SYSTEMTEST_CMD_TEMPLATE.  

Also, this file is not named following the convention of the others  (named after the test group, in which case would be called system.mk).  Not asking for it to be changed in this PR, but noting it is not like openjdk/openjdk.mk, perf/perf.mk, jck/jck.mk.

llxia:
I think you need `:=` in 

`APPLICATION_OPTIONS := -jvmArgs $(Q)$(JVM_OPTIONS)$(Q)`

Mesbah-Alam:
> append or replace JVM_OPTIONS via the Grinder parameters

Supplying a value in JVM_OPTIONS in Grinder parameter doesn't seem to cause $(JVM_OPTIONS) in systemtest.mk to have that value. 

@llxia - any idea why this is happening? 

I tried with your suggestion in https://github.com/adoptium/aqa-tests/pull/3968#issuecomment-1247218532 too : Grinder/27559.

Mesbah-Alam:
Tried with a "variation" defined in playlist too (Grinder/27560). Same result. I believe the logic in systemtest.mk is wrong: 

```
APPLICATION_OPTIONS :=
ifdef JVM_OPTIONS
  APPLICATION_OPTIONS := -jvmArgs $(Q)$(JVM_OPTIONS)$(Q)
endif
```

llxia:
systemtest.mk is not included in jcstress/playlist.xml

Mesbah-Alam:
> systemtest.mk is not included in jcstress/playlist.xml

That was it! Thanks @llxia ! 

The JVM_OPTION is now being passed properly: Grinder/27561. 

We still have "FATAL: No matching tests" though. But I feel we can remove the "place holder" target `jcstress_SampleTestBench` altogether after the "real" targets get added.

Mesbah-Alam:
> other clean up notes (not related to this PR).. 

I will have a separate PR for these updates.

llxia (approved):
LGTM

smlambert (approved):
thanks @Mesbah-Alam !

PR #3971, opened 2022-09-16
Depend on getDependentLibs to fix local build of jcstress targets

smlambert:
Fixes #3956

Issue #3921, opened 2022-08-15
3rd party libs should be downloaded via getDependency job


llxia:
Some 3rd libs get downloaded at runtime (see the links below). 

https://github.com/adoptium/aqa-tests/blob/1898c35e22a6100c7be9d9270cad110bfe9b2d52/perf/dacapo/build.xml#L37
https://github.com/adoptium/aqa-tests/blob/f9c915a44322fab060e3dc13c5db0ce4b5dddc52/perf/renaissance/build.xml#L37

Depending on the network status and 3rd party server availability, we may run into the issue of having Invalid or corrupt jarfile (see runtimes/backlog/issues/872). We should pre-download the 3rd party lib via https://ci.adoptopenjdk.net/job/test.getDependency/

code is in TKG: https://github.com/adoptium/TKG/blob/ae395bd9f8dff8ab73e39608916955147308e1aa/scripts/getDependencies.pl#L62

PR #3987, opened 2022-09-22
Add SHA1SUMS to list of files to archive in getDependency script

Mesbah-Alam:
Add SHA1SUMS to the list of files to be archived 

### DIFF
--- a/buildenv/jenkins/getDependency
+++ b/buildenv/jenkins/getDependency
@@ -37,7 +37,7 @@ def testBuild() {
 			} else {
 				sh 'curl -Os https://raw.githubusercontent.com/adoptium/TKG/master/scripts/getDependencies.pl'
 				sh 'perl ./getDependencies.pl -path . -task default'
-				archiveArtifacts '*.jar, *.zip, *.txt, *.gz'
+				archiveArtifacts '*.jar, *.zip, *.txt, *.gz, SHA1SUMS'
 			}
 		} finally {
 			cleanWs()


Mesbah-Alam:
Test : Grinder/27996.

llxia:
It is correct to have getDependentLibs. We should not remove it. Please see the [internal issue](https://github.ibm.com/runtimes/openj9-openjdk-jdk11-zos/issues/1623) for details.

smlambert:
Note my comment https://github.com/adoptium/aqa-tests/pull/3971#issuecomment-1249412586

I think that it'd be an update to [getDependentLibs ant target](https://github.com/adoptium/TKG/blob/master/scripts/getDependencies.xml#L26) in TKG that should check if they are already present and not re-fetch.

llxia:
That is correct, Shelley. It is correct to have getDependentLibs. Mesbah has updated the PR to revert getDependentLibs change. And this PR is about archiving the SHA1SUMS file in getDependency.

llxia (approved):
LGTM

Mesbah-Alam:
> I guess that won't be the place to do it, easier to put the logic into the jcstress/build.xml file, something along the lines of this https://github.com/adoptium/aqa-tests/pull/3697 (check if file exists, if so do not trigger download again).

This could be a second pr?

llxia:
The 3rd party libs download logic is in getDependentLibs in TKG. We have a pre-stage job (test.getDependency) to call it and pre-download all libs to prevent any network/server hiccups. And if the pre-staged copy does not exist or is not up to date (sha check), it will download it again. 
https://github.com/adoptium/aqa-tests/pull/3697 sets us back to the download at runtime. It causes problems when the network is unstable. There is an open issue to update the logic: https://github.com/adoptium/aqa-tests/issues/3921

Just to clarify, the current problem is because get dependency did not set up properly for jcstress.

Mesbah-Alam:
Tested this PR in test.getDependency/10391.

llxia:
Thanks @Mesbah-Alam to deliver the changes via https://github.com/adoptium/aqa-tests/pull/3990 and https://github.com/adoptium/TKG/pull/359. This PR is no longer needed for now.

Issue #3992, opened 2022-09-26
jdk11_tier1_cipher(sanity.openjdk) is subset of jdk_security2(extened.openjdk)


sendaoYan:
**Describe the bug**
jdk11_tier1_cipher is subset of jdk_security2

jdk11_tier1_cipher:

![image](https://user-images.githubusercontent.com/24123821/192240925-ad0127b9-6925-4744-a5b0-dae8ebc8fae0.png)

jdk_security2:

![image](https://user-images.githubusercontent.com/24123821/192241057-a1dfb449-8fdf-4d9c-b932-ea0ae3526d32.png)

The test/jdk/TEST.groups descripte jdk_security2 test group include all testcases of jdk11_tier1_cipher(com/sun/crypto/provider/Cipher)

![image](https://user-images.githubusercontent.com/24123821/192242189-ad080be5-d27f-4367-89ad-cda7a046500e.png)

**Expected behavior**
The sanity.openjdk and extended.openjdk test target has no any duplicate testcases.

smlambert:
Indeed we should not duplicate testcases.  

Some options:
- simplest: remove jdk11_tier1_cipher target (if we really want to be running these tests more regularly, thus keep in sanity level)
- alternate: move the other 3 dirs in jdk_security2 into a different target or targets that stay in extended level
- alternate: remove jdk11_tier1_cipher and move jdk_security2 into sanity level

PR #3994, opened 2022-09-26
Delete jdk11_tier1_cipher for duplicate testcases

sendaoYan:
jdk11_tier1_cipher(sanity.openjdk) is subset of jdk_security2(extened.openjdk), which cause there are 91 duplicate testcases in sanity.openjdk and extended.openjdk. Delete jdk11_tier1_cipher in openjdk/playlist.xml to avoid duplicate testcases

Fixes: #3992

### DIFF
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -1755,34 +1755,6 @@
 		</groups>
 	</test>
 	<test>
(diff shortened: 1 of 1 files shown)


llxia (approved):
LGTM

PR #4028, opened 2022-10-11
Allow any project to copyArtifact from test jobs

AdamBrousseau:
When a Jenkins instance doesn't allow anonymous read by default, the copyArtifact plugin step will fail unless the job with the artifacts has copyArtifactPermission enabled with a set of projects listed or a wildcard to allow any job.

### DIFF
--- a/buildenv/jenkins/testJobTemplate
+++ b/buildenv/jenkins/testJobTemplate
@@ -195,6 +195,9 @@ ARCH_OS_LIST.each { ARCH_OS ->
 								}
 							}
 						}
+						copyArtifactPermission {
+							projectNames('*')
+						}
 					}
 
 					definition {


AdamBrousseau:
fyi @llxia 
Tested internally
Test_Job_Auto_Gen/48769/
Which generated
job/Test_openjdknext_j9_sanity.openjdk_x86-64_windows_Nightly

I'm not sure if the children will automatically get this as well or if we need to force child regeneration. We can discuss tomorrow.

llxia:
> I'm not sure if the children will automatically get this as well or if we need to force child regeneration. We can discuss tomorrow.

No, test jobs will not get autogenerated by default if they already exist. To populate this change, we need to force child regeneration.

smlambert (approved):

Issue #4006, opened 2022-10-04
Could we have spotted the missing `jvmti` support in JDK19 on non-x64/aarch64 platforms?


sxa:
Under some of the testing performed during the release cycle for JDK19 we discovered that the `--enable-preview` option resulted in `jvmti` support not being available on certain platforms (Notably the ones where a full Virtual Threads implementation was not available - everything except the x86-64 and aarch64 platforms).

Could we have pre-emptively caught this as part of the regular test cycles in order to avoid hitting the problem during the release cycle, and if not is there anything we could put in place to get early warning of such a problem in the future?

smlambert:
The answer is yes, if someone had been triaging the jdk19 pipelines, those failures were present in nightly and weekly runs. 

Example: 
https://trss.adoptium.net/output/test?id=63130ae993ac60470b274021

```
[2022-09-03T04:15:39.295Z] Exception in thread "main" java.lang.UnsupportedOperationException: Unsupported os, arch, or address size: Linux, s390x, 64
[2022-09-03T04:15:39.295Z] 	at java.base/jdk.internal.foreign.CABI.current(CABI.java:69)
[2022-09-03T04:15:39.295Z] 	at java.base/jdk.internal.foreign.abi.SharedUtils.getSystemLinker(SharedUtils.java:237)
[2022-09-03T04:15:39.295Z] 	at java.base/java.lang.foreign.Linker.nativeLinker(Linker.java:198)
[2022-09-03T04:15:39.295Z] 	at ImplicitAttach.main(ImplicitAttach.java:48)
```

Also if the advisement of the [release checklist](https://github.com/adoptium/adoptium/blob/main/.github/ISSUE_TEMPLATE/release-checklist.md) to **run a trial release pipeline** a week ahead of the release had been followed, these failures would have been spotted.

<img width="1151" alt="Screen Shot 2022-10-04 at 12 52 56 PM" src="https://user-images.githubusercontent.com/2836948/193879870-38e900b0-7864-4481-a2d7-db0dfb7bc62f.png">

Looks like it was removed from https://github.com/adoptium/adoptium/issues/171.

Closing this issue, as the 'action' to take is to monitor the test pipelines, and follow the release checklist.  

Noting that testing is still disabled, so any nightly monitoring will happen once testing is re-enabled, which likely needs to be added to the checklist to remind release champions. (see https://github.com/adoptium/adoptium/issues/177).

sxa:
[...]

PR #3970, opened 2022-09-16
Add serviceability for JDK19+ openj9

llxia:
Enable hotspot serviceability jvmti test to run with openj9.
Set `-nativepath:"$(TESTIMAGE_PATH)/hotspot/jtreg/native"`
Since this is a hotspot test, some subtests may require minor tweaks.

related: https://github.com/eclipse-openj9/openj9/issues/15748
Signed-off-by: lanxia <lan_xia@ca.ibm.com>

PR #4035, opened 2022-10-13
Enable serviceability_jvmti tests for 19+

sophia-guo:
Related with #4006 

### DIFF
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -218,6 +218,37 @@
 		</impls>
 	</test>
 	<test>
+		<testCaseName>hotspot_serviceability_jvmti</testCaseName>
+		<variations>
+			<variation>Mode150</variation>
+			<variation>Mode650</variation>
+			<variation>Mode1000</variation>
+		</variations>
+		<command>$(JAVA_COMMAND) -Xmx512m -jar $(Q)$(TEST_RESROOT)$(D)jtreg$(D)lib$(D)jtreg.jar$(Q) \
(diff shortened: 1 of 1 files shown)


sophia-guo:
Grinder on different platforms

https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5774/
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5774/
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5774/
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5774/

sophia-guo:
This test case `hotspot_serviceability_jvmti` can be combined with test case [serviceability_jvmti_j9](https://github.com/adoptium/aqa-tests/blob/master/openjdk/playlist.xml#L220-L250), which is specifying nativepath by `-nativepath:"$(TESTIMAGE_PATH)/hotspot/jtreg/native" ` and I don't think it works. However even if updated as $(JVM_NATIVE_OPTIONS) it probably won't work too for openj9, as openj9 may not build corresponding test libs. 

As mentioned in https://github.com/adoptium/aqa-tests/pull/3970#issuecomment-1250999986 there are 94 tests failed, which is same number of tests with hotspot if nativepath is not set correctly or testimage is missing https://ci.adoptopenjdk.net/job/Grinder/5731/testReport/

sophia-guo:
As we may gradually enable other Loom related tests https://github.com/adoptium/aqa-tests/issues/4006#issuecomment-1278170329 in one test case, which might not work with openj9 or openj9 might not be interested in. I will leave it as is for now, which can be combined easily later if required.

smlambert (approved):

sxa (approved):
LGTM - I had assumed we were running all tests with `--enable-preview` which would cause failures in 19 but on the basis of the grinders that's not the case, so I approvel

PR #4065, opened 2022-10-20
update jdk19 tag and repo for 2022 oct CPU

zdtsw:


zdtsw:
is the change we want for AQA? @sophia-guo @ShelleyLambert

sophia-guo (approved):

sophia-guo:
> is the change we want for AQA? 

Are you talking about the AQA master branch? If yes, no need to update it.

smlambert:
We always want to ensure that the aqa-tests release branch for a particular release has the same tags in testenv.properties file as is used by the build pipelines

We do not know for certain what the tag will be until they are announced.

Issue #4093, opened 2022-10-27
Jenkins: Parallel option should be ignored when num machines is 1


AdamBrousseau:
**Describe the bug**
If the number of machines, and therefore the number of child jobs, equals 1, I don't think we need to bother launching a child job.

**To Reproduce**
Launch a test with a parallel option and a build node label that only matches 1 machine

**Expected behavior**
Don't do parallel as it will create unnecessary overhead

**Screenshots**

**Additional context**

pshipton:
If we make this change, we could also change the parallel default in a grinder from "None" to (my preferred) "NodesByIterations".

PR #4102, opened 2022-10-31
Update logic for Parallel option with num_machines=1

llxia:
When running tests with Parallel option and num_machines=1, run tests serially.
Don't do parallel as it will create unnecessary overhead

resolve: https://github.com/adoptium/aqa-tests/issues/4093
Signed-off-by: Lan Xia <Lan_Xia@ca.ibm.com>

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -87,6 +87,7 @@ def setupEnv() {
 	env.DOCKER_REGISTRY_DIR = params.DOCKER_REGISTRY_DIR ? params.DOCKER_REGISTRY_DIR : ""
 	env.DOCKER_REGISTRY_URL_CREDENTIAL_ID = params.DOCKER_REGISTRY_URL_CREDENTIAL_ID ? params.DOCKER_REGISTRY_URL_CREDENTIAL_ID : ""
 	ITERATIONS = params.ITERATIONS ? "${params.ITERATIONS}".toInteger() : 1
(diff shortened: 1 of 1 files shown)


llxia:
- PARALLEL=Dynamic and NUM_MACHINES is empty: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6077 ===> no parallel was launched as expected
- PARALLEL=Dynamic and NUM_MACHINES=1: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6082 ===> no parallel was launched as expected
- PARALLEL=NodesByIterations and NUM_MACHINES=1: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6081 ===> no parallel was launched as expected

LongyuZhang (approved):
LGTM

smlambert (approved):
thanks @llxia !

AdamBrousseau:
Is NUM_MACHINES calculated or taken only from the param?
Thinking of the case where parallelx4 is requested but only 1 machine satisfies the label(s). Will we end up launching 1 child job or will this change catch that case?

llxia:
> Is NUM_MACHINES calculated or taken only from the param?
Thinking of the case where parallelx4 is requested but only 1 machine satisfies the label(s). Will we end up launching 1 child job or will this change catch that case?


We take NUM_MACHINES from the param and we will do check. It cannot be more than the number of available machines. In the above case, we end up launching 1 child job.

PR #4119, opened 2022-11-07
Add special concurrency for jdk_vector tests

Akira1Saitoh:
Introduce a new parameter `SPECIAL_CONCURRENCY` to set the jtreg concurrency for jdk_vector tests. `SPEICAL_CONCURRENCY` is set to 1 for OpenJ9 Linux AArch64. For other platforms, the calculated value is used as before.

Signed-off-by: Akira Saitoh <saiaki@jp.ibm.com>

### DIFF
--- a/openjdk/openjdk.mk
+++ b/openjdk/openjdk.mk
@@ -89,7 +89,6 @@ JTREG_BASIC_OPTIONS += $(JTREG_TIMEOUT_OPTION)
 JTREG_XML_OPTION = -xml:verify
 JTREG_BASIC_OPTIONS += $(JTREG_XML_OPTION)
 # Add any extra options
-JTREG_BASIC_OPTIONS += $(EXTRA_JTREG_OPTIONS)
 JTREG_KEY_OPTIONS :=
 VMOPTION_HEADLESS :=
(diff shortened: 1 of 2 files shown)


smlambert (commented):
Could this be accomplished by inserting a block between openjdk.mk/L65 and L66 to be something like:

# set JTREG_CONC to 1 if the jdk is openj9 and the platform is linux_aarch64.
ifneq ($(filter openj9 ibm, $(JDK_IMPL)),)
	ifneq ($(filter linux_aarch64, $(SPEC)),)
	    JTREG_CONC := 1
	endif
endif

and leaving the playlist file alone?

llxia:
We only want to set a different concurrency number for `jdk_vector`. The rest of the test should still use the default logic. Therefore, a change has to be made to the playlist for `jdk_vector`.

smlambert:
I see, thanks I should have read the comment more closely.  And presumably this is temporary during development.  👍

llxia:
For testing, please see https://github.com/eclipse-openj9/openj9/issues/15268#issuecomment-1305231257 and https://github.com/eclipse-openj9/openj9/issues/15268#issuecomment-1306031520

llxia (approved):
LGTM

gdams (approved):

smlambert (approved):

PR #4147, opened 2022-11-14
Add hotspot_container and jdk_container test groups

### LINKED ISSUE #4143: Add hotspot_container tests to the openjdk test group
We should be adding the hotspot_containers target to the openjdk/[playlist](https://github.com/adoptium/aqa-tests/blob/master/openjdk/playlist.xml) nearby the other [hotspot_misc](https://github.com/openjdk/jdk17u-dev/blob/master/test/hotspot/jtreg/TEST.groups#L58-L66) targets that we already have (see [hotspot_compiler, hotspot_gc, hotspot_runtime, hotspot_serviceability](https://github.com/adoptium/aqa-tests/blob/master/openjdk/playlist.xml#L93-L219))

```
	<test>
		<testCaseName>hotspot_container</testCaseName>
		<variations>
			<variation>Mode150</variation>
			<variation>Mode650</variation>
			<variation>Mode1000</variation>
		</variations>
	$(JTREG_BASIC_OPTIONS) $(JVM_NATIVE_OPTIONS) -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS) $(VMOPTION_HEADLESS)$(Q) \
	-r $(Q)$(REPORTDIR)$(D)report$(Q) \
	-jdk:$(Q)$(TEST_JDK_HOME)$(Q) \
	-exclude:$(Q)$(JTREG_HOTSPOT_TEST_DIR)$(D)ProblemList.txt$(Q) \
	-exclude:$(Q)$(TEST_RESROOT)$(D)$(PROBLEM_LIST_FILE)$(Q) \
	${FEATURE_PROBLEM_LIST_FILE} \
	${VENDOR_PROBLEM_LIST_FILE} \
	$(Q)$(JTREG_HOTSPOT_TEST_DIR):hotspot_container$(Q); \
	$(TEST_STATUS)</command>
		<versions>
			<version>17+</version>
		</versions>
		<levels>
			<level>extended</level>
		</levels>
		<groups>
			<group>openjdk</group>
		</groups>
		<impls>
			<impl>hotspot</impl>
		</impls>
	</test>
```

There may be an implication here that we need the test machines to have Docker installed, so presume we may need to check whether we need an infrastructure issue also created and of course disable these container tests from the appropriate platforms where it is not available).  For now we can create a PR and run some tests across platforms to see how these tests behave.

The tests may skip themselves from running if machine canTestDocker in which case we may not need to add restrictions in the playlist as to where they can run, TBD.

jerboaa:
Closes #4143

### DIFF
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -91,6 +91,124 @@
 		</impls>
 	</test>
 	<test>
+		<testCaseName>hotspot_container</testCaseName>
+		<variations>
+			<variation>NoOptions</variation>
+		</variations>
+		<command>$(JAVA_COMMAND) -Xmx512m -jar $(Q)$(TEST_RESROOT)$(D)jtreg$(D)lib$(D)jtreg.jar$(Q) \
+	$(JTREG_BASIC_OPTIONS) $(JVM_NATIVE_OPTIONS) -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS) $(VMOPTION_HEADLESS)$(Q) \
+	-w $(Q)$(REPORTDIR)$(D)work$(Q) \
+	-r $(Q)$(REPORTDIR)$(D)report$(Q) \
+	-jdk:$(Q)$(TEST_JDK_HOME)$(Q) \
+	-exclude:$(Q)$(JTREG_HOTSPOT_TEST_DIR)$(D)ProblemList.txt$(Q) \
+	-exclude:$(Q)$(TEST_RESROOT)$(D)$(PROBLEM_LIST_FILE)$(Q) \
+	${FEATURE_PROBLEM_LIST_FILE} \
+	${VENDOR_PROBLEM_LIST_FILE} \
+	${OPENJDK_CONTAINER_TEST_OPTS} \
+	$(Q)$(JTREG_HOTSPOT_TEST_DIR):hotspot_containers$(Q); \
+	$(TEST_STATUS)</command>
+		<versions>
+			<version>11+</version>
+		</versions>
+		<levels>
+			<level>dev</level>
+		</levels>
+		<groups>
+			<group>openjdk</group>
+		</groups>
+		<impls>
+			<impl>hotspot</impl>
+		</impls>
+		<platformRequirements>os.linux</platformRequirements>
+	</test>
+	<test>
+		<testCaseName>hotspot_container_jdk8</testCaseName>
+		<variations>
+			<variation>NoOptions</variation>
+		</variations>
+		<command>$(JAVA_COMMAND) -Xmx512m -jar $(Q)$(TEST_RESROOT)$(D)jtreg$(D)lib$(D)jtreg.jar$(Q) \
+	$(JTREG_BASIC_OPTIONS) $(JVM_NATIVE_OPTIONS) -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS) $(VMOPTION_HEADLESS)$(Q) \
+	-w $(Q)$(REPORTDIR)$(D)work$(Q) \
+	-r $(Q)$(REPORTDIR)$(D)report$(Q) \
+	-jdk:$(Q)$(TEST_JDK_HOME)$(Q) \
+	-exclude:$(Q)$(JTREG_HOTSPOT_TEST_DIR)$(D)ProblemList.txt$(Q) \
+	-exclude:$(Q)$(TEST_RESROOT)$(D)$(PROBLEM_LIST_FILE)$(Q) \
+	${FEATURE_PROBLEM_LIST_FILE} \
+	${VENDOR_PROBLEM_LIST_FILE} \
+	${OPENJDK_CONTAINER_TEST_OPTS} \
+	$(Q)$(OPENJDK_DIR)$(D)hotspot$(D)test$(D)runtime$(D)containers$(Q); \
+	$(TEST_STATUS)</command>
+		<versions>
+			<version>8</version>
+		</versions>
+		<levels>
+			<level>dev</level>
+		</levels>
+		<groups>
+			<group>openjdk</group>
+		</groups>
+		<impls>
+			<impl>hotspot</impl>
+		</impls>
+		<platformRequirements>os.linux</platformRequirements>
+	</test>
+	<test>
+		<testCaseName>jdk_container</testCaseName>
+		<variations>
+			<variation>NoOptions</variation>
+		</variations>
+		<command>$(JAVA_COMMAND) -Xmx512m -jar $(Q)$(TEST_RESROOT)$(D)jtreg$(D)lib$(D)jtreg.jar$(Q) \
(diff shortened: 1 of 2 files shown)


jerboaa:
@ShelleyLambert @sophia-guo I'm going to need help with testing this patch. I've forgotten all about how to run those custom targets in a grinder. Note that those need to be run as a user that can do `docker build` and `docker run`. It would also be good to only run those if the test host is a physical system (not already in a container). Thanks for your help!

openjdk/playlist.xml
```diff
@@ -90,6 +90,99 @@
...
+		</levels>
+		<groups>
+			<group>openjdk</group>
+		</groups>
+		<impls>
+			<impl>hotspot</impl>
+		</impls>
+	</test>
+	<test>
+		<testCaseName>hotspot_container</testCaseName>
```
smlambert, line 126:
```suggestion
		<testCaseName>hotspot_container_jdk8</testCaseName>
```

smlambert (commented):
We may want to add these new tests into `<level>dev</level>` and make an update to the ci-jenkins-pipeline code so that we can direct openjdk.dev test jobs to nodes that are not 'test-docker*' nodes but rather nodes that have docker installed (label = sw.tool.docker).  Will chat with @sophia-guo to discuss options for ways we can ensure we do not send container tests to containers.

openjdk/playlist.xml
jerboaa: Thanks, updated.

jerboaa:
> We may want to add these new tests into `<level>dev</level>` and make an update to the ci-jenkins-pipeline code

OK, I've used `<level>dev</level>` in the updated patch.

smlambert:
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6187/

jerboaa:
> https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6187/

Thank you! Looks like it failed the `TestMemoryWithCgroupV1` test. That in turn seems to be missing https://bugs.openjdk.org/browse/JDK-8289146 backport?

jerboaa:
> > https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6187/
> 
> Thank you! Looks like it failed the `TestMemoryWithCgroupV1` test. That in turn seems to be missing https://bugs.openjdk.org/browse/JDK-8289146 backport?

Can that be, though? It's supposed to be in 11.0.18+1 onwards. How would I check?

jerboaa:
> > > https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6187/
> > 
> > 
> > Thank you! Looks like it failed the `TestMemoryWithCgroupV1` test. That in turn seems to be missing https://bugs.openjdk.org/browse/JDK-8289146 backport?
> 
> Can that be, though? It's supposed to be in 11.0.18+1 onwards. How would I check?

Nah, it's there. As it suggests to run this on a system with `'cgroup_enable=memory swapaccount=1'` on the kernel command line.

See https://github.com/adoptium/jdk11u/blob/dde20e471dd77528baa922dde484e9ea92c005aa/test/hotspot/jtreg/containers/docker/TestMemoryWithCgroupV1.java#L94

This will require some more infra setup.

jerboaa:
> Single test target, hotspot_container: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6187/
> 
> All dev.openjdk targets (which currently should only be these 2, hotspot_container and jdk_container x 2 modes, nocompressedoops and compressedoops): https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6188/

For these tests it probably would make more sense to have two different modes. cgroup v1 and cgroup v2 :)

openjdk/playlist.xml
```diff
@@ -90,6 +90,99 @@
 			<impl>hotspot</impl>
 		</impls>
 	</test>
+	<test>
+		<testCaseName>hotspot_container</testCaseName>
+		<variations>
+			<variation>Mode150</variation>
+			<variation>Mode650</variation>
+			<variation>Mode1000</variation>
+		</variations>
```
smlambert, line 99: This currently is set to running [Mode1000](https://github.com/adoptium/TKG/blob/master/resources/modes.xml#L4617-L4625) NoOptions for 32bit platforms and [Mode150](https://github.com/adoptium/TKG/blob/master/resources/modes.xml#L954-L962) `-XX:+UseCompressedOops` / [Mode650](https://github.com/adoptium/TKG/blob/master/resources/modes.xml#L4279-L4287) `-XX:-UseCompressedOops` for 64 bit platforms (so will be run twice on 64bit platforms).  If this is not a useful set of variations for these targets, this could be changed to 
```
<variations>
        <variation>NoOptions</variation>
</variations>
```
To run once, with no additional JVM options added.

jerboaa: OK, thanks! Those variations don't seem useful. I'll update.

jerboaa:
Latest grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6192/

jerboaa:
I think this is good to go, once the infra ticket has been handled. `TestMemoryWithCgroupV1.testMemoryLimitWithSwappiness` is expected to fail on hosts disabling memory swap accounting.

smlambert:
We could also limit this target to only run on x64 linux via the playlist (either with `<platformRequirements>` or `<disables>` for other platforms, but we can also restrict where this gets launched in the build pipeline itself.

related: https://github.com/adoptium/ci-jenkins-pipelines/issues/501

jerboaa:
Note that there shouldn't be any limitation on supported architectures if `podman` is being used as container engine. At least `container-tools` module providing podman is available for RHEL 8 hosts. So this would largely be an infra related limitation.

jerboaa:
See also https://bugs.openjdk.org/browse/JDK-8297274 for the `TestMemoryWithCgroupV1` test issue.

jerboaa:
> We could also limit this target to only run on x64 linux via the playlist (either with `<platformRequirements>`

I've added `<platformRequirements>os.linux</platformRequirements>` to the latest push as that seemed most appropriate.

smlambert:
If you want to exclude xxx as per https://github.com/adoptium/aqa-tests/pull/4147#issuecomment-1323537830 then we could additionally add an entry for TestMemoryWithCgroupV1  to the [ProblemLists](https://github.com/adoptium/aqa-tests/blob/master/openjdk/excludes/) for it (likely depends on how long you thing 
https://bugs.openjdk.org/browse/JDK-8297274 will take to fix upstream).  I am happy to merge this now and assess the build pipeline runs afterwards.

smlambert (approved):

jerboaa:
I'll follow-up with exclude list updates once this is merged. We need to remove them from JDK 8 and the said cgroup v1 test without swap accounting enabled on the kernel level. Thanks!

sophia-guo (approved):

Issue #4155, opened 2022-11-17
Initial problem listing for Windows ARM64


joe-braley:
Please let me know if this should be broken up into separate GitHub issues!

Initial problem listing for Window AArch64 for JDK11 and JDK17:
```
- JDK11u:
    - tools/jlink/plugins/VendorInfoPluginsTest.java 
    - jdk/jfr/jvm/TestDumpOnCrash.java
    - java/lang/ProcessBuilder/Basic.java#id0
    - java/lang/ProcessBuilder/PipelineTest.java
- JDK17u:
    - jdk/jfr/event/runtime/TestDumpReason.java
    - jdk/jfr/event/runtime/TestShutdownEvent.java
    - jdk/jfr/jvm/TestDumpOnCrash.java
    - java/foreign/valist/VaListTest.java
    - java/foreign/malloc/TestMixedMallocFree.java
    - java/foreign/TestNative.java
    - java/foreign/TestVarArgs.java
    - java/foreign/StdLibTest.java
```
With respect to `java/foreign/*` tests there is ongoing work to update the foreign function and memory API to work with Windows ARM64 so I believe that Temurin might be affected by these tests as well.

Example output:
```
**JDK11**

Test: tools/jlink/plugins/VendorInfoPluginsTest.java
Reasoning: Expecting strings from stdout. Might be a possible port issue.
Output:
java.lang.RuntimeException: '^# +https://bugs.xyzzy.com/crash/$' missing from stdout 
  at jdk.test.lib.process.OutputAnalyzer.stdoutShouldMatch(OutputAnalyzer.java:300)
  at VendorInfoPluginsTest.main(VendorInfoPluginsTest.java:101)
  at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
  at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.base/java.lang.reflect.Method.invoke(Method.java:566)
  at com.sun.javatest.regtest.agent.MainActionHelper$AgentVMRunnable.run(MainActionHelper.java:312)
  at java.base/java.lang.Thread.run(Thread.java:829)
-------------

Test: jdk/jfr/jvm/TestDumpOnCrash.java
[...]

PR #4159, opened 2022-11-18
Initial Windows ARM64 problem listing

joe-braley:
Hey team!

Related GH issue: https://github.com/adoptium/aqa-tests/issues/4155. Please let me know if this issue requires splitting up.

`java/foreign/*` tests might impact the temurin builds as well so included those in the main problem lists.  I also believe that `jdk/sun/tools/jhsdb/JShellHeapDumpTest.java` might impact that build as well since there have been similar issues with that test on windows in the past (see: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8230731). But as always, I could be mistaken.

Please let me know if there are any questions or concerns with this PR!

Thanks!

CC @karianna / @gdams / @ShelleyLambert for visibility.


sophia-guo:
@joe-braley Will Microsoft run openjdk tests with with jdk8 or other versions? If yes the corresponding vendor ProblemList file will also need to be added even if it might be empty file without tests excluded.

joe-braley:
@sophia-guo We currently run openjdk tests against 11 and 17. I believe the next version we will support will be 21 which comes out September 2023 if I am not mistaken. I can add the empty problem lists if that is still a requirement!

sophia-guo (approved):
LGTM

smlambert (approved):

Issue #4092, opened 2022-10-27
Capture TKG test result logic may not work with ITERATIONS


llxia:
**Describe the bug**
With the capture TKG test result, we can visualize the test summary in the Jenkins job (see below). This is really handy for the users. 

 
![image](https://user-images.githubusercontent.com/19273206/198327679-6319182b-fa84-4533-80c7-0ebbad884137.png)

However, this feature may give false positives or false negatives when running with ITERATIONS. In the following case ([Grinder link](https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/29046/)), the test has an intermittent failure. It passed 3/5. `TEST TARGETS SUMMARY` shows a false positive. 

![image](https://user-images.githubusercontent.com/19273206/198328107-58235ca3-6441-40e0-abe3-61ff0aad1364.png)

I think we should update the logic to capture all iteration results.
https://github.com/adoptium/aqa-tests/blob/9166b61bddc5156f6d3182ff536a5e035293452a/buildenv/jenkins/JenkinsfileBase#L673-L710

As we enable [Rerun failed tests automatically feature](https://github.com/adoptium/aqa-tests/issues/3431), this will affect the nightly runs with rerun enabled.
Also, the good news is that the job status is correct (i.e., UNSTABLE) in this case because we also check TAP files.

FYI @renfeiw

renfeiw:
Currently we are using manager.getLogMatcher to get the match, but it only checks the first occurence. I will try parsing manager.build.logFile to see it works.

PR #4160, opened 2022-11-18
Remove tap plugin

### LINKED ISSUE #4154: Remove Tap plugin as Jenkins job status is supported by exit code of TKG and checkTestResults()
Originally we are using Tap plugin to support test result files to Jenkins, which has caused some issues. Right now jenkins job status is set TKG exit code and checkTestResults(). Suggest to remove the tap plugin step `step([$class: "TapPublisher", testResults: "aqa-tests/TKG/**/*.tap", outputTapToConsole: false, failIfNoResults: true])` for following reasons

- tap plugin step is in post() stage, which is after checkTestResults(). There might be the risk of overriding the build result setting by checkTestResult() ( just suspect). If it is not used  it should be removed
- openjdk diagnostic level has been updated to default level. That means if there are no test failures but have jvm crash those information will be added to tap file. In this case if there are special characters which can't be parsed by tap plugin it will tag jenkins as failed instead of unstable.

sophia-guo:
Close #4154 

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -708,8 +708,6 @@ def post(output_name) {
 			tar_cmd_suffix = ""
 		}
 
-		step([$class: "TapPublisher", testResults: "aqa-tests/TKG/**/*.tap", outputTapToConsole: false, failIfNoResults: true])
-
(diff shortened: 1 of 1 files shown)


llxia:
Add @renfeiw as a reviewer as I am not sure if we are ready to remove the TAP plugin. See the issue with the TKG test result: https://github.com/adoptium/aqa-tests/issues/4092

renfeiw:
I don't think we can remove TAP plugin. It contains more detailed information compared to the result summary captured by Jenkins.

PR #4161, opened 2022-11-18
Ignore dev jtx when using custom target

Mesbah-Alam:
We should ignore dev jtx when using custom target so that developers can run excluded tests if needed.


Mesbah-Alam:
Test with this PR : Grinder_jck/917. 
Test without this PR: Grinder_jck/916 (test excluded in dev.jtx is skipped).

llxia (approved):
LGTM

smlambert (approved):

PR #4176, opened 2022-12-01
Jck vm/jdwp & jvmti needs to run concurrency=1 as not thread safe

andrew-m-leonard:
Jck19 jdwp is potentially hanging or encountering:
```
java.lang.IllegalStateException: This section of the test result is now immutable.
```
I believe this is because jdwp by design is not thread/process safe.
This PR sets the vm/jdwp concurrency to 1.

### DIFF
--- a/jck/runtime.vm/playlist.xml
+++ b/jck/runtime.vm/playlist.xml
@@ -51,7 +51,7 @@
 		<variations>
 			<variation>NoOptions</variation>
 		</variations>
-		<command>$(JCK_CMD_TEMPLATE) tests=vm/jdwp testsuite=RUNTIME; \
+		<command>$(JCK_CMD_TEMPLATE) tests=vm/jdwp testsuite=RUNTIME concurrency=1; \
 	$(TEST_STATUS)</command>
 		<levels>
 			<level>sanity</level>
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
Suspecting vm/jvmti has same problem.

sxa (approved):
I wonder what the performance impact of doing this is, especially on higher core machines.

Mesbah-Alam:
Hi @andrew-m-leonard , Is this a JDK19 only issue? In that case it'd be better to set the concurrency value in JavaTestRunner.java.

andrew-m-leonard:
> Hi @andrew-m-leonard , Is this a JDK19 only issue? In that case it'd be better to set the concurrency value in JavaTestRunner.java.

@Mesbah-Alam I don't think so, for two reasons, we see intermittent behaviour for jdwp with all versions, and also from working on the jdwp code a while back I know it is not thread safe by design.

andrew-m-leonard:
Just to clarify my thoughts, jdwp supports multi-threading as a feature, the jdwp transport initialization is not. So if in a "test framework" you try and concurrently construct multiple transports things can sometimes go wrong.

Mesbah-Alam (approved):
LGTM

Mesbah-Alam:
Thanks @andrew-m-leonard !

Issue #4184, opened 2022-12-05
End of test job PROCESSCATCH does not terminate all hung/leftover processes


andrew-m-leonard:
The beginning and end of test job cleanup does not always terminate all test jobs, and on Windows results in a 30 iteration loop continuously terminating launched java processes. The reason is the logic does not terminate the "make" processes that is launching the rogue processes. Also related jck test job processes are not terminated.

PR #4185, opened 2022-12-05
Enhanced pre/post test job hung/leftover process checking

andrew-m-leonard:
Fixes: https://github.com/adoptium/aqa-tests/issues/4184

Enhances the pre/post PROCESSCATCH logic checking for more left over processes, in particular:
- "make" : stops repetitive process killing at end of failed/aborted test jobs
- "jck" extended processes
- more specific "java" process selection

Enhanced to better filter Jenkins Agent process, and utilise Windows enhanced powershell queries.

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -26,12 +26,7 @@ def setupEnv() {
 	env.JOBSTARTTIME = sh(script: 'LC_TIME=C date +"%a, %d %b %Y %T %z"', returnStdout: true).trim()
 
 	// Terminate any previous test related processes that are running
-	echo "PROCESSCATCH: Terminating previous test related processes which are currently on the machine:"
-	def statusCode=sh(script:"aqa-tests/terminateTestProcesses.sh ${env.USER} 2>&1", returnStatus:true)
-	echo "returnCode = ${statusCode}"
-	if (statusCode != 0) {
-		echo "Unable to terminate all processes."
-	}
+        terminateTestProcesses()
 
 	if ( params.JDK_VERSION ) {
 		env.ORIGIN_JDK_VERSION = params.JDK_VERSION
@@ -792,8 +787,8 @@ def testBuild() {
 				testExecution()
 			}
 		} finally {
-			// Terminate any left over java processes
-			terminateJavaProcesses()
+			// Terminate any left over test job processes
+			terminateTestProcesses()
 
 			if (!params.KEEP_WORKSPACE) {
 				// cleanWs() does not work in some cases, so set opts below
@@ -829,7 +824,7 @@ def testExecution() {
 	}
 }
 
-def terminateJavaProcesses() {
+def terminateTestProcesses() {
 	echo "PROCESSCATCH: Terminating any hung/left over test processes:"
 
         def statusCode=sh(script:"aqa-tests/terminateTestProcesses.sh ${env.USER} 2>&1", returnStatus:true)
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -828,9 +828,8 @@ def terminateTestProcesses() {
 	echo "PROCESSCATCH: Terminating any hung/left over test processes:"
 
         def statusCode=sh(script:"aqa-tests/terminateTestProcesses.sh ${env.USER} 2>&1", returnStatus:true)
-        echo "returnCode = ${statusCode}"
         if (statusCode != 0) {
-                echo "Unable to terminate all processes."
+                echo "rc = ${statusCode}, Unable to terminate all processes."
         } 
 }
 
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -27,7 +27,7 @@ def setupEnv() {
 
 	// Terminate any previous test related processes that are running
 	echo "PROCESSCATCH: Terminating previous test related processes which are currently on the machine:"
-	def statusCode=sh(script:"aqa-tests/terminateTestProcesses.sh 2>&1", returnStatus:true)
+	def statusCode=sh(script:"aqa-tests/terminateTestProcesses.sh ${env.USER} 2>&1", returnStatus:true)
 	echo "returnCode = ${statusCode}"
 	if (statusCode != 0) {
 		echo "Unable to terminate all processes."
@@ -832,7 +832,7 @@ def testExecution() {
 def terminateJavaProcesses() {
 	echo "PROCESSCATCH: Terminating any hung/left over test processes:"
 
-        def statusCode=sh(script:"aqa-tests/terminateTestProcesses.sh 2>&1", returnStatus:true)
+        def statusCode=sh(script:"aqa-tests/terminateTestProcesses.sh ${env.USER} 2>&1", returnStatus:true)
         echo "returnCode = ${statusCode}"
         if (statusCode != 0) {
                 echo "Unable to terminate all processes."
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -25,12 +25,12 @@ def makeTest(testParam) {
 def setupEnv() {
 	env.JOBSTARTTIME = sh(script: 'LC_TIME=C date +"%a, %d %b %Y %T %z"', returnStdout: true).trim()
 
-	// Check what existing java processes are running
-	echo "PROCESSCATCH: Java processes which are currently on the machine:"
-	if (PLATFORM.contains("windows")) {
-		sh(script:'powershell -c "Get-Process java | select id,processname,starttime"', returnStatus:true)
-	} else {
-		sh(script:'ps -fu '+"${env.USER}"+' | grep java | egrep -v "remoting.jar|agent.jar|grep"', returnStatus: true)
+	// Terminate any previous test related processes that are running
+	echo "PROCESSCATCH: Terminating previous test related processes which are currently on the machine:"
+	def statusCode=sh(script:"aqa-tests/terminateTestProcesses.sh 2>&1", returnStatus:true)
+	echo "returnCode = ${statusCode}"
+	if (statusCode != 0) {
+		echo "Unable to terminate all processes."
 	}
 
 	if ( params.JDK_VERSION ) {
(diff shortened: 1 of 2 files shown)


buildenv/jenkins/JenkinsfileBase
```diff
@@ -25,12 +25,12 @@ def makeTest(testParam) {
...
-	if (PLATFORM.contains("windows")) {
-		sh(script:'powershell -c "Get-Process java | select id,processname,starttime"', returnStatus:true)
-	} else {
-		sh(script:'ps -fu '+"${env.USER}"+' | grep java | egrep -v "remoting.jar|agent.jar|grep"', returnStatus: true)
+	// Terminate any previous test related processes that are running
+	echo "PROCESSCATCH: Terminating previous test related processes which are currently on the machine:"
+	def statusCode=sh(script:"aqa-tests/terminateTestProcesses.sh 2>&1", returnStatus:true)
+	echo "returnCode = ${statusCode}"
+	if (statusCode != 0) {
+		echo "Unable to terminate all processes."
```
llxia, line 33: Should we just call `terminateJavaProcesses()`?

andrew-m-leonard: yeah :-)

andrew-m-leonard:
@llxia @sxa ready for review please

llxia:
@AdamBrousseau could you also review this PR? Thanks

terminateTestProcesses.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/sh
+
+# Arg S1 : Unix current env.User
+
+# If Windows use PowerShell
```
karianna, line 5:
```suggestion
# If Windows uses PowerShell
```

buildenv/jenkins/JenkinsfileBase
```diff
@@ -829,50 +824,14 @@ def testExecution() {
...
-	echo "PROCESSCATCH: Removed processes - here is what is remaining:"
-	if (PLATFORM.contains("windows")) {
-		sh(script:'powershell -c "Get-Process java | select id,processname,starttime"', returnStatus:true)
-	} else {
-		sh(script:'ps -fu '+"${env.USER}"+' | grep java | egrep -v "remoting.jar|agent.jar|grep"', returnStatus:true)
-	}
+        def statusCode=sh(script:"aqa-tests/terminateTestProcesses.sh ${env.USER} 2>&1", returnStatus:true)
+        echo "returnCode = ${statusCode}"
+        if (statusCode != 0) {
+                echo "Unable to terminate all processes."
```
karianna, line 833: Print out the error code here?

terminateTestProcesses.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/sh
+
```
karianna, line 2: set -ex ?

terminateTestProcesses.sh
```diff
@@ -0,0 +1,84 @@
...
+# Arg S1 : Unix current env.User
+
+# If Windows use PowerShell
+if [ "$OSTYPE" = "cygwin" ]; then
+  echo "Windows machine, using powershell queries..."
+
+  # Match anything that most likely test job or process related
```
karianna, line 9:
```suggestion
  # Match anything that is most likely a test job or process related
```

terminateTestProcesses.sh
```diff
@@ -0,0 +1,84 @@
...
+             CommandLine like '%rmiregistry%' or \
+             CommandLine like '%tnameserv%' or \
+             CommandLine like '%make.exe%')"
+
+  # Ignore Jenkins agent and grep cmd
+  ignore_str="not CommandLine like '%remoting.jar%' and \
+              not CommandLine like '%agent.jar%' and \
+              not CommandLine like '%grep%'"
+
+  C=`powershell -c "(Get-WmiObject Win32_Process -Filter {${ignore_str} and ${match_str}} | measure).count" | tr -d "\\\\r"`
```
karianna, line 25: What is a `C` :-)

terminateTestProcesses.sh
```diff
@@ -0,0 +1,84 @@
...
+  
+  if [ $C -gt 0 ]; then
+      echo Windows rogue processes detected, attempting to stop them..
+      powershell -c "Get-WmiObject Win32_Process -Filter {${ignore_str} and ${match_str}}"
+      powershell -c "(Get-WmiObject Win32_Process -Filter {${ignore_str} and ${match_str}}).Terminate()"
+      uname | grep CYGWIN 2>/dev/null || uptime
+      sleep 10
```
karianna, line 32: Is that seconds?  Probably worth echoing that so users expect it

terminateTestProcesses.sh
```diff
@@ -0,0 +1,84 @@
...
+      uname | grep CYGWIN 2>/dev/null || uptime
+      C=`powershell -c "(Get-WmiObject Win32_Process -Filter {${ignore_str} and ${match_str}} | measure).count" | tr -d "\\\\r"`
+      if [ $C -gt 0 ]; then
+        echo "Cleanup failed, processes still remain..."
+        exit 127
+      fi 
+      echo "Processes stopped successfully"
+  else
+      echo Woohoo - no rogue processes detected
```
karianna, line 41:
```suggestion
      echo Woohoo - no rogue processes detected!
```

terminateTestProcesses.sh
```diff
@@ -0,0 +1,84 @@
...
+  # Fix For Issue https://github.com/adoptium/infrastructure/issues/2442 - Solaris PS Command Truncation
+  if [ `uname` = "SunOS" ]; then
+      PSCOMMAND="/usr/ucb/ps -uxww"
+  else
+      PSCOMMAND="ps -fu $1"
+  fi
+
+  if $PSCOMMAND | egrep "${match_str}" | egrep -v "${ignore_str}"; then
+      echo Boooo - there are rogue processes kicking about
+      echo Issuing a kill to all processes shown above - though this will not terminate vfb processes on AIX as we are trying to debug source of leftovers there
```
karianna, line 63: Seems very AIX specific, perhaps wrap in a detection block?

terminateTestProcesses.sh
```diff
@@ -0,0 +1,84 @@
...
+      $PSCOMMAND | egrep "${match_str}" | egrep -v "${ignore_str}" | awk '{print $2}' | xargs -n1 kill
+      uname | grep CYGWIN 2>/dev/null || uptime
+      sleep 10
```
karianna, line 66: Same as above, consider echoing.

karianna (changes requested):

terminateTestProcesses.sh
andrew-m-leonard: not here, don't want to exit at first failed shell command

terminateTestProcesses.sh
andrew-m-leonard: a count variable, i'll rename it!

terminateTestProcesses.sh
andrew-m-leonard: That's a copy&paste from somewhere else, i'll remove it

terminateTestProcesses.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/sh
+
+# Arg S1 : Unix current env.User
+
+# If Windows uses PowerShell
+if [ "$OSTYPE" = "cygwin" ]; then
```
sxa, line 6: It looks like this was `PLATFORM.contains("windows")` instead of an explicitly cygwin check previously. What is the reason for this change as there are alternative environments that can be used to build on Windows other than cygwin that would presumably not be covered after this.

terminateTestProcesses.sh
```diff
@@ -0,0 +1,83 @@
...
+  echo "Windows machine, using powershell queries..."
+
+  # Match anything that is most likely a test job or process related
+  match_str="(CommandLine like '%java%jck%' or \
+             CommandLine like '%openjdkbinary%java%' or \
+             CommandLine like '%java%javatest%' or \
+             CommandLine like '%java%-Xfuture%' or \
+             CommandLine like '%X%vfb%' or \
```
sxa, line 14: `X%vfb` is a UNIX process, so should not be relevant on Windows.

terminateTestProcesses.sh
```diff
@@ -0,0 +1,83 @@
...
+else
+  echo "Unix type machine.."
+
+  # Match anything that is most likely a jck test job or process related
+  match_str="java.*jck|openjdkbinary.*java|java.*javatest|java.*-Xfuture|X.*vfb|rmid|rmiregistry|tnameserv|make"
+
+  # Ignore Jenkins agent and grep cmd
+  ignore_str="remoting.jar|agent.jar|grep"
+
+  if [ `uname` = "SunOS" ]; then
```
sxa, line 53: Note: This seems to have changed from the previous version. The new one will likely not work on Solaris 11.
(This code looks familiar ;-) I'm going to have to look separately on adjusting this at the source to be ok on Solaris 11!)

terminateTestProcesses.sh
```diff
@@ -0,0 +1,83 @@
...
+             CommandLine like '%rmiregistry%' or \
+             CommandLine like '%tnameserv%' or \
+             CommandLine like '%make.exe%')"
+
+  # Ignore Jenkins agent and grep cmd
+  ignore_str="not CommandLine like '%remoting.jar%' and \
+              not CommandLine like '%agent.jar%' and \
+              not CommandLine like '%grep%'"
+
+  count=`powershell -c "(Get-WmiObject Win32_Process -Filter {${ignore_str} and ${match_str}} | measure).count" | tr -d "\\\\r"`
```
sxa, line 25: Not sure why this has changed from the `GetProcess` used in the previous version of this function but this seems to be filtering on `CommandLine` which isn't the object returned by `Get-WmiObject Win32_Process`. Am I missing something?
image

sxa (changes requested):
The Windows process check in particular is going to need checking, the others are suggestions and superfluous clauses.

terminateTestProcesses.sh
andrew-m-leonard: ok thanks?

andrew-m-leonard: @sxa maybe other places that use uname SunOS check need changing? https://github.com/adoptium/aqa-tests/blob/d089897c5c9a7d87ba6def8efd53a4cb786c2e99/buildenv/jenkins/JenkinsfileBase#L425

terminateTestProcesses.sh
sxa: I suspect that doing `if [ "$OS" = "Windows_NT" ]` would trap all situations since that's defined by default on Windows, and wouldn't require any additional parameters. `SXA-ProcessCheck` was a bit of a hack so I wouldn't necessarily assume that its way of doing things is a good example of how to do things ;-)

terminateTestProcesses.sh
andrew-m-leonard: Ah ok! yes, so this and your other scripts will need altering for Sol11

terminateTestProcesses.sh
```diff
@@ -0,0 +1,82 @@
...
+  
+  if [ $count -gt 0 ]; then
+      echo Windows rogue processes detected, attempting to stop them..
+      powershell -c "Get-WmiObject Win32_Process -Filter {${ignore_str} and ${match_str}}"
+      powershell -c "(Get-WmiObject Win32_Process -Filter {${ignore_str} and ${match_str}}).Terminate()"
+      uname | grep CYGWIN 2>/dev/null || uptime
+      echo Sleeping for 10 seconds...
+      sleep 10
+      uname | grep CYGWIN 2>/dev/null || uptime
```
sxa, line 33: Suggest removing this line as it doesn't achieve anything as far as I can tell

terminateTestProcesses.sh
andrew-m-leonard: updated

sxa (approved):

andrew-m-leonard:
@AdamBrousseau Hi Adam, if you could take a look as well please, as another CI user, thanks

llxia:
zos [Grinder](https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/30138/console)

buildenv/jenkins/JenkinsfileBase
```diff
@@ -829,50 +824,13 @@ def testExecution() {
...
+def terminateTestProcesses() {
+	echo "PROCESSCATCH: Terminating any hung/left over test processes:"
 
-	echo "PROCESSCATCH: Removed processes - here is what is remaining:"
-	if (PLATFORM.contains("windows")) {
-		sh(script:'powershell -c "Get-Process java | select id,processname,starttime"', returnStatus:true)
-	} else {
-		sh(script:'ps -fu '+"${env.USER}"+' | grep java | egrep -v "remoting.jar|agent.jar|grep"', returnStatus:true)
-	}
+        def statusCode=sh(script:"aqa-tests/terminateTestProcesses.sh ${env.USER} 2>&1", returnStatus:true)
```
karianna, line 830:
```suggestion
        def statusCode = sh(script:"aqa-tests/terminateTestProcesses.sh ${env.USER} 2>&1", returnStatus:true)
```

karianna (approved):
LGTM

llxia (approved):
LGTM

AdamBrousseau:
Sorry I didn't get a chance to finish reviewing before merge. I am happy with it assuming it's been tested on all the platforms. We'll keep an eye out. We have a "sanitize" job we run a few times a day. This might allow us to disable that. Thanks!

PR #4209, opened 2022-12-17
Only display left over processes during end of test process check

andrew-m-leonard:
The end of test left over process check that kills any remaining test processes, has the unfortunate consequence in an host environment where test sub-containers are also running tests, whereby the container processes are visible and can be accidentally terminated. So for the moment just display the left over processes, and not terminate them.

Resolves temurin-compliance issue: 282


smlambert:
Is this PR still needed @andrew-m-leonard ?

andrew-m-leonard:
Yes, this has been resolved.

Issue #4228, opened 2023-01-05
Support clone repo by ssh protocol in testJob


Accelerator1996:
Add a parameter USE_GIT_SSH to the testJob to make the clone repo by ssh protocol

PR #4229, opened 2023-01-05
Support clone all repo by ssh protocol in testJob (#4228)

Accelerator1996:


buildenv/jenkins/JenkinsfileBase
```diff
@@ -77,6 +80,7 @@ def setupEnv() {
 	env.EXTERNAL_TEST_CMD=params.EXTERNAL_TEST_CMD ? params.EXTERNAL_TEST_CMD : "mvn clean install"
 	env.DYNAMIC_COMPILE=params.DYNAMIC_COMPILE ? params.DYNAMIC_COMPILE : false
 	env.USE_JRE=params.USE_JRE ? params.USE_JRE : false
+        env.USE_GIT_SSH = params.USE_GIT_SSH ? params.USE_GIT_SSH : false
```
hgeraldino, line 83:
```suggestion
        env.USE_GIT_SSH = params.USE_GIT_SSH ?: false
```

hgeraldino:
I'm not too familiar on how these tests are run, but an alternative to changing the git clone command everywhere would be to set the URL replacement via `git config`. Something like:

```
git config --global url."ssh://".insteadOf "https://"
```

Now, I don't know if everything is executed within a single Jenkins context or not, so this suggestion might not be valid.

Accelerator1996:
> I'm not too familiar on how these tests are run, but an alternative to changing the git clone command everywhere would be to set the URL replacement via `git config`. Something like:
> 
> ```
> git config --global url."ssh://".insteadOf "https://"
> ```
> 
> Now, I don't know if everything is executed within a single Jenkins context or not, so this suggestion might not be valid.

Thank you for your reply. I think your configuration is fine to run on the machine, but if the test job runs in docker, additional configuration is required.

llxia:
@Accelerator1996 could you add more details to the issue https://github.com/adoptium/aqa-tests/issues/4228? I would like to understand the case that you are trying to cover.

Also, we have much more repos than the 3 repo parameters updated in the PR. In theory, a generic solution is much preferred (Hector's suggestion above). What scenario did you encounter in docker so the git config would not work?

Accelerator1996:
> @Accelerator1996 could you add more details to the issue #4228? I would like to understand the case that you are trying to cover.
> 
> Also, we have much more repos than the 3 repo parameters updated in the PR. In theory, a generic solution is much preferred (Hector's suggestion above). What scenario did you encounter in docker so the git config would not work?

I'm sorry, my answer may be a bit wrong, because I thought his suggestion was to directly modify the machine environment with instructions, obviously I misunderstood. There is no doubt that his advice is very good @hgeraldino ! I think this still needs to increase the parameters of the test job, I will modify this. Thank you very much!

Accelerator1996:
Hello, @hgeraldino @llxia @renfeiw , I modified the code according to your suggestion. I tested locally and it was fine. Please review my PR again. Thank you very much!

buildenv/jenkins/JenkinsfileBase
```diff
@@ -141,6 +141,15 @@ def setupEnv() {
 	} else {
 		sh 'printenv'
 	}
+        if ( params.USE_GIT_SSH ) {
+                def gitGlobalConfig = fileExists('~/.gitconfig') ? sh(script:"git config --global -l", returnStdout:true) : ''
+                if (!gitGlobalConfig || !gitGlobalConfig.contains('url.git@github.com:.insteadof')) {
```
hgeraldino, line 146: I think overwriting should be fine, no need to check for this condition. Or even better, do `git config --local ...` to install the preferences on the current cloned directory (Jenkins workspace for the current build)

Local (repository) level configs take precedence over global ones.

hgeraldino (commented):
Left one comment.

One question I had was if there's any situation in which HTTPS is preferred over git+ssh; if not, would it make sense to make git+ssh the default?

Accelerator1996:
> One question I had was if there's any situation in which HTTPS is preferred over git+ssh; if not, would it make sense to make git+ssh the default?

Maybe this situation will only exist in mainland China. Since we couldn't get overseas arm machines before, we had to use arm machines in mainland China. However, the network between mainland China and Github is very unsmooth. Although we modified our aqa-tests downstream repo to solve the problem, we think that there may be other vendors who also encounter this problem.

karianna:
@Accelerator1996 Will need a rebase

PR #4242, opened 2023-01-13
Disable TestDockerMemoryMetrics for OpenJ9 JDK8

babsingh:
OpenJDK needs to fix the implementation and test code.

Related: https://github.com/eclipse-openj9/openj9/issues/16460


karianna (approved):
LGTM

smlambert:
Has the problem been reported upstream?  If so, we can / should use the upstream issue number in the problem list.

babsingh:
> Has the problem been reported upstream? If so, we can / should use the upstream issue number in the problem list.

Not sure. In which upstream project should it be reported?

It was fixed in OpenJDK19: https://github.com/ibmruntimes/openj9-openjdk-jdk19/commit/ce7bf63e67eff780630c114a9a6999f897c1e303. It needs to be back-ported to older JDK versions.

smlambert (approved):
ok, given that, we leave it pointing to the openj9 issue.  👍

PR #4264, opened 2023-01-20
Disable LoadLibraryUnloadTest for OpenJ9 JDK19

babsingh:
OpenJDK fix for the testcase has not been backported to JDK19.

See https://github.com/eclipse-openj9/openj9/issues/16337#issuecomment-1398708803 for more details.


babsingh:
@pshipton @JasonFengJ9 @dmitripivkine Please thumbs up this message if there are no objections to permanently disable `LoadLibraryUnloadTest` for OpenJ9 JDK19 as per the details in https://github.com/eclipse-openj9/openj9/issues/16337#issuecomment-1398708803.

Mesbah-Alam (approved):
LGTM

pshipton:
Why don't we just backport the fix? Does the test pass with the fix?

babsingh:
> Why don't we just backport the fix? Does the test pass with the fix?

Just chose this solution since it was taken for https://github.com/eclipse-openj9/openj9/issues/16460 (precedent). I don't mind backporting the fix since it is also a feasible solution. The test should pass with the fix as per the details in https://github.com/openjdk/jdk/pull/9474. We can verify with a grinder as well.

babsingh:
Closing: backporting the fix instead of disabling the test.

pshipton:
https://github.com/eclipse-openj9/openj9/issues/16460 would have been a feature backport, not a test fix. It's better to fix and run the tests rather than exclude them, in case OpenJ9 problems are found by the test. Backporting to jdk19 comes with little cost, in that we won't need to maintain the backport since jdk19 will be out of service soon.

PR #4286, opened 2023-01-30
Exclude api/java/util target from aarch64 mac on jdk11

Mesbah-Alam:
- Exclude api/java/util target from aarch64 mac on jdk11
- Related backlog/issues/489


pshipton:
A quick fix is preferred to excluding the test, if possible. It seems like the file system contains a folder with the wrong case-sensitive name.

Mesbah-Alam:
> A quick fix is preferred to excluding the test, if possible. It seems like the compiled class contains the wrong case-sensitive name.

Agreed. Fix is in progress (details in backlog/issues/489).  

Closing this PR without merging.

Issue #2415, opened 2021-03-26
jdk8 hs Windows com/sun/jdi tests fail in ShellScaffold.sh: C:\Windows/tmp.6102: Permission denied


andrew-m-leonard:
This test also fails on zLinux for a different problem here: https://github.com/AdoptOpenJDK/openjdk-tests/issues/2408

https://trss.adoptopenjdk.net/output/test?id=605ba25b5730424dbcb67d31

The ShellScaffold.sh script tries to create a tmp file using the wrong slashes for CYGWIN environment:
https://github.com/AdoptOpenJDK/openjdk-jdk8u/blob/a1d853a5d797f38f7be8c260f4a83c1b3fe48adc/jdk/test/com/sun/jdi/ShellScaffold.sh#L271

causing:
```
[2021-03-24T19:20:22.349Z] ShellScaffold.sh: Version
[2021-03-24T19:20:22.349Z] E:/workspace/Test_openjdk8_hs_sanity.openjdk_x86-64_windows/openjdk-tests/openjdk/openjdk-jdk/jdk/test/com/sun/jdi/ShellScaffold.sh: line 277: C:\Windows/tmp.6102: Permission denied
[2021-03-24T19:20:22.349Z] cat: 'C:\Windows/tmp.6102': No such file or directory
[2021-03-24T19:20:22.349Z] mkdir: cannot create directory '/aa6102': Permission denied
[2021-03-24T19:20:22.349Z] touch: cannot touch '/aa6102/jxdbOutput.txt': No such file or directory
[2021-03-24T19:20:22.349Z] E:/workspace/Test_openjdk8_hs_sanity.openjdk_x86-64_windows/openjdk-tests/openjdk/openjdk-jdk/jdk/test/com/sun/jdi/ShellScaffold.sh: line 380: cd: /aa6102: No such file or directory
```
The logic needs to use correct slashes depending on Windows or Cygwin.

andrew-m-leonard:
The ShellScaffold.sh usage of a tmpFile here:
https://github.com/AdoptOpenJDK/openjdk-jdk8u/blob/a1d853a5d797f38f7be8c260f4a83c1b3fe48adc/jdk/test/com/sun/jdi/ShellScaffold.sh#L270
Probably needs changing to do something more like this example here for Windows/Cygwin:
[...]

PR #4288, opened 2023-01-31
Unexclude com/sun/jdi tests on windows for JDK8

zzambers:
Problem with jdi tests on windows has been fixed by  JDK-8295164 [1]. These tests are part of jdk/tier1 and are being ran as part of testing in GitHub CI on jdk8u-dev since.  Also comment in referenced aqa-tests issue says, that tests can be removed from list, once that JDK issue is fixed.

Fixes https://github.com/adoptium/aqa-tests/issues/2415

[1] https://bugs.openjdk.org/browse/JDK-8295164
[2] https://github.com/adoptium/aqa-tests/issues/2415#issuecomment-1277685526

### DIFF
--- a/openjdk/excludes/ProblemList_openjdk8.txt
+++ b/openjdk/excludes/ProblemList_openjdk8.txt
@@ -380,8 +380,8 @@ sun/jvmstat/monitor/MonitoredVm/CR6672135.java https://github.com/adoptium/aqa-t
 ###############################################################################
 # jdk_jdi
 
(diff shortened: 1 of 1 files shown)


zzambers:
There are 2 tests, which were also excluded on linux-s390x. I kept them excluded on linux-s390x, as that is separate issue, just updated links to that other issue (windows issue -> s390x issue).

smlambert:
Verify in https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6538

smlambert (approved):

sendaoYan (approved):

PR #4290, opened 2023-01-31
Unexclude compiler/rtm tests on JDK8

zzambers:
Compiler/rtm tests can be removed from exclude list of JDK8 in aqa-tests. They were excluded in jdk8u itself as part of effort to fix hotspot/tier1. See: https://bugs.openjdk.org/browse/JDK-8295915

### DIFF
--- a/openjdk/excludes/ProblemList_openjdk8.txt
+++ b/openjdk/excludes/ProblemList_openjdk8.txt
@@ -88,10 +88,6 @@ runtime/SharedArchiveFile/SpaceUtilizationCheck.java https://github.com/adoptium
 serviceability/jvmti/TestRedefineWithUnresolvedClass.java	https://github.com/adoptium/aqa-tests/issues/124	generic-all
 serviceability/jvmti/GetObjectSizeOverflow.java	https://github.com/adoptium/aqa-tests/issues/124	linux-all
 serviceability/sa/jmap-hashcode/Test8028623.java	https://github.com/adoptium/aqa-tests/issues/124	linux-all
-compiler/rtm/locking/TestUseRTMAfterLockInflation.java https://bugs.openjdk.java.net/browse/JDK-8180723 generic-all
-compiler/rtm/locking/TestUseRTMForInflatedLocks.java https://bugs.openjdk.java.net/browse/JDK-8180723 generic-all
-compiler/rtm/locking/TestUseRTMForStackLocks.java https://bugs.openjdk.java.net/browse/JDK-8180723 generic-all
-compiler/rtm/method_options/TestUseRTMLockElidingOption.java https://bugs.openjdk.java.net/browse/JDK-8180723 generic-all
 runtime/7110720/Test7110720.sh https://github.com/adoptium/aqa-tests/issues/2818 aix-all
 compiler/6894807/IsInstanceTest.java https://github.com/adoptium/aqa-tests/issues/2818 aix-all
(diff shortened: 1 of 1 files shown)


karianna (approved):

sophia-guo:
Those tests will be running on all platforms with `@requires vm.rtm.cpu & vm.rtm.compiler` Upstream jdk8u only excludes specific arch `generic-x64,generic-i586` https://github.com/openjdk/jdk8u-dev/commit/fb98418773b63dcbf9fea17d5e660bf4a135017e, which would be good for github actions. In Adoptium we have much more different platforms to support - x64, i586, ppc64, ppc64le, s390x, aarch64 etc.

sophia-guo:
Could upstream exclude as generic-all? Otherwise it might fail again in Adoptium on other platforms.

zzambers:
@sophia-guo There should be no need to exclude them on other archs. RTM (Restricted Transactional Memory) is extension of x86 ISA (See: [1]). Checks should get skipped on processors not supporting this extension, which should always be the case for non-x86 (See: [2][3]).

[1] https://en.wikipedia.org/wiki/Transactional_Synchronization_Extensions#RTM
[2] https://github.com/openjdk/jdk8u-dev/blob/784f71f218bdb0e885a03fae35105d2fc96d56ec/hotspot/test/compiler/rtm/locking/TestUseRTMAfterLockInflation.java#L66
[3] https://github.com/openjdk/jdk8u-dev/blob/784f71f218bdb0e885a03fae35105d2fc96d56ec/hotspot/test/compiler/testlibrary/rtm/predicate/SupportedCPU.java#L34

sophia-guo:
Recently found that one of our ppc64le has rtm feature https://github.com/adoptium/infrastructure/issues/2896

zzambers:
@sophia-guo, Interesting, didn't know ppc64 can also supports RTM. But seems you are right, there is some rtm code added for ppc by JDK-8077838 [1]. Do rtm tests fail there?

Anyway JDK-8077838 was not backported to JDK8 and I could not find arch specific code for rtm other than x86 in JDK8. I think backport of ppc rtm code to JDK8 is unlikely. Deprecation of RTM was actually considered [2] (which btw lists x86 and ppc).

[1] https://bugs.openjdk.org/browse/JDK-8077838
[2] https://bugs.openjdk.org/browse/JDK-8292096

sophia-guo:
I did't pay much attention to those rtm tests on jdk8 as they were excluded. For recent jdk19 I can say those rtm tests fail intermittently on some ppc64le.

sophia-guo:
Interesting I did test all 4 tests on linux64 and ppc64le with rtm or without rtm features all tests passed. https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6565/ https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6566/
https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6567/

Good to remove the exclusion and let upstream to decide exclude or not.

sophia-guo (approved):

zzambers:
@sophia-guo In case of JDK 8 I don't think excluding upstream other architectures than x86 (32,64) is necessary (as JDK8 seems to only support transactional memory on x86, if there is some problem in ppc code, it should only affect newer JDKs).

I checked our results, as we also do some testing on ppc64le (rpms), but I have not found any rtm test failures (8, 17). But it is possible, that we don't have right HW to reproduce your issue.

Issue #4295, opened 2023-02-01
Some files may fail to download or download incompletely


Accelerator1996:
For the network in mainland China, it may be somewhat unsmooth, and some files will fail to download or download incompletely at this time. I think it is possible to add the resuming option to the curl command.

Such as dacapo.jar, it cost 5h.
![image](https://user-images.githubusercontent.com/38688194/216054941-d175dcdc-aa3f-4b6c-bfec-bd30ad5531a1.png)

But finally, it was incomplete.
![image](https://user-images.githubusercontent.com/38688194/216055387-1a751cce-b427-4eff-bb8c-21997b688407.png)

llxia:
IMO, 5hrs+ to download the 3rd party jar is not acceptable. I am not sure if adding resuming option is a good workaround. 

The logic for 3rd party jar needs to be updated so that we can leverage the stage job - [test.getDependency](https://ci.adoptopenjdk.net/job/test.getDependency/). Please see https://github.com/adoptium/aqa-tests/issues/3921

Accelerator1996:
> IMO, 5hrs+ to download the 3rd party jar is not acceptable. I am not sure if adding resuming option is a good workaround.
> 
> The logic for 3rd party jar needs to be updated so that we can leverage the stage job - [test.getDependency](https://ci.adoptopenjdk.net/job/test.getDependency/). Please see #3921

I once thought about putting it in test.getDependency, but it is only 180+MB. I thought you didn't have this problem. I think your modification can better solve our problem, thank you very much!

llxia:
`test.getDependency` should be set up on your local Jenkins, so it should be fast for you.

Accelerator1996:
> `test.getDependency` should be set up on your local Jenkins, so it should be fast for you.

Thank you! We've add it into test.getDependency in our local Jenkins, and it was fine.

PR #4297, opened 2023-02-01
Add resuming option to make sure download dependency successfully(#4295)

Accelerator1996:
Issue: #4295


sendaoYan (approved):

karianna (approved):

smlambert:
Not sure if either reviewers saw the conversation in the issue, https://github.com/adoptium/aqa-tests/issues/4295, but I believe another solution was proposed for the problem reported.  

Hopefully this PR will not cause problems on certain platforms where these curl options may not be available (where older versions of curl are in use).  (cross that bridge when we come to it).

Issue #3059, opened 2021-10-20
Various hotspot_jre_1 and jdk_management test failures on ppc64 linux JDK 8


Haroon-Khel:
https://ci.adoptopenjdk.net/job/Test_openjdk8_hs_extended.openjdk_ppc64le_linux_testList_2/18/testReport/

`compiler/6865265/StackOverflowBug.java.StackOverflowBug`
`compiler/8009761/Test8009761.java.Test800976`
`compiler/exceptions/TestRecursiveReplacedException.java.TestRecursiveReplacedException`
`compiler/uncommontrap/StackOverflowGuardPagesOff.java.StackOverflowGuardPagesOff`
`compiler/uncommontrap/TestStackBangMonitorOwned.java.TestStackBangMonitorOwned`
`compiler/uncommontrap/TestStackBangRbp.java.TestStackBangRbp`

all fail with the following error `The stack size specified is too small, Specify at least 328k`

`gc/arguments/TestCMSHeapSizeFlags.java.TestCMSHeapSizeFlags`
`gc/arguments/TestG1HeapSizeFlags.java.TestG1HeapSizeFlags`
`gc/arguments/TestParallelHeapSizeFlags.java.TestParallelHeapSizeFlags`
`gc/arguments/TestSerialHeapSizeFlags.java.TestSerialHeapSizeFlags`

all fail with the following error `Execution failed: `main' threw exception: java.lang.RuntimeException: 'Error' missing from stdout/stderr`

PR #4317, opened 2023-02-06
Unexclude compiler tests with 64K page size fix on JDK8

zzambers:
Few compiler tests used to fail on systems with 64K page size as requested stack size was too small. (We have also seen these failures on rhel-8 aarch64, which uses 64K page size.) Problem has been fixed for JDK8 by [JDK-8067941](https://bugs.openjdk.org/browse/JDK-8067941) backport.

This fixes one category of failures described in: https://github.com/adoptium/aqa-tests/issues/3059


smlambert (approved):
👍

gdams (approved):

Issue #3829, opened 2022-07-07
Add new security tests (Part 1: CryptoTest)


smlambert:
We will like to add some new security tests contributed from the Red Hat team.  We have a couple of options for how to include them in the AQA test suite:
1. Add them to the functional group
2. Create a new test group called security and add them there.

The 2nd option implies expanding the test grid by adding a new column, and also as we have security tests scattered in most of the groups (system, openjdk, external, functional), we could consider moving them into the security group.  Unfortunately, this is not necessarily a good idea, as it implies that the setup and compilation time will be high, as each of the existing tests has their own set of prereqs that would need to be downloaded and compiled.  So the course of action if we take option 2 is to use the new group for new tests that do not have a large set of prereqs and requirements.

The new tests of concern for this issue can be found here:
https://github.com/rh-openjdk/CryptoTest - #4356 
https://github.com/rh-openjdk/ssl-tests

We will plan to do this after branching in prep for the July release (so these new tests will not be part of the July release branch).

We will intend to add these tests as 2 unique new test targets into the functional group, at the dev level.

The work required is to clone the test material via the [functional/security/build.xml](https://github.com/adoptium/aqa-tests/blob/master/functional/security/build.xml) file and add the 2 new targets in the [functional/security/playlist.xml](https://github.com/adoptium/aqa-tests/blob/master/functional/security/playlist.xml) file.

llxia:
I see pros and cons in both options. Personally, I think I prefer option 1 a little more because the existing security tests are scattered. Once we have the new security tests added, we can then evaluate and adjust them if needed.

smlambert:
Agree, I think that is a good approach too, as we do not delay adding the tests, and we can consider implications of Option 2 on execution time if we want a future re-org of test material.

lumuchris256:
hi @smlambert i would like to take a spin on this , more detail is welcome

smlambert:
Thanks @lumuchris256 - this is a much desired feature, and it will be great if you help to add these new security tests!

It will be helpful for you to look at the README of each of the 2 repositories of test material we wish to add:
https://github.com/rh-openjdk/CryptoTest/blob/master/README.md
tells how to build the test material and how to run tests.  

To incorporate new test material, we typically put the information for how to download and build test material into the build.xml file which is an ant script, and information about how to run the test in the playlist.xml file.  Then we would try to run the test locally or in Docker locally or by running them via Github actions.  

[...]

PR #4319, opened 2023-02-08
Add Crypotests

sophia-guo:
Add Crypotest #3829 

- Move existing security tests to sub dir
- Update gitignore
- Add task getDependentLibs to support compiling by subdir functional/security/AQAvit


sophia-guo:
CryptoTest will be added to functional security as a sub test. Would prefer to move the current security tests net.adoptopenjdk.test to sub directory. @ShelleyLambert is that ok? If it's ok any preference for the tests name? ( in my draft  PR I just name it Adoptium by random).

smlambert:
Since this test material originates from rh-openjdk, then that would be an appropriate name to use.

sophia-guo:
@smlambert I'm talking the tests already exists in aqa-tests/functional/security, which was added by you

functional/build.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0"?>
+
+<!--
```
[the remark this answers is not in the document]
sophia-guo: The file was added by accident. Removed.

sophia-guo:
Grinder https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/6606/

functional/security/AQAvit/build.xml
```diff
@@ -0,0 +1,72 @@
...
+		<echo>============COMPILER SETTINGS============</echo>
+		<echo>===fork:                         yes</echo>
+		<echo>===executable:                   ${compiler.javac}</echo>
+		<echo>===debug:                        on</echo>
+		<echo>===destdir:                      ${DEST}</echo>
+		<javac srcdir="${src}" destdir="${build}" debug="true" fork="true" executable="${compiler.javac}" includeAntRuntime="false" encoding="ISO-8859-1">
+			<src path="${src}" />
+			<classpath>
+				<pathelement location="${LIB_DIR}/testng.jar"/>
+				<pathelement location="${LIB_DIR}/jcommander.jar"/>
```
llxia, line 48: testng.jar and jcommand.jar are used, but they are not set before line 23. This will result in failure for the local runs. 

Please add the following line before line 23:
```
<property name="LIB" value="testng,jcommander" />
```

llxia:
https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/1986/

functional/security/AQAvit/build.xml
sophia-guo: The default value of LIB is all. This should not fail the local run though it's alway good to limit to the lib needed.

llxia (approved):
LGTM

smlambert (commented):
thanks @sophia-guo - there is a very recent upstream update to convert this test material for running with jtreg (see https://github.com/rh-openjdk/CryptoTest/pull/29/files, and cc-ed you on an email today).  In light of that update, can you update this PR to try and use jtreg to launch the test material please?

sophia-guo:
Move to https://github.com/adoptium/aqa-tests/pull/4327 using jtreg

PR #4357, opened 2023-02-22
Exclude java/lang/annotation/LoaderLeakTest on OpenJ9 jdk11

pshipton:
It's excluded on jdk17+. The test is added to jdk11 in 11.0.19.

Issue https://github.com/eclipse-openj9/openj9/issues/13201


smlambert (approved):

llxia (approved):
LGTM

PR #4363, opened 2023-02-23
Initial support for the SPECjbb perf testing benchmark 

ahmedmuhsin:
Initial support for SPECjbb will allow users who already have a licensed  local copy of the benchmark to launch a simple run in multi-jvm mode from within aqa-test. 

Please see `perf/specjbb/README.md` for instructions on how to use this feature.


perf/specjbb/README.md
```diff
@@ -0,0 +1,34 @@
...
+See the License for the specific language governing permissions and
+-->
+
+
+## SPECjbb 2015
+
+SPECjbb is a Java Client/Server performance testing benchmark. 
+Please visit its product page for more information, including a user guide: https://www.spec.org/jbb2015/
+
+
```
karianna, line 18:
```suggestion
```

perf/specjbb/README.md
```diff
@@ -0,0 +1,34 @@
...
+## Prerequisites
+Since SPECjbb requires a license to run, this aqa-test config requires that a licensed copy of the benchmark already exists on the machine that you are running on.
+
+## Setup
+
+Set the `SPECJBB_SRC` environment variable to the absolute path of your SPECjbb benchmark directory
+```
```
karianna, line 25:
```suggestion
```sh
```

perf/specjbb/README.md
```diff
@@ -0,0 +1,34 @@
...
+SPECjbb is a Java Client/Server performance testing benchmark. 
+Please visit its product page for more information, including a user guide: https://www.spec.org/jbb2015/
+
+
+## Prerequisites
+Since SPECjbb requires a license to run, this aqa-test config requires that a licensed copy of the benchmark already exists on the machine that you are running on.
+
+## Setup
+
+Set the `SPECJBB_SRC` environment variable to the absolute path of your SPECjbb benchmark directory
```
karianna, line 24:
```suggestion
Set the `SPECJBB_SRC` environment variable to the absolute path of your SPECjbb benchmark directory

```

perf/specjbb/run/run_multi.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
```
karianna, line 1: This shell is set differently to your other shell script

perf/specjbb/run/options/specjbb-multijvm-simple.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+# This will configure a basic run of SPECjbb in multi-jvm mode
+# Therefore, the topography of the run includes 1 group consisting of
+# 1 TransactionInjector and 1 Backend
+
+export GROUP_COUNT=1
```
karianna, line 6: Worth commenting 'why' these parameters have been set?

*Comment applies to all parameters

perf/specjbb/run/run_multi.sh
```diff
@@ -0,0 +1,80 @@
...
+for ((n=1; $n<=$NUM_OF_RUNS; n=$n+1)); do
+
+  # Create result directory                
+  temp="${RESULTS_DIR%\"}"
+  temp="${temp#\"}"
+  timestamp=$(date '+%y-%m-%d_%H%M%S')
+  result="${temp}/${timestamp}"
+  mkdir -pv $result
+  
+
```
karianna, line 17:
```suggestion
```

perf/specjbb/run/run_multi.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+
+echo "Backend opts: $JAVA_OPTS_BE"
+echo "Number of runs: ${NUM_OF_RUNS}"
+echo "Results dir: ${RESULTS_DIR}"
+
+for ((n=1; $n<=$NUM_OF_RUNS; n=$n+1)); do
```
karianna, line 8: for loop is packed with goodness.  Can we extract any of that into functions?

karianna (changes requested):
Not sure if shellcheck was run, but if not then that's worth doing so as well.

perf/specjbb/run/options/specjbb-multijvm-simple.sh
ahmedmuhsin: I included a comment on lines 2-4 explaining some of the parameters. However, I see that params on lines 14-16 need more detail since they imply a hardware config  requirement for a successful run. I'll also surface that hardware requirement to the playlist.xml file so that it is explicit.

karianna (approved):
Still a few of my comments left to go, but none are blockers so I'm approving

perf/specjbb/playlist.xml
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+<playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../TKG/playlist.xsd">
+    <include>run/testenv.mk</include>
+    <test>
+        <!-- The minimum hardware requirement to run this test is 8GB of memory and 8 CPU cores -->
+        <!-- see specjbb/run/options/specjbb-multijvm-simple.sh for more details -->
+
+        <testCaseName>SPECjbb2015-multijvm-simple</testCaseName>
```
smlambert, line 21:
```suggestion
        <testCaseName>SPECjbb2015-multijvm-simple</testCaseName>
        	<disables>
			<disable>
				<comment>Requires SPECjbb 2015 License</comment>
				<vendor>eclipse</vendor>
			</disable>
		</disables>
```

perf/specjbb/playlist.xml
```diff
@@ -0,0 +1,30 @@
...
+        <command>. $(RUN_OPTIONS)/specjbb-multijvm-simple.sh; bash $(RUN_SCRIPTS)/run_multi.sh; $(TEST_STATUS)</command>
+        <levels>
+            <level>extended</level>
```
smlambert, line 24:
```suggestion
            <level>dev</level>
```

smlambert: We can add this test to the dev level so it will never get included as a necessary test to run as part of AQAvit verification.

perf/specjbb/playlist.xml
```diff
@@ -0,0 +1,30 @@
...
+        </levels>
+        <groups>
+            <group>perf</group>
+        </groups>
```
smlambert, line 28:
```suggestion
        </groups>
        <vendors>
            <vendor>microsoft</vendor>
         </vendors>
```

smlambert: This will ensure it only gets run for the listed vendors.  Once this is in, other vendors may want to add this to their suites by adding another `<vendor>` entry.  If all vendors wanted to run it, and only we at Adoptium could not, we could instead add a `<disables><disable><comment>Requires SPECjbb 2015 License</comment><vendor>eclipse</vendor></disable></disables>`

perf/specjbb/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+```sh
+export SPECJBB_SRC=/Users/aqa-tester/SPECjbb2015-1.03
+```
+
+## Run
+
+The configuration currently includes one group, `extended`, and that group contains one test, `SPECjbb2015-multijvm-simple`. You can use the following commands to run it:
```
smlambert, line 31:
```suggestion
This test target is part of the `dev` level, and the test target name is, `SPECjbb2015-multijvm-simple`. You can use the following commands to run it:
```

perf/specjbb/run/run_multi.sh
gdams: agreed 
```suggestion
#!/usr/bin/env bash
```

karianna:
@ahmedmuhsin JIC the GH notifications didn't trigger through the MSFT firewall.

perf/specjbb/playlist.xml
ahmedmuhsin: Should I leave this suggested change out since I've already the other suggested change that added the `<disables>` tag: https://github.com/ahmedmuhsin/aqa-tests/blob/specjbb-initial-support/perf/specjbb/playlist.xml#L22 ?

smlambert: I think we should **not** use the `<disables><disable><comment>Requires SPECjbb 2015 License</comment><vendor>eclipse</vendor></disable></disables>` approach for now (which says everyone except Eclipse runs this test).  I mentioned we could move to that approach if we find that all other vendors (which include IBM, Alibaba, Huawei, Azul, Red Hat) would want to run this test.  Since we do not know that answer yet, we would only use `<vendors><vendor>microsoft</vendor></vendors>` approach (which says MSFT runs this test).

gdams (approved):

perf/specjbb/playlist.xml
gdams: @smlambert out of interest where does this value get set? Trying to work out how we (msft) would enable this test in our CI system

smlambert: its [auto-detected by TKG](https://github.com/adoptium/TKG/blob/master/src/org/openj9/envInfo/JavaInfo.java#L192), you won't have to change anything to enable this...

Mesbah-Alam:
This PR seems incomplete and it breaks the internal perf test build compilation due to `SPECJBB_SRC` not being defined (ref: backlog/issues/1051). 

From README of this PR : 

```
Set the `SPECJBB_SRC` environment variable to the absolute path of your SPECjbb benchmark directory
``` 

This test is assuming `${env.SPECJBB_SRC}/specjbb2015.jar` will be pre-staged on the machine where it will run and that `SPECJBB_SRC` will point to it. 

Error detail: 

```
23:03:45  printInput:
23:03:45       [echo] SPECJBB_SRC ${env.SPECJBB_SRC}
23:03:45       [echo] SPECJBB_DEST C:/Users/jenkins/workspace/Test_openjdk8_j9_sanity.perf_x86-32_windows/aqa-tests///..//jvmtest/perf/specjbb/suite
23:03:45       [echo] SPECJBB_JAR specjbb2015.jar
23:03:45       [echo] JAVA_BIN C:/Users/jenkins/workspace/Test_openjdk8_j9_sanity.perf_x86-32_windows/openjdkbinary/j2sdk-image/bin/java
23:03:45  
23:03:45  srcIsValid:
23:03:45       [echo] ##### Validating source SPECjbb dir
23:03:45  
23:03:45  dirIsValid:
23:03:45  
23:03:45  exists:
23:03:45  
23:03:45  isRunnable:
23:03:45       [exec] Error: Unable to access jarfile ${env.SPECJBB_SRC}/specjbb2015.jar
23:03:45       [exec] Result: 1
23:03:45  
```

PR #4370, opened 2023-02-27
Exclude EC curves fo RH ojdk 17

RadekCap:
We are removing ECCurves on our version.

I'm not sure about proper exclusion syntax, pls feel free to point me to some source if exists. I guess it from the existing exclusions around.


karianna (approved):

smlambert (commented):
thanks @RadekCap - my only comment to this is that these 2 testcases are part of jdk_security3 target, so to keep things organized by target, you may want to create a section between jdk_security and jdk_security4 in this file, and move these 2 lines into that section.

smlambert (approved):

PR #4372, opened 2023-02-27
Implement MultiJVM tck in one node

Mesbah-Alam:
- For AIX, z/OS, zLinux and pLinux, we'd run the multijvm compiler tests using two nodes. 
- For the rest of the platforms, for which there exist RI JDKs, we'd use a single-node approach. 

Implements backlog/issues/1035

### DIFF
--- a/jck/compiler.api/playlist.xml
+++ b/jck/compiler.api/playlist.xml
@@ -129,7 +129,7 @@
 		<variations>
 			<variation>NoOptions</variation>
 		</variations>
-		<command>$(GEN_JTB) tests=api/java_rmi testsuite=COMPILER'; \
+		<command>$(GEN_JTB) tests=api/java_rmi testsuite=COMPILER$(USEQ); \
 			$(START_MULTI_JVM_COMP_TEST); \
 			$(GEN_SUMMARY) tests=api/java_rmi testsuite=COMPILER; \
 		$(TEST_STATUS)</command>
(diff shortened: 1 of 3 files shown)


Mesbah-Alam:
Test on Linux: Grinder_JCK/1099.

jck/compiler.api/playlist.xml
```diff
@@ -129,7 +129,7 @@
 		<variations>
 			<variation>NoOptions</variation>
 		</variations>
-		<command>$(GEN_JTB) tests=api/java_rmi testsuite=COMPILER'; \
+		<command>$(GEN_JTB) tests=api/java_rmi testsuite=COMPILER$(USEQ); \
```
llxia, line 132: Instead of creating a new variable, can use use `$(Q)`?

jck/jck.mk
```diff
@@ -107,15 +107,37 @@ endif
...
+ifneq ($(filter aix os/390, $(OS)),)
+   USE_MULTINODE = true
+endif
+ifeq ($(OS),linux)
+   ifeq ($(ARCH),ppc64le)
+      USE_MULTINODE = true
+   else ifeq ($(ARCH),s390x)
+      USE_MULTINODE = true
+   endif
+endif
```
llxia, line 122: We should be able to use `$(SPEC)`. See https://github.com/adoptium/aqa-tests/blob/master/openjdk/openjdk.mk#L110 for example.

jck/compiler.api/playlist.xml
Mesbah-Alam: We can not, since that extra quote needs to be used only for the two-node case.

jck/jck.mk
Mesbah-Alam: That's simpler! Updated. Tested with AIX: Grinder_JCK/1125, xLinux: Grinder_JCK/1126.

llxia (approved):
Thanks @Mesbah-Alam

karianna (approved):

PR #4380, opened 2023-03-01
Disable ContinuationTest

babsingh:
ContinuationTest is being disabled for OpenJ9 since it is specific to
the reference implementation and it does not apply to OpenJ9.

Related: https://github.com/eclipse-openj9/openj9/issues/16792


babsingh:
@Mesbah-Alam @llxia Request for review.

karianna (approved):

llxia (approved):

smlambert (commented):
You may want to exclude it under this issue https://github.com/adoptium/aqa-tests/issues/1297 if it is a permanent exclude.  (We have tooling that looks at problem list files and will try to determine if the referenced issue https://github.com/eclipse-openj9/openj9/issues/16792 is closed, it will attempt to reinclude the test.  Our tooling will ignore issue 1297).

PR #4402, opened 2023-03-08
Do not change params.LABEl, which is used to specify node to run

### LINKED ISSUE #4401: params.LABEL is used to specify Jenkins node label and should not be changed.
params.LABEL is used to set to node name to run on a particular machine. Default value specified in PLATFORM_MAP. For cases we need to update accordingly(e.g. special requirements : FIPS, Dockers, label_addition, etc) the change should be limited to default.

That it https://github.com/adoptium/aqa-tests/blob/541d1bed40933936f0d0251680cb89eaad1e866f/buildenv/jenkins/openjdk_tests#L173-L180 should be moved to if (params.LABEL) else block https://github.com/adoptium/aqa-tests/blob/541d1bed40933936f0d0251680cb89eaad1e866f/buildenv/jenkins/openjdk_tests#L173-L180

sophia-guo:
Fix #4401 

### DIFF
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -170,30 +170,31 @@ timestamps{
                 LABEL = params.LABEL
             } else {
                 LABEL = PLATFORM_MAP[params.PLATFORM]["LABEL"]
-                    if (params.BUILD_LIST.contains("perf")) {
-                        def perfLabel = LABEL.minus("ci.role.test&&").concat("&&ci.role.perf")
-                        if (areNodesWithLabelOnline(perfLabel)) {
+                if (params.BUILD_LIST.contains("perf")) {
+                    def perfLabel = LABEL.minus("ci.role.test&&").concat("&&ci.role.perf")
+                    if (areNodesWithLabelOnline(perfLabel)) {
                         LABEL = perfLabel
-                        }
                     }
                 }
+                if (params.TEST_FLAG == "FIPS") {
+                    LABEL = LABEL.minus("ci.role.test&&").concat("&&ci.role.test.fips")
+                }
+                // Temporarily exclude ubuntu 22 machines for criu sanity.external pipeline
+                if (JOB_NAME.contains ("sanity.external") && JOB_NAME.contains("_criu")) {
+                    LABEL += "&&!sw.os.ubuntu.22"
+                }
+
+                if (params.DOCKER_REQUIRED) {
+                    LABEL += "&&sw.tool.docker"
+                }
+
+                if (params.LABEL_ADDITION) {
+                    LABEL += "&&${params.LABEL_ADDITION}"
+                }
+            }
 
             if (params.TEST_FLAG == "FIPS") {
-                LABEL = LABEL.minus("ci.role.test&&").concat("&&ci.role.test.fips")
                 env.EXTRA_OPTIONS += " -Dsemeru.fips=true"
-            }
-
-            // Temporarily exclude ubuntu 22 machines for criu sanity.external pipeline
-            if (JOB_NAME.contains ("sanity.external") && JOB_NAME.contains("_criu")) {
-                LABEL += "&&!sw.os.ubuntu.22"
-            }
-
-            if (params.DOCKER_REQUIRED) {
-                LABEL += "&&sw.tool.docker"
-            }
-
-            if (params.LABEL_ADDITION) {
-                LABEL += "&&${params.LABEL_ADDITION}"
             }
 
             println "SPEC: ${SPEC}"


buildenv/jenkins/openjdk_tests
```diff
@@ -170,30 +170,31 @@ timestamps{
...
+                if (params.BUILD_LIST.contains("perf")) {
+                    def perfLabel = LABEL.minus("ci.role.test&&").concat("&&ci.role.perf")
+                    if (areNodesWithLabelOnline(perfLabel)) {
                         LABEL = perfLabel
-                        }
                     }
                 }
+                if (params.TEST_FLAG == "FIPS") {
+                    LABEL = LABEL.minus("ci.role.test&&").concat("&&ci.role.test.fips")
+                }
```
llxia, line 181: This is needed regardless `params.LABEL` is provided by the users or not.

buildenv/jenkins/openjdk_tests
```diff
@@ -170,30 +170,31 @@ timestamps{
...
-            }
+                if (params.DOCKER_REQUIRED) {
+                    LABEL += "&&sw.tool.docker"
+                }
 
-            if (params.DOCKER_REQUIRED) {
-                LABEL += "&&sw.tool.docker"
+                if (params.LABEL_ADDITION) {
+                    LABEL += "&&${params.LABEL_ADDITION}"
+                }
```
llxia, line 193: Will this change the behavior? If `params.LABEL` is not provided by the users, `params.LABEL_ADDITION` will not work.

buildenv/jenkins/openjdk_tests
```diff
@@ -170,30 +170,31 @@ timestamps{
...
-                        }
                     }
                 }
+                if (params.TEST_FLAG == "FIPS") {
+                    LABEL = LABEL.minus("ci.role.test&&").concat("&&ci.role.test.fips")
+                }
+                // Temporarily exclude ubuntu 22 machines for criu sanity.external pipeline
+                if (JOB_NAME.contains ("sanity.external") && JOB_NAME.contains("_criu")) {
+                    LABEL += "&&!sw.os.ubuntu.22"
+                }
```
llxia, line 185: This is no longer needed as CRIU is moved out of sanity.external. @LongyuZhang could you remove this in a separate PR? Thanks.

buildenv/jenkins/openjdk_tests
```diff
@@ -170,30 +170,31 @@ timestamps{
...
-                env.EXTRA_OPTIONS += " -Dsemeru.fips=true"
-            }
-
-            // Temporarily exclude ubuntu 22 machines for criu sanity.external pipeline
-            if (JOB_NAME.contains ("sanity.external") && JOB_NAME.contains("_criu")) {
-                LABEL += "&&!sw.os.ubuntu.22"
-            }
+                if (params.DOCKER_REQUIRED) {
+                    LABEL += "&&sw.tool.docker"
+                }
```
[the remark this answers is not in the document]
sophia-guo: If user specify the agent why we need those extra label? Unless user is specify the label not the agent, which is not the params.LABEL is designed for.

If use specify the specific machine say test-docker-ubi8-x64-2 and also set extra label the label would be test-docker-ubi8-x64-2&&sw.tool.docker and there is no machine named that like. https://ci.adoptium.net/view/Test_grinder/job/Grinder/6817/console

```
[Pipeline] nodesByLabel
15:42:39  Could not find any nodes with 'test-docker-ubi8-x64-2&&sw.tool.docker' label
[Pipeline] timeout
15:42:39  Timeout set to expire in 15 min
[Pipeline] {
[Pipeline] node
15:42:54  Still waiting to schedule task
15:42:54  There are no nodes with the label ‘[test-docker-ubi8-x64-2&&sw.tool.docker](https://ci.adoptium.net/label/test-docker-ubi8-x64-2&&sw.tool.docker/)’
```

buildenv/jenkins/openjdk_tests
sophia-guo: If user would like to specify the agent or machine I think it should be user's responsibility to first  figure the correct the agent which satisfy all the label requirements for the jenkins job.

buildenv/jenkins/openjdk_tests
sophia-guo: I'm not sure why? Same as other parameters. User should figure out the correct machine, which has the label `ci.role.test.fips` and hasn't the label `ci.role.test`, to manually run FIPS test with specific agents. Unless the params.LABEL is also used to specify the labels instead of specific agent, that is the only reason I can think of.

sophia-guo:
Grinder specify the agent and with extra '&&sw.tool.docker' will cause no machine. https://ci.adoptium.net/view/Test_grinder/job/Grinder/6817/console

With fix. Those extra label won't affect the specified machine: https://ci.adoptium.net/view/Test_grinder/job/Grinder/6815/parameters/

Other normal cases:
No LABEL no extra label: https://ci.adoptium.net/view/Test_grinder/job/Grinder/6813/
Specify agents: https://ci.adoptium.net/view/Test_grinder/job/Grinder/6814/
No LABEL , with extra labels: https://ci.adoptium.net/view/Test_grinder/job/Grinder/6816/

buildenv/jenkins/openjdk_tests
llxia: This `user` can be our test pipeline. In the Semeru test pipeline, it sets `params.LABEL_ADDITION` in some special cases and leave `params.LABEL` empty. This change will break our pipeline.

llxia (changes requested):
Since this PR could break the Semeru pipeline, I would like to make sure the concerns are addressed.

smlambert:
I've not reviewed this at all, but I do recall thinking that it was odd/funny to create an additional boolean parameter called DOCKER_REQUIRED when really its just redirecting to append a label.  If we are to be careful about label bloat, we would consider dropping it.

buildenv/jenkins/openjdk_tests
llxia: For FIPS testing, all tests have to run on the `ci.role.test.fips` machines. This is to prevent people from running fips testing on non-fips machines.

PR #4407, opened 2023-03-08
Reorganize comment and remove unnessary message to reduce confusion

### LINKED ISSUE #4404: Only check if dynamic agents are available when none permanent agents are available
Currently the dynamic agents will be always be asked.

sophia-guo:
Close #4404 

### DIFF
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -209,12 +209,8 @@ timestamps{
                         }
                     }
                 } else {
-                    // IF no nodes are idle we will check if there is supported virtual agent
-                    // When Parallel the race condition could happen. Say the number of multiply jobs is larger than the available nodes the query's result may be delayed and wrong
-                    // In this case jobs will be fooled to fall back to wait local busy nodes. 
                     dynamicAgents = PLATFORM_MAP[params.PLATFORM]["DynamicAgents"] ? PLATFORM_MAP[params.PLATFORM]["DynamicAgents"] : []
-                    println "dynamicAgents: ${dynamicAgents}"
-
+                    // If virtual agent isn't supported fall back to normal LABEL check regardless if there are nodes available or not
                     if (params.CLOUD_PROVIDER != null && params.CLOUD_PROVIDER in dynamicAgents && LABEL == PLATFORM_MAP[params.PLATFORM]["LABEL"]) {
                         boolean isNodeIdle = false
                         node {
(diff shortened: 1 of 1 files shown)


buildenv/jenkins/openjdk_tests
```diff
@@ -230,8 +226,11 @@ timestamps{
                                 }
                             }
                         }
+                        // No agents available, start VM agent. 
+                        // When Parallel the race condition could happen. Say the number of multiply jobs is larger than the available nodes the query's result may be delayed and wrong
+                        // In this case jobs will be fooled to fall back to wait local busy nodes.
                         if (!isNodeIdle) {
-                            println "Cannot find any idle nodes. Starting dynamic vm"
+                            println "Cannot find any idle nodes. Starting dynamic vm ${params.CLOUD_PROVIDER}"
```
smlambert, line 233:
```suggestion
                            println "Cannot find any idle nodes. Starting dynamic agent from CLOUD_PROVIDER=${params.CLOUD_PROVIDER}"
```

buildenv/jenkins/openjdk_tests
```diff
@@ -230,8 +226,11 @@ timestamps{
                                 }
                             }
                         }
+                        // No agents available, start VM agent. 
+                        // When Parallel the race condition could happen. Say the number of multiply jobs is larger than the available nodes the query's result may be delayed and wrong
+                        // In this case jobs will be fooled to fall back to wait local busy nodes.
```
karianna, line 231: I don't quite understand this comment.

smlambert (commented):
We have too many different terms in the comments for just 2 main things that we are working with in this piece of code: 
- `static Jenkins nodes` 
- `dynamic agents`
We could stick to those terms, and replace anything else to avoid confusion.   Where `virtual agent`,  `VM agent`,  `vm` terms are used, we could replace with one term, `dynamic agent`.  Where `idle nodes`, `agents`, `nodes` terms are used, could replace with one term, `static Jenkins nodes`.

The idea of this code is: 
"If dynamic agents are supported for this platform, then we consider to use them, but we prefer to utilize the static Jenkins nodes first, if they are available.  If no static Jenkins nodes are available, spin up a dynamic agent.  

If a bunch of jobs get launched at once, because a parent job triggered parallel jobs to be run, there may be a situation, where all of those jobs ask the same question at the same time, _are there static Jenkins nodes available_?  The response to all of those simultaneous questions may be delayed and no longer valid.  In this situation, this code will fall back to wait for static Jenkins nodes to become available, instead of spinning up a dynamic agent."

Not suitable for code comments, but for separate documentation that we should also write, it may be easier to understand with an example: 
3 static Jenkins nodes are available, a parent job requests to run tests in parallel across 4 nodes.  All 4 child jobs run this setup code and ask if static nodes are available.  Because they all ask at the same time, the Jenkins server may respond to all 4 child jobs that static nodes are available, but only 3 will get deployed onto those nodes.  The 4th child job, will now wait for a static Jenkins node to become available, rather than spinning up a dynamic agent.

smlambert:
@sophia-guo - do you still want to try to update this one to get it in?

karianna:
Needs a rebase

sophia-guo:
Close this one as outdated.

PR #4437, opened 2023-03-17
Remove podman prune in pingPerf

llxia:
podman prune in pingPerf will clean all images. Some of them are needed for other tests

related: runtimes/backlog/issues/1057


renfeiw (approved):
LGTM

PR #4461, opened 2023-03-21
Explicitly exit the jenkins job if setting the wrong owner branch format

sophia-guo:
[owner]:[branch] format is required to set user specified repo and branch. In case of the wrong format is set, explicitly exit the jenkins job with ERROR message to help user get quickly notice the issue.

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -1291,6 +1291,8 @@ def getGitRepoBranch(ownerBranch, defaultOwnerBranch, repo) {
 		String[] tokens = ownerBranch.split(":")
 		if (tokens.size() == 2 ) {
 			actualRepoBranch = tokens;
+		} else {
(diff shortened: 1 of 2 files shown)


karianna (approved):

buildenv/jenkins/JenkinsfileBase
```diff
@@ -1291,6 +1291,8 @@ def getGitRepoBranch(ownerBranch, defaultOwnerBranch, repo) {
 		String[] tokens = ownerBranch.split(":")
 		if (tokens.size() == 2 ) {
 			actualRepoBranch = tokens;
+		} else {
+			error "[ERROR]: Wrong format of Owner and Branch. Please use the format as [owner]:[branch]"
```
llxia, line 1295: Since we print out the error message, please also print the `ownerBranch` value. For example:
```
error "[ERROR]: Wrong format of XXX_OWNER_BRANCH value: ${ownerBranch}. The expected format is [owner]:[branch]"
```

llxia (approved):

PR #4474, opened 2023-03-27
Enable CryptoTests on jdk19+

zzambers:
CryptoTests can be enabled on jdk19+.
Issue has been fixed by: https://github.com/rh-openjdk/CryptoTest/pull/31

### DIFF
--- a/functional/security/Crypto/playlist.xml
+++ b/functional/security/Crypto/playlist.xml
@@ -19,6 +19,7 @@
 			<disable>
 				<comment>https://github.com/eclipse-openj9/openj9/issues/16712</comment>
 				<version>19+</version>
+				<impl>openj9</impl>
 			</disable>
 		</disables>
 		<command>$(JAVA_COMMAND) -Xmx512m -jar $(Q)$(TEST_RESROOT)$(D)jtreg$(D)lib$(D)jtreg.jar$(Q) \


pshipton:
I'd prefer we don't enable Crypto tests for OpenJ9 until they are working. They fail consistently on Windows due to access denied https://github.com/eclipse-openj9/openj9/issues/16710. It would be wonderful if we could disable OpenJ9 Crypto tests on Windows across all versions.

They fail sometimes on Linux machines which don't have a proper /dev/random https://github.com/eclipse-openj9/openj9/issues/16720. I don't mind the occasional failure on Linux, if we enable them for Java 19 there.

zzambers:
@llxia @pshipton thanks for pointing out openj9 issues, I was not aware of some of them, I'll take a look.

karianna (approved):

zzambers:
CryptoTest being disabled on JDK19+ does not seem to be connected to openj9 issues.

Problems there seem to be infra related and not specific to JDK19+. (Blocking `/dev/random` causing timeouts, "access denied" in windows MSCAPI)

smlambert (commented):
Instead of deleting this block of code, please leave it and add 
`<impl>openj9</impl>` 
after the `<version>19+</version>` section

which effectively means it will remain disabled only for that implementation, and not for hotspot implementations.

zzambers:
@smlambert Updated

JDK19 linux x86_64: OK
https://ci.adoptium.net/view/Test_grinder/job/Grinder/7293/

smlambert (approved):

PR #4497, opened 2023-04-05
Extract JDK_VERSION from java binary in JAVA_HOME

Luigi96:
Extract JDK_VERSION from java binary in JAVA_HOME if JAVA_HOME is defined and JDK_VERSION is not


karianna (approved):

llxia:
If JDK_VERSION  is not defined, it should be auto-detected before reaching this code. May I know what is your use case?

smlambert:
I am suspecting that TKG is not being used in this case (which would be an odd choice, but noting that MSFT is running in AzDo and in past cases, skirted around use of TKG), otherwise, auto-detection and JDK_VERSION assignment is done by [TKG](https://github.com/adoptium/TKG/blob/master/src/org/openj9/envInfo/JavaInfo.java#L213) and propagated via https://github.com/adoptium/TKG/blob/master/settings.mk.

PR #4499, opened 2023-04-06
jck: Use multiJVM=group

sxa:
As per latest TC retrospective there are significant benefits to be had by avoiding the startup time of the JVM for each test. This switches the testing to use multJVM group instead of just multijvm.


andrew-m-leonard (approved):
Looks good, we shall try this in a dry-run

smlambert (approved):

PR #4506, opened 2023-04-12
Ensure multiJVM set to plain Yes for jck17+ Windows

andrew-m-leonard:
MultiJVM=group option for jck17+ causes memory exhaustion on Windows platform, so ensure for multiJVM execution type that MultiJVM is set to just "Yes"

### DIFF
--- a/jck/jtrunner/JavaTestRunner.java
+++ b/jck/jtrunner/JavaTestRunner.java
@@ -565,6 +565,11 @@ public class JavaTestRunner {
 			extraJvmOptions += " -Dfile.encoding=US-ASCII";
 		}
 
+                // testExecutionType of multiJVM_group on Windows causes memory exhaustion, so limit to plain multiJVM
+                if (getJckVersionInt(jckVersionNo) >= 17 && platform.contains("win")) {
+			fileContent += "set jck.env.testPlatform.multiJVM \"Yes\";\n";
+                }
+
 		// Set the operating system as 'Windows' for Windows and 'other' for all other operating systems.
 		// If 'other' is specified when executing on Windows, then Windows specific settings such
 		// as systemRoot are rejected, and the JCK harness assumes that DISPLAY is required for GUI tests.


smlambert (approved):
Everyone using these files picks up this change (where before they had the ability to set things in their copy of default config files).  I am approving, but ideally we sequester such changes to vendor default config so as not to force a change on others if they did not need it (like if they have beefier Windows machines).

sxa (approved):
Agree with Shelley's comment - we should revisit this post release to diagnose and attempt to test with this option enabled across environments (I'll also note that we switched it to multiJVM_group without too much discussion with external parties to my knowledge)

jck/jtrunner/JavaTestRunner.java
```diff
@@ -565,6 +565,11 @@ public static boolean jckConfigurationForMultijvmWithNoAgent() throws Exception
 			extraJvmOptions += " -Dfile.encoding=US-ASCII";
 		}
 
+                // testExecutionType of multiJVM_group on Windows causes memory exhaustion, so limit to plain multiJVM
```
Mesbah-Alam, line 568: Since the property is being set in JavaTestRunner, meaning it’d appear in the generated jtb, it will override whatever setting we have for this property in the jti file we use at IBM. If we haven't tested this change on IBM JDKs already, could we pls ensure that this is only done for HotSpot for the moment? We can then do some further testing with IBM before enabling it generally.

andrew-m-leonard: @Mesbah-Alam Yes, agree.
However, you may not realize the default runtime & compiler.jti files have changed the multiJVM default to "group"
jck.env.testPlatform.multiJVM=Yes_group

https://github.com/adoptium/aqa-tests/commit/44ea825538f9b07398ccd460bfec53d957428203

So you may want me to actually change this to below ?: 
```
                if (getJckVersionInt(jckVersionNo) >= 17 && (platform.contains("win") || isIbmJvm())) {
                        fileContent += "set jck.env.testPlatform.multiJVM \"Yes\";\n";
                }
```

andrew-m-leonard: @Mesbah-Alam using multiJVM=Yes_group (now the default) causes Windows memory exhaustion in our tests, but vastly increases Jck execution on other platforms.
However, it's probably best for openj9 to use multiJVM=Yes until you have tested with it.

andrew-m-leonard: Either way, we probably need to apply above modified change or revert https://github.com/adoptium/aqa-tests/commit/44ea825538f9b07398ccd460bfec53d957428203

Mesbah-Alam: I had not realized earlier that the jti file we use has `jck.env.testPlatform.multiJVM=yes` anyway. So, adding `isIbmJvm()` is not needed. Your original change in this PR should work fine at our end. Please ignore my previous comment. @andrew-m-leonard .

PR #4512, opened 2023-04-13
Use JavaTestUtil for all tck

Mesbah-Alam:
This PR replaces JavaTestRunner with JavaTestUtil as the tck automation wrapper. 

**Impact:**

1. _Faster Execution:_ Absence of JavaTestRunner wrapper means we are using on less JVM instance. The JVM running the actual tests via tck harness jar is not run as a sub-process spawned by the JVM running JavaTeatRunner, but executed directly from playlist. This reduces overall time taken to run JCK builds.

2. _Easy maintenance:_ Having the ability to specify tck harness parameters directly in the playlist makes each tck playlist command line clearer, easier to maintain and update. 

3. _Enhanced debug-ability:_ Absence of JavaTestRunner wrapper means less redirection hence more debug-ability for failures.

4. _Less overhead in running additional services:_ Since additional services like rmi registry, rmid, etc, will be launched directly, as opposed to via ProcessBuilder in JavaTestRunner, it should also contribute to overall improved performance of tck builds in Jenkins.

Related to backlog/issues/1013

### DIFF
--- a/jck/agent-drive.sh
+++ b/jck/agent-drive.sh
@@ -1,14 +1,84 @@
-#!/bin/bash
+#!/bin/bash -ue
 
-echo "Starting JCK agent.."
-eval $1
+jckAgentPID=0
+rmiRegistryPID=0
+rmidPID=0
+tnameservPID=0
 
-pid=$!
+startJCKAgent() {
+	echo "Starting JCK agent.."
+	eval $1
+	jckAgentPID=$!
+	echo "Agent started with PID $jckAgentPID"
+}
 
-echo "Agent started with PID $pid" 
+stopJCKAgent() {
+	echo "Test complete. Stopping JCK Agent.."
+	kill -9 $jckAgentPID
+}
 
-echo "Starting JCK harness.."
-eval $2
+startJCKHarness() {
+	echo "Starting JCK harness.."
+	eval $1
+}
 
-echo "Test complete. Stopping JCK Agent.." 
-kill -9 $pid
+startRMIRegistry() {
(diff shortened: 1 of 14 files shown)


Mesbah-Alam:
Test with sanity group on OSX: Grinder_JCK/1311.

jck/agent-drive.sh
```diff
@@ -1,14 +1,80 @@
 #!/bin/bash
 
-echo "Starting JCK agent.."
```
karianna, line 3: No set -eu?

jck/agent-drive.sh
```diff
@@ -1,14 +1,80 @@
...
-eval $1
+if [ $# -eq 2 ]; then
```
karianna, line 3: could split this into two functions?

jck/agent-drive.sh
Mesbah-Alam: Sounds good, added!

jck/agent-drive.sh
Mesbah-Alam: Cleaned up code-duplication in agent-drive.sh. Organized code in functions (test: Grinder/32994).

karianna (approved):

llxia (approved):
Thanks @Mesbah-Alam

Mesbah-Alam:
Was it tested at @smlambert 's end before merging?

smlambert:
Yes, I ran tests at the temurin-compliance server, thanks for checking @Mesbah-Alam - I should have added a note!

PR #4523, opened 2023-04-20
Set jdk8u testenv.properties release tag

andrew-m-leonard:
Since we've already built jdk8u372-b07 preemptively, but looking likely a b08 will arrive, so update to prep for that..


zdtsw (approved):

smlambert:
We will not be needing this PR. Good to have prepped in case though.  :)

PR #4541, opened 2023-04-26
Move the junit publisher step to the end of the post stage

### LINKED ISSUE #4540: Move junit publisher after other post steps
Current post steps order is: Tap publisher -> junit publisher -> archive artifacts -> add rerun failed link -> trigger rerun job.

There are cases TAP Reports finish but feels like there are some issues with junit `09:26:01  Recording test results
09:32:29  java.io.IOException: Unexpected EOF ` , which stopped the jenkins job with other following steps are disabled.

For example it happened during  April release #4525 , which we have to manually get the tap file and can't use the rerun failed | jdk_custom link to triage. 
https://ci.adoptium.net/job/Test_openjdk8_hs_extended.openjdk_ppc64le_linux_testList_2/100/console
https://ci.adoptium.net/job/Test_openjdk17_hs_extended.openjdk_aarch64_linux_testList_1/8/

Not sure if it is a junit plugin issue, but there is no harm to do the change.

sophia-guo:
Avoid the junit publisher failures affecting other functions.

Fix #4540 

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -839,8 +839,6 @@ def post(output_name) {
 
 		step([$class: "TapPublisher", testResults: "aqa-tests/TKG/**/*.tap", outputTapToConsole: false, failIfNoResults: true])
 
-		junit allowEmptyResults: true, keepLongStdio: true, testResults: '**/work/**/*.jtr.xml, **/result/**/*.jtr.xml, **/junitreports/**/*.xml, **/external_test_reports/**/*.xml'
-
 		//call the archive function for each file
 		archiveFile("aqa-tests/testenv/testenv.properties", true)
 		archiveFile("aqa-tests/TKG/**/*.tap", true)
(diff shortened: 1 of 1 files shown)


buildenv/jenkins/JenkinsfileBase
```diff
@@ -848,6 +846,8 @@ def post(output_name) {
 		}
 		String failedTests = addFailedTestsGrinderLink()
 		triggerRerunJob(failedTests)
+
+		junit allowEmptyResults: true, keepLongStdio: true, testResults: '**/work/**/*.jtr.xml, **/result/**/*.jtr.xml, **/junitreports/**/*.xml, **/external_test_reports/**/*.xml'
```
llxia, line 850: I think we should use try catch instead. Other steps can fail as well. If they fail, we do not want junit report to be missing.

sophia-guo: Other steps failures means bugs we need to fix. This failure might be the issue with the junit plugin, which we don't have control.

llxia: > Other steps failures mean bugs we need to fix.

That is not always the case. For example, we could have a network issue that results in a failure to archive the artifacts.

smlambert (commented):
The suggestion of adding a try/catch is good, expecting to see that and then this will be good to merge.

smlambert:
@sophia-guo - would you be willing to update your branch based on the review comments for a try/catch?

buildenv/jenkins/JenkinsfileBase
sophia-guo: The Jenkins Junit Plugin itself doesn't generate/throw exception, it only prepare/display unit tests report. Not sure if the try catch can catch the exception or not. Updated and see how it works.

karianna (approved):

smlambert (approved):

PR #4607, opened 2023-06-05
Limit rmi services to only rmi test + Add agent startup to jdwp + fix false negative 

Mesbah-Alam:
- Limit rmi services startup to only rmi test. 
- Add agent startup to jdwp.
- Remove `-ue` from agent-drive.sh to not exit script when harness fails (allows agent-drive.sh to stop all services started-- fixing the hang issue on java_net and java_util). 
- Capture and return exit value of harness from agent-drive.sh. Additionally, in the playlists, move `$(TEST_STATUS);` immediately below the javatest command, ahead of result summary generation. This ensures TKG can mark the test as 'failed' when it fails-- fixing the false negative issue.

### DIFF
--- a/jck/agent-drive.sh
+++ b/jck/agent-drive.sh
@@ -1,9 +1,10 @@
-#!/bin/bash -ue
+#!/bin/bash
 
 jckAgentPID=0
 rmiRegistryPID=0
 rmidPID=0
 tnameservPID=0
+harnessExitCode=0
 
 startJCKAgent() {
 	echo "Starting JCK agent.."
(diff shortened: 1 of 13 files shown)


sxa (approved):
This seems to work for my use case, so I'm approving on the basis it makes it significantly less broken than it is at the moment ;-)

Mesbah-Alam:
Test : 

- sanity.jck with jdk 8 (x64 Linux) : Grinder_JCK/1435 (Green) 
- extended.jck with jdk11 (s390x) : Grinder_JCK/1436 (Amber - with some real test failures) 
- Jdwp with temirun 17 (aarch64 Linux): Grinder_JCK/1392 (green)
-

smlambert (approved):
LGTM, thanks @Mesbah-Alam !

PR #4609, opened 2023-06-07
Update pingPerf test replacing OpenLiberty base image

LongyuZhang:
- Update pingPerf test replacing OpenLiberty base image

### DIFF
--- a/external/criu/pingPerf.sh
+++ b/external/criu/pingPerf.sh
@@ -86,7 +86,8 @@ prepare() {
         curCommitID=$(git rev-parse HEAD)
         echo "Using dockerfile from OpenLiberty/ci.docker repo branch $openLibertyBranch with commit hash $curCommitID"
         libertyDockerfilePath="releases/latest/beta/Dockerfile.ubi.openjdk17"
-        sed -i 's;FROM icr.io\/appcafe\/ibm-semeru-runtimes:open-17-ea-jdk-ubi-amd64;FROM local-ibm-semeru-runtimes:latest;' $libertyDockerfilePath
+        # replace OpenLiberty dockerfile base image
+        sed -i 's;FROM icr.io\/appcafe\/ibm-semeru-runtimes:open-17-jdk-ubi;FROM local-ibm-semeru-runtimes:latest;' $libertyDockerfilePath
     )
 }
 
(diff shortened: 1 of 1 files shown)


LongyuZhang:
Test Links:
- checkpoint image create: hyc_grinder 33221
- restore: hyc_grinder 33222

external/criu/pingPerf.sh
```diff
@@ -86,7 +86,8 @@ prepare() {
         curCommitID=$(git rev-parse HEAD)
         echo "Using dockerfile from OpenLiberty/ci.docker repo branch $openLibertyBranch with commit hash $curCommitID"
         libertyDockerfilePath="releases/latest/beta/Dockerfile.ubi.openjdk17"
-        sed -i 's;FROM icr.io\/appcafe\/ibm-semeru-runtimes:open-17-ea-jdk-ubi-amd64;FROM local-ibm-semeru-runtimes:latest;' $libertyDockerfilePath
+        # replace OpenLiberty dockerfile base image
+        sed -i 's;FROM icr.io\/appcafe\/ibm-semeru-runtimes:open-17-jdk-ubi;FROM local-ibm-semeru-runtimes:latest;' $libertyDockerfilePath
```
llxia, line 90: In case the OpenLiberty changes this again, we can add a check? If the string that we are looking for does not exist, fail the build? For example:

```
imageFrom="FROM icr.io/appcafe/ibm-semeru-runtimes:open-17-jdk-ubi;"
        grep "${imageFrom}" $libertyDockerfilePath
        rtCode=$?
        if [ $rtCode == 0 ]; then
            echo "Replace ${imageFrom} to FROM local-ibm-semeru-runtimes:latest;"
            sed -i 's;FROM icr.io\/appcafe\/ibm-semeru-runtimes:open-17-jdk-ubi;FROM local-ibm-semeru-runtimes:latest;' $libertyDockerfilePath
        else
            echo "Cannot find ${imageFrom} in https://github.com/OpenLiberty/ci.docker/blob/instanton/releases/latest/beta/Dockerfile.ubi.openjdk17"
            exit 1
        fi
```

llxia: The above code will not work as we have space in imageFrom. Maybe we can use `q` command in sed:
https://stackoverflow.com/questions/15965073/return-value-of-sed-for-no-match/15965681#15965681

llxia: We have decided to do this in a separate PR so the weekly run will fail.

llxia (approved):
LGTM

PR #4611, opened 2023-06-08
Disable multijvm targets on hotspot

smlambert:
These targets are experimental dev level and duplicate of targets already being run.  Disable for now so as not to attempt to run them without required extra machine/network configuration.


smlambert:
Related: https://github.com/adoptium/aqa-tests/pull/4590#issuecomment-1564312254

karianna (approved):

sxa (approved):
> 3487

3487 went a bit wobbly but I think that was something specific to the machine and not related to this PR. Approving so we can get this in for the weekend runs.

Mesbah-Alam (approved):
LGTM

PR #4626, opened 2023-06-15
Enable aqavit-test in alpine-linux container(#4625)

Accelerator1996:


smlambert (changes requested):
As we begin to add more and more of these examples of using containers instead of agents, we need to parameterize some things.  For certain, we need to parameterize the image being used.

Accelerator1996:
> As we begin to add more and more of these examples of using containers instead of agents, we need to parameterize some things. For certain, we need to parameterize the image being used.
> 
> Instead of doing each platform individually, riscv and alpine-linux and whatever is next, we need to design the change to allow for us to do this with any platform where containers are supported, and be able to pass in an image or file that suits the purpose.

What kind of parameterization do you expect? Does it refer to the container images selected and used according to various configurations on nodes of a specific architecture? But like riscv involves qemu and two images, I feel that its execution method will not be too different from the current one.

smlambert:
Parameterizing image and args, instead of hardcoding to alibabadragonwelljdk images `alibabadragonwelljdk/alpine_test_image` etc, as there may be others using these scripts that will want to use their own images.  

I am going to have a discussion with @sophia-guo and @llxia - it relates to broadening support for this for all platforms that can support containers and also relates to my desire to investigate the use of Testcontainers.

Accelerator1996:
> Parameterizing image and args, instead of hardcoding to alibabadragonwelljdk images `alibabadragonwelljdk/alpine_test_image` etc, as there may be others using these scripts that will want to use their own images.
> 
> I am going to have a discussion with @sophia-guo and @llxia - it relates to broadening support for this for all platforms that can support containers and also relates to my desire to investigate the use of Testcontainers.

All right. Looking forward to the outcome of your discussion

smlambert:
close as stale

PR #4631, opened 2023-06-19
Exclude overrunning SPARCV9 tests

adamfarley:
These tests take between 10 mins and 2 hours to run, so I'm excluding them to avoid timeouts. Anyone who wants them put back in should feel free to propose faster machines, make the case for 5 extra hours of runtime per execution, or find a solution for the cause of the timeouts.

### DIFF
--- a/openjdk/excludes/ProblemList_openjdk8.txt
+++ b/openjdk/excludes/ProblemList_openjdk8.txt
@@ -14,6 +14,7 @@
 
 # hotspot_all
 
+compiler/7184394/TestAESMain.java https://github.com/adoptium/infrastructure/issues/2893 solaris-sparcv9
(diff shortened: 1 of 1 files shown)


karianna (approved):

sxa (commented):
@adamfarley Can you adjust the title of the PR+Commit message to say "SPARC" rather than "Spark" please

adamfarley:
Sure thing Stewart. Will remember that for the future.

sxa (approved):
I'm happy with this but it would be good to have an explicit 👍🏻 on this from @smlambert too before merging.

smlambert (approved):
I approve it, though I think excluding the 10 min tests is likely more aggressive than required.  We may revisit those to reinclude some at some point in the future.

PR #4633, opened 2023-06-19
Exclude all multijvm dev level targets against hotspot

smlambert:
Missed this playlist when doing https://github.com/adoptium/aqa-tests/pull/4611


Mesbah-Alam (approved):
LGTM

karianna (approved):

PR #4644, opened 2023-06-22
Exclude more multijvm dev level targets against hotspot

sxa:
Follow up to https://github.com/adoptium/aqa-tests/pull/4633 and https://github.com/adoptium/aqa-tests/pull/4611

@smlambert I assume this one is ok to exclude as well - it failed in Grinder 3518

Testing in 3519

### DIFF
--- a/jck/compiler.lang/playlist.xml
+++ b/jck/compiler.lang/playlist.xml
@@ -729,6 +729,29 @@
 		</groups>
 	</test>
 	<test>
+		<testCaseName>jck-compiler-lang-EXPR-group1_multijvm</testCaseName>
+		<disables>
+			<disable>
+				<comment>Disabled temporarily in multijvm mode</comment>
+				<impl>hotspot</impl>
+			</disable>
+		</disables>
+		<variations>
+			<variation>NoOptions</variation>
+		</variations>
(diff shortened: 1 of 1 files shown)


Mesbah-Alam (approved):
LGTM

jck/compiler.lang/playlist.xml
```diff
@@ -728,6 +728,29 @@
...
+		$(TEST_STATUS); \
+		$(GEN_SUMMARY) tests=lang/EXPR testsuite=COMPILER
+		</command>
+		<levels>
+			<level>dev</level>
+		</levels>
+		<groups>
+			<group>jck</group>
+		</groups>
+	</test>
```
llxia, line 753: There is `jck-compiler-lang-EXPR-group1_multijvm` at https://github.com/adoptium/aqa-tests/blob/1305f48d1d779427b8c9b8a5050e93bed9cc7463/jck/compiler.lang/playlist.xml#L1144 
Why do we add `jck-compiler-lang-EXPR-group1_multijvm` here again and exclude it?

sxa: I hadn't spotted that it was there (I don't usually modify this file) - fixed.

smlambert (approved):

Issue #4650, opened 2023-06-29
Adoptium exclude files contain duplicate tests


adamfarley:
**Summary**

Duplicate tests in the exclude files are causing failures. Will fix and take action to help prevent future duplicates.

**Details**

When a test appears in an exclude file multiple times, the platforms in the final instance override the platforms in the other instances.

Scenario:
- Test is excluded on platform A.
- Test is excluded on platform B.
- Platform B's exclude file entry overrides the entry for platform A.
- Test subsequently re-enters the test bucket for platform A, despite the exclusion.

However, as the exclude files are not always kept alphabetical, duplicate entries can be overlooked when excluding tests.

Several instances of this have been found, including a long-running test that I tried to exclude last week, so I'm just as guilty of this as anyone.

**Plan of action**

I plan to do this:
- [x] 1) Comment out the duplicate entries in the exclude files.
- [x] 2) Ensure that the commented-out tests have their platforms appended to the non-excluded duplicate entries.
- [x] 3) Tidy up the exclude files to make them easier to read.
- [x] 4) Add a comment to Slack to advise people of the situation and help them avoid the issue in future.
- [ ] 5) Create a Git PR Check that identifies duplicate entries in the future.
Notes for 5: Cannot simply search exclude files for duplicate lines, as the hash lines are all duplicates, and duplicate *test names* may still have different platforms and issue IDs in the same line. Parse by whitespace? Must tolerate tabs and spaces.

adamfarley:
Actions 1, 2, and 3 are being worked here: https://github.com/adoptium/aqa-tests/pull/4651
Action 4 was executed here: https://adoptium.slack.com/archives/C5219G28G/p1687957144892759
Action 5 is pending.

smlambert:
5) is partially addressed (part of https://github.com/adoptium/aqa-tests/issues/1853, there is a script written as part of a CanOSP project that parses all problemlists and does some checking for correct format and duplicate entries).  Noting, it is not enabled to run regularly, and it does more than check for correct formatting and duplicate entries. (related: https://github.com/adoptium/aqa-tests/tree/master/scripts/disabled_tests#dry-run-to-see-errors-and-warnings).

adamfarley:
Looks like point 5 is handled by Shelley's issue, so I'll defer there to avoid duplication.

As for points 1-3, the PR is ready for review [here](https://github.com/adoptium/aqa-tests/pull/4651).

adamfarley:
[...]

PR #4434, opened 2023-03-16
Update problemlist files to refer to upstream issue

smlambert:
We prefer to reference the 'real' issue and close our placeholder issues when possible.  https://github.com/adoptium/aqa-tests/issues/3034 was a placeholder but now that the correct upstream issue is found, refer to it in ProblemList files instead.

https://bugs.openjdk.org/browse/JDK-8277361 appears to be fixed, which would imply we could reenable the test.  I want to make this update and then use this example to test our re-enablement automation (upcoming feature).

Closes #3034

Issue #2299, opened 2021-02-23
Trigger workflow after pushing changes to a PR


llxia:
We limited automatically triggered PR testing to `ready_for_review`. 
https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/.github/workflows/directoriesFilesChangePR.yml#L4

If people do not create a draft, the workflow does not get triggered. The workflow should be triggered after pushing changes to a PR. Maybe we should use `synchronize `. We should check the gitbub docs.

patkarns:
Changes are in pull request #2292

PR #4651, opened 2023-06-29
Consolidating duplicate excluded tests

adamfarley:
Commenting out duplicate tests from the exclude files, ensuring the platforms are appended to their unexcluded duplicates, and tidying the targets to help avoid this problem in the future.

See here for details:

https://github.com/adoptium/aqa-tests/issues/4650

### DIFF
--- a/openjdk/excludes/ProblemList_openjdk11.txt
+++ b/openjdk/excludes/ProblemList_openjdk11.txt
@@ -17,17 +17,19 @@
 ############################################################################
 
 # jdk_beans
-java/beans/PropertyEditor/6380849/TestPropertyEditor.java	https://bugs.openjdk.java.net/browse/JDK-8173082 macosx-all
-java/beans/PropertyEditor/TestColorClass.java	https://bugs.openjdk.java.net/browse/JDK-8173082 macosx-all
-java/beans/PropertyEditor/TestColorClassJava.java	https://bugs.openjdk.java.net/browse/JDK-8173082 macosx-all
-java/beans/PropertyEditor/TestColorClassNull.java	https://bugs.openjdk.java.net/browse/JDK-8173082 macosx-all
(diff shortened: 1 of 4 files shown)


adamfarley:
Duplicates have been removed. Alphabetizing now.

smlambert (approved):

sendaoYan (approved):

karianna:
@adamfarley Lots of failing test?

smlambert:
We need to update our workflows (captured in https://github.com/adoptium/aqa-tests/issues/2299)
- to not run all the tests when these files change
- to ensure Github runners are configured with up-to-date prereqs

Will not block this PR on those failures.  (Jenkins [Grinder run](https://ci.adoptium.net/view/Test_grinder/job/Grinder/7321) as a better test for this PR than using run-aqa in a workflow).

adamfarley:
> @adamfarley Lots of failing test?

Ok, got a theory vis-a-vis the failures:

Looking at a bunch of other PRs in this repo (including [this one](https://github.com/adoptium/aqa-tests/pull/4434), which is also an exclude file change), it seems that these tests aren't run for any other PRs.

My theory is that, since this PR started out as a WIP PR, it triggered the test jobs (via the github action "Directories/Files Change Test PR") when I took the WIP out of the title and clicked the "ready for review" button.

So, in short, I think we should merge this PR as-is, and raise 2 separate issues for the following problems (to be addressed AFTER the release ofc):

1. That the Git action for these tests appears to have been designed to run on every PR, and is instead only run when someone changes a PR from draft to ready-for-review. If we are meant to be running these tests for every PR, this needs fixing.
2. That there are a number of issues in these tests which need to be solved. Some issues look familiar, so perhaps they're all already solved and the fixes just need to be integrated into the Git action.

PR #4667, opened 2023-07-11
Update JCK repo default branch to 'main'

AdamBrousseau:
Internal infra 8166

### DIFF
--- a/jck/build.xml
+++ b/jck/build.xml
@@ -38,7 +38,7 @@
 		<isset property="isZOS"/>
 	</condition>
 	
-	<condition property="jck_branch" value="${env.JCK_GIT_BRANCH}" else="master">
+	<condition property="jck_branch" value="${env.JCK_GIT_BRANCH}" else="main">
 		<isset property="env.JCK_GIT_BRANCH"/>
 	</condition>
 	
@@ -176,7 +176,7 @@
 										<exec executable="git" dir="${JCK_ROOT_USED}" failonerror="true">
 											<arg value="reset" />
 											<arg value="--hard" />
-											<arg value="origin/master" />
+											<arg value="origin/main" />
 										</exec>
 										<exec executable="git" dir="${JCK_ROOT_USED}" failonerror="true">
 											<arg value="clean" />


karianna (approved):

llxia (approved):

llxia:
@smlambert I do not think this will affect adoptium, but could you run a Grinder to check?

smlambert (changes requested):
```
16:55:33       [exec] hint: You can replace "git config" with "git config --global" to set a default
16:55:33       [exec] hint: preference for all repositories. You can also pass --rebase, --no-rebase,
16:55:33       [exec] hint: or --ff-only on the command line to override the configured default per
16:55:33       [exec] hint: invocation.
16:55:35       [exec] fatal: couldn't find remote ref main
16:55:35       [exec] Result: 1
16:55:35       [echo] Performing hard reset...
16:55:35       [exec] fatal: ambiguous argument 'origin/main': unknown revision or path not in the working tree.
16:55:35       [exec] Use '--' to separate paths from revisions, like this:
16:55:35       [exec] 'git <command> [<revision>...] -- [<file>...]'
```
I believe it implies we would have to change our *-unzipped repositories (which we do not have read or write access to do).  We do not use the tagged aqa-tests repo for running TCKs, we use the master branch.  We will need to raise a EF helpdesk and I would not want to be attempting to do it prior to our July release.

llxia:
What is your JCK_GIT_BRANCH value in Grinder? Is it set to `master`?

smlambert:
I do not have that parameter in our Grinders, so it was unset.  I see in our test jobs it is set to master.  I am going to create a metric called NumberOfAQAvitParams which must be nearing 100 by now.  What can be done to reduce the number, or at least slow the steady growth.

llxia:
I think we should raise an EF helpdesk ticket to update Adoptium *-unzipped repos to `main` after the July release so that we do not have to maintain code for both `master` and `main`.

smlambert:
> I think we should raise an EF helpdesk ticket to update Adoptium *-unzipped repos to main after the July release so that we do not have to maintain code for both master and main.

💯 agree

PR #4669, opened 2023-07-13
Update `testenv.properties` in v0.9.8 branch with GA level guesses

sxa:
JDK20 could be anything since we don't have any tags for 20.0.2 yet.
For the others I'm going to guess that we'll get updates compared to where we are now.


llxia (changes requested):
This PR is created against master branch, not v0.9.8 branch

sxa:
> This PR is created against master branch, not v0.9.8 branch

Well spotted - superceding with https://github.com/adoptium/aqa-tests/pull/4671

PR #4675, opened 2023-07-17
Update versions for jck-compiler-api-java_rmi

renfeiw:
- remove the versions tag as it should run against all versions


llxia:
Please make sure this test passes on JDK17 and backport the change into v0.9.8-release branch.

PR #4685, opened 2023-07-25
Thread/virtual/PreviewFeaturesNotEnabled no longer exists in JDK21

babsingh:
java/lang/Thread/virtual/PreviewFeaturesNotEnabled no longer exists in
https://github.com/ibmruntimes/openj9-openjdk-jdk21 since Virtual
Threads is no longer a preview and will GA in JDK21.

Related: eclipse-openj9/openj9#16044


babsingh:
@llxia @LongyuZhang Requesting your review.

llxia (approved):

PR #4725, opened 2023-08-24
Add c99 CFLAG option for JDK21 p/zLinux natives

jdekonin:
@llxia


jdekonin:
I modified the condition to work for jdk21+ using the method that is consistent with how its performed in the rest of the script.  Ready for review.

llxia (approved):
Thanks @jdekonin

PR #4728, opened 2023-08-25
Set failIfNotNew to false

llxia:
If the test run and the plugin run are more than 3 seconds apart, the plugin will throw NoNewTestReportException. Therefore, set failIfNotNew to false

resolves: backlog/issues/1187


LongyuZhang (approved):
LGTM

renfeiw (approved):
LGTM

PR #4750, opened 2023-09-07
Update jtrunner code message and keyword

### LINKED ISSUE #4637: Error message enhancement: Multijvm tests with Agent turned off
### Problem 1: Wrong Error

In JavaTestRunner.java, there's an error that reads:

```
System.out.println("Only non-multijvm tests are supported with Agent turned off");
```

[Link.](https://github.com/adoptium/aqa-tests/blob/4a4f3aa15ef338ed8feba3a1353ba922612755c2/jck/jtrunner/JavaTestRunner.java#L1024)

However, the if statement that must fail in order to produce this error is:

```
if (testExecutionType.equals("multijvm") && withAgent.equals("off")) {
```

[Link.](https://github.com/adoptium/aqa-tests/blob/4a4f3aa15ef338ed8feba3a1353ba922612755c2/jck/jtrunner/JavaTestRunner.java#LL901C3-L901C73)

@Mesbah-Alam - Could you please review this? These seem to contradict each other. Either Multijvm tests **can** run with the agent turned off, or they **cannot**, surely?

### Problem 2: Unused keyword

On [line 608](https://github.com/adoptium/aqa-tests/blob/21c8dd1eeb39e90c8eac00cbcdba25129a80ca4f/jck/jtrunner/JavaTestRunner.java#L608) the string "&!robot" is appended to the "keyword" variable.

The keyword/s have already been added to fileContent for RUNTIME, so this keyword will never be used.

@Mesbah-Alam - Can you verify this? Perhaps we should move that logic further up?

smlambert:
Fixes #4637

### DIFF
--- a/jck/jtrunner/JavaTestRunner.java
+++ b/jck/jtrunner/JavaTestRunner.java
@@ -565,6 +565,10 @@ public class JavaTestRunner {
 				return false; 
 			}
 			
+			if ( tests.contains("api/java_awt") || tests.contains("api/javax_swing") || tests.equals("api") ) {
+				keyword += "&!robot";
+			}
+
 			fileContent += "concurrency " + concurrencyString + ";\n";
 			fileContent += "timeoutfactor 4" + ";\n";	// 4 base time limit equal 40 minutes
 			fileContent += keyword + ";\n";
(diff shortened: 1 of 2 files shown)


smlambert:
Run some checks on TC Jenkins 
- Grinder/3718
- Grinder/3719

tellison (approved):

Issue #4701, opened 2023-08-14
Check build.xml files to ensure they fail fast when one of their targets fail


smlambert:
This came up in context of JavaTestUtil.java failed to compile, but we continued on to try and run tests.  See: https://github.com/adoptium/aqa-tests/blob/master/jck/build.xml

smlambert:
Noting that the build-jtrunner target which is how JavaTestUtil is compiled has failonerror="true" set, looking at if it propagated properly.

Rajeevjewar:
Dear @smlambert,

I would like to request that you kindly assign this issue to me. As a beginner open-source contributor, I am eager to take on the challenge and work towards fixing it. Your support and guidance would be greatly appreciated.

Thank you in advance.

Best regards,
Rajeev Kumar

PR #4763, opened 2023-09-18
Change all failonerror="false" to true

Rajeevjewar:
Hi @smlambert,
I wanted to let you know that I have made some changes to address the #4701 issue we discussed earlier. I would appreciate it if you could review my pull request and merge it into the master branch if everything looks good to you.

Thank you for your time and attention.

### DIFF
--- a/jck/build.xml
+++ b/jck/build.xml
@@ -162,7 +162,7 @@
 									</then>
 								</if>
 								<echo message="Updating ${JCK_ROOT_USED} with latest..." />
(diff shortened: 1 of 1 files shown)


smlambert:
Hi @Rajeevjewar - this was not exactly what we need to have done.  Everything that is explicitly marked as failonerror=false should stay that way.

Other targets that do not explicitly have failonerror defined need to be considered.

Also, in order for you to contribute, you will need to sign the ECA (details on how to do that are here: https://adoptium.net/docs/eca-sign-off/).

PR #4787, opened 2023-09-26
add ProblemList_openjdk21.txt for Microsoft

gdams:
We noticed this was missing so I've copied the Adoptium problem list. Not sure if I need to copy this to other vendor directories too?

### DIFF
--- a/openjdk/excludes/vendors/microsoft/ProblemList_openjdk21.txt
+++ b/openjdk/excludes/vendors/microsoft/ProblemList_openjdk21.txt
@@ -1,0 +1,520 @@
+############################################################################
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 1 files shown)


sophia-guo:
Yes, the problem list is needed per jdk version for all vendors. However, if there are no specific tests need to be excluded per vendor an empty problem list file as a placeholder should be enough.

sophia-guo (approved):

smlambert (approved):

smlambert:
@sophia-guo - I was not aware that the vendor problem list file was actually needed, I thought it was used if it is there and not used if not present.

sophia-guo:
> I was not aware that the vendor problem list file was actually needed, I thought it was used if it is there and not used if not present.

We can definitely enhance this.

gdams:
> @sophia-guo - I was not aware that the vendor problem list file was actually needed, I thought it was used if it is there and not used if not present.

Yeah it currently produces an error if it doesn't exist

sophia-guo:
Issue address this. https://github.com/adoptium/aqa-tests/issues/4790

PR #4791, opened 2023-09-27
Include vendor's problem list only if there is one

### LINKED ISSUE #4790: Only includes vendor's openjdk problem list if there is one available.
The vendor's problem list is needed per jdk version for all vendors. Ideally it should work as 'it was used if it is there and not used if not present'.

sophia-guo:
Close #4790 


llxia (approved):

karianna (approved):

karianna (changes requested):
Even though those problems lists are empty, downstream providers may rely on them?

sophia-guo:
> Even though those problems lists are empty, downstream providers may rely on them?

If Vendors do need to exclude tests per their specific reason per JDK they can create directly add the files.

smlambert:
@karianna - this PR contains a change in openjdk.mk that means empty files are no longer needed as placeholders.  
Screenshot 2023-09-28 at 10 02 39 AM

If downstream vendors have specific tests that need to be excluded, that are not in the [common set of ProblemList](https://github.com/adoptium/aqa-tests/tree/master/openjdk/excludes) files applicable to all vendors, then you can add the file back into the [vendor specific ProblemList files](https://github.com/adoptium/aqa-tests/tree/master/openjdk/excludes/vendors) with an entry in it.  It makes no sense for us to force vendors to add empty files in order to run without error, so the change in the openjdk.mk file of the PR fixes that behaviour.

smlambert (approved):

karianna (approved):

PR #4792, opened 2023-09-28
Fix concurrency on non-cgroups linux

sxa:
There appears to be a bug in the logic when detecting the amount of RAM available on a Linux machine which does not have cgroups enabled.

In this case, the amount of RAM is coming back in GB instead of MB so ends up looking very small when the concurrency calculation is performed. This should resolve it, however it needs to be checked on more machines since I found that `/sys/fs/cgroup/memory/memory.limit_in_bytes` on one of out Linux/ppc64le machines (`test-osuosl-ubuntu1604-ppc64le-2`) was `9223372036854710272` which doesn't seem right.

Fixing this has the potential to dramatically decrease the openjdk test job times on some Linux machines (Spotted on RISC-V)

### DIFF
--- a/openjdk/openjdk.mk
+++ b/openjdk/openjdk.mk
@@ -46,7 +46,7 @@ endif
 # Upstream OpenJDK, roughly, sets concurrency based on the
 # following: min(NPROCS/2, MEM_IN_GB/2).
 MEM := $(shell expr $(MEMORY_SIZE) / 2048)
-CORE := $(shell expr $(NPROCS) / 2)
+CORE := $(shell expr $(NPROCS))
 CONC := $(CORE)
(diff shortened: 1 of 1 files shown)


sxa:
This change (resulting in `concurrency:2` improves the execution time on the RISC-V systems it was spotted on by around a third for sanity.openjdk. Increasing it to 3 provides a slight improvement (In the ballpark of around 10%) but does result in a CPU load in excess of the number of cores for most of the execution so I would not necessarily advocate increasing it further, at least for systems with 4 cores.

sxa:
Superceded by other PRs - most recently https://github.com/adoptium/aqa-tests/pull/5063 - so closing.

Issue #1636, opened 2020-02-19
JTReg Fail on arm32: java/net/URLPermission/URLTest.java


M-Davies:
https://github.com/AdoptOpenJDK/openjdk-tests/issues/1548

- test category, sanity.openjdk.jdk_net_0
- OS/architecture, aarch32_linux
- public build SHAs (i.e. java -version output), in Grinder provides JDK_VERSION/JDK_IMPL info and helps identify point at which regression introduced: 
```
02:48:13  openjdk version "1.8.0_242"
02:48:13  OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_242-202001150439-b06)
02:48:13  OpenJDK Client VM (AdoptOpenJDK)(build 25.242-b06, mixed mode)
```
- Created via non-JIT VM for the JDK8-hotspot-arm32 GA release (see linked issue)
- Known to fail on JDK11-arm32 #802 

Test itself passes 4/5 times. Example:
```
STDOUT:
04:48:50  url=http://127.0.0.1:12567/foo.html method=GET header1=X-Foo header2 = null exceptionExpected=false
04:48:50  OK
04:48:50  url=http://127.0.0.1:12567/foo.html method=GET header1=Z-Bar header2 = X-Foo exceptionExpected=false
04:48:50  OK
04:48:50  url=http://127.0.0.1:12567/foo.html method=GET header1=X-Foo header2 = Z-Bar exceptionExpected=false
04:48:50  OK
04:48:50  url=http://127.0.0.1:12567/foo.html method=GET header1=Z-Bar header2 = null exceptionExpected=false
04:48:50  OK
04:48:50  url=https://127.0.0.1:12568/foo.html method=POST header1=X-Fob header2 = null exceptionExpected=false
04:48:50  OK
04:48:50  url=http://127.0.0.1:12567/foo.html method=POST header1=X-Foo header2 = null exceptionExpected=true
04:48:50  OK
04:48:50  url=https://127.0.0.1:12568/foo.html method=GET header1=X-Fob header2 = null exceptionExpected=true
04:48:50  OK
04:48:50  url=http://127.0.0.1:12567/bar.html method=GET header1=X-Foo header2 = null exceptionExpected=true
04:48:50  OK
04:48:50  url=https://127.0.0.1:12568/bar.html method=POST header1=X-Fob header2 = null exceptionExpected=true
04:48:50  OK
04:48:50  STDERR:
[...]

Issue #1635, opened 2020-02-19
JTReg Fail on arm32: com/sun/net/httpserver/Test9a.java


M-Davies:
https://github.com/AdoptOpenJDK/openjdk-tests/issues/1548

- test category, sanity.openjdk.jdk_net_0
- OS/architecture, aarch32_linux
- public build SHAs (i.e. java -version output), in Grinder provides JDK_VERSION/JDK_IMPL info and helps identify point at which regression introduced: 
```
02:48:13  openjdk version "1.8.0_242"
02:48:13  OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_242-202001150439-b06)
02:48:13  OpenJDK Client VM (AdoptOpenJDK)(build 25.242-b06, mixed mode)
```
- Created via non-JIT VM for the JDK8-hotspot-arm32 GA release (see linked issue)
- Known to fail on JDK11-arm32 #802 

```
02:55:34  STDOUT:
02:55:34  Test9a: 
02:55:34  STDERR:
02:55:34  javax.net.ssl.SSLException: java.lang.ClassCastException: [B cannot be cast to [B
02:55:34  	at sun.security.ssl.Alerts.getSSLException(Alerts.java:208)
02:55:34  	at sun.security.ssl.SSLSocketImpl.fatal(SSLSocketImpl.java:1946)
02:55:34  	at sun.security.ssl.SSLSocketImpl.fatal(SSLSocketImpl.java:1903)
02:55:34  	at sun.security.ssl.SSLSocketImpl.handleException(SSLSocketImpl.java:1886)
02:55:34  	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1402)
02:55:34  	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1379)
02:55:34  	at sun.net.www.protocol.https.HttpsClient.afterConnect(HttpsClient.java:559)
02:55:34  	at sun.net.www.protocol.https.AbstractDelegateHttpsURLConnection.connect(AbstractDelegateHttpsURLConnection.java:185)
02:55:34  	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1570)
02:55:34  	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1498)
[...]

PR #4806, opened 2023-10-11
Workflow enhancement to handle Git failures

KapilPowar:
runtimes/backlog/issues/1184

Update test workflow to handle Git failures.
   -  fixes issue of corrupted git directory resulting in test failure

Signed-off-by: Kapil Powar [kapil.powar@in.ibm.com](mailto:kapil.powar@in.ibm.com)

CC: @llxia

### DIFF
--- a/jck/build.xml
+++ b/jck/build.xml
@@ -92,31 +92,42 @@
 			<available file="${JCK_ROOT_USED}" type="dir" />
 			<then>
 				<!--Obtain local and remote SHA to figure out if local materials are up-to-date--> 
-				<exec executable="git" dir="${JCK_ROOT_USED}" outputproperty="localSHA">
+				<exec executable="git" dir="${JCK_ROOT_USED}" failonerror="false" outputproperty="localSHA" resultproperty="return.code">
 					<arg line="rev-parse"/>
 					<arg line="HEAD"/>
 				</exec>
-				<exec executable="git" dir="${JCK_ROOT_USED}" outputproperty="remoteSHA">
-					<arg line="ls-remote"/>
-					<arg line="${JCK_GIT_REPO_USED}"/>
-					<arg line="${jck_branch}"/>
-				</exec>
-				<loadresource property="localSHATrimmed">
-					<propertyresource name="localSHA"/>
-					<filterchain>
-						<tokenfilter>
-							<trim/>
-						</tokenfilter>
-						<striplinebreaks/>
-					</filterchain>
-				</loadresource>
-					
-				<echo message="LocalSHA = --${localSHATrimmed}--"/>
-				<echo message="RemoteSHA= --${remoteSHA}--"/>
+				<if>
+					<not>
+						<equals arg1="${return.code}" arg2="0"/>
+					</not>
+					<then>
+						<echo message="Error: Git command failed with return code ${return.code}. Delete the directory and continue." />
+						<delete dir="${JCK_ROOT_USED}" failonerror="true" />
+					</then>
+					<else>
+						<exec executable="git" dir="${JCK_ROOT_USED}" failonerror="true" outputproperty="remoteSHA">
+							<arg line="ls-remote"/>
+							<arg line="${JCK_GIT_REPO_USED}"/>
+							<arg line="${jck_branch}"/>
+						</exec>
+						<loadresource property="localSHATrimmed">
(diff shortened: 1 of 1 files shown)


KapilPowar:
Here are test runs for the changes
[Grinder #1635](https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder_JCK/1635/console)
[Grinder #1636](https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder_JCK/1636/console)

jck/build.xml
```diff
@@ -92,31 +92,42 @@
...
-				<echo message="LocalSHA = --${localSHATrimmed}--"/>
-				<echo message="RemoteSHA= --${remoteSHA}--"/>
+				<if>
+					<not>
+						<equals arg1="${return.code}" arg2="0"/>
+					</not>
+					<then>
+						<echo message="fatal: not a git repository.... Delete the directory and retry" />
+						<delete dir="${JCK_ROOT_USED}" failonerror="false" />
+					</then>
```
jdekonin, line 106: I suggest a more generic error message in this case since the error check is for anything non-zero.  I believe we also want to fail on error for deleting the directory, otherwise if there is a failure with the existing repo it will just resume and fail in the update materials section with the same error.

jck/build.xml
```diff
@@ -92,31 +92,42 @@
 			<available file="${JCK_ROOT_USED}" type="dir" />
 			<then>
 				<!--Obtain local and remote SHA to figure out if local materials are up-to-date--> 
-				<exec executable="git" dir="${JCK_ROOT_USED}" outputproperty="localSHA">
+				<exec executable="git" dir="${JCK_ROOT_USED}" failonerror="false" outputproperty="localSHA" resultproperty="return.code">
 					<arg line="rev-parse"/>
 					<arg line="HEAD"/>
 				</exec>
-				<exec executable="git" dir="${JCK_ROOT_USED}" outputproperty="remoteSHA">
```
jdekonin, line 99: Having failonerror default (true) here is right, as if there is a connection to github issue we don't want the testing to continue.

KapilPowar: Added failonerror(true) as per comment.

jck/build.xml
KapilPowar: Changed to more generic error message and changed fail on error to true for deleting the directory.

jdekonin: Technically its not "retry" but "continue".  Other than that I believe this does make recovery in a "corrupt" local cache possible and better than before.  While I can approve I should note that I am not a regular Adopium reviewer/committer that is intimately familiar with how this might have a larger impact.  @llxia can you please also advise.

jdekonin (approved):

llxia (approved):
LGTM

llxia:
FYI @AdamBrousseau @pshipton

jck/build.xml
KapilPowar: Changed to `continue`. Thanks

AdamBrousseau:
I would recommend a better commit comment. You have linked to an internal IBM issue which nobody outside of IBM will be able to see. Your commit comment should explain what you're changing and why. This repo has a contributing.md ([here](https://github.com/adoptium/aqa-tests/blob/master/Contributing.md)). I try to follow, and by default I direct people to [OpenJ9's guidelines on commit comments](https://github.com/eclipse-openj9/openj9/blob/master/CONTRIBUTING.md#commit-guidelines). You will thank yourself later when you're scrolling through the history trying to track down a change.

AdamBrousseau (commented):
Will this only handle the issue where the `.git` dir is missing? I was originally thinking it would cover all cases where the git update fails.

llxia:
This PR will check the first git cmd, that is `git rev-parse HEAD`. If this cmd does not run successfully (regardless of the error), the program will delete the test dir and re-clone. This should cover most of the cases that we have seen so far. 
However, it will not catch the case where it fails at later git cmd.

I think we should get this change in to ease the problems. Meanwhile, working on a solution to handle all git cases. To be honest, I think we should re-write the logic in this file.

btw, If the material is in a bad state (i.e., jenkins user does not have permission to delete the material), there is not much we can do.

LongyuZhang (approved):
LGTM

PR #4854, opened 2023-11-03
(v0.9.9) Update 0.9.9 testenv for z/OS to use the release branches

pshipton:
ibm-jdk11-zOS-11.0.21
ibm-jdk17-zOS-17.0.9


pshipton:
Some day we may need to have a different OpenJ9 branch for z/OS 11 vs 17, they build from different branches, although today these branches have the same test material.

llxia (approved):
LGTM

PR #4869, opened 2023-11-15
Automate API check for updates

KapilPowar:
Automate API check for updates

Check if update is available in Artifactory for any JDK version and trigger sync job to update repository.

### DIFF
--- a/buildenv/jenkins/API_check
+++ b/buildenv/jenkins/API_check
@@ -1,0 +1,128 @@
+#!groovy
+
+timestamps{
+    stage('Setup') {
+        def CHECK_LABEL = params.CHECK_LABEL ?: "ci.role.test&&hw.arch.x86&&sw.os.linux"
+        println "CHECK_LABEL: ${CHECK_LABEL}"
+        node(CHECK_LABEL) {
+            //echo ("clone the SCM GIT repo")
+            try {
+                def gitConfig = scm.getUserRemoteConfigs().get(0)
+                timeout(time: 1, unit: 'HOURS') {
+                        forceCleanWS()
+                    }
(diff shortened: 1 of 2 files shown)


buildenv/jenkins/API_check
```diff
@@ -0,0 +1,128 @@
...
+                    echo ("Automate API script output:")
+                    echo scriptOutput
+
+                    jdkList= getJDKList(scriptOutput)
+                    jdkList.each { key, value ->
+                        println("JDK number: $key")
+                        def CHILD_JOB_NAME="JCK_Sync"
+                        def BUILD_URL="${JENKINS_URL}job/${CHILD_JOB_NAME}/"
+                        def CHILD_PARAMS = []
+                        def JCK_GIT_REPO = "git@github.ibm.com:runtimes/JCK${key}-unzipped.git"
```
jdekonin, line 80: The script is also used by other vendors (e.g. Adoptium), not only ibm, so the link repo cannot be hardcoded.  Suggest a root location be passed into the script (jenkins env var) and JCK_GIT_REPO be constructed.

jck/jck-semiauto-updater/update-check.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+
+# File to store the previous state
+PREVIOUS_STATE_FILE="/home/jenkins/previous_state.txt"
```
jdekonin, line 4: Shouldn't assume that the system where this would be run will have a /home/jenkins folder.  If appropriate, would the use of ${WORKSPACE} be better?  Or maybe setting to the location of the script if the file is only intended for temporary use.

PR #4887, opened 2023-11-22
Add ant/unzip: command not found in the error list

llxia:
If the error is found in the console output, the node will be marked offline.

related: backlog/issues/983#issuecomment-67423294


llxia:
[Grinder](https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/36388)
The problematic node is removed, so we can test the scenario. The above Grinder is just to ensure this PR does not break anything.

pshipton:
Could this be more generic and just look for `command not found`? Or if that is too scary, we should check for `ant: command not found` as well, which was another earlier problem we saw.

smlambert (approved):

llxia:
Yes, `command not found` is too scary. I will add `ant: command not found`.

llxia:
PR is updated to include `ant: command not found`

pshipton (approved):

Issue #4846, opened 2023-11-01
Update git clone logic in build.xml 


llxia:
We have encountered lots of test material staging problems (see [build.xml](https://github.com/adoptium/aqa-tests/blob/master/jck/build.xml#L89)). In the build.xml, the logic is complicated as we try to handle lots of different cases. 

To make this simple, if git cmd fails, we should remove the pre-staged material and re-clone.

llxia:
@annaibm can you take a look at this one? Thanks

llxia:
Here is the general logic:
- if the material does not exist, git clone it
- if it exists, check SHA
- if SHA does not match, update
- if the update fails, git hard reset
- if any of the above cmd fails, remove the test material and git clone again

PR #4891, opened 2023-11-28
Update git clone logic in build.xml

annaibm:
Removes prestaged jck material and reclones it, if git command fails.

resolves: https://github.com/adoptium/aqa-tests/issues/4846

### DIFF
--- a/jck/build.xml
+++ b/jck/build.xml
@@ -22,7 +22,7 @@
 	<property name="DEST" value="${BUILD_ROOT}/jck" />
 	<property environment="env" />
 
-	<condition property="isZOS" value="true">
+	<condition property="isZOS" value="true" else="false">
 		<os family="z/os" />
 	</condition>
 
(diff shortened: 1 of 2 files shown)


annaibm:
Grinder_jck test: 2010/2011

jck/build.xml
```diff
@@ -101,8 +118,13 @@
 						<equals arg1="${return.code}" arg2="0"/>
 					</not>
 					<then>
+						<!-- Git command failed, remove pre-staged material and re-clone -->
 						<echo message="Error: Git command failed with return code ${return.code}. Delete the directory and continue." />
-						<delete dir="${JCK_ROOT_USED}" failonerror="true" />
+						<delete dir="${JCK_ROOT_USED}" failonerror="false" />
```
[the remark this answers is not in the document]
LongyuZhang: Also, if you delete ${JCK_ROOT_USED}, then git clone directly, I think the folder structure will be different than what line 163 is. maybe should delete ${JCK_GIT_REPO_USED} instead?

jck/build.xml
```diff
@@ -139,14 +161,8 @@
...
-				<exec executable="git" dir="${JCK_ROOT_USED}/.." failonerror="true">
-					<arg value="clone" />
-					<arg value="--single-branch" />
-					<arg value="-b"/>
-					<arg value="${jck_branch}"/>
-					<arg value="${JCK_GIT_REPO_USED}" />
-					<arg value="${JCK_ROOT_USED}" />
-				</exec>
+				<!-- Use the git-clone macro -->
+					<git-clone branch="${jck_branch}" repo="${JCK_GIT_REPO_USED}" target-dir="${JCK_ROOT_USED}" />
```
karianna, line 165:
```suggestion
				<git-clone branch="${jck_branch}" repo="${JCK_GIT_REPO_USED}" target-dir="${JCK_ROOT_USED}" />
```

karianna (approved):

LongyuZhang (approved):
LGTM. Thanks @annaibm

llxia (changes requested):
This problem is not limited to git clone. It is for general git cmd. I think it would be easier to rewrite the logic from scratch. ant script may not be the best way to do it.

llxia:
Converted the PR in draft. Please change the PR in `Ready for review` when it is ready.

jck/build.xml
```diff
@@ -15,14 +15,15 @@
 -->
 
 <project name="jck" default="build" basedir=".">
+	<!-- Define a macro for Git clone with error handling -->
```
LongyuZhang, line 18: This comment could be removed

jck/build.xml
```diff
@@ -15,14 +15,15 @@
...
 	<taskdef resource="net/sf/antcontrib/antlib.xml" />
 	<description>Build JCK Tests</description>
 
 	<!-- set global properties for this build -->
 	<property name="DEST" value="${BUILD_ROOT}/jck" />
 	<property environment="env" />
 
-	<condition property="isZOS" value="true">
+	<condition property="isZOS" value="true" else="">
```
LongyuZhang, line 26: Setting `else` to `false` would be better to avoid parameter pass in errors.

jck/build.xml
```diff
@@ -87,156 +88,16 @@
...
-		</if> 
+		<exec executable="bash" failonerror="false">
+			<arg value="update_git_material.sh"/>
+			<!-- Add other arguments as needed -->
+			<!-- Pass all properties as environment variables -->
+			<!-- Pass all properties as environment variables -->
+			<env key="JCK_ROOT_USED" value="${JCK_ROOT_USED}" />
+			<env key="JCK_GIT_REPO_USED" value="${JCK_GIT_REPO_USED}" />
+			<env key="jck_branch" value="${jck_branch}" />
+			<env key="isZOS" value="${isZOS}" />
```
LongyuZhang, line 99: Setting `env` variables are not necessary here, you can pass in them as parameters of the shell script, you can refer this test for it ([link](https://github.com/adoptium/aqa-tests/blob/c089b7328234947ed6cc7d24632c084049da5e5c/external/criu-portable-checkpoint/build.xml#L21-L27))

jck/build.xml
```diff
@@ -87,156 +88,16 @@
...
-						</exec>
-						<exec executable="git" dir="${JCK_ROOT_USED}" failonerror="false">
-							<arg value="reset" />
-							<arg value="--hard" />
-						</exec>
-					</else>
-				</if>
-			</then>
-		</if> 
+		<exec executable="bash" failonerror="false">
```
LongyuZhang, line 91: Not sure if we need to set `true` here?

jck/update_git_material.sh
```diff
@@ -0,0 +1,145 @@
...
+    if [ $? -eq 0 ]; then
+        # Change into the parent directory
+        cd "$(dirname "$clone_directory")" || exit 1
+        # Clone the repository into the target directory
+        git clone --single-branch -b "${jck_branch}" "${JCK_GIT_REPO_USED}" "$(basename "$clone_directory")"
+        # Check if the clone was successful
+        if [ $? -eq 0 ]; then
+            echo "Repository cloned successfully into $clone_directory"
+        else
+            echo "Failed to clone repository"
```
LongyuZhang, line 43: If git clone fails, we need to exit too, not only print the error message

jck/update_git_material.sh
```diff
@@ -0,0 +1,145 @@
...
+        fi
+    else
+        echo "Failed to create parent directory: $(dirname "$clone_directory")"
```
LongyuZhang, line 46: Same here for exit.

jck/update_git_material.sh
```diff
@@ -0,0 +1,145 @@
...
+    if [ -d "$JCK_ROOT_USED/.git" ] || git -C "$JCK_ROOT_USED" rev-parse --is-inside-work-tree &>/dev/null; then
+        echo "The directory is a valid Git repository."
+        check_and_handle_updates
+    else
+        echo "The directory is not a valid Git repository."
+        remove_and_clone
+    fi    
+else
+    # JCK materials don't exist, git clone
+    git_clone || remove_and_clone
```
LongyuZhang, line 144: If the directory doesn't exist, `git_clone` once would be enough, no need `remove_and_clone`.

jck/build.xml
```diff
@@ -87,156 +87,16 @@
...
-						</exec>
-					</else>
-				</if>
-			</then>
-		</if> 
+		<exec executable="bash" failonerror="true">
+			<arg value="update_git_material.sh"/>
+			<!-- Add other arguments as needed -->
+			<!-- Pass all properties as environment variables -->
+			<!-- Pass all properties as environment variables -->
```
karianna, line 94: Duplicate comment

jck/update_git_material.sh
```diff
@@ -0,0 +1,151 @@
+#!/bin/bash
+# Now you can use these variables in your script
```
karianna, line 2:
```suggestion
# Now you can use these variables in your script

```

jck/update_git_material.sh
```diff
@@ -0,0 +1,151 @@
...
+JCK_ROOT_USED="$1"
+JCK_GIT_REPO_USED="$2"
+jck_branch="$3"
+isZOS="$4"
+echo "JCK_ROOT_USED is: $JCK_ROOT_USED"
+echo "JCK_GIT_REPO_USED is: $JCK_GIT_REPO_USED"
+echo "jck_branch is: $jck_branch"
+echo "isZOS is: $isZOS"
+
```
karianna, line 11:
```suggestion
```

jck/update_git_material.sh
```diff
@@ -0,0 +1,151 @@
...
+        if [ -n "$local_material_uptodate" ]; then
+            echo "zOS Local JCK materials up-to-date. Skipping hard reset"
+        else
+            # Perform hard reset for z/OS
+            perform_hard_reset_zos
+        fi
+    fi
+}
+
+# Function to perform hard reset for z/OS
```
karianna, line 123: Why is this needed?

jck/update_git_material.sh
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    # Clean untracked files and directories
+    git clean -dfx || {
+        echo "Failed to clean untracked files and directories."
+        remove_and_clone
+    }
+
+    # Pull the latest changes
+    git pull "$JCK_GIT_REPO_USED" "$jck_branch" || {
```
llxia, line 77: I know this logic is inherited, but I think there is no need to do git pull again after git reset. 
https://stackoverflow.com/questions/43037293/what-is-the-difference-between-git-pull-and-git-reset-hard-origin-branch
Could you verify if `git pull` is still needed here?

jck/update_git_material.sh
```diff
@@ -0,0 +1,150 @@
...
+
+# Function to remove test material and git clone
+remove_and_clone() {
+    echo "Removing test material and git clone again"
+    rm -rf "$JCK_ROOT_USED"
+    git_clone
+}
+
+# Function to perform hard reset and pull
+perform_hard_reset_and_pull() {
```
llxia, line 63: This should handle both zos and non-zos. We do not need perform_hard_reset_zos to handle zos separately.

llxia (changes requested):

llxia:
The current PR only adds the error check in the places that we think it may fail. This is not sufficient. The better way to do this is to use `try/catch`. In shell, we mimic `try/catch` as follows. 

```
{ # try

  # commands

} || { # catch

}
```
https://stackoverflow.com/questions/22009364/is-there-a-try-catch-command-in-bash

Also, this can help us to write clean code to avoid repeatedly calling the same functions. For example, `remove_and_clone` is called 6 times in the PR.

jck/update_git_material.sh
annaibm: Will need to remove this function as `perform_hard_reset_and_pull` needs to handle both ZOS and non-ZOS.

jck/update_git_material.sh
annaibm: @llxia So I need to move the logic of `perform_hard_reset_zos` into  `perform_hard_reset_and_pull` based on ZOS flag or is it general logic which is already there? As I had inherited the logic from build.xml here, so not sure.

jck/update_git_material.sh
annaibm: Sure will test and let you know

jck/update_git_material.sh
llxia: Based on the ZOS flag, ZOS needs to replace the attribute file.

jck/update_git_material.sh
```diff
@@ -0,0 +1,127 @@
...
+    fi
+else
+    # JCK materials don't exist, git clone
+    echo "JCK materials don't exist."
+    git_clone
+fi
+# Catch block
+if [ "$global_error" -eq 1 ]; then
+    handle_error
+fi
```
llxia, line 127: Did the following work?
```
{ 
     check_and_handle_updates
} || {
     git_clone
}
```

jck/update_git_material.sh
```diff
@@ -0,0 +1,127 @@
...
+	fi
+}
+
+# Function to perform hard reset and pull
+perform_hard_reset() {
+	echo "Performing hard reset..."
+	# Fetch changes from the remote repository
+	git fetch "$JCK_GIT_REPO_USED"
+	# Reset the local branch to match the remote branch
+     git reset --hard "FETCH_HEAD"
```
llxia, line 59: Please fix the indentation. Please check the rest of PR.

jck/update_git_material.sh
```diff
@@ -0,0 +1,127 @@
...
+		git clean -fdx
+		local_material_uptodate="true"
+	fi
+	# If SHA match or update succeeded
+   	echo "Local JCK materials updated"
+}
+
+# Try block
+if [ -d "$JCK_ROOT_USED" ] && { [ -d "$JCK_ROOT_USED/.git" ] || git -C "$JCK_ROOT_USED" rev-parse --is-inside-work-tree &>/dev/null; }; then
+	echo "The directory is a valid Git repository."
```
llxia, line 109: We need to clean all unstaged change.
```
git -C $JCK_ROOT_USED clean -fd
```

annaibm: Yes we are handling it under `check_and_handle_updates` method, does that suffice?

jck/update_git_material.sh
annaibm: I had tried the above method but it doesnt catch errors happening under functions within `perform_hard_reset` or other places as such. So need to add error handling to check the error code returned in addition to trap statement.

jck/update_git_material.sh
llxia: no, it is not. In some cases, someone can manually change the file on disk without committing. Therefore, the SHA is the same. check_and_handle_updates does not catch this case.

PR #4894, opened 2023-11-30
Fix portable aot tests

LongyuZhang:
- Some pipelines have $DOCKER_REGISTRY_CREDENTIALS in jenkins-pipeline but not $BASE_DOCKER_REGISTRY_CREDENTIAL, in this case login is not needed.
- Related Issue: github_ibm/runtimes/backlog/issues/1262

### DIFF
--- a/external/external.sh
+++ b/external/external.sh
@@ -286,7 +286,7 @@ if [ $command_type == "prepare" ]; then
 	# Specify docker.io or internal registry to prepare base image with login to increase pull limit or authenticate; Redhat Registry no login.
 	if [[ $base_docker_registry_url != "default" ]]; then
 		# Container credential check. 
-		# Temporarily criu-ubi image with criu binary is only available internally
+		# Temporarily host criu-ubi image with criu binary on internal/private hub.  In that case, provide USR/PSW for access 
 		if [[ "${test}" == *"criu-ubi"* && ! -z $DOCKER_REGISTRY_CREDENTIALS_USR ]]; then
 			BASE_DOCKER_REGISTRY_CREDENTIAL_USR=$DOCKER_REGISTRY_CREDENTIALS_USR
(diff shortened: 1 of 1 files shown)


external/external.sh
```diff
@@ -285,14 +285,15 @@ parseCommandLineArgs "$@"
 if [ $command_type == "prepare" ]; then
 	# Specify docker.io or internal registry to prepare base image with login to increase pull limit or authenticate; Redhat Registry no login.
 	if [[ $base_docker_registry_url != "default" ]]; then
-		# Container credential check. Temporarily criu-ubi image with criu binary is only available internally
-		if [[ ! -z $BASE_DOCKER_REGISTRY_CREDENTIAL_USR || ! -z $DOCKER_REGISTRY_CREDENTIALS_USR ]]; then
+		# Container credential check. 
+		# Temporarily criu-ubi image with criu binary is only available internally
```
karianna, line 289: This comment is a bit hard to read :-).  Not quite sure what logic it is trying to convey.

LongyuZhang: In this step, external tests get the initial base container image. Most tests uses `BASE_DOCKER_REGISTRY_CREDENTIAL` to get this base image from Dockerhub (this login will increase pull limit), but `criu-ubi` base image temporarily get its base image from hyc-dockerhub instead since we can only set up its dependent package internally with ubi subscription. Once we are able to publish this image on Dockerhub later, we will remove this logic then.

smlambert, line 289:
```suggestion
		# Temporarily host criu-ubi image with criu binary on internal/private hub.  In that case, provide USR/PSW for access 
```

LongyuZhang: The suggested change has been made. Thanks.

smlambert (approved):

Issue #4864, opened 2023-11-10
Remove JavaTestRunner


llxia:
[JavaTestRunner](https://github.com/adoptium/aqa-tests/blob/master/jck/jtrunner/JavaTestRunner.java) and [JavatestUtil](https://github.com/adoptium/aqa-tests/blob/master/jck/jtrunner/JavatestUtil.java) are similar file. 

And most of tests are using JavatestUtil except https://github.com/adoptium/aqa-tests/blob/4a1eca103c3a40e98f36328ca03f2eef02de05d9/jck/jck.mk#L129

https://github.com/search?q=repo%3Aadoptium%2Faqa-tests%20JCK_CMD_TEMPLATE&type=code

I think if we update these two tests to run via JavatestUtil, we can remove JavaTestRunner.

smlambert:
Yes, that was the next step in the 'evolution' of this automation (stripping away unneeded layers).

llxia:
@annaibm can you take a look at this one? Thanks

PR #4898, opened 2023-12-04
Update tests to use JavatestUtil

annaibm:
- This code updates the custom tests to use JavatestUtil instead of JavaTestRunner.

- Removes JavaTestRunner file

resolves: https://github.com/adoptium/aqa-tests/issues/4864


LongyuZhang (approved):
Related to reverted  PR https://github.com/adoptium/aqa-tests/pull/4893. 
It was reverted because the release test failed after that PR merged with error information related to the packages modified in the PR, so reverted to unblock the releases process, but it was a coincidence, the failure was actually not related to this PR.

llxia (approved):

PR #4911, opened 2023-12-08
Add sbom file as part of downloaded sdk files

sophia-guo:
Add sbom file as part of downloaded sdk files to support reproducible comparison test


karianna (approved):

sophia-guo:
Grinder https://ci.adoptium.net/view/Test_grinder/job/Grinder/8212/

smlambert (approved):

PR #4915, opened 2023-12-12
Exclude MulticastSendReceiveTests for All Platforms

annaibm:
Cherry picked the commit for excluding MulticastSendReceiveTests for all Platforms.


llxia (changes requested):
This is incorrect. The cherry pick should be delivered in https://github.com/adoptium/aqa-tests/tree/j9.0.42-release, not master

PR #4929, opened 2023-12-20
Use adoptium/ubuntu2004_build_image:linux-riscv64 image for testing on RISC-V

luhenry:


sxa (commented):
This needs to be discussed with Alibaba since they are using this functionaly and we need to make sure we don't negatively impact them.

luhenry:
@sxa is there a variant flag here similarly to the build-side? The current image is not working for testing temurin so it's not working for Rivos.

sxa:
From discussion with Ludovic this morning it seems the issue with the current image is that it's using a very old version of qemu. In the short term I'm probably going to look at not using this dynamic functionality on ci.adoptium.net (since it doesn't work and stalls the jobs) in the short term, but then looking to get usable images fired up to test with RVV support. 

From https://github.com/adoptium/aqa-tests/pull/4307#issuecomment-1436484037 is seems that the dockerfile for that image is available at https://github.com/dragonwell-project/docker-qemu-riscv64/blob/master/Dockerfile so we may be able to bump the qemu version in there and get a new image (The default parameter in there seems to be for 7.2.0 just now)

> is there a variant flag here similarly to the build-side? 

It may be possible to do it on the build variant, but I suspect that making a change to not use dynamic containers if the jenkins server is ci.adoptium.net is probably the simplest solution here in the short term as that should be easy to determine.

luhenry:
Also, as discussed on the call this morning, the plan is to avoid using the dynamic pool, and instead queue up on our existing RISC-V boards. We would then not need to run inside a Docker container. Closing this PR until it's deemed necessary or better than current approach. Please ping me when/if any wants to discuss. Thanks!

PR #4931, opened 2023-12-21
Do not use dockerAgents on ci.adoptium.net

sxa:
This is a change to prevent the changes from https://github.com/adoptium/aqa-tests/pull/4307 from taking effect on the adoptium CI. It can still be used elsewhere, but for now the functionality does not work on our jenkins instances and causes jobs to be queued up indefinitely waiting for an image which will not be created if all of the real machines are in use.

It may be preferable to have an override at the `PLATFORM_MAP` level earlier in the file where this is defined, and ultimately we want to have this functionality enabled (For RVV support) but for now this is the easiest way to stop the test failing to run.


sxa:
Verified in https://ci.adoptium.net/job/Grinder/8327 which queued successfully when given a specific node label  to run on

luhenry:
Do I understand correctly that this is related to "Disable dynamic pool for testing, and queue test jobs for execution on boards"?

sxa:
> Do I understand correctly that this is related to "Disable dynamic pool for testing, and queue test jobs for execution on boards"?

Correct. This should resolve that problem.

luhenry (approved):

sxa:
Noting that https://ci.adoptium.net/job/Grinder/8349/console seemed to work ok without this change on on [dockerhost-equinix-ubuntu2004-x64-1](https://ci.adoptium.net/computer/dockerhost%2Dequinix%2Dubuntu2004%2Dx64%2D1/)

buildenv/jenkins/openjdk_tests
```diff
@@ -267,7 +267,13 @@ timestamps{
             env.EXTRA_OPTIONS = params.EXTRA_OPTIONS ? params.EXTRA_OPTIONS : ""
             env.BUILD_LIST = params.BUILD_LIST ? params.BUILD_LIST : ""
             SPEC = PLATFORM_MAP[params.PLATFORM]["SPEC"]
-            dockerAgents = PLATFORM_MAP[params.PLATFORM]["DockerAgents"] ? PLATFORM_MAP[params.PLATFORM]["DockerAgents"] : []
+            // Block dockerAgents on adoptium CI for riscv64 until we make it work
+            // This will allow RISC-V test jobs to queue up and not hang
+            if ( env.JENKINS_URL.contains("ci.adoptium.net")) {
```
sophia-guo, line 272: This will stop other linux test jobs using dynamic agents. Is this expected?

sophia-guo: For RISC-V specifically, maybe limit to `if ( env.JENKINS_URL.contains("ci.adoptium.net") && SPEC.equals('linux_riscv64'))` ?

sxa: What other platforms do we do this for? I was under the impression it was only RISC-V

sxa:
For now I've disabled the machines in the set [ci.role.test&&sw.os.linux&&sw.tool.docker&&hw.arch.x86](https://ci.adoptium.net/label/ci.role.test&&sw.os.linux&&sw.tool.docker&&hw.arch.x86/) which are not able to perform this function (verified with the smoke test, so fairly basic) which should allow the test jobs to run through successfully

smlambert:
Please do not disable https://ci.adoptium.net/label/ci.role.test&&sw.os.linux&&sw.tool.docker&&hw.arch.x86/ machines, 
if you do, then external test jobs will fail.  We have recently re-enabled them and in 1Q will be working on fixing them to run green.  https://ci.adoptium.net/job/Test_openjdk17_hs_sanity.external_x86-64_linux/

The purpose of external tests:
https://github.com/adoptium/aqa-tests/tree/master/external#roadmap

sxa:
> Please do not disable https://ci.adoptium.net/label/ci.role.test&&sw.os.linux&&sw.tool.docker&&hw.arch.x86/ machines, if you do, then external test jobs will fail. 

Yep I'm aware of that I've ensured that I'm only changing a subset of them which are being problematic. There should still be enough to run the x64 tests - bear in mind that one of the remaining ones has five executors on it. If you still think that's a problem let me know.

luhenry:
@sxa @smlambert how can we can that forward? It's blocking testing for RISC-V on boards exclusively, which is impossible today for the weekly runs. Happy to have that disabled for RISC-V explicitly. Thanks!

smlambert:
As a temporary approach, incorporating Sophia's review comments would move this forward.

As the actual approach for this feature, this is not the path I would like us to take.

As per this comment, I would like the images to be an input parameter to the test jobs 
https://github.com/adoptium/aqa-tests/pull/4626#issuecomment-1594653371

karianna (approved):
Approving since Shelly is OK with it :-)

Issue #4930, opened 2023-12-21
trapme: Xvfb startup timeout not obvious in logs


sxa:
This is something I'm sure I've hit before (or certainly similar problems) but I don't think we've addressed it, so I'm raising an issue so it doesn't get lost. I've added `trapme` to the title so I can track any improvements to error conditions.

In https://ci.adoptium.net/job/Test_openjdk21_hs_sanity.openjdk_riscv64_linux/108/console we hit an issue where, for no clear reason, the tests didn't seem to start running. @sophia-guo spotted that it might be an issue with the Xvfb startup. Here is the log from a failing run:
```
8:54:41  ITERATION: 1/1
[Pipeline] wrap
18:54:41  $ Xvfb -displayfd 2 -screen 0 1024x768x24 -fbdir /home/jenkins/workspace/Test_openjdk21_hs_sanity.openjdk_riscv64_linux/.xvfb-108-..fbdir4921422941276978355
[Pipeline] // wrap
[Pipeline] }
[Pipeline] // stage
[Pipeline] stage
[Pipeline] { (Post)
[Pipeline] step
18:55:21  TAP Reports Processing: START
18:55:21  Looking for TAP results report in workspace using pattern: aqa-tests/TKG/**/*.tap
18:55:21  Did not find any matching files. Setting build result to FAILURE.
```
And this is from a good run of the same thing on another machine:
```
10:04:20  ITERATION: 1/1
[Pipeline] wrap
10:04:20  $ Xvfb -displayfd 2 -screen 0 1024x768x24 -fbdir /home/jenkins/workspace/Test_openjdk21_hs_sanity.openjdk_riscv64_linux/.xvfb-115-..fbdir4816611008607352758
[Pipeline] {
[Pipeline] sh
10:04:32  + ps -f
10:04:32  + awk {print $9}
10:04:32  + grep [X]vfb
[Pipeline] echo
10:04:32  env.DISPLAY is :0
```
Ideally we should have something showing in the log to indicate that a timeout occurred (if indeed it can be confirmed that it is the cause of the failure)

adamfarley:
[...]

PR #4932, opened 2023-12-21
adjust Xvfb timeout to 20s for HiFive Unleashed

sxa:
Increase timeout as some of the machines are taking longer thean 10 seconds to fully initiate the Xvfb process, causing a failure to kick off the tests but no obvious reason why.

Ref: https://github.com/adoptium/aqa-tests/issues/4930 (This does not fix the lack of error message, but should stop things failing. I've verified that it seems ok (sample size of one test run) on the unl1 node which this issue was seen on.

Ref general RISC-V plan at https://github.com/adoptium/temurin-build/issues/3591


sophia-guo (approved):
Rerun on same agents 20 secs works https://ci.adoptium.net/job/Test_openjdk21_hs_sanity.openjdk_riscv64_linux/117/parameters/

smlambert (approved):

Issue #4964, opened 2024-01-15
orka x64 Mac test jobs being terminated by terminateTestProcesses.sh


andrew-m-leonard:
It looks like the x64 Mac Orka ndoes run in a docker type container host environment, as it seems one test job can see the processes of another test job, and subsequently the terminateTestProcesses.sh cleanup terminates them!

eg.
https://ci.adoptium.net/job/Test_openjdk17_hs_extended.functional_x86-64_mac/371/console
Is killing processes from job "extended.perf" : https://ci.adoptium.net/job/Test_openjdk17_hs_extended.perf_x86-64_mac/145/console

smlambert:
Related infra issue: https://github.com/adoptium/infrastructure/issues/2536

andrew-m-leonard:
Following the Orka update to the VM provisioning for this issue, I have run a manual test of a job to query test processes and kill any it finds, while at the same time running a separate build&Test job.
Unfortunately the manual job was still able to kill the seperate Jenkins Test Job processes:
This hand crafted job running on Orka provisioned node **test-orka-macos14-x64-54xfp** : https://ci.adoptium.net/view/work-in-progress/job/andrew-exec/21/console
Was able to Terminate the processes running under this Test job running on Orka provisioned node **test-orka-macos14-x64-6ckv8** : https://ci.adoptium.net/job/Test_openjdk17_hs_extended.system_x86-64_mac/394/console

PR #4967, opened 2024-01-15
Do not cleanup processes on Mac due to no Orka VM isolation

andrew-m-leonard:
Orka x64 VMs do not seem to be process isolated, see: https://github.com/adoptium/aqa-tests/issues/4964
Which means currently the terminateProcess.sh pre&post test job will terminate any other test job processes that maybe visible on the Orka macstadium node that is being used.

### DIFF
--- a/terminateTestProcesses.sh
+++ b/terminateTestProcesses.sh
@@ -38,6 +38,8 @@ if [ "$OS" = "Windows_NT" ]; then
   else
       echo Woohoo - no rogue processes detected!
   fi
+elif [ `uname` = "Darwin" ]; then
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
Test Grinders:
- x64Mac https://ci.adoptium.net/view/Test_grinder/job/Grinder/8476/
- x64Linux https://ci.adoptium.net/view/Test_grinder/job/Grinder/8477/
- aarch64Mac https://ci.adoptium.net/view/Test_grinder/job/Grinder/8478/

smlambert (approved):
Approve with the note below. 

We intend to look for a resolution with our Orka configuration but put this workaround in place in the meantime.  While this may be specific to the Adoptium project and not other vendors who may be using this material, we know many vendors have Cleanup scripts that run separately to kill rogue processes and we did not previously have this script in place at all, so regressing functionality to that point.

sophia-guo (approved):

Issue #4789, opened 2023-09-26
[aix]: java/lang/String/concat/IntegerMinValue.java jdk21 failed 


sophia-guo:
java/lang/String/concat/IntegerMinValue.java - https://bugs.openjdk.org/browse/JDK-8306007 , which ask XLC17 compiler. Current xlc version in adoptium: Version: 13.01.0003.0007.

Should we update the xlc version @sxa or we will exclude the test @smlambert ?

See https://github.com/adoptium/aqa-tests/issues/4773#issuecomment-1726762777

smlambert:
This likely ends up being excluded until our AIX 7.2 machines pick up the xlc17 compiler.  

@sxa - are there any plans to update the xlc version on AIX machines?

sxa:
Short answer: No
Longer answer: We have been looking at installing the newer compiler on some machines (See https://github.com/adoptium/infrastructure/issues/3208) although it hasn't had the full playbooks run on it so is likely not yet functional. There have been some upstream discussions relating to updating to xlc17 for the builds but they were not concluded last time I looked, so we're not deploying it by default yet.

On that basis I'm in agreement with something you mentioned on our chat earlier that would suggest we disable that test for now.

smlambert:
Test is disabled and linked to this issue.  The ProblemList entry should really point to an issue in a different repository where a future fix to the issue would be landed, infrastructure, as this test will not be re-enabled until machine updates occur, which is currently not in plan.

All that to be said, this issue has to stay open until an infra issue is created and the ProblemList is updated to point to the open issue.  Alternatively, if this will be a permanent exclusion, then can exclude via an umbrella issue for all permanent exclusions.

Issue #4956, opened 2024-01-10
Update disabledTestsLinter to account for missing lxml in environment


smlambert:
https://github.com/pyinstaller/pyinstaller/issues/5431 describes similar issues to what is being seen when running as a github actiion on a default runner.

@Haroon-Khel - FYI, this is the first exercising of the disabled_tests linter, see that it has an error: 
```
Run find . -name "playlist.xml" | python3 ./scripts/disabled_tests/playlist_parser.py -v > /dev/null
Traceback (most recent call last):
  File "/home/runner/work/aqa-tests/aqa-tests/./scripts/disabled_tests/playlist_parser.py", line 9, in <module>
    import lxml.etree
ModuleNotFoundError: No module named 'lxml'
Error: Process completed with exit code 1.
```

<img width="785" alt="Screenshot 2024-01-10 at 1 57 08 PM" src="https://github.com/adoptium/aqa-tests/assets/2836948/5a7f5c02-16b6-46e4-89c7-b5e77c18e425">

PR #4968, opened 2024-01-15
aix: exclude java/lang/String/concat/IntegerMinValue.java

sxa:
This is a failure on JDK21+ sanity.openjdk runs and is apparently a problem if you do not use the xlc17 compiler
Ref: https://github.com/adoptium/aqa-tests/issues/4789 and https://bugs.openjdk.org/browse/JDK-8306007

Noting that a test by the same name is ok on JDK17 e.g. : https://ci.adoptium.net/job/Test_openjdk17_hs_sanity.openjdk_ppc64_aix/349/testReport/java_lang_String_concat_IntegerMinValue/

Also the failure can apparently be bypassed by running with ` -XX:-OptimizeStringConcat` although I haven't verified this myself and I'm not sure if it's practical to add that into the run instead of excluding.


smlambert (approved):

smlambert:
Noting https://github.com/adoptium/aqa-tests/issues/4956 was raised already to address the linter failure.

karianna (approved):

smlambert:
Needs cherry-picking to v1.0.0-release branch.

Issue #4948, opened 2024-01-09
Update v1.0.0-release branch with JDKx_BRANCH info


smlambert:
testenv/testenv.properties files in the v1.0.0-release branch still need to be updated to include correct info for JDKx_BRANCH entries (derived/guessed from https://www.java.com/releases/)

Related: https://github.com/adoptium/aqa-tests/issues/4373 where we intend to add a mechanism that automates this, similar to what is done to trigger the build pipelines, adding JDKx_BRANCH_PREFIX entries to the testenv.properties file, and looking for the GA tag with that prefix, if not found, takes the most recent tag with that prefix.  Logic can be updated in https://github.com/adoptium/TKG/blob/master/scripts/getTestenvProperties.sh#L65 or a helper function in that file.

smlambert:
Closed via https://github.com/adoptium/aqa-tests/pull/4970

PR #4970, opened 2024-01-15
[v1.0.0-release] Update tags with predictions

Haroon-Khel:
No need to merge this evening. Can merge tomorrow morning


karianna (approved):

smlambert:
We may want to 'guess' at what 21.0.2+? is going to be, in case you are right in your guess.  

Fixes #4948

Haroon-Khel:
@ShelleyLambert I'm going with jdk-21.0.2+1 as the last tag was +0

Haroon-Khel:
JDK21 tag should be +13 when GA tag comes

PR #4974, opened 2024-01-16
Update release tag

sophia-guo:
Update release tag


smlambert:
This is a dup of https://github.com/adoptium/aqa-tests/pull/4970

PR #4986, opened 2024-01-19
Use the sane defaults for the $TKG_BRANCH

RadekCap:
The current release branch isn't working with the default values and requires manual setup of the TKG_BRANCH variable on Windows. This change proposes a fix to eliminate it.


RadekCap:
I've updated my Eclipse Account status from RadoslavCap to the current more easily pronounceable RadekCap I'm using. Can someone please run a re-run of the ECA check as I already signed it. The mismatch in names is the probable reason why it is red. Thank you.

smlambert (changes requested):
This is not a correct fix for the issue you are experiencing @RadekCap.

RadekCap:
@smlambert Can you please elaborate more?
The issue is, that the defaults is failing on Windows when the default configuration is used.
To be able to successfully run there you need to use an external configuration.
What do you suggest to handle this?

smlambert (approved):
Revising my original statement, while I do not think this is ideal, it is a PR specific to the v1.0.0-release branch so not harming/affecting others, and will help move you forward.

After release activities end, I would like to understand how you are triggering testing, so we can not have to do this type of workaround.

smlambert:
Additional background in this [Slack thread](https://adoptium.slack.com/archives/C5219G28G/p1704231165068949)

judovana:
> Revising my original statement, while I do not think this is ideal, it is a PR specific to the v1.0.0-release branch so not harming/affecting others, and will help move you forward.

Its not strictky for us. We can workaround it by sessting TKG_BRANCH to desired value, but why would release branch clone depndencies it can not work with?
> 
> After release activities end, I would like to understand how you are triggering testing, so we can not have to do this type of workaround.

It is not a workaround, the aqa-tests should be pulling a tkg (well generally anything)  in version, it can work with. If you have to set up TKG_BRANCH before running get.sh, then it is agains any principle of reasonable defaults, because default run do not work.

Thanx a lot for merging!

smlambert:
You do not have to set up TKG_BRANCH before running get.sh

You have to set USE_TESTENV_PROPERTIES=true before running get.sh, then the expected behaviour is that it picks up the TKG_BRANCH set in the testenv.properties file the aqa-tests release branch you are running from.

judovana:
I see. In that case, the USE_TESTENV_PROPERTIES should be on by default or at least mentioned in README.md.

As it is not, I guess many values in testenv/testenv.properties  are duplicated as defaults (as they are used even if USE_TESTENV_PROPERTIES is not used)

smlambert:
USE_TESTENV_PROPERTIES is only needed for AQAvit verification, for example when you want to verify to push to the marketplace, as detailed on this page, https://adoptium.net/docs/aqavit-verification/ (which has been shared and walked-through with all marketplace vendors).  In day-to-day running, USE_TESTENV_PROPERTIES=false by intention to set the default to pull from master/main branches.

judovana:
Hmm.  I have to miss something.  I do not want to run certification.  And even if so, I want to keep it as standalone as possible due to testing rpms/srpms/providing custom native libraries and so on.  I would like  to run some single random testsuite.

The issue(?) is "fixable" by setting TKG_BRANCH. Your advice was to  "you have to set USE_TESTENV_PROPERTIES=true" . Now you are saying  "USE_TESTENV_PROPERTIES is only needed for AQAvit verification" . Those three contradict each other (or I really miss something). Generally speaking non of that is mandatory, but without that it do not work.  The sane defaults are a foundation from my naive point of view.

Maybe then the USE_TESTENV_PROPERTIES can be mandatory, instead of duplicated defaults in other scripts?

PR #4993, opened 2024-01-22
Cherry pick exclude jck aarch64mac

annaibm:


llxia:
@annaibm Please do a cherry-pick for the related PR only.

Issue #5022, opened 2024-01-30
Mac x64 test failure: java/lang/management/MemoryMXBean/LowMemoryTest2.sh


d3r3kk:
Related to: #157 
Described in detail by: [JDK-8209743](https://bugs.openjdk.org/browse/JDK-8209743) (test framework problem)
   _Note: The fix for this issue was not backported to JDK11_

Here's the results of the run without altering the test on our end:
[jdk_management_0.tar.gz](https://github.com/adoptium/aqa-tests/files/14102466/jdk_management_0.tar.gz)

Here's the results of the run after applying the 'band-aid' of increasing memory to 24m:
[jdk_management_0.success.24m.tar.gz](https://github.com/adoptium/aqa-tests/files/14102465/jdk_management_0.success.24m.tar.gz)

Since the conclusion of the JDK TESTBUG is:

> In conclusion, this is a test bug. Increasing the metaspace space size can change the proportion (class and non-class allocation types) of the meta space usage and allow the test to pass, although it is not a real fix to the test issue.

...the fix on the Microsoft end will likely be to problemlist it.

PR #5023, opened 2024-01-30
Added problematic LowMemoryTest2 to problemlist for Microsoft

d3r3kk:
Related to #5022


d3r3kk:
We will propose problem listed tests for Microsoft in a single upcoming PR to make it easier on our Adoptium friends.

PR #5034, opened 2024-02-02
Unset CUSTOMIZED_SDK_URL if SDK_RESOURCE is not customized

### LINKED ISSUE #2097: Setting CUSTOMIZED_SDK_URL and SDK_RESOURCE  as non-customized ignores the SDK_RESOURCE value
In the grinder setting it says _Customized SDK URL, only set when SDK_RESOURCE customized._ However if the user accidentally set both CUSTOMIZED_SDK_URL and SDK_RESOURCE ( non-customized) the system will take CUSTOMIZED_SDK_URL and ignore the SDK_RESOURCE's setting.

Senario:
1. First use CUSTOMIZED_SDK_URL
2. Rerun use SDK_RESOURCE = nightly to compare the difference ( CUSTOMIZED_SDK_URL is not reset as empty)
3. Build will rerun with CUSTOMIZED_SDK_URL, which ignores SDK_RESOURCE as nightly ( also in get.sh, SDK_RESOURCE default is nightly).

Solution:
Enhance the check condition(https://github.com/adoptium/aqa-tests/blob/master/get.sh#L180) with`if "$SDK_RESOURCE" == "customized" `  in two places. 

1. https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L371-L373
2. https://github.com/adoptium/aqa-tests/blob/master/get.sh#L180

example：
https://ci.adoptopenjdk.net/view/work-in-progress/job/grinder_sandbox/162/console

sophia-guo:
CUSTOMIZED_SDK_URL is secondary parameter determined by primary parameter SDK_RESOURE. If SDK_RESOURCE is set as non-customized, CUSTOMIZED_SDK_URL should be set as none( except nightly)

Fix #2097

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -397,32 +397,37 @@ def createJob( TEST_JOB_NAME, ARCH_OS ) {
 def setup() {
 	stage('Setup') {
 		setupEnv()
-
-		if (params.SDK_RESOURCE == 'nightly' && params.CUSTOMIZED_SDK_URL) {
-			// remove single quote to allow variables to be set in CUSTOMIZED_SDK_URL
-			CUSTOMIZED_SDK_URL_OPTION = "-c ${params.CUSTOMIZED_SDK_URL}"
-		} else if (params.SDK_RESOURCE != 'customized') {
-			CUSTOMIZED_SDK_URL_OPTION = ""
-		} else if (params.CUSTOMIZED_SDK_URL) {
-			SDK_RESOURCE = "customized"
-			CUSTOMIZED_SDK_URL_OPTION = "-c '${params.CUSTOMIZED_SDK_URL}'"
-			if (params.ADDITIONAL_ARTIFACTS_REQUIRED == "RI_JDK") {
-				def server = Artifactory.server params.ARTIFACTORY_SERVER
-				def artifactoryUrl = server.getUrl()
-				def repoForRi = ''
-				def riURL = ''
-				if (params.ARTIFACTORY_REPO.contains(',')) {
-					String[] repos = params.ARTIFACTORY_REPO.split(",")
-					// Assumption: ARTIFACTORY_REPO=X,Y (X=to upload results, Y=to download ri)
-					repoForRi = repos[1].trim()
-					if (!env.SPEC.startsWith('aix') && !env.SPEC.startsWith('zos')) {
-						riURL = "${artifactoryUrl}/${repoForRi}/Latest/${PLATFORM}/${JDK_VERSION}"
+		if (params.CUSTOMIZED_SDK_URL) {
+			if (params.SDK_RESOURCE == 'nightly') {
+				// remove single quote to allow variables to be set in CUSTOMIZED_SDK_URL
+				CUSTOMIZED_SDK_URL_OPTION = "-c ${params.CUSTOMIZED_SDK_URL}"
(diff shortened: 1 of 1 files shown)


karianna (approved):

smlambert:
Asking for a review from @llxia to check it works for the ways that this field may be overwritten to pickup nightly openj9 builds and when used as part of AQA_Test_Pipeline.

sophia-guo:
I believe that nightly openj9 builds is in the category `if (params.SDK_RESOURCE == 'nightly' && params.CUSTOMIZED_SDK_URL) `, which won't be affected.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -401,6 +401,8 @@ def setup() {
 		if (params.SDK_RESOURCE == 'nightly' && params.CUSTOMIZED_SDK_URL) {
 			// remove single quote to allow variables to be set in CUSTOMIZED_SDK_URL
 			CUSTOMIZED_SDK_URL_OPTION = "-c ${params.CUSTOMIZED_SDK_URL}"
+		} else if (params.SDK_RESOURCE != 'customized') {
+			CUSTOMIZED_SDK_URL_OPTION = ""
```
llxia, line 405: Here, we assume if user provides SDK_RESOURCE != 'customized' and CUSTOMIZED_SDK_URL, we will ignore CUSTOMIZED_SDK_URL value and proceed with SDK_RESOURCE value only.
IMO, we should just fail in this case.

I would like to rewrite the whole logic to be more readable:

```
if (params.CUSTOMIZED_SDK_URL) {
	if (params.SDK_RESOURCE == 'nightly') {
		// remove single quote to allow variables to be set in CUSTOMIZED_SDK_URL
        	CUSTOMIZED_SDK_URL_OPTION = "-c ${params.CUSTOMIZED_SDK_URL}"
	} else if (params.SDK_RESOURCE == 'customized' || !params.SDK_RESOURCE) {
		SDK_RESOURCE = "customized"
		CUSTOMIZED_SDK_URL_OPTION = "-c '${params.CUSTOMIZED_SDK_URL}'"
                … // RI_JDK code
        } else {
		error("SDK_RESOURCE: ${params.SDK_RESOURCE} and CUSTOMIZED_SDK_URL: ${params.CUSTOMIZED_SDK_URL} combo is not supported!)
        }
} else {
        if (params.SDK_RESOURCE == 'customized') {
		error("SDK_RESOURCE: ${params.SDK_RESOURCE}, please provide CUSTOMIZED_SDK_URL")
        }
}
```

sophia-guo: This is a more strict user story. Updated.

llxia:
Thanks @sophia-guo . Could you run some Grinder to verify?

sophia-guo:
https://ci.adoptium.net/view/Test_grinder/job/Grinder/8722/
https://ci.adoptium.net/view/Test_grinder/job/Grinder/8723/
https://ci.adoptium.net/view/Test_grinder/job/Grinder/8725/

llxia:
Thanks @sophia-guo .
I have also tested the case of the SDK_RESOURCE=nightly and CUSTOMIZED_SDK_URL for openj9 ([internal link](https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/37842/))

llxia (approved):

PR #5035, opened 2024-02-02
Use nproc instead of /proc/cpuinfo to find number of processors for concurrency

Haroon-Khel:
See https://github.com/adoptium/infrastructure/issues/3360#issuecomment-1924438777

For docker containers, `grep -c ^processor /proc/cpuinfo` will return the number of processors of the host, not the container. Whereas `nproc` will return the number of processors of the container (given the container is run using the `--cpuset-cpus="0-n"` option.

`nproc` works on linux hosts so should not affect non docker nodes and it comes pre installed

### DIFF
--- a/openjdk/openjdk.mk
+++ b/openjdk/openjdk.mk
@@ -64,7 +64,7 @@ ifeq ($(JTREG_CONC), 0)
 		JTREG_CONC := 1
 	endif
 endif
-EXTRA_JTREG_OPTIONS += -concurrency:$(JTREG_CONC)
+EXTRA_JTREG_OPTIONS += -concurrency:2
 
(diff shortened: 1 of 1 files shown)


Haroon-Khel:
More importantly, jdk_time_2 was failling on some of our arm32 docker static containers because the test scripts read the cpu number of the host and so set `-concurrency:81`, https://ci.adoptium.net/job/Grinder/8700/console, causing the tests to overwhelm the container and timeout. `nproc` find the correct number of cpus for the container, allowing the test to pass on our arm32 docker nodes https://ci.adoptium.net/job/Grinder/8710/console

sophia-guo (approved):

karianna (approved):

PR #5101, opened 2024-02-26
Added churn testsuite

judovana:
churn testsuite is powerful suite to test individual GCs in OpenJDK

As it is added heere, it tests all GCs in given version of jdk.
Initial pass is https://ci.adoptium.net/view/Test_grinder/job/Grinder/8946/
As it is now - for review process, it runs for some 20 seconds - that means aprox 4 seconds per GC, which i moreover useless. In GH actions (https://github.com/rh-openjdk/churn/actions) I run it  on all platforms for aprox 3 minutes per GC, whcih is still moreover useless.
Exept running as it is, it can also run in disabled compressed oops mode. Not sure if it is required here. To run both, should be easy.  To make it configurable.. not sure...

Few questions:
 * the site generates archive full of logs, which are sueprusefull in case of failure. How to propagate them? Or maybe it is done correctly? If so how to see them in grinder?-)
 * what is acceptable time for you to  run churn? I left in comment suggesting 5hours for all GCS (so aprox 1hour per gc - close to be usefull)
 *  there is archive with results for jtreg plugin. It contains single juni xml file. Can it be somehow publised?
 * Are you ok with running all gcs in one task?
   * the individual GCs are reported separately in both logs archive and in junit xml file
   * it woudl be nice to have them reported as individual taps or at least individual targets or somethng liek that.. but It is not strightforward to select correct set of GCs per JDK, so I would rather remain in current ALL
   *  Despite I would liek to see above, I'm not sure it is possible. The logs from several hours long run are huge

### DIFF
--- a/functional/churn/README.md
+++ b/functional/churn/README.md
@@ -1,14 +1,36 @@
 # churn
-Test targeting garbage collector(s)
+Test targeting garbage collector(s).
+The suite is able to operate with all GCs in OpenJDK. If your implementation is missing some, the calls to them will fall. If you have some additional one, an enabling wrapper must be created first.
 
 ## Running churn locally
 clone https://github.com/rh-openjdk/churn read  https://github.com/rh-openjdk/churn/blob/master/README and run run.sh
 
 ### Setting up the environment:
-Java required. Maven recommended, churn can be compiled via direct javac if needed.
-Java is used from JAVA_HOME or guessed from PATH
-run.sh can use OTOOL_garbageCollector and OTOOL_JDK_VERSION instead of gc argument. See https://github.com/rh-openjdk/churn/blob/master/README for details
+Java required. Maven recommended, churn can be compiled via direct javac if needed
+Java is used from JAVA_HOME or guessed from PATH.
+run.sh can use OTOOL_garbageCollector and OTOOL_JDK_VERSION instead of GC argument. See https://github.com/rh-openjdk/churn/blob/master/README for details.
+See https://github.com/rh-openjdk/churn/tree/master/bin for list of supported GCs. The compressed ops in upstream run can be controlled by runner or NOCOMP=-nocoops variable (for run.sh)
+
+## Running via aqavit
+
+The support for compressed ops is handled by aqavit itself. It is currently not sure if churn will be able to honor it. If not, churn will be fixed.
+Similarly the java version and JAVA_HOME are handled by aqavit.
 
 ### Executing the testsuite
-Now it is as simple as running bash script run.sh.
+The `BUILD_LIST of functional/churn` contains three targets:
+ * _churn_1m_allGCs
+ * _churn_5h_allGCs
+ * _churn_custom
+
+There are major differences in them:
+ * churn_1m_allGCs - is testing ground, which runs each GC only for aprox 10 seconds, to simply see if the setup works
+ * churn_5h_allGCs - Is running each GC a bit over, which runs each GC for aprox hour and half. A minimum, which can find some real GC issue.
+ * churn_custom - this one needs you to set and export CHURN_GCS and CHURN_DURATION to select GC(or GCs) and duration. So it allows you to test your custom GC - if churn supports that, despite of anybody else (eg `default` or `ALL` thinks)
+
+churn_1m_allGCs and churn_5h_allGCs are using pony `ALL` keyword, which is interpreted (based on hardcoded list) as all GC in tested JVM. The set time is divided among them.
+Note, that if you use `churn_custom` and enumeration, eg `CHURN_GCS="zgc g1"` then the time will not be divided. The `CHURN_DURATION` is in seconds
+The `CHURN_GCS="defaultgc"`will set the tested GC to defautl GC as run.sh think is right. So be aware. Although it is maintained, if your custom JDK have custom GC, it is unlikely to be known
+
+### reading results:
+tap file and compressed xmlfile is genrated. Use eg https://github.com/jenkinsci/report-jtreg-plugin or https://plugins.jenkins.io/tap/ to read them.
 
--- a/functional/churn/README.md
+++ b/functional/churn/README.md
@@ -1,0 +1,14 @@
+# churn
+Test targeting garbage collector(s)
+
+## Running churn locally
+clone https://github.com/rh-openjdk/churn read  https://github.com/rh-openjdk/churn/blob/master/README and run run.sh
+
+### Setting up the environment:
+Java required. Maven recommended, churn can be compiled via direct javac if needed.
+Java is used from JAVA_HOME or guessed from PATH
+run.sh can use OTOOL_garbageCollector and OTOOL_JDK_VERSION instead of gc argument. See https://github.com/rh-openjdk/churn/blob/master/README for details
+
+### Executing the testsuite
+Now it is as simple as running bash script run.sh.
+
--- a/functional/churn/README.md
+++ b/functional/churn/README.md
@@ -11,10 +11,10 @@ Java is used from JAVA_HOME or guessed from PATH.
 run.sh can use OTOOL_garbageCollector and OTOOL_JDK_VERSION instead of GC argument. See https://github.com/rh-openjdk/churn/blob/master/README for details.
 See https://github.com/rh-openjdk/churn/tree/master/bin for list of supported GCs. The compressed ops in upstream run can be controlled by runner or NOCOMP=-nocoops variable (for run.sh)
 
-## Running via aqavit
+## Running via AQAvit
 
-The support for compressed ops is handled by aqavit itself. It is currently not sure if churn will be able to honor it. If not, churn will be fixed.
-Similarly the java version and JAVA_HOME are handled by aqavit.
+The support for compressed ops is handled by AQAvit itself. It is currently not sure if churn will be able to honor it. If not, churn will be fixed.
+Similarly the java version and JAVA_HOME are handled by AQAvit.
 
 ### Executing the testsuite
 The `BUILD_LIST of functional/churn` contains three targets:
@@ -25,12 +25,12 @@ The `BUILD_LIST of functional/churn` contains three targets:
 There are major differences in them:
  * churn_1m_allGCs - is testing ground, which runs each GC only for aprox 10 seconds, to simply see if the setup works
  * churn_5h_allGCs - Is running each GC a bit over, which runs each GC for aprox hour and half. A minimum, which can find some real GC issue.
- * churn_custom - this one needs you to set and export CHURN_GCS and CHURN_DURATION to select GC(or GCs) and duration. So it allows you to test your custom GC - if churn supports that, despite of anybody else (eg `default` or `ALL` thinks)
+ * churn_custom - this one is to support development, when run on the commandline you need to export at least DURATION and/or OTOOL_garbageCollector (+ many more optional, see [upstream readme](https://github.com/rh-openjdk/churn/blob/master/README)  to select DURATION in seconds and GC(or GCs). So it allows you to test your custom GC - if churn supports that, despite what other churn options suggest (eg `default` or `ALL` thinks).  When running in a Jenkins Grinder job, those are wrapped in TODO_CHURN_GCS and TODO_CHURN_DURATION
 
 churn_1m_allGCs and churn_5h_allGCs are using pony `ALL` keyword, which is interpreted (based on hardcoded list) as all GC in tested JVM. The set time is divided among them.
 Note, that if you use `churn_custom` and enumeration, eg `CHURN_GCS="zgc g1"` then the time will not be divided. The `CHURN_DURATION` is in seconds
-The `CHURN_GCS="defaultgc"`will set the tested GC to defautl GC as run.sh think is right. So be aware. Although it is maintained, if your custom JDK have custom GC, it is unlikely to be known
+The `CHURN_GCS="defaultgc"`will set the tested GC to default GC as run.sh think is right. So be aware. Although it is maintained, if your custom JDK have custom GC, it is unlikely to be known
 
-### reading results:
-tap file and compressed xmlfile is genrated. Use eg https://github.com/jenkinsci/report-jtreg-plugin or https://plugins.jenkins.io/tap/ to read them.
+### Reading results
+tap file and compressed junit xmlfile are generated. Use eg https://github.com/jenkinsci/report-jtreg-plugin or https://plugins.jenkins.io/tap/ to read them.
 
--- a/functional/churn/build.xml
+++ b/functional/churn/build.xml
@@ -17,7 +17,7 @@
 <project name="churn" default="build" basedir=".">
 	<taskdef resource="net/sf/antcontrib/antlib.xml" />
 	<description>
-		Test of correct behavior of swing components in both headful and headless jre environment.
+		Test targeting garbage collector(s)
 	</description>
 	
 	<!-- set properties for this build -->
--- a/functional/churn/build.xml
+++ b/functional/churn/build.xml
@@ -1,0 +1,64 @@
+<?xml version="1.0"?>
+
+<!--
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+
+<project name="churn" default="build" basedir=".">
+	<taskdef resource="net/sf/antcontrib/antlib.xml" />
+	<description>
+		Test of correct behavior of swing components in both headful and headless jre environment.
+	</description>
+	
+	<!-- set properties for this build -->
+	<property name="TEST" value="churn" />
+	<property name="DEST" value="${BUILD_ROOT}/functional/churn" />
+	<property name="src" location="./churn" />
+
+	<target name="THC.check">
+		<condition property="THCdir.exists">
+			<available file="churn" type="dir"/>
+		</condition>
+	</target>
+
+	<if>
+		<contains string="${SPEC}" substring="zos"/>
+		<then>
+			<property name="GIT_REPO" value="git@github.com:" />
+		</then>
+		<else>
+			<property name="GIT_REPO" value="https://github.com/" />
+		</else>
+	</if>
+
+	<target name="getChurn" depends="THC.check" unless="THCdir.exists">
+		<exec executable="git" failonerror="true">
+			<arg value="clone" />
+			<arg value="${GIT_REPO}rh-openjdk/churn.git" />
+		</exec>
+	</target>
+
+	<target name="init">
+		<mkdir dir="${DEST}"/>
+	</target>
+
+	<target name="dist" depends="getChurn" description="generate the distribution">
+		<copy todir="${DEST}">
+			<fileset dir="${src}" includes="*.xml, *.mk"/>
+		</copy>
+	</target>
+
+	<target name="build">
+		<antcall target="dist" inheritall="true" />
+	</target>
+</project>
--- a/functional/churn/playlist.xml
+++ b/functional/churn/playlist.xml
@@ -61,4 +61,28 @@
 			<group>system</group>
 		</groups>
 	</test>
+	<test>
+		<testCaseName>churn_custom</testCaseName>
+		<command> 
+            export OTOOL_JDK_VERSION="$(JDK_VERSION)" ; \
+            export JREJDK="jdk" ; \
+            export OTOOL_garbageCollector="${CHURN_GCS}" ; \
+            export DURATION="${CHURN_DURATION}" ; \
+            export TMPRESULTS="$(Q)$(REPORTDIR)$(D)report$(Q)"; \
+            cd $TMPRESULTS ; \
+              bash "$(TEST_ROOT)$(D)functional$(D)churn$(D)churn$(D)run.sh"; \
+            $(TEST_STATUS)
+		</command>
+		<levels>
+			<level>dev</level>
+		</levels>
+		<platformRequirementsList>
+			<platformRequirements>os.linux</platformRequirements>
+			<platformRequirements>os.win</platformRequirements>
+			<platformRequirements>os.osx</platformRequirements>
+		</platformRequirementsList>
+		<groups>
+			<group>system</group>
+		</groups>
+	</test>
 </playlist>
--- a/functional/churn/playlist.xml
+++ b/functional/churn/playlist.xml
@@ -71,6 +71,11 @@
 	</test>
 	<test>
 		<testCaseName>churn_custom</testCaseName>
+		<disables>
+			<disable>
+				<comment>Can run disabled.churn_custom in Grinder, meant for dev use not regular automated tests.</comment>
+			</disable>
+		</disables>
 		<command> 
             export OTOOL_JDK_VERSION="$(JDK_VERSION)" ; \
             export JREJDK="jdk" ; \
--- a/functional/churn/playlist.xml
+++ b/functional/churn/playlist.xml
@@ -20,6 +20,7 @@
             export JREJDK="jdk" ; \
             export OTOOL_garbageCollector="ALL" ; \
             export DURATION="60" ; \
+            export CHURN_TAP=false ; \
             export TMPRESULTS="$(Q)$(REPORTDIR)$(D)report$(Q)"; \
             cd $TMPRESULTS ; \
               bash "$(TEST_ROOT)$(D)functional$(D)churn$(D)churn$(D)run.sh"; \
@@ -48,6 +49,7 @@
             export JREJDK="jdk" ; \
             export OTOOL_garbageCollector="ALL" ; \
             export DURATION="18000" ; \
+            export CHURN_TAP=false ; \
             export TMPRESULTS="$(Q)$(REPORTDIR)$(D)report$(Q)"; \
             cd $TMPRESULTS ; \
               bash "$(TEST_ROOT)$(D)functional$(D)churn$(D)churn$(D)run.sh"; \
--- a/functional/churn/playlist.xml
+++ b/functional/churn/playlist.xml
@@ -16,13 +16,13 @@
 	<test>
 		<testCaseName>churn_1m_all</testCaseName>
 		<command> 
-            export OTOOL_JDK_VERSION=$(JDK_VERSION) ; \
+            export OTOOL_JDK_VERSION="$(JDK_VERSION)" ; \
             export JREJDK="jdk" ; \
-            export OTOOL_garbageCollector=ALL ; \
-            export DURATION=60 ; \
-            export TMPRESULTS=$(Q)$(REPORTDIR)$(D)report$(Q); \
+            export OTOOL_garbageCollector="ALL" ; \
+            export DURATION="60" ; \
+            export TMPRESULTS="$(Q)$(REPORTDIR)$(D)report$(Q)"; \
             cd $TMPRESULTS ; \
-              bash $(TEST_ROOT)$(D)functional$(D)churn$(D)churn$(D)run.sh; \
+              bash "$(TEST_ROOT)$(D)functional$(D)churn$(D)churn$(D)run.sh"; \
             $(TEST_STATUS)
 		</command>
 		<levels>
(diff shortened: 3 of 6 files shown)


judovana:
Grinder with disabled compressed oops https://ci.adoptium.net/view/Test_grinder/job/Grinder/8948/

functional/churn/build.xml
```diff
@@ -0,0 +1,64 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+
+<project name="churn" default="build" basedir=".">
+	<taskdef resource="net/sf/antcontrib/antlib.xml" />
+	<description>
+		Test of correct behavior of swing components in both headful and headless jre environment.
```
smlambert: Needs to be updated (copy/pasted from headless tests)

functional/churn/playlist.xml
```diff
@@ -0,0 +1,67 @@
...
+			<platformRequirements>os.linux</platformRequirements>
+			<platformRequirements>os.win</platformRequirements>
+			<platformRequirements>os.osx</platformRequirements>
+		</platformRequirementsList>
+		<groups>
+			<group>functional</group>
+		</groups>
+	</test>
+	<test>
+		<testCaseName>churn-nocoops</testCaseName>
```
smlambert, line 42: We already have a mechanism to automatically run all test targets twice, once with compressedrefs JVM option, and a second time as noncompressedrefs JVM option.  

I have to look closer at what NOCOMP does in the churn code, is it used for other decision-making within the test code to best advise.

smlambert (changes requested):
Going to put a request changes on this PR, so it doesn't get merged before all your questions get answered.  I have added some review comments in-line, and will answer some other questions shortly.

My immediate question is what happens when this testcase is run against distributions that do not ship with certain GCs.  This likely also relates to your question about running each GC as a separate testcase.

judovana:
> 
>     * the site generates archive full of logs, which are sueprusefull in case of failure. How to propagate them? Or maybe it is done correctly? If so how to see them in grinder?-)
> 

It seems I'm handling resutls correctly. The only disadvantage of run in grinder is, that logs got packed second time. I will fix this in the usptream runner.

judovana:
Hello! Bit more bussy day then expected:(

* I had remvoed the disbaled coompressed ops variant, and left it to the framework to deal with it. If you would guide me a bit more, I will adjsut upstream to honor your setup I believe
* I had renamed the targets and split them as you suggested
* the results seems correct now
*   answer to iminent question is that it will fail ig the GC is msising, or not test, if there is additional GC. Any suggestion on the vendors which gamble so much with GCs? Simialrly with default GC.
    * Currently the decision is based simply on jdk major version, I can add more, or add some custom wrapper

smlambert:
re: https://github.com/adoptium/aqa-tests/pull/5101#issuecomment-1966681217 - great, chat with you tomorrow about final steps

judovana:
pls do not merge (yet). The https://github.com/adoptium/aqa-tests/commit/050fb07dd56547b55b6728a7bbbea0cd04643705 is missbehavg in my test setup. But maybe my testsetup is wrong

smlambert:
re: https://github.com/adoptium/aqa-tests/pull/5101#issuecomment-1969259343 - I will convert this to Draft state, until all reviews and testing is completed (that way it can not get merged prematurely)

judovana:
The https://github.com/adoptium/aqa-tests/commit/050fb07dd56547b55b6728a7bbbea0cd04643705 , overwritten as https://github.com/adoptium/aqa-tests/commit/2249bc0ddbadfd0c2040597a78fa285ef048717d now works.

Am now updating readme to reflect the state. Will lift draft (thanx for setting it up) once done

Two new questions
 - the suite now generates next to jtreg archive also nice tap file. Is any special step needed to make it visible?
   - https://ci.adoptium.net/view/Test_grinder/job/Grinder/8994/ do not publish them
 - if any GC fails, or (in case of churn_custom) user put in ninsense, it will make the job red.
   -  Is it ok, or should some other logic be added, to make it unstable?
 - Will this suite ( I guess the churn_5h_allGCs target) be part of any regular aqa runs in public jenkins?
   - if so, to which, if not by default, but is desired can I add it somewhere?

judovana:
> The [050fb07](https://github.com/adoptium/aqa-tests/commit/050fb07dd56547b55b6728a7bbbea0cd04643705) , overwritten as [2249bc0](https://github.com/adoptium/aqa-tests/commit/2249bc0ddbadfd0c2040597a78fa285ef048717d) now works.
> 
> Am now updating readme to reflect the state. Will lift draft (thanx for setting it up) once done
> 
> Two new questions
> 
>     * the suite now generates next to jtreg archive also nice tap file. Is any special step needed to make it visible?
>       
>       * https://ci.adoptium.net/view/Test_grinder/job/Grinder/8994/ do not publish them


hmm.. https://ci.adoptium.net/view/Test_grinder/job/Grinder/8994/tapResults/they are here! thoughts? I'm not sure what I did...

smlambert:
re: https://github.com/adoptium/aqa-tests/pull/5101#issuecomment-1969567366 - you did not need to do anything, our test pipelines always generate TAP files that include all test targets ( all `<test><testCaseName> ... </test>` entries defined in playlists).  It is a feature of our TKG layer to create the TAP file, and a feature of our Jenkins groovy scripts to then attach the TAP files to the Jenkins job as an artifact.

functional/churn/playlist.xml
```diff
@@ -0,0 +1,88 @@
...
+		<groups>
+			<group>system</group>
+		</groups>
+	</test>
+	<test>
+		<testCaseName>churn_custom</testCaseName>
+		<command> 
+            export OTOOL_JDK_VERSION="$(JDK_VERSION)" ; \
+            export JREJDK="jdk" ; \
+            export OTOOL_garbageCollector="${CHURN_GCS}" ; \
```
smlambert, line 69:
```suggestion
            export OTOOL_garbageCollector="${APPLICATION_OPTIONS}" ; \
```

functional/churn/README.md
```diff
@@ -0,0 +1,36 @@
...
+## Running churn locally
+clone https://github.com/rh-openjdk/churn read  https://github.com/rh-openjdk/churn/blob/master/README and run run.sh
+
+### Setting up the environment:
+Java required. Maven recommended, churn can be compiled via direct javac if needed
+Java is used from JAVA_HOME or guessed from PATH.
+run.sh can use OTOOL_garbageCollector and OTOOL_JDK_VERSION instead of GC argument. See https://github.com/rh-openjdk/churn/blob/master/README for details.
+See https://github.com/rh-openjdk/churn/tree/master/bin for list of supported GCs. The compressed ops in upstream run can be controlled by runner or NOCOMP=-nocoops variable (for run.sh)
+
+## Running via aqavit
```
smlambert, line 14:
```suggestion
## Running via AQAvit
```

functional/churn/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+The support for compressed ops is handled by aqavit itself. It is currently not sure if churn will be able to honor it. If not, churn will be fixed.
```
smlambert, line 16:
```suggestion
The support for compressed ops is handled by AQAvit itself through the use of the `<variation>` tag in playlist.xml. It is currently not sure if churn will be able to honor it. If not, churn will be fixed.
```

functional/churn/README.md
```diff
@@ -0,0 +1,36 @@
...
+Similarly the java version and JAVA_HOME are handled by aqavit.
```
smlambert, line 17:
```suggestion
Similarly the java version and JAVA_HOME are handled by AQAvit.
```

functional/churn/playlist.xml
```diff
@@ -0,0 +1,88 @@
...
+            export DURATION="${CHURN_DURATION}" ; \
```
smlambert, line 70:
```suggestion
            export DURATION="300" ; \
```

functional/churn/README.md
```diff
@@ -0,0 +1,36 @@
...
+### Executing the testsuite
+The `BUILD_LIST of functional/churn` contains three targets:
+ * _churn_1m_allGCs
+ * _churn_5h_allGCs
+ * _churn_custom
+
+There are major differences in them:
+ * churn_1m_allGCs - is testing ground, which runs each GC only for aprox 10 seconds, to simply see if the setup works
+ * churn_5h_allGCs - Is running each GC a bit over, which runs each GC for aprox hour and half. A minimum, which can find some real GC issue.
+ * churn_custom - this one needs you to set and export CHURN_GCS and CHURN_DURATION to select GC(or GCs) and duration. So it allows you to test your custom GC - if churn supports that, despite of anybody else (eg `default` or `ALL` thinks)
```
smlambert, line 28:
```suggestion
 * churn_custom - this one is to support development, when run on the commandline you need to export CHURN_GCS and CHURN_DURATION to select GC(or GCs) and duration. So it allows you to test your custom GC - if churn supports that, despite what other churn options suggest (eg `default` or `ALL` thinks).  When running in a Jenkins Grinder job, can use the APPLICATION_OPTIONS parameter to set CHURN_GCS value and CHURN_DURATION defaults to 5 minutes.
```

functional/churn/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+churn_1m_allGCs and churn_5h_allGCs are using pony `ALL` keyword, which is interpreted (based on hardcoded list) as all GC in tested JVM. The set time is divided among them.
+Note, that if you use `churn_custom` and enumeration, eg `CHURN_GCS="zgc g1"` then the time will not be divided. The `CHURN_DURATION` is in seconds
+The `CHURN_GCS="defaultgc"`will set the tested GC to defautl GC as run.sh think is right. So be aware. Although it is maintained, if your custom JDK have custom GC, it is unlikely to be known
```
smlambert, line 32:
```suggestion
The `CHURN_GCS="defaultgc"`will set the tested GC to default GC as run.sh think is right. So be aware. Although it is maintained, if your custom JDK have custom GC, it is unlikely to be known.
```

functional/churn/README.md
```diff
@@ -0,0 +1,36 @@
+# churn
+Test targeting garbage collector(s).
+The suite is able to operate with all GCs in OpenJDK. If your implementation is missing some, the calls to them will fall. If you have some additional one, an enabling wrapper must be created first.
+
+## Running churn locally
+clone https://github.com/rh-openjdk/churn read  https://github.com/rh-openjdk/churn/blob/master/README and run run.sh
+
+### Setting up the environment:
```
smlambert, line 8:
```suggestion
### Setting up the environment
```

functional/churn/README.md
```diff
@@ -0,0 +1,36 @@
...
+There are major differences in them:
+ * churn_1m_allGCs - is testing ground, which runs each GC only for aprox 10 seconds, to simply see if the setup works
+ * churn_5h_allGCs - Is running each GC a bit over, which runs each GC for aprox hour and half. A minimum, which can find some real GC issue.
+ * churn_custom - this one needs you to set and export CHURN_GCS and CHURN_DURATION to select GC(or GCs) and duration. So it allows you to test your custom GC - if churn supports that, despite of anybody else (eg `default` or `ALL` thinks)
+
+churn_1m_allGCs and churn_5h_allGCs are using pony `ALL` keyword, which is interpreted (based on hardcoded list) as all GC in tested JVM. The set time is divided among them.
+Note, that if you use `churn_custom` and enumeration, eg `CHURN_GCS="zgc g1"` then the time will not be divided. The `CHURN_DURATION` is in seconds
+The `CHURN_GCS="defaultgc"`will set the tested GC to defautl GC as run.sh think is right. So be aware. Although it is maintained, if your custom JDK have custom GC, it is unlikely to be known
+
+### reading results:
```
smlambert, line 34:
```suggestion
### Reading results
```

functional/churn/README.md
```diff
@@ -0,0 +1,36 @@
...
+tap file and compressed xmlfile is genrated. Use eg https://github.com/jenkinsci/report-jtreg-plugin or https://plugins.jenkins.io/tap/ to read them.
```
smlambert, line 35:
```suggestion
TAP file and compressed xmlfile are generated. Use eg https://github.com/jenkinsci/report-jtreg-plugin or https://plugins.jenkins.io/tap/ to read them.
```

functional/churn/playlist.xml
```diff
@@ -0,0 +1,88 @@
...
+			<level>dev</level>
+		</levels>
+		<platformRequirementsList>
+			<platformRequirements>os.linux</platformRequirements>
+			<platformRequirements>os.win</platformRequirements>
+			<platformRequirements>os.osx</platformRequirements>
+		</platformRequirementsList>
+		<groups>
+			<group>system</group>
+		</groups>
```
smlambert, line 38: Let's restrict to particular vendors for now, as I am not sure which vendors include the entire list of GCs that are in 'allGCs' in their distribution (in particular shenandoah) and we do not want to break others.
Add this bit into the `<test>` block, can be right after the `</groups>` tag
```
<vendors>
        <vendor>eclipse</vendor>
        <vendor>redhat</vendor>
</vendors>
```

functional/churn/playlist.xml
```diff
@@ -0,0 +1,88 @@
...
```
smlambert, line 62:
```
<vendors>
        <vendor>eclipse</vendor>
        <vendor>redhat</vendor>
</vendors>
```

functional/churn/playlist.xml
```diff
@@ -0,0 +1,88 @@
...
+	</test>
+	<test>
+		<testCaseName>churn_custom</testCaseName>
```
smlambert, line 65:
```
<disables>
	<disable>
		<comment>Can run disabled.churn_custom in Grinder, meant for dev use not regular automated tests.</comment>
	</disable>
</disables>
```

smlambert (changes requested):
Review comments and suggestions are added inline

smlambert:
> if any GC fails, or (in case of churn_custom) user put in ninsense, it will make the job red.
Is it ok, or should some other logic be added, to make it unstable?

Yes, we will add this test target with a `disable` tag, since it is there to support development, it can be used in Grinder jobs, and will be skipped in automated / regular test runs.

> Will this suite ( I guess the churn_5h_allGCs target) be part of any regular aqa runs in public jenkins?
if so, to which, if not by default, but is desired can I add it somewhere?

Yes, it will get run against our EA builds which get triggered whenever a new upstream tag is available (so approximately weekly for each JDK version).  Since we are adding this into dev level, and system group, the 2 enabled test targets in your playlist, churn_1m_allGCs and churn_5h_allGCs, will get run in our `*dev.system*` test jobs.  I have asked you to restrict the test targets to just eclipse and redhat vendors for now, as I do not want to impact other vendors test execution times and also do not think other vendors are shipping with Shenandoah enabled.

judovana:
> re: [#5101 (comment)](https://github.com/adoptium/aqa-tests/pull/5101#issuecomment-1969567366) - you did not need to do anything, our test pipelines always generate TAP files that include all test targets ( all `<test><testCaseName> ... </test>` entries defined in playlists). It is a feature of our TKG layer to create the TAP file, and a feature of our Jenkins groovy scripts to then attach the TAP files to the Jenkins job as an artifact.

Well, yes. With all pros and cons. There is usually one tap file, which consider the whole testsuite as on test.  Then find properly what really run/failed inside is quite a detective task(but most liekly it is lack of my insight). Where the one default generated tap file is *awesome* feature, the suites usually generates much better outputs, and should be allowed to publish it.

In this case, the tap file churn generates is pretty good, and as the tkg toolchain is picking up all taps, it is pretty usable (feel free to correct me, or tell me to not generate it).

functional/churn/playlist.xml
judovana: I do not understand this requirements:
```
- export OTOOL_garbageCollector="${CHURN_GCS}" ; \
+export OTOOL_garbageCollector="${APPLICATION_OPTIONS}" ; \
...
- export DURATION="${CHURN_DURATION}" ; \
+ export DURATION="300" ; \
```
The APPLICATION_OPTIONS is something what is custom in aqavit? Isn't it some more general setup?
The "zgc" or "shenandaoh" can be hardly considered as generic.

To remove custom duration s moreover killing the purpise of the changeset and the suite. If you are developing custom GC or porting JDK to new paltform, you want to run churn on each GC you care about (usually default or the single custom one) for several hours. 300s is useles.  Generally any hardcoded value in the "custom" setupo is (IMO) bad.

Can we stay with originals pelase? Or maybe an APPLICATION_OPTIONS contain key=value pairs? (but as GC can be space separated list of GCs, it sounds bad). Or maybe APPLICATION_OPTIONS_GC and APPLICATION_OPTIONS_DURATION ?

functional/churn/playlist.xml
judovana: Answered in
```
           export OTOOL_garbageCollector="${CHURN_GCS}" ; \
            export OTOOL_garbageCollector="${APPLICATION_OPTIONS}" ; \
```
section

functional/churn/README.md
judovana: I had adapt this only partially, as I had not (yet!) followed the APPLICATION_OPTIONS and 5 minutes as you requested (subject of another discussion)

functional/churn/playlist.xml
judovana: Seeing your touch to readme - is there eal fear, that some ccrazy eprson will start 20days long churn run? If so, then I would lke to still keep the duration adjsutable, and only eg limit it from top.

if such logic would be desirable, then also the  churn_1m_allGCs and churn_5h_allGCs may actually take GC as parameter. And fallback to "ALL" if nothing provided. wdyt?

functional/churn/playlist.xml
judovana: I will add this element, But I think the generic openjdk should be for sure here.  I now recall, there may be issue with shenandoah on jdk8, but I would rather workaround it in CHURN itself, then exclude generic oepnjdk. Thoughts?

functional/churn/playlist.xml
judovana: cool!

judovana:
> > if any GC fails, or (in case of churn_custom) user put in ninsense, it will make the job red.
> > Is it ok, or should some other logic be added, to make it unstable?
> 
> Yes, we will add this test target with a `disable` tag, since it is there to support development, it can be used in Grinder jobs, and will be skipped in automated / regular test runs.
> 
> > Will this suite ( I guess the churn_5h_allGCs target) be part of any regular aqa runs in public jenkins?
> > if so, to which, if not by default, but is desired can I add it somewhere?
> 
> Yes, it will get run against our EA builds which get triggered whenever a new upstream tag is available (so approximately weekly for each JDK version). Since we are adding this into dev level, and system group, the 2 enabled test targets in your playlist, churn_1m_allGCs and churn_5h_allGCs, will get run in our `*dev.system*` test jobs.

Thanx! 

>  I have asked you to restrict the test targets to just eclipse and redhat vendors for now, as I do not want to impact other vendors test execution times and also do not think other vendors are shipping with Shenandoah enabled.

IIRC, shenandoah is enabled everywhere except jdk8 by default, so they would need to manually disable it during build.   But why would anybody disable best GC around?

In case of shenandoah, especially because of its 'not-present' in jdk8, I will make it more dynamic. I can exclude GCs from default list based on `:XX` listings Will elaborate. It do not need to block this PR, only once it is created, other vendors may be onboarded.

smlambert:
> It do not need to block this PR, only once it is created, other vendors may be onboarded.
Yes, my thought with all of my review comments was to get the PR in and then come back later to adjust after we have run it for a while at the project.

smlambert:
> There is usually one tap file, which consider the whole testsuite as on test. Then find properly what really run/failed inside is quite a detective task(but most liekly it is lack of my insight). Where the one default generated tap file is awesome feature, the suites usually generates much better outputs, and should be allowed to publish it.  In this case, the tap file churn generates is pretty good, and as the tkg toolchain is picking up all taps, it is pretty usable (feel free to correct me, or tell me to not generate it).

The TAP file is produced to 'normalize' test results from all suites.  Having TAP files that do something different moves us away from normalization.  For more granular information, we would look to test suites producing some sort of JUnit style output and then also present that.

functional/churn/playlist.xml
judovana: I had made the listing of GCs dynamic: https://github.com/rh-openjdk/churn/commit/17f1b833249028f9efee13a4411f4b39ed592b0a

it is still based on known values, but at least will not enforce not-existing GC. Any advice on enabling more vendors? (compelltyh up to you :) tyvm!!)

judovana:
> > There is usually one tap file, which consider the whole testsuite as on test. Then find properly what really run/failed inside is quite a detective task(but most liekly it is lack of my insight). Where the one default generated tap file is awesome feature, the suites usually generates much better outputs, and should be allowed to publish it.  In this case, the tap file churn generates is pretty good, and as the tkg toolchain is picking up all taps, it is pretty usable (feel free to correct me, or tell me to not generate it).
> 
> The TAP file is produced to 'normalize' test results from all suites. Having TAP files that do something different moves us away from normalization. For more granular information, we would look to test suites producing some sort of JUnit style output and then also present that.

I see. The churn generates both tap file and packed junit file. So Your recomndation is to stop generating tap file? (no problem on my side, only I liked it :) )

functional/churn/playlist.xml
smlambert: Please see that I have asked that this churn_custom target is disabled (using the `<disables>` block I suggested), in that way it does not get run during our automated runs.  

Also in our world, no one is allowed to go onto test machines and set CHURN_DURATION and CHURN_GCS environment variables, and we would not want them to do that (for lots of reasons, not the least of which is transparency of machine config).

This leaves me to the point to look at this test target as a useful thing when someone wants to run a Grinder to test a particular GC for a short duration, which is why I suggest leveraging the existing APPLICATION_OPTIONS mechanism for pinpointed testing of a particular GC failure.

All that to be said, feel free to not take those review comments and apply them.  

> if such logic would be desirable, then also the churn_1m_allGCs and churn_5h_allGCs may actually take GC as parameter. And fallback to "ALL" if nothing provided. wdyt?

Yes, the other option is to remove the churn_custom target altogether from the playlist.xml file in this PR.  That would be the expedient way to get this PR through.

Could consider then leveraging APPLICATION_OPTIONS to pass a subset of GCs and if its blank, expect the churn test suite to default to all.

judovana:
> > It do not need to block this PR, only once it is created, other vendors may be onboarded.
> 
> Yes, my thought with all of my review comments was to get the PR in and then come back later to adjust after we have run it for a while at the project.

Sure. Thank you very much! Highly appreciated.
 *  I had followed you review everywhere except the CHURN_DURATION. How do you feel about it?
 * I think we may enable more venodrs, but it is up to you, the requested chnage is done
 * readme updated
 * custom tap is no longer geenrated (/me very sad)
 * GH actions are passing all three OSes. But it is msys2 for windows on them, in grinder it is cygwin, and it [fails](https://ci.adoptium.net/view/Test_grinder/job/Grinder/9034/tapResults/) there. Thoughts? 
 * no mac run finished up to now (still waiting for executor).
   *   thus saying, I would :
```
 		<platformRequirementsList>
 			<platformRequirements>os.linux</platformRequirements>
-			<platformRequirements>os.win</platformRequirements>
-			<platformRequirements>os.osx</platformRequirements>
 		</platformRequirementsList>
```
wdyt?

Do you se enay more obstacles?

sophia-guo:
The tests might need to move to system/ folder instead of functional/ folder if we'd like to trigger the tests automatically as part of jenkins test builds of dev.system. It  does work in grinder under functional/ folder as in grinder we can manually set buildList=functional/churn, however in jenkins dev.system jobs buildList=system.

functional/churn/playlist.xml
```diff
@@ -0,0 +1,103 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+-->
+<playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../TKG/resources/playlist.xsd">
+	<test>
+		<testCaseName>churn_1m_allGCs</testCaseName>
```
sophia-guo, line 17: churn_1m_allGCs is for testing only. When the PR is ready for merge can it be disabled by adding tag <disables/> ?

smlambert:
> The tests might need to move to system/ folder instead of functional/ folder if we'd like to trigger the tests automatically as part of jenkins test builds of dev.system. It does work in grinder under functional/ folder as in grinder we can manually set buildList=functional/churn, however in jenkins dev.system jobs buildList=system.

Good catch, we started off with this in functional group, but because it is a stress/load test, it is more appropriate to add it to system, so better to shift the churn folder into the system folder. +1

sophia-guo:
> GH actions are passing all three OSes. But it is msys2 for windows on them, in grinder it is cygwin, and it [fails](https://ci.adoptium.net/view/Test_grinder/job/Grinder/9034/tapResults/) there. Thoughts?

Does the test require MVN? The output says `07:15:38  + which mvn
07:15:38  which: no mvn ` so no classes generated and failed.

judovana:
> > GH actions are passing all three OSes. But it is msys2 for windows on them, in grinder it is cygwin, and it [fails](https://ci.adoptium.net/view/Test_grinder/job/Grinder/9034/tapResults/) there. Thoughts?
> 
> Does the test require MVN? The output says `07:15:38 + which mvn 07:15:38 which: no mvn ` so no classes generated and failed.

Default build system is maven. But if it is not found, javac would do. So that is this case, it is compiled directly by javac.

sophia-guo:
I see. It's a issue with window path and cygwin path.

```
+ E:/workspace/Grinder/jdkbinary/j2sdk-image/bin/javac -d /cygdrive/e/workspace/Grinder/aqa-tests/functional/churn/churn/target/classes ./org/jboss/churn/LogHistogram.java ./org/jboss/churn/TestRunner.java ./org/jboss/churn/WorkItem.java ./org/jboss/churn/WorkItemMap.java
07:15:40  + pushd /cygdrive/e/workspace/Grinder/aqa-tests/functional/churn/churn/target/classes
07:15:40  E:/workspace/Grinder/aqa-tests/\functional\churn\churn\run.sh: line 183: pushd: /cygdrive/e/workspace/Grinder/aqa-tests/functional/churn/churn/target/classes: No such file or directory
```

sophia-guo:
You might need to use `cygpth -w` to print Windows form of directory of `/cygdrive/e/workspace/Grinder/aqa-tests/functional/churn/churn/target/classes` and probably other cygwin paths too.

sophia-guo:
Mac looks good https://ci.adoptium.net/view/Test_grinder/job/Grinder/9033/console

judovana:
> For cygwin `"$OS" =~ CYGWIN* ` you might need to use `cygpth -w` to print Windows form of directory of `/cygdrive/e/workspace/Grinder/aqa-tests/functional/churn/churn/target/classes` and also for any other cygwin paths.

Right yiou are, https://github.com/rh-openjdk/churn/pull/24 was all waht was needed. Rerunning windows.
Thanx a lot for info of mack! saved me some search :)

judovana:
> > > It do not need to block this PR, only once it is created, other vendors may be onboarded.
> > 
> > 
> > Yes, my thought with all of my review comments was to get the PR in and then come back later to adjust after we have run it for a while at the project.
> 
> Sure. Thank you very much! Highly appreciated.
> 
>     * I had followed you review everywhere except the CHURN_DURATION. How do you feel about it?
> 
>     * I think we may enable more venodrs, but it is up to you, the requested chnage is done
> 
>     * readme updated
> 
>     * custom tap is no longer geenrated (/me very sad)
> 
>     * GH actions are passing all three OSes. But it is msys2 for windows on them, in grinder it is cygwin, and it [fails](https://ci.adoptium.net/view/Test_grinder/job/Grinder/9034/tapResults/) there. Thoughts?
> 
>     * no mac run finished up to now (still waiting for executor).
>       
>       * thus saying, I would :
> 
> 
> ```
>  		<platformRequirementsList>
>  			<platformRequirements>os.linux</platformRequirements>
> -			<platformRequirements>os.win</platformRequirements>
> -			<platformRequirements>os.osx</platformRequirements>
>  		</platformRequirementsList>
> ```

^ no longer necessary.
> 
> wdyt?
> 
> Do you se enay more obstacles?

I guess the disable of  `churn_custom` target will happen as last thing in this commit once everything else is tunned

sophia-guo:
> [Rerunning windows.](https://ci.adoptium.net/view/Test_grinder/job/Grinder/9037/)

The new failure was related with migration from functional to system, I'm hitting the same issue with reproducible comparing tests. I can take a look next Monday.

judovana:
Gosh. Rerunning windows here: https://ci.adoptium.net/view/Test_grinder/job/Grinder/9038/
mac: https://ci.adoptium.net/view/Test_grinder/job/Grinder/9039/ 
linux: https://ci.adoptium.net/view/Test_grinder/job/Grinder/9040/

sophia-guo (approved):

judovana:
@sophia-guo  TY!
@smlambert  one approval reached, do you have any more hints?

system/churn/build.xml
```diff
@@ -0,0 +1,64 @@
...
+			<arg value="clone" />
+			<arg value="${GIT_REPO}rh-openjdk/churn.git" />
+		</exec>
+	</target>
+
+	<target name="init">
+		<mkdir dir="${DEST}"/>
+	</target>
+
+	<target name="dist" depends="getChurn" description="generate the distribution">
```
sophia-guo, line 55: Should add `init` task as depends `depends="getChurn, init"`  unless the non-exist directory will be created when doing the `<copy todir="${DEST}">`? Not sure about this looking at the grinder the $DEST was created.

smlambert (approved):
thanks @judovana !  I may come back for an enhancement on how we run the churn_custom target, but this is good as is.

PR #5102, opened 2024-02-27
Check testenv.properties JDKxx_BRANCH tags

### LINKED ISSUE #4373: Automate or semi-automate update of testenv.properties file
This is an enhancement feature to automate or semi-automate the final update to the testenv.properties file.

### Background
We currently create a release branch 1 - 2 weeks ahead of an OpenJDK release and populate the contents of the testenv.properties file with the tags that we think will be closest to the GA tags to use for that release.  This allows for vendors, including Adoptium with its Temurin distribution, to begin AQAvit verification testing dry runs prior to the OpenJDK GA tags being available.

Upon their availability, we update the testenv.properties file with the new tag.

When anyone subsequently runs AQAvit verification on a JDK binary, they will be pulling test material from the same branch / tag, and the subsequent SHAs are captured in the test result files (TAP files) that are publicly shared in order to list in the marketplace.

### Enhancement
Create a workflow that monitors the mirror repos for GA tags availability and automatically create a PR to update the testenv.properties file with the new tag, if it represents a different SHA than the tag that currently in the file for a particular version.   Noting that the testenv.properties file contains all of the active JDK_VERSIONS, so this workflow would be triggered multiple times to create multiple PRs to update the file, as GA tags become available, are compared against the current tag in the file and either initiates to create a PR or not (depending on git diff of the 2 tags).

### Considerations
The build pipelines now also watch for the GA tags and launch automatically when they are present, however, ideally we would like to launch them automatically after the aqa-tests release branch has its testenv.properties file updated.

Note:
- ensure we know whether JDKxx_BRANCH_PREFIX can be predicted from https://www.java.com/releases/
- January CPU / 21.0.2, 17.0.10, 11.0.22, 8u401

smlambert:
Fixes #4373 
Add a script to check the existence of testenv.properties tags for the JDKxx_BRANCH values when USE_TESTENV_PROPERTIES=true.  If the value is a valid branch or tag name, then use it, otherwise get the latest tag and use that.

A second pass of this should be done to cover the case where the value in the testenv.properties files are SHAs (instead of branch or tag names).


smlambert:
Various tests

[Grinder/8973](https://ci.adoptium.net/job/Grinder/8973) (JDK8, dev branch)
```
21:09:05  Running checkTags with ./testenv/testenv.properties and 8
21:09:05  Use branch name dev from ./testenv/testenv.properties
```

[Grinder/8979](https://ci.adoptium.net/job/Grinder/8979/) (JDK11, non-existent jdk-11.0.23-ga tag)
```
21:55:45  Running checkTags with ./testenv/testenv.properties and 11
21:55:46  Override non-existent jdk-11.0.23-ga with JDK11_BRANCH=jdk-11.0.23+4 and write to ./testenv/testenv.properties 
```

[Grinder/8980](https://ci.adoptium.net/job/Grinder/8980/) (JDK11, non-existent blahblah tag)
```
22:13:50  Running checkTags with ./testenv/testenv.properties and 11
22:13:51  Unable to resolve the latest JDK11_BRANCH=blahblah tag from JDK11_REPO
```
and exits immediately

smlambert:
[Grinder/8983](https://ci.adoptium.net/job/Grinder/8983) (JDK17 with JDK17_BRANCH=latestSHA)
```
23:06:57  Running checkTags with ./testenv/testenv.properties and 17
23:06:59  SHA 37cf29ab1f33c827b7cf67f44f7698ab8c292b88 exists on https://github.com/adoptium/jdk17u.git
```

Grinder fails later, as presumably, we do not handle the case for getting the OpenJDK test material for a particular commit SHA at present.  So this may come in as a separate PR.

llxia:
From the console output, cmd for a branch is expected (see below). We need to update openjdk logic.

```
23:07:21       [exec] git clone --depth 1 -q --reference-if-able /home/jenkins/openjdk_cache -b 37cf29ab1f33c827b7cf67f44f7698ab8c292b88 https://github.com/adoptium/jdk17u.git openjdktemp
23:07:21       [exec] info: Could not add alternate for '/home/jenkins/openjdk_cache': reference repository '/home/jenkins/openjdk_cache' is not a local repository.
```

smlambert:
Given https://github.com/adoptium/aqa-tests/pull/5102#issuecomment-1966842379, I will not worry about the SHA edge case for this PR.  This PR does not break anything, since the functionality is not there.  This PR does goes as far as identifying when we have a SHA which is good, and we can address the update to the logic in the openjdk/build.xml and openjdk.mk file in a subsequent PR.

llxia (approved):

sophia-guo:
@ShelleyLambert just to double check the following scenario of 

> [Grinder/8979](https://ci.adoptium.net/job/Grinder/8979/) (JDK11, non-existent jdk-11.0.23-ga tag)

There is  the output of `Override non-existent jdk-11.0.23-ga with JDK11_BRANCH=jdk-11.0.23+4 and write to ./testenv/testenv.properties `， which is contradictory to originally story  `21:57:28  USE_TESTENV_PROPERTIES was set, not writing to testenv.properties`. Might need to update some print message? And if the ./testenv/testenv.properties is overwritten do we still need to manually create the PR to do the update? Otherwise the overwritten will be done everytime called during release?

smlambert:
> There is the output of Override non-existent jdk-11.0.23-ga with JDK11_BRANCH=jdk-11.0.23+4 and write to ./testenv/testenv.properties ， which is contradictory to originally story 21:57:28  USE_TESTENV_PROPERTIES was set, not writing to testenv.properties. Might need to update some print message? And if the ./testenv/testenv.properties is overwritten do we still need to manually create the PR to do the update? Otherwise the overwritten will be done everytime called during release?

This PR was created so that we will no longer need to manually update the testenv.properties file.  One less step in release checklist!

The bit of code below, which updates the contents of the testenv.properties call only gets called in a very specific case, it is called when all of the following criteria are met:
- the current value for JDKxx_BRANCH in the testenv.properties file is a tag that does not exist (example if this entry is in the properties file, JDK11_BRANCH=jdk-11.0.23-ga which doesn't exist)
- we strip off the -ga substring of the tag value (if present), take the remaining string and try and fetch the latest tag that starts with that substring (example, take the prefix of the non-existent tag, jdk-11.0.23 and find the latest tag with that prefix)
```
        if [[ $latestTag ]]; then
          echo "Override non-existent $workingTag with $branchFromPropFile=$latestTag and write to $teFile "
          export $branchFromPropFile=$latestTag
          setProperty $branchFromPropFile $latestTag 
```
if $latestTag is not found, we do not write it to testenv.properties file, we exit / fail.

You are correct that I should update the output message to state why in this case we are writing to the testenv.properties file when we said we would not be doing so, if USE_TESTENV_PROPERTIES=true.  testenv.properties should contain the property values that were used in the test run.  When USE_TESTENV_PROPERTIES=false, we grab the SHAs that were used and write them to the file.  When USE_TESTENV_PROPERTIES=true and we find that the tag value does not exist, but we find a suitable fallback tag and proceed to use it, we need to capture that in the testenv.properties file.

Let me see if I can improve the print message a bit.

smlambert:
@sophia-guo - I have updated my output message, and can create a TKG pull request to change the wording of the other output message, which could read something like "USE_TESTENV_PROPERTIES was set, not overwriting the testenv.properties file with SHAs"

sophia-guo (approved):

Issue #5117, opened 2024-03-01
Sanity.external openliberty-mp-tck test fails on unreleased JDK versions like 22 and 23


adamfarley:
**Summary**

openliberty-mp-tck cannot find a docker image manifest file for unreleased jdk versions.

- JDK22: eclipse-temurin:22-jdk
- JDK23: eclipse-temurin:23-jdk

**Error message**

```
00:27:48       [exec] The test in the build_image() function is openliberty-mp-tck
00:27:48       [exec] #####################################################
00:27:48       [exec] INFO: docker build  --no-cache -t adoptopenjdk-openliberty-mp-tck-test:22-jdk-ubuntu-hotspot-full -f /home/jenkins/workspace/Test_openjdk22_hs_sanity.external_x86-64_linux/jvmtest/external/openliberty-mp-tck/dockerfile/22/jdk/ubuntu/Dockerfile.hotspot.full /home/jenkins/workspace/Test_openjdk22_hs_sanity.external_x86-64_linux/jvmtest/external/
00:27:48       [exec] #####################################################
00:27:48       [exec] Sending build context to Docker daemon  71.68kB
00:27:48       [exec] Step 1/25 : ARG IMAGE=eclipse-temurin:22-jdk
00:27:48       [exec] Step 2/25 : ARG OS=ubuntu
00:27:48       [exec] Step 3/25 : ARG IMAGE_VERSION=nightly
00:27:48       [exec] Step 4/25 : ARG TAG=22-jdk
00:27:48       [exec] Step 5/25 : FROM $IMAGE
00:27:48       [exec] manifest for eclipse-temurin:22-jdk not found: manifest unknown: manifest unknown
```

**Links**

https://ci.adoptium.net/job/Test_openjdk22_hs_sanity.external_x86-64_linux/6/console
https://hub.docker.com/_/eclipse-temurin

**Details**

Looks like we don't make docker images for jdk22 and 23 yet. Will ask George if we need to create the relevant folders in the [docker repo](https://github.com/adoptium/containers) so we can start creating + uploading these images.

adamfarley:
@gdams - Could you take a look at this, please?

smlambert:
We now only produce official Docker Temurin images for versions that are 'officially released' (GA builds).  As per https://www.java.com/releases/, since JDK22 will only have its first official release in March 2024, and JDK23 in Sept 2024, we should disable the external tests on those versions.

[...]

PR #5118, opened 2024-03-01
Exclude sanity.external targets for 22+

smlambert:
Related: https://github.com/adoptium/aqa-tests/issues/5117

Temporarily disable these test targets (re-enable once official Temurin Docker images become available)


smlambert:
Actually will close this PR and make an update in ci-jenkins-pipeline repository instead (since if we keep running the sanity.external test jobs in 22+, they will run and since there will be zero test targets enabled, those jobs would fail with an error message that reported that "no tests were found".

PR #5125, opened 2024-03-06
Disabled NaN-related tests on RISC-V more broadly

luhenry:
The issue is not related with the version of the JDK, but rather with how the RISC-V architecture handles NaN and the expectations of the test.


luhenry:
cc @smlambert @sxa @gdams

gdams (approved):

llxia (approved):

PR #5135, opened 2024-03-12
Adoptopenjdk_repo can only be changed in grinders

adamfarley:
This change is to alert jenkins users that non-grinder test jobs will ignore any changes to these two variables.

### DIFF
--- a/buildenv/jenkins/testJobTemplate
+++ b/buildenv/jenkins/testJobTemplate
@@ -227,7 +227,7 @@ ARCH_OS_LIST.each { ARCH_OS ->
 							    sectionHeader('Repositories where we pull test material from. Unless you are testing test code, these do not need to be changed.')
 							    sectionHeaderStyle(sectionHeaderHelpTextStyleCss)
 							}
-							stringParam('ADOPTOPENJDK_REPO', ADOPTOPENJDK_REPO, "AdoptOpenJDK git repo. Please use ssh for zos. Can only be changed in grinders and generated child jobs.")
+							stringParam('ADOPTOPENJDK_REPO', ADOPTOPENJDK_REPO, "AdoptOpenJDK git repo. Please use ssh for zos. Changes will only be used in grinders and generated child jobs.")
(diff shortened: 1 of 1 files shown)


sendaoYan (approved):
/LGTM

smlambert (changes requested):
I do not believe this is technically true (and want to check how that is enforced if is true).  ~~Also why drop the information about zos, used by one of our members.~~ ignore that, see that its retained.

You may be referring to when Test jobs triggered by build pipelines override the values of these, or when USE_TESTENV_PROPERTIES=true, and not the case when running a Test job manually with the appropriate parameters set.

adamfarley:
Hi Shelley,

I think this enforced by the "Pipeline script from SCM" bit in the job config. If you're rerunning a test job manually, the Git repository URL is hard-coded in a normal test job, but set dynamically to the value of the ADOPTOPENJDK_REPO variable in a grinder job.

I recall the team had a discussion about this during one of our calls (bit vague, sorry), and the conclusion was that this was intentional. I proposed adding a comment to clarify the functionality.

adamfarley:
Ok, here's the chat: https://adoptium.slack.com/archives/C5219G28G/p1707139690344389

adamfarley:
I think what tripped me up at the time was that the aqa-test branch is set dynamically, but the repo is hard-coded.

In normal test jobs, I mean. Like this one: https://ci.adoptium.net/view/Failing%20Temurin%20Test%20Jobs/job/Test_openjdk23_hs_extended.perf_aarch64_windows/

llxia (changes requested):
This statement is incorrect. This PR should be closed without merging.

adamfarley:
> This statement is incorrect. This PR should be closed without merging.

Hi Lan. I agree that the comment could be more accurate, so I've updated it.

Let me know if this works for you.

karianna (approved):

smlambert:
Close as undesired / not needed.

PR #5138, opened 2024-03-12
Jdk22 testenv.properties update for dryrun

andrew-m-leonard:


smlambert (approved):
I guess this is the only way to test dryrun tags, once dry-run complete, we can switch line 20 to jdk-22-ga if that is the expected tag for March (which presume it is, based on https://github.com/adoptium/mirror-scripts/blob/master/releasePlan.cfg#L15).

PR #5165, opened 2024-03-21
riscv64 should not attempt robot tests

sxa:
This fixes failures caused by invalid `jti` files being created on the platform, preventing some of the suites from executing correctly.

I'm open to options here but this is the simplest that fits in with the current format of this "switch" statement.

Alternatives if reviewers wish to consider them:
- Split out the if clauses to set robot and libpath separately which might be neater
- Add the riscv64 check as an `or` into the Alpine one

### DIFF
--- a/jck/jtrunner/JavatestUtil.java
+++ b/jck/jtrunner/JavatestUtil.java
@@ -486,10 +486,7 @@ public class JavatestUtil {
 			if (spec.contains("win")) {
 				libPath = "PATH";
 				robotAvailable = "Yes";
-			} else if (spec.contains("alpine-linux")) {
-				libPath = "LD_LIBRARY_PATH";
-				robotAvailable = "No";
-			} else if (spec.contains("riscv64")) {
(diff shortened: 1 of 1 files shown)


gdams (approved):

jck/jtrunner/JavatestUtil.java
```diff
@@ -489,6 +489,9 @@ private static boolean generateJTB() throws Exception {
 			} else if (spec.contains("alpine-linux")) {
```
llxia, line 489: can we do it in one line?
```
else if (spec.contains("alpine-linux") || spec.contains("riscv64")) {
```

sxa: Yes we can if you think that's best - that was one of the two alternative options that I put in the initial description as I wasn't clear if it has been deliberately set up with only one condition in each branch of the if/else section :-)

smlambert: Ultimately in a future revision, this code can be improved/updated to check if a platform under test is headless, and set 'no' to robot.

karianna (approved):

PR #5202, opened 2024-04-06
Set testenv.properties to dry run tags

sxa:
dry run release triggers do not work properly unless the triggered tags exist in testenv.properties. This caused a knock on issue which prevents subsequent triggers from working properly.

I have used jdk-22.0.0-dryrun-ga instead of 22.0.1 because there are no viisble tags we can use so I've used an earlier version number to avoid confusion and attached it to a recent commit in the repository.

More details in the [#build slack thread](https://adoptium.slack.com/archives/C09NW3L2J/p1712403713338159?thread_ts=1712334507.581379&cid=C09NW3L2J).


smlambert (approved):

smlambert:
We will proceed with this temporary change for April dryruns, but created https://github.com/adoptium/ci-jenkins-pipelines/issues/994 as we should be able to handle the dryrun case in such a way that does not need us to make these late-in-game changes to this file.

PR #5248, opened 2024-04-19
Crypto test original target now allows to run kereberos tests

judovana:
if both
   SKIP_AGENT_TESTS=false
   AGENT_HOSTNAME=some.krb.enabled.agent
are provided, the 5 kerberos based tests will run too

part of https://github.com/adoptium/aqa-tests/issues/5246

PR #5249, opened 2024-04-19
Made both CryptoTests_jtreg and CryptoTests targets vendor neutral

judovana:
It is passing for all vendors, the differences lies in underlyng OS, not JDK

### DIFF
--- a/functional/security/Crypto/playlist.xml
+++ b/functional/security/Crypto/playlist.xml
@@ -36,8 +36,7 @@
 				<platform>.*solaris</platform>
 			</disable>
 			<disable>
-				<comment
-					>This target is for those using GH actions or the Jenkins jtreg-plugin. The run.sh script provides extra functionality, tarring of results, downloading of bespoke jtreg.jar from a particular untagged commit SHA that is not yet upstreamed to a release version of jtreg that is used see complete output in jtr.xml files. In addition it simplify enable/disable of kerberos-based tests which needs remote, pre-set KDC.
+				<comment>This target is for those using GH actions or the Jenkins jtreg-plugin. The run.sh script provides extra functionality, tarring of results, downloading of bespoke jtreg.jar from a particular untagged commit SHA that is not yet upstreamed to a release version of jtreg that is used to see complete output in jtr.xml files. In addition, it simplifies enable/disable of kerberos-based tests which needs remote, pre-set KDC.
 				</comment>
 			</disable>
 		</disables>
--- a/functional/security/Crypto/playlist.xml
+++ b/functional/security/Crypto/playlist.xml
@@ -35,6 +35,11 @@
 				<comment>https://github.com/adoptium/aqa-tests/issues/4336</comment>
 				<platform>.*solaris</platform>
 			</disable>
+			<disable>
+				<comment
+					>This target is for those using GH actions or the Jenkins jtreg-plugin. The run.sh script provides extra functionality, tarring of results, downloading of bespoke jtreg.jar from a particular untagged commit SHA that is not yet upstreamed to a release version of jtreg that is used see complete output in jtr.xml files. In addition it simplify enable/disable of kerberos-based tests which needs remote, pre-set KDC.
+				</comment>
+			</disable>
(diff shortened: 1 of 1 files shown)


judovana:
However, as pointed out in https://github.com/adoptium/aqa-tests/pull/5248#pullrequestreview-2011754569   they must have no intersection on platforms/OSes

I would like to return CryptoTest target to be vendor neutral, and correclty run on platforms where it can.
On all other platforms - or generally anywhere where @llxia  needs it - it should run the new CryptoTests_jtreg

Can you help me to set up platforms properly or is this out of cope?

judovana:
Thanx, that is alredy much more frienldy, Will continue with it. TY!

functional/security/Crypto/playlist.xml
```diff
@@ -35,6 +35,14 @@
 				<comment>https://github.com/adoptium/aqa-tests/issues/4336</comment>
 				<platform>.*solaris</platform>
 			</disable>
+			<disable>
+				<comment>
+					This target is for those who would like to run upstream fully supported way of how to use this suite.
+					This launcher runs in GH, can enable custom jtreg.jar with upstreamed but unreleased stdout fixes,
+					can correctly enable/disable/setup kerberos tests and works in any CWD.
```
smlambert, line 42: This comment is inappropriate, please fix it with the suggested sentence instead of using loaded terms and phrases.  For the record, it is not only for the issue on s390x that others do not want to use run.sh, it is precisely because it has a bunch of extra behaviour that is unneeded and unwanted by those using AQAvit Jenkins pipelines.

judovana: Sure, I will try to elaborate on it. However there is no single word lying. Maybe except  ` ancient s390x platforms` . Although the s390x was the only platform I was able to get close reproduce the issue. On MACs it was working fine. 

I can not agree that ability to enable the kerberos based thest is unwanted. The only disputable part was tarring of  jtr.xml files, where I offered that I will add support for both.
Similarly I think that any possible user should be warned that `run.sh` is what is used in git hub actions.

As I stated before, I would much more likely create the original target with the original launcher to be usefull for anybody (and I committed myself to it) then do any forking. But it have been already decided.

smlambert: Fact-based:
`<comment>This target is for those using GH actions or the Jenkins jtreg-plugin.  The run.sh script provides extra functionality, tarring of results, downloading of bespoke jtreg.jar from a particular untagged commit SHA that is not yet upstreamed to a release version of jtreg that is used to set up some kerberos tests.</comment>`

functional/security/Crypto/playlist.xml
```diff
@@ -35,6 +35,11 @@
 				<comment>https://github.com/adoptium/aqa-tests/issues/4336</comment>
 				<platform>.*solaris</platform>
 			</disable>
+			<disable>
+				<comment
+					>This target is for those using GH actions or the Jenkins jtreg-plugin. The run.sh script provides extra functionality, tarring of results, downloading of bespoke jtreg.jar from a particular untagged commit SHA that is not yet upstreamed to a release version of jtreg that is used see complete output in jtr.xml files. In addition it simplify enable/disable of kerberos-based tests which needs remote, pre-set KDC.
+				</comment>
+			</disable>
```
smlambert, line 42:
```suggestion
			<disable>
				<comment>This target is for those using GH actions or the Jenkins jtreg-plugin. The run.sh script provides extra functionality, tarring of results, downloading of bespoke jtreg.jar from a particular untagged commit SHA that is not yet upstreamed to a release version of jtreg that is used to see complete output in jtr.xml files. In addition, it simplifies enable/disable of kerberos-based tests which needs remote, pre-set KDC.
				</comment>
			</disable>
```

judovana:
Thanx a lot for patience! Ok by me if I have permissions to say anything to it. TY!

smlambert (approved):
This is still in draft mode.  I can move it out to merge @judovana if you have made your final edits.

llxia (approved):

PR #5279, opened 2024-05-02
Exclude TestHandshake in JDK17 and JDK21

babsingh:
The new async handshake approach to fix synchronization issues was
introduced only in JDK22.

See https://bugs.openjdk.org/browse/JDK-8310644 for more details on
the async handshake approach.

JDK17 and JDK21 still use the old implementation, which has
synchronization issues. Thus, TestHandshake is being excluded in
JDK17 and JDK21.

Related: https://github.com/eclipse-openj9/openj9/issues/13211


babsingh:
fyi @pshipton

pshipton (approved):

JasonFengJ9 (approved):

Issue #5295, opened 2024-05-08
Error message `set: Illegal option -o pipefail`


sophia-guo:
Running grinder with USE_TESTENV_PROPERTIES=true got error message

```
16:56:42  + MAKE=make
16:56:42  + cd ./aqa-tests
16:56:42  + . ./scripts/testenv/testenvSettings.sh
16:56:42  + set +x
16:56:42  /home/jenkins/workspace/Grinder@tmp/durable-b016861e/script.sh.copy: 3: set: Illegal option -o pipefail
[Pipeline] sh
```
https://ci.adoptium.net/view/Test_grinder/job/Grinder/9890/console

Might be something related with #5290 ? @llxia

judovana:
This was fixed wrongly, and should be reopened. See; https://github.com/adoptium/aqa-tests/pull/5296#issuecomment-2102633008

PR #5296, opened 2024-05-08
remove set -eo pipefail

llxia:
fixes: https://github.com/adoptium/aqa-tests/issues/5295

### DIFF
--- a/scripts/testenv/testenvSettings.sh
+++ b/scripts/testenv/testenvSettings.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 set +x
-set -eo pipefail
 
 if [[ $USE_TESTENV_PROPERTIES == true ]]; then
     testenv_file="./testenv/testenv.properties"


llxia:
reran on the failed machine: https://ci.adoptium.net/view/Test_grinder/job/Grinder/9891/

LongyuZhang (approved):
LGTM

smlambert (approved):

judovana:
`set -eo pipefail` is quite crucial part of the script. Its removal may be breakig other platforms.

In addition, this PR is wrong. The message clearly says "set: Illegal option -o pipefail"  so this PR should have remove **only**  `-o pipefail`, and keep the `e` on. 

Seeing also the `#!/usr/bin/env bash` in the script, is is supposed to be bash, bash should know `-o pipefail`  unless it is ancient bash (or fake bash).

@llxia   - please update your systems. You have to be run exceptionally legacy configurations, and it is enforcing breaking changes  for everybody else.


In addition, the https://github.com/llxia/openjdk-tests/blob/cb86453541d2cedb78dab55e0dac63905c14ad6d/scripts/testenv/testenvSettings.sh do not have isngle *pipe*  so

```
- set -eo pipefail
+ set -e
``
would be very correct fix.  But

```
- set -eo pipefail
``
Thus removing whole -e is usualy very bad fix.  Please  revert if you can.

judovana:
Then at least return the `-e`

Issue #5302, opened 2024-05-09
create MiniMix_10m test


llxia:
copy `MiniMix_5m`,  rename it `MiniMix_10m` and change timeLimit=`10m`:
https://github.com/adoptium/aqa-tests/blob/83bec274a10eb52ac812af2acc50bb3ec4bc4673/system/otherLoadTest/playlist.xml#L34-L62

This is the first step of adding https://github.com/eclipse-openj9/openj9/issues/16428

coolujain:
I'll do this

Hajer-Dahmani:
I'll take this

PR #5337, opened 2024-05-20
Changed the test case name and the time limit

coolujain:
Fixes: #5302 
Signed-off-by: Lajeen <lhasna@andrew.cmu.edu>


llxia:
dup of https://github.com/adoptium/aqa-tests/pull/5338

PR #5349, opened 2024-05-27
Support nightly or release sbom api

### LINKED ISSUE #5348: Download sbom file for jdk21+ if sdk source is nightly/release
part of https://github.com/adoptium/temurin-build/issues/3683

sophia-guo:
Close #5348


sophia-guo:
Grinder https://ci.adoptium.net/view/Test_grinder/job/Grinder/10095/parameters/ with release, sbom is downloaded

smlambert (commented):
Any reason to limit this to JDK21+?  I know you are only using it right now for the reproducibility test (which is only supported for 21+), but SBOMs are produced for all versions 8+, and we may later want to add tests for other things?

smlambert (approved):
I guess we can change it later when new testcases which involve SBOMs are added for earlier versions (does not need to be a reproducibility test, could be tests for SBOM contents, etc).

karianna (approved):

PR #5359, opened 2024-05-30
Update MockitoTest's build.xml to use the same jdk version as to playlist.xml

sophiaxu0424:
- Issue to https://github.com/adoptium/infrastructure/issues/3043

### DIFF
--- a/functional/MockitoTests/build.xml
+++ b/functional/MockitoTests/build.xml
@@ -41,14 +41,17 @@ limitations under the License.
 		<echo>===executable:                   ${compiler.javac}</echo>
 		<echo>===debug:                        on</echo>
 		<echo>===destdir:                      ${DEST}</echo>
+		<echo>testing: sophia0 ${JDK_VERSION}</echo>
 		<if>
 			<not>
 				<matches string="${JDK_VERSION}" pattern="^(8|9|10)$$" />
 			</not>
 			<then>
 				<antcall target="clean" inheritall="true" />
+				<fail message="JDK 8, 9, or 10 is required to build this project."/>
 			</then>
 		</if>
+		<echo>testing: sophia1 ${JDK_VERSION}</echo>
 		<javac srcdir="${src}" destdir="${build}" debug="true" fork="true" executable="${compiler.javac}" includeAntRuntime="false" encoding="ISO-8859-1">
 			<src path="${src}" />
 			<classpath>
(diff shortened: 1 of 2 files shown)


sophiaxu0424:
Running with nightly jdk11, and it passed:
https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/40988/console

Still need to dive into the case with jdk8 for fixing the 55.0 version error in build.xml

sophiaxu0424:
Trying to testing on jdk8, but both failed with origin master branch, and this PR branch, expected SKIPPED, but keeps failing.
Testing with origin master branch, nightly jdk8, for expected: SKIPPED
https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/41000/ failed

Testing with this PR issue3043 branch, nightly jdk8 expected: SKIPPED
https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/41001/ failed

sophiaxu0424:
@llxia @LongyuZhang Would you please kindly check and see if other test should be prepared? TIA!

LongyuZhang (commented):
Since this PR is to fix JDK8 build failure, the grinder test needs to run with JDK8 other tests (reproduce what caused failure in the origin issue) to prove this change fixed it.

sophiaxu0424:
> Since this PR is to fix JDK8 build failure, the grinder test needs to run with JDK8 other tests (reproduce what caused failure in the origin issue) to prove this change fixed it.

Hi Longyu,

The original mockito test case with jdk11+ limitations, has compile failure and can't be skipped.
while when we remove the jdk11+ limitation, the reported issue ``` class file has wrong version 55.0, should be 52.0``` could be reproduced, and this reported issue came from the mockito (third-party) not our side, and we are waiting for them to fix.

I am wondering that maybe we should add this jdk8 to disabled list to avoid this compile failure? please advice as well. thanks!

LongyuZhang:
@sophiaxu0424 To validate this changes fixed the JDK8 problem, you can run your branch in a grinder with
- DYNAMIC_COMPILE set to unchecked;
- JDK is JDK 8
- Target is `floatSanityTests`
Then make sure this grinder passed, and the console contains info of either MockitoTest skipped or not shown at all. Thanks.

llxia (approved):
LGTM

LongyuZhang (approved):
LGTM

PR #5402, opened 2024-06-24
Exclude 2 ipv6 jdk8 tests on linux-all for Adoptium

Haroon-Khel:
Both tests are failing consistently on JDK8 linux platforms, https://github.com/adoptium/infrastructure/issues/3614.


sophia-guo (approved):

smlambert (approved):
Approving as we had done this for mac, but this is quite specific to our Adoptium infrastructure not having IPv6 enabled, so perhaps we should have these excludes in a [vendor/adoptium](https://github.com/adoptium/aqa-tests/tree/master/openjdk/excludes/vendors) excludes folder, instead of excluding testcases for all vendors.

Haroon-Khel:
> Approving as we had done this for mac, but this is quite specific to our Adoptium infrastructure not having IPv6 enabled, so perhaps we should have these excludes in a [vendor/adoptium](https://github.com/adoptium/aqa-tests/tree/master/openjdk/excludes/vendors) excludes folder, instead of excluding testcases for all vendors.

I didn't realise that was an option, it's a better idea. I've changed the pr accordingly

Haroon-Khel:
I dont have write access so I can't merge this despite the 2 approvals

sophia-guo:
@Haroon-Khel I didn't realize that you have updated to [vendor/adoptium](https://github.com/adoptium/aqa-tests/tree/master/openjdk/excludes/vendors) folder and merged. I don't think currently we support adoptium vendor yet

sophia-guo:
Checked TKG, for now If `System.getProperty('java.vendor')=Eclipse Adoptium` jdk_vendor is set as `eclipse`. Could we use /vendor/eclipse @smlambert ? If yes @Haroon-Khel we need to update https://github.com/adoptium/aqa-tests/blob/master/openjdk/openjdk.mk#L238 to include eclipse

PR #5447, opened 2024-07-16
Exclude TestUTCParse for JDK11 Linux for Adoptium only

Haroon-Khel:
ref https://github.com/adoptium/infrastructure/issues/3658#issuecomment-2220728689

Excluding this test just for Adoptium, unless others think it should be excluded for all vendors.

Will cherry pick this into the release branch once merged


Haroon-Khel:
Oops ive added it to the wrong group of tests. GIve me a sec to find its correct group

Haroon-Khel:
Now its in the correct group

smlambert (approved):

llxia (approved):

PR #5450, opened 2024-07-16
Add testcase exclusions for July

smlambert:
- Add Temurin specific exclude files (17+)
- For JDK21+ exclude AbsPathsInImage due to OOM on public Adoptium nodes
- For JDK11+ exclude SecureDS on linux-all as not supported on alpine-linux and jtreg currently does not differentiate alpine-linux from other linux platforms. Will plan to raise an issue to fix in upstream jtreg, so can have a more specific exclude later.
- VersionCheck exclude on JDK8/x64Linux
- TestHumongousCodeCacheRoots JDK8/aix exclusion


LongyuZhang (approved):
LGTM

llxia (approved):

PR #5467, opened 2024-07-31
Exclude DaaLoadTest special Mode351 on AIX

pshipton:
Issue https://github.com/eclipse-openj9/openj9/issues/19947


pshipton:
@llxia can I disable both a variation and a platform?

llxia:
Yes, you can, but platform should be `<platform>ppc64_aix</platform>`

pshipton:
There are too many failures to exclude, I ended up turning off special.system testing for AIX and zlinux until the regression is resolved.

Issue #3115, opened 2021-11-02
Multiple tests from jdk_jfr_2 fail on arm32 due to dump related issues


Haroon-Khel:
Many failed tests in jdk_jfr_2 in the extended openjdk suite on arm32

The list of failed tests are in https://ci.adoptopenjdk.net/job/Test_openjdk8_hs_extended.openjdk_arm_linux_testList_2/16/testReport/
There are 335 failed tests in total so wont be posting all of them here

All of the test failures have a similar error log

```
[GC (Allocation Failure) [DefNew: 4416K->512K(4928K), 0.0075890 secs] 4416K->1241K(15872K), 0.0077235 secs] [Times: user=0.00 sys=0.00, real=0.00 secs] 
[GC (Allocation Failure) [DefNew: 4928K->463K(4928K), 0.0157184 secs] 5657K->1697K(15872K), 0.0157842 secs] [Times: user=0.00 sys=0.00, real=0.02 secs] 
[GC (Allocation Failure) [DefNew: 4879K->344K(4928K), 0.0102744 secs] 6113K->2030K(15872K), 0.0105317 secs] [Times: user=0.02 sys=0.00, real=0.01 secs] 
[Full GC (System.gc()) [Tenured: 1685K->1798K(10944K), 0.0291196 secs] 6292K->1798K(15872K), [Metaspace: 4583K->4583K(5424K)], 0.0292266 secs] [Times: user=0.02 sys=0.00, real=0.03 secs] 
[thread -759499664 also had an error]
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGBUS (0x7) at pc=0xf5f48698, pid=423216, tid=0xd3038470
#
# JRE version: OpenJDK Runtime Environment (8.0_312-b07) (build 1.8.0_312-b07)
# Java VM: OpenJDK Client VM (25.312-b07 mixed mode linux-aarch32 )
# Problematic frame:
# V  [libjvm.so+0x333698]  write_checkpoint_header(unsigned char*, long long, long long, bool, unsigned int)+0xe8
#
# Core dump written. Default location: /home/jenkins/workspace/Test_openjdk8_hs_extended.openjdk_arm_linux_testList_2/aqa-tests/TKG/output_16358029827425/jdk_jfr_2/work/scratch/15/core or core.423216
#
# An error report file with more information is saved as:
[...]

PR #5469, opened 2024-07-31
Exclude jfr tests on aarch32 linux

roberttoyonaga:
JFR tests will fail on aarch32. We should exclude them to avoid false positives. See explanation here: https://github.com/adoptium/aqa-tests/issues/3115#issuecomment-2256561374

### DIFF
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -1476,6 +1476,10 @@
 		<testCaseName>jdk_jfr</testCaseName>
 		<disables>
 			<disable>
+				<comment>https://github.com/adoptium/aqa-tests/issues/3115#issuecomment-2256561374</comment>
+				<vendor>eclipse</vendor>
+			</disable>
+			<disable>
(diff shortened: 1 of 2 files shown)


karianna (approved):

openjdk/playlist.xml
```diff
@@ -1475,6 +1475,10 @@
 	<test>
 		<testCaseName>jdk_jfr</testCaseName>
 		<disables>
+			<disable>
+				<comment>https://github.com/adoptium/aqa-tests/issues/3115#issuecomment-2256561374</comment>
+				<vendor>eclipse</vendor>
```
smlambert, line 1480: needs a `<platform>` element to specify just disabling for a particular platform

sophia-guo:
Should this be limited to jdk8?  

```
	
			<version>8</version>
	
```

roberttoyonaga:
Updated to specify platform and JDK version. Thanks for the reviews!

sophia-guo (approved):

PR #5480, opened 2024-08-02
Exclude jdk/internal/vm/Continuation/OSRTest for OpenJ9 jdk21

pshipton:
Issue https://github.com/eclipse-openj9/openj9/issues/19961


llxia (approved):

JasonFengJ9 (approved):
LGTM

This test was excluded for JDK 23/24.
Assuming no need to do so for JDK 22 which is close to its EOL.

Issue #5428, opened 2024-07-10
Allow JRE testing for other testsuite aside from JCK


andrlos:
So basically what the problem is, is that we are unable to run JRE testing with anything but JCK, since JAVA and JAVAC command variables are only defined and propagated from the jck.mk...
if we want to expand JRE testing on other testsuites we should find a suitable place to move this declarations to.
This also targets the issue of headless testsuite https://github.com/adoptium/aqa-tests/issues/5224 , that was designed with jre testing in mind and is circumventing this in its own way. We should align this with the mechanics used elsewhere in Adoptium pipeline, in this case use the same mechanic for jre testing as used in JCK.

andrlos:
Hi, so here is a thought: within aqa-tests with have JenkinsfileBase groovy script that seems to be handling the basic test setup.. within this script there is a function called "setup_jck_interactives" I would split this function into setting the CUSTOMIZED_SDK_URL processing and setup of JAVA_TO_TEST that we would move from the jck/jck.mk to a new function inside the script and the other jck specific procedures that would be left inside the mentioned function. This would allow execution of the new function outside of the jck context.

andrlos:
for example setup_customized_sdk could be the name..

andrlos:
or better yet.. what would be the harm in running the "setup_jck_interactives" function in non-jck test? just renaming a few things there.. there is suspiciously small amount of code that needs to be jck specific other than name-wise in my opinion

smlambert:
We also support running from command line for those who do not use our Jenkins scripts, so the setup logic can not be specifically at that '2nd layer' of our [3 layer](https://github.com/adoptium/aqa-tests/blob/master/docs/pages/LayeredDesign.md) design.

The logic in jck.mk will be moved to a makefile that can then be included by all of the test group makefiles (jck.mk, openjdk.mk, system.mk, etc).  Setting USE_JRE=true will be sufficient to utilize the supplied SDK for compiling tests, but the supplied JRE for executing tests.

andrlos:
yea, but I thought that the goal was to not have to introduce a new mk file for functional testing... we would have to create a functional.mk (non-existent now AFAIK) to have a mk file to include it by.

andrlos:
https://github.com/adoptium/aqa-tests/pull/5490 so this is how you would see it smlambert

smlambert:
[...]

PR #5490, opened 2024-08-07
Issue 5428: Allow JRE testing for other testsuite aside from JCK

andrlos:
JAVA_TO_TEST definition moved to allow for JRE testing of other than just JCK testsuite

https://github.com/adoptium/aqa-tests/issues/5428


andrlos:
abandoned approach

Issue #5491, opened 2024-08-07
hotspot_custom is broken with jdk8


sophia-guo:
hotspot_custom is broken with jdk8 due to change of   JVM_NATIVE_OPTIONS in the playlist to `-nativepath:"$(TESTIMAGE_PATH)/hotspot/jtreg/native"` in https://github.com/adoptium/aqa-tests/pull/5370.  This change meant to use TARGET=disabled.hotspot_custom to run **openj9 custom serviceability jvmti test**  https://github.com/adoptium/aqa-tests/issues/5357 , which works by disabled.* target instead of hotspot_custom directly. I would prefer to separate a custom target to standardize the use of custom target.  https://github.com/adoptium/aqa-tests/issues/5357#issuecomment-2197503446  https://github.com/adoptium/aqa-tests/issues/5357#issuecomment-2197510570 

https://ci.adoptium.net/view/Test_grinder/job/Grinder/10733/

```
11:08:52  TESTING:
11:08:52  Error: The -nativepath path does not exist: /hotspot/jtreg/native
11:08:52  Usage:
```

@LongyuZhang @llxia

LongyuZhang:
Will separate the custom test for hotspot and j9 then.

PR #5493, opened 2024-08-07
Separate hotspot_custom test for j9

LongyuZhang:
Separate hotspot_custom and hotspot_custom_j9 tests

Issue: https://github.com/adoptium/aqa-tests/issues/5491

### DIFF
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -112,7 +112,7 @@
 			<variation>Mode1000</variation>
 		</variations>
 		<command>$(JAVA_COMMAND) -Xmx512m -jar $(Q)$(TEST_RESROOT)$(D)jtreg$(D)lib$(D)jtreg.jar$(Q) \
-	$(JTREG_BASIC_OPTIONS) -nativepath:"$(TESTIMAGE_PATH)/hotspot/jtreg/native" -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS) $(VMOPTION_HEADLESS)$(Q) \
+	$(JTREG_BASIC_OPTIONS) $(JVM_NATIVE_OPTIONS) -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS) $(VMOPTION_HEADLESS)$(Q) \
 	$(TIMEOUT_HANDLER) \
 	-w $(Q)$(REPORTDIR)$(D)work$(Q) \
 	-r $(Q)$(REPORTDIR)$(D)report$(Q) \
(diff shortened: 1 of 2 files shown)


openjdk/playlist.xml
```diff
@@ -125,6 +115,38 @@
...
+	</test>
+	<test>
+		<testCaseName>hotspot_custom_j9</testCaseName>
+		<variations>
+			<variation>Mode150</variation>
+			<variation>Mode650</variation>
+			<variation>Mode1000</variation>
+		</variations>
+		<command>$(JAVA_COMMAND) -Xmx512m -jar $(Q)$(TEST_RESROOT)$(D)jtreg$(D)lib$(D)jtreg.jar$(Q) \
+	$(JTREG_BASIC_OPTIONS) -nativepath:"$(TESTIMAGE_PATH)/hotspot/jtreg/native" -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS) $(VMOPTION_HEADLESS)$(Q) \
```
sophia-guo, line 130: This only works for jvmti tests. Would be better to change the testCaseName.

sophia-guo (dismissed):

openjdk/playlist.xml
llxia: Yes, we only support serviceability tests for now. But this provides the ability to test/run other tests if needed. As the name suggested `hotspot_custom_j9`, users have the option to choose what tests to run.

sophia-guo: For other tests which nativepath `-nativepath:"$(TESTIMAGE_PATH)/hotspot/jtreg/native"` or `-nativepath:"$(TESTIMAGE_PATH)$(D)openj9" ` work? I thought for openj9 serviceability tests is different as it need hotspot lib while others still need openj9 lib. I might be wrong, just to confirm.

sophia-guo: I can see that `[hotspot_native_sanity](https://github.com/adoptium/aqa-tests/blob/master/openjdk/playlist.xml#L424)` and `[ASGCTBaseTest_j9](https://github.com/adoptium/aqa-tests/blob/master/openjdk/playlist.xml#L389)` are using openj9 lib path nativepath:"$(TESTIMAGE_PATH)$(D)openj9". That is this new openj9 hotspot custom target might not work for those tests. I'm suggesting to restore the hotspot_custom as is and add this new jvmti custom for openj9.

LongyuZhang: This test is to offer openj9 the capability to test with hotspot native path, the `JvmtiGetSystemPropertyTest` is only an example to show it works, but we can also specify other hotspot native tests as the custom_target when needed.
  The `hotspot_native_sanity` and `ASGCTBaseTest_j9` is to test j9 native path, instead of the hotspot native path as this test. They serve different purposes.
  The `hotspot_custom` test is already restored for impl hotspot only in this PR. Thanks.

sophia-guo: My suggestion is restore hotspot_custom for hotspot and openj9 as before, so openj9 hotspot_native_sanity and ASGCTBaseTest_j9 can also use the hotspot_custom as nativepath has been correspondingly set to j9 native path.

This new hotspot_custom_j9 won't work with openj9 hotspot_native_sanity and ASGCTBaseTest_j9

LongyuZhang: The hotspot_custom can be restored for both hotspot and openj9, but it still needs to be disabled for j9, since no default custom target will pass all versions and platforms so far ( `hotspot_native_sanity` is disabled now, and `ASGCTBaseTest_j9` is xlinux and 16+ only). Shall we do it this way?

sophia-guo: yes, in that way hotspot_custom can be disabled for j9.

sophia-guo: For the naming issue with `hotspot_custom_j9` and `nativepath -nativepath:"$(TESTIMAGE_PATH)/hotspot/jtreg/native"`  ,  my question is if it works for hotspot_native_sanity and ASGCTBaseTest_j9 (those are also hotspot tests)? My answer is no.  I don't think this provide the ability to run other openjdk hotspot tests for openj9 if openj9 use openj9 native path. Are you going to add another target for openj9 to run hotspot tests using openj9 native libs?

llxia: I do not understand why we are so focused on hotspot_native_sanity and ASGCTBaseTest_j9? As Longyu mentioned, hotspot_custom_j9 is designed for tests that hotspot native lib, not for tests that use openj9 native lib. That is, it is not designed for hotspot_native_sanity and ASGCTBaseTest_j9.

btw, hotspot_native_sanity has not been enabled for openj9 since the beginning. ASGCTBaseTest_j9 runs a single java file, so it does not need to use custom target to run.

LongyuZhang: Updated the commit to restore `hotspot_custom` as suggested. Thanks.

smlambert (approved):
As per the discussion this afternoon, approving, you may want to update openj9 docs regarding the nuance of the use of the hotspot_custom_j9 target.

PR #5498, opened 2024-08-12
Propagate JAVA_TOOL_OPTIONS down to external tests in container

judovana:
Currently the external tests are swallowng the JAVA_TOOL_OPTIONS and thus the tested JDK does miss the global config. This pr should be fixing that

### DIFF
--- a/external/dockerfile_functions.sh
+++ b/external/dockerfile_functions.sh
@@ -197,11 +197,11 @@ print_jdk_install() {
             "\n\t PATH=\"/opt/java/openjdk/bin:\$PATH\" " \
             "\n" >> ${file}
 
-    echo -e "\nENV JAVA_TOOL_OPTIONS=\"-XX:+IgnoreUnrecognizedVMOptions -XX:+IdleTuningGcOnIdle\" " \
+    echo -e "\nENV JAVA_TOOL_OPTIONS=\"$JAVA_TOOL_OPTIONS -XX:+IgnoreUnrecognizedVMOptions -XX:+IdleTuningGcOnIdle\" " \
             "\n" >> ${file}
 
     echo -e "\nENV RANDFILE=/tmp/.rnd  \\" \
-            "\n\t OPENJ9_JAVA_OPTIONS=\"-XX:+IgnoreUnrecognizedVMOptions -XX:+IdleTuningGcOnIdle -Dosgi.checkConfiguration=false\" " \
+            "\n\t OPENJ9_JAVA_OPTIONS=\"$OPENJ9_JAVA_OPTIONS  -XX:+IgnoreUnrecognizedVMOptions -XX:+IdleTuningGcOnIdle -Dosgi.checkConfiguration=false\" " \
             "\n" >> ${file}
 
 }
@@ -445,7 +445,7 @@ print_maven_install() {
 print_java_tool_options() {
     local file=$1
 
-    echo -e "ENV JAVA_TOOL_OPTIONS=\"-Dfile.encoding=UTF8 -Djava.security.egd=file:/dev/./urandom\"\n" >> ${file}
+    echo -e "ENV JAVA_TOOL_OPTIONS=\"$JAVA_TOOL_OPTIONS -Dfile.encoding=UTF8 -Djava.security.egd=file:/dev/./urandom\"\n" >> ${file}
 }
 
 print_environment_variable() {
--- a/external/jenkins/test.sh
+++ b/external/jenkins/test.sh
@@ -16,7 +16,7 @@ source $(dirname "$0")/test_base_functions.sh
 # Set up Java to be used by the jenkins test
 echo_setup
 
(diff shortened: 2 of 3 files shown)


judovana:
https://ci.adoptium.net/view/Test_grinder/job/Grinder/10758/console 

looks good

external/dockerfile_functions.sh
```diff
@@ -197,11 +197,11 @@ print_jdk_install() {
             "\n\t PATH=\"/opt/java/openjdk/bin:\$PATH\" " \
             "\n" >> ${file}
 
-    echo -e "\nENV JAVA_TOOL_OPTIONS=\"-XX:+IgnoreUnrecognizedVMOptions -XX:+IdleTuningGcOnIdle\" " \
+    echo -e "\nENV JAVA_TOOL_OPTIONS=\"$JAVA_TOOL_OPTIONS -XX:+IgnoreUnrecognizedVMOptions -XX:+IdleTuningGcOnIdle\" " \
             "\n" >> ${file}
 
     echo -e "\nENV RANDFILE=/tmp/.rnd  \\" \
-            "\n\t OPENJ9_JAVA_OPTIONS=\"-XX:+IgnoreUnrecognizedVMOptions -XX:+IdleTuningGcOnIdle -Dosgi.checkConfiguration=false\" " \
+            "\n\t OPENJ9_JAVA_OPTIONS=\"$OPENJ9_JAVA_OPTIONS  -XX:+IgnoreUnrecognizedVMOptions -XX:+IdleTuningGcOnIdle -Dosgi.checkConfiguration=false\" " \
```
karianna, line 204:
```suggestion
            "\n\t OPENJ9_JAVA_OPTIONS=\"$OPENJ9_JAVA_OPTIONS -XX:+IgnoreUnrecognizedVMOptions -XX:+IdleTuningGcOnIdle -Dosgi.checkConfiguration=false\" " \
```

karianna (approved):

judovana:
hi! I will be now away for a week, and @pmikova and @jandrlik are waiting for this change. Unless the concept is fundamentaly wrong and needs full revisit, pleae merge it. We are using JAVA_TOOL_OPTIONS to test different runtiem variants..

sophia-guo:
I'm not sure why we need to pass the global `JAVA_TOOL_OPTIONS` of the host environment to test containers. No matter what the host environment is tests will only run in the containers. Tests specific settings is set up by Image file( external/thirdparty/test.properties) and shell script test.sh (external/thirdparty/test.sh). If tests need special export JAVA_TOOL_OPTIONS setting, it should be done by test.sh?

judovana:
JAVA_TOOL_OPTIONS is there to ensue, that you can orchestrate all your JVMs in same way. We use it eg to pass everywhere -XX+UseShenandoahGC and similar. BY not having it in contrainers means, that the JDK in container is tested differently then the jdsk on localhost. THis PR is removing this difference.

If you will disaprove this, then a different way how to pass global settings down through container is needed

judovana:
> If you will disaprove this, then a different way how to pass global settings down through container is needed


On top of that - whatever will be implement to workaround this, is simply reimplementing wheel. because this is exactly what JAVA_TOOL_OPTIONS is for, and unless there is real  strong reason to not forward it in, it should be considered as bug.

sophia-guo:
That means tests will run with  different JAVA_TOOL_OPTIONS on different environments but it's same for all other tests. Thanks @judovana

external/dockerfile_functions.sh
```diff
@@ -197,11 +197,11 @@ print_jdk_install() {
             "\n\t PATH=\"/opt/java/openjdk/bin:\$PATH\" " \
             "\n" >> ${file}
 
-    echo -e "\nENV JAVA_TOOL_OPTIONS=\"-XX:+IgnoreUnrecognizedVMOptions -XX:+IdleTuningGcOnIdle\" " \
+    echo -e "\nENV JAVA_TOOL_OPTIONS=\"$JAVA_TOOL_OPTIONS -XX:+IgnoreUnrecognizedVMOptions -XX:+IdleTuningGcOnIdle\" " \
             "\n" >> ${file}
 
     echo -e "\nENV RANDFILE=/tmp/.rnd  \\" \
-            "\n\t OPENJ9_JAVA_OPTIONS=\"-XX:+IgnoreUnrecognizedVMOptions -XX:+IdleTuningGcOnIdle -Dosgi.checkConfiguration=false\" " \
+            "\n\t OPENJ9_JAVA_OPTIONS=\"$OPENJ9_JAVA_OPTIONS -XX:+IgnoreUnrecognizedVMOptions -XX:+IdleTuningGcOnIdle -Dosgi.checkConfiguration=false\" " \
```
sophia-guo, line 204: Should this be $JAVA_TOOL_OPTIONS too?

sophia-guo: Merge now if need can be updated later.

PR #5521, opened 2024-08-22
Exclude jtreg test sun/security/pkcs11/fips/TestTLS12.java in jdk8u

sendaoYan:
Hi all,
We have not configure NSS lib in ours test environments for now, so the test `sun/security/pkcs11/fips/TestTLS12.java` will failed in jdk8u. In jdk11u, this test will skip automatically. So before we configure NSS lib in ours test environments, we need to exclude this testcase in jdk8u.
Change only happen in alibaba excludes files, no risk.


smlambert (approved):

llxia (approved):

sendaoYan:
Thanks all for the review.

Issue #5510, opened 2024-08-16
Improve AQA_Test_Pipeline for release testing


llxia:
Due to Semeru release process change, we do not use Semeru pipeline to launch testing. AQA_Test_Pipeline is used for launching release testing. As we have more variations (regular + FIPS140-2 + FIPS140-3 weak profile + FIPS140-3 strict profile), each one comes with its own unique twist (i.e., slightly different jdk versions, test targets, test flags, vender repos, platforms, etc). Currently, 20 separate AQA_Test_Pipeline builds (5 JDK versions x 4 variations) have to be launched manually. If we have multiple Jenkins SDK build links, the number needs to be multiplied accordingly.

To streamline the release testing process and reduce errors, we will enhance the AQA_Test_Pipeline for release testing.
A JSON config will be defined to store parameters that are needed to launch the release testing.

In the initial version, we will support launching release testing based on a JSON configuration. This will later be enhanced to allow customized values alongside the remaining values in the JSON files.

llxia:
Changes are delivered.

PR #5535, opened 2024-08-29
Support custom config json file

llxia:
related: https://github.com/adoptium/aqa-tests/issues/5510

### DIFF
--- a/buildenv/jenkins/aqaTestPipeline.groovy
+++ b/buildenv/jenkins/aqaTestPipeline.groovy
@@ -37,15 +37,21 @@ fail = false
 JDK_VERSIONS.each { JDK_VERSION ->
     if (params.PLATFORMS == "release") {
         def configJson = []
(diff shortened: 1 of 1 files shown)


llxia:
- [Internal run](https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/AQA_Test_Pipeline_Release/86/)
- [Internal run](https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/AQA_Test_Pipeline_Release/87/)

LongyuZhang (approved):
LGTM

karianna (commented):
Are we also able to do an external run to test for regression?

llxia:
This does not affect external runs. The feature is only enabled if the special keyword is set.
If the external run wants this feature, we can address it in a separate issue.

karianna (approved):

PR #3363, opened 2022-02-17
Update external tests to use different Docker images

jnjerin:
External tests have been updated to use different Docker images.
Now using **eclipse-temurin** for hotspot and **ibm-semeru-runtimes** for openj9

Fixes #3079
Fixes #3080  

Signed-off-by: jnjerin <njerijoan24@gmail.com>

PR #5553, opened 2024-09-04
Global image for external

### LINKED ISSUE #5555: provide a way to allow to run external tests on any container
Currently the runtime of containers supports some kind of lets say "default" image, which is temurin-jdk/jre image (so a custom jdk passed in  have all deps), and that is based on ubuntu.  As an option on openj9 images have several hardcoded occurences and are again based on ubuntu. In addition there is hardcoded version of  openj9 on Red Hat UBI images - https://github.com/adoptium/aqa-tests/blob/2ee9988dedef51534a8afdb15827167e389443b6/external/dockerfile_functions.sh#L87-L105 .  There are then several occurences of hardcoded ubi/ubuntu:, eg:https://github.com/adoptium/aqa-tests/blob/2ee9988dedef51534a8afdb15827167e389443b6/external/dockerfile_functions.sh#L617 (note also the `!package` expansion).  Note, that the resulting image is *jsut*  `ARG` in  rsulting dockerfile. eg:
```
ARG IMAGE=docker.io/library/eclipse-temurin:21-jdk
ARG OS=ubuntu
ARG IMAGE_VERSION=nightly
ARG TAG=21-jdk
FROM $IMAGE
```
There are two ways how to run the tests:
 * Through https://github.com/adoptium/aqa-tests/blob/2ee9988dedef51534a8afdb15827167e389443b6/external/build_all.sh which seems to be only building images and TBH I'm not sure what it is for in wider scope
 * through more usual `TEST_JDK_HOME`, `BUILD_LIST` and `make compile` which builds image and `make _testname`  which runs

The initial change should affect at least the second. To affect the build_all.sh on top of that, should be refactoring, which would remove all the `for test in ${supported_tests} do for vm in ${supported_jvms} do for os in ${supported_os} do for package in ${supported_packages} do for build in ${supported_builds} do build_image.sh ${test} ${version} ${vm} ${os} ${package} ${build}`  to actually do all the looping through `${supported_jvms}, ${supported_os}, ${supported_packages},  ${supported_builds}`  first, to prepare set of (fully) qualified container IDs and then loop through them, via shortened `for test in ${supported_tests} do build_image.sh ${test} ${image}` I would probably like to addres it in different issue and different purpose as usage of build-all.sh is unclear to me.

Second part of this work is to properly adjust dependencies. Currently the test.properties have hardcoded `ubuntu_packages` and/or `ubi_packages`. This  is to clumsy. A better way how to handle generic dependencies, and per os-family dependencies  down to os dependencies and os:version dependencies must be done. It would be also nice to have some generic deps whcih are installed always, so unzip and curl are not repeated in each test.properties. Hoefully this is aligned also with one of the future goals, to have ability to run external tests on local machine without container.
The second part may sound separated, but is not. If it should go  alone, it would need to go before all other, and without all other it would be change just for itself.

judovana:
This PR is enabling any possible image to be fed to external tests.
User can set it via `EXTERNAL_AQA_IMAGE` variable. Eg `EXTERNAL_AQA_IMAGE=my.local.repo/on/hdd/fedora:40`. Any legal  subset -eg just `fedora` is allowed.
The properties model was  quite enhanced to work properly in multi distroenviroenment. This may be also good base for future localhost runs. I had tested all features on default ubuntu based temurin and on fedora:39. To tune it for centos/rhel will be quite a fun. But full tuning should not be part of this PR.

The build_all.sh is not affected by this change.

this is currently draft, becasue for some reasonnot yet found by me, the mounted jdk in fedora ... is not here. The default works. I wil continue to work on this, but any feedback appreciated.

Close #5555


sophia-guo:
@judovana thanks for the work on this PR. Those enhancements are nice. They are addressing two different issues would it be nice to separate the PR into two ( one for global image, one for common property)? Also as we talked recently to keep our workflow organized and ensure each change is properly  tracked and reviewed, could you please open two issues and reference them in the respective PRs?

judovana:
The two parts can not go separtely in parallel. Maybe the properties part can go in first, but the geenric images depends on it. I would rather keep it togehther. They do not have sense one by one. (the btohering with first, if the second will not go in or if it will need major changes is not worthy) 

Sure, I will create an issue. Thanx a lot for reminder.

judovana:
Few external tests passes on few jdks on few images.... I will start to roll up some test matrix on this but tbh not sure how it is complete... Many external tests needs forking for newer jdks, also tuning up the depndencies will be fun.... 

What is required to double check to make this pass through?

external/dockerfile_functions.sh
```diff
@@ -84,34 +84,50 @@ print_image_args() {
...
         else
+            if isExternalImageEnabled ; then
+              echo "openj9 ubi and custom EXTERNAL_AQA_IMAGE are not compatible"
+              exit 1
+            fi
             # os is ubi, and test is criu
             # temporarily all ubi based testing use internal base image
             image_name="$DOCKER_REGISTRY_URL/$base_docker_registry_dir"
             tag="latest"
+            EXTERNAL_AQA_IMAGE="${image_name}:${tag}"
```
sophia-guo, line 110: If EXTERNAL_AQA_IMAGE needs to be set? If yes, should move to line 114 for both temurin and openj9 ?

sophia-guo:
> What is required to double check to make this pass through?

Based on the changes, for  temurin default image I would suggest to run a build with tomcat tests. The test job doesn't need to pass. It should be good as long as the generated Dockerfile looks good. Running locally it should be good with default or EXTERNAL_AQA_IMAGE=my.local.repo/on/hdd/fedora:40. For openj9 external tests you may need @LongyuZhang help to run a few grinder to ensure it works for openj9.

judovana:
> > What is required to double check to make this pass through?
> 
> Based on the changes, for temurin default image I would suggest to run a build with tomcat tests. The test job doesn't need to pass. It should be good as long as the generated Dockerfile looks good. Running locally it should be good with both default and EXTERNAL_AQA_IMAGE=my.local.repo/on/hdd/fedora:40. For openj9 external tests you may need @LongyuZhang help to run a few grinder to ensure it works for openj9.

The default image(s) were not changed, and generated docekrfile remains identical!-) tat was moreover primary goal during this PR. The only chnage which is affecting the defaults is the search in /opt first, but it did not have anmy siode effect as far as I was able to check.  I will try tomcat.

external/dockerfile_functions.sh
judovana: To construct the EXTERNAL_AQA_IMAGE artificially and globaly, is quite a good idea, and I was playing with it a lot.  The reason why I have not included it, are non deterministic OSes.  Ubuntu was hardcoded, to be there for jdk-xy-temurin/openj9, because there is no record of ubuntu in name.  From my limited knowledge, this is the only case where it is actually missing. Eg redhat java images still have ubi in name, and so on.

If there will be more usage of EXTERNAL_AQA_IMAGE then current  baseurl, path, name, os variables, then I guess such a crossroad table would be necessary, but for now I decided to not implement it.

judovana:
jacoco jdk21: fails coorectly (needs different jacaoco sources: https://ci.adoptium.net/view/Test_grinder/job/Grinder/10846/console
```
15:36:05  [ERROR] Source option 7 is no longer supported. Use 8 or later.
15:36:05  [ERROR] Target option 7 is no longer supported. Use 8 or later.
```  
(todo, fix??)
jacoco jdk17:  https://ci.adoptium.net/view/Test_grinder/job/Grinder/10842/tapResults/
jacoco jdk11:  https://ci.adoptium.net/view/Test_grinder/job/Grinder/10845/tapResults/
jacoco jdk8:  https://ci.adoptium.net/view/Test_grinder/job/Grinder/10844/tapResults/

judovana:
> jacoco jdk21: fails coorectly (needs different jacaoco sources: https://ci.adoptium.net/view/Test_grinder/job/Grinder/10846/console
> 
> ```
> 15:36:05  [ERROR] Source option 7 is no longer supported. Use 8 or later.
> 15:36:05  [ERROR] Target option 7 is no longer supported. Use 8 or later.
> ```
> 
> (todo, fix??)

https://github.com/adoptium/aqa-tests/issues/5575

LongyuZhang (commented):
Tests for CRIU failed, looks like the container os and package install was not aligned after the change. It tries to use `apt-get install` inside `ubi` image (hyc_grinder_43087).
```
16:44:37       [exec] STEP 1/23: FROM sys-rt-docker-local/ubi8-with-criu/linux_x86-64-ubi8-criu:latest
16:44:37       [exec] STEP 2/23: ENV RESULT_COMMENT="IN CONTAINER(not-as-root/podman)"
16:44:37       [exec] --> fb4b96608433
16:44:37       [exec] STEP 3/23: ARG CRIU-UBI-PORTABLE-CHECKPOINT_TAG=
16:44:37       [exec] --> b2cd85d4a861
16:44:37       [exec] STEP 4/23: RUN apt-get update 	&& apt-get install -qq -y --no-install-recommends software-properties-common 	&& apt-get install -qq -y --no-install-recommends gnupg 	&& add-apt-repository ppa:ubuntu-toolchain-r/test 	&& apt-get update 	&& apt-get install -y --no-install-recommends git wget unzip tar curl   	&& rm -rf /var/lib/apt/lists/*
16:44:39       [exec] /bin/sh: apt-get: command not found
16:44:39       [exec] Error: building at STEP "RUN apt-get update 	&& apt-get install -qq -y --no-install-recommends software-properties-common 	&& apt-get install -qq -y --no-install-recommends gnupg 	&& add-apt-repository ppa:ubuntu-toolchain-r/test 	&& apt-get update 	&& apt-get install -y --no-install-recommends git wget unzip tar curl   	&& rm -rf /var/lib/apt/lists/*": while running runtime: exit status 127
16:44:39  
16:44:39  BUILD FAILED
``

**Comment by judovana** [comment]
> Tests for CRIU failed, looks like the container os and package install was not aligned after the change. It tries to use `apt-get install` inside `ubi` image (hyc_grinder_43087).
> 
> ```
> 16:44:37       [exec] STEP 1/23: FROM sys-rt-docker-local/ubi8-with-criu/linux_x86-64-ubi8-criu:latest
> 16:44:37       [exec] STEP 2/23: ENV RESULT_COMMENT="IN CONTAINER(not-as-root/podman)"
> 16:44:37       [exec] --> fb4b96608433
> 16:44:37       [exec] STEP 3/23: ARG CRIU-UBI-PORTABLE-CHECKPOINT_TAG=
> 16:44:37       [exec] --> b2cd85d4a861
> 16:44:37       [exec] STEP 4/23: RUN apt-get update 	&& apt-get install -qq -y --no-install-recommends software-properties-common 	&& apt-get install -qq -y --no-install-recommends gnupg 	&& add-apt-repository ppa:ubuntu-toolchain-r/test 	&& apt-get update 	&& apt-get install -y --no-install-recommends git wget unzip tar curl   	&& rm -rf /var/lib/apt/lists/*
> 16:44:39       [exec] /bin/sh: apt-get: command not found
> 16:44:39       [exec] Error: building at STEP "RUN apt-get update 	&& apt-get install -qq -y --no-install-recommends software-properties-common 	&& apt-get install -qq -y --no-install-recommends gnupg 	&& add-apt-repository ppa:ubuntu-toolchain-r/test 	&& apt-get update 	&& apt-get install -y --no-install-recommends git wget unzip tar curl   	&& rm -rf /var/lib/apt/lists/*": while running runtime: exit status 127
> 16:44:39  
> 16:44:39  BUILD FAILED
> ``
> ```

Thanx! Will elaborate!   If I do not  resolve this  until your timezone,  can you please provide mre details how to reproduce it locally?

**Comment by judovana** [comment]
@LongyuZhang   I'm trying something like:
```
sh build_image.sh zookeeper 21 openj9 ubi jdk full linux_x86-64 default 0 
```

And you seems to be right, it is `ARG IMAGE=docker.io/library/eclipse-temurin:openj9-full` and then of course the `apt` is called.   Do you also see wrong `ARG_IMAGE`  pelase?

**Comment by judovana** [comment]
it seems the parameters are wrongly passed:
```
+ print_image_args /home/jvanek/git/aqa-tests/external/criu-functional/dockerfile//jdk/ubi/Dockerfile.openj9.full criu-functional jdk openj9 ubi full linux_x86-64 0
+ local file=/home/jvanek/git/aqa-tests/external/criu-functional/dockerfile//jdk/ubi/Dockerfile.openj9.full
+ local test=criu-functional
+ local os=jdk
+ local version=openj9
+ local vm=ubi
+ local package=full
+ local build=linux_x86-64
+ local base_docker_registry_dir=0
```

**Comment by judovana** [comment]
Nope, that was correct. Already call to `build_image.sh ` is wrong:
```
+ generate_dockerfile /home/jvanek/git/aqa-tests/external/criu-functional/dockerfile//jdk/ubi/Dockerfile.openj9.full criu-functional openj9 ubi jdk full linux_x86-64 default 0
```
You can see, the version of jdk is missing, abnd that is shifting the args:
https://github.com/adoptium/aqa-tests/blob/2ee9988dedef51534a8afdb15827167e389443b6/external/build_all.sh#L41

It seems, that: https://github.com/adoptium/aqa-tests/pull/1601/files  was originally traversing also through versions, but that was reemoved, and the removal was incorrect

**Comment by judovana** [comment]
@LongyuZhang  now it behaves correctly. May you try please?

**Comment by LongyuZhang** [comment]
Tested with the updated commits, the `ubi` with `apt-get` issue got fixed. But got another failure:
```
13:09:20  /aqa-tests/TKG/scripts/getTestenvProperties.sh --repo_dir /aqa-tests/openj9 --output_file /aqa-tests/testenv/testenv.properties --repo_name OPENJ9
13:09:20  Check sha in /aqa-tests/openj9 and store the info in /aqa-tests/testenv/testenv.properties
13:09:21  OPENJ9_REPO=https://github.com/eclipse-openj9/openj9.git
13:09:21  OPENJ9_SHA=216f34562d5b1a7253f8dbb33773c5bb8adfde27
13:09:21  Generating criu checkpoint
13:09:21  /test.sh: line 42: make: command not found
13:09:22  -----------------------------------
13:09:22  criu-ubi-portable-checkpoint_test_0_FAILED
```
I suspect it is because the `make` install on `ubi` got affected with the changes somewhere. 
We don't run criu test locally since it requires special host machine settings done by our infra team, so we mainly use internal jenkins for it.

**Comment by judovana** [comment]
> Tested with the updated commits, the `ubi` with `apt-get` issue got fixed. But got another failure (hyc_grinder_43142):
> 
> ```
> 13:09:20  /aqa-tests/TKG/scripts/getTestenvProperties.sh --repo_dir /aqa-tests/openj9 --output_file /aqa-tests/testenv/testenv.properties --repo_name OPENJ9
> 13:09:20  Check sha in /aqa-tests/openj9 and store the info in /aqa-tests/testenv/testenv.properties
> 13:09:21  OPENJ9_REPO=https://github.com/eclipse-openj9/openj9.git
> 13:09:21  OPENJ9_SHA=216f34562d5b1a7253f8dbb33773c5bb8adfde27
> 13:09:21  Generating criu checkpoint
> 13:09:21  /test.sh: line 42: make: command not found
> 13:09:22  -----------------------------------
> 13:09:22  criu-ubi-portable-checkpoint_test_0_FAILED
> ```
> 
> I suspect it is because the `make` install on `ubi` got affected with the changes somewhere. We don't run criu test locally since it requires special host machine settings done by our infra team, so we mainly use internal jenkins for it.

Hello! Thanx a lot! Which test (set of tests) is this?

https://github.com/adoptium/aqa-tests/pull/5553/commits/5f8891585f459882e8f8a9c7eea3c43771f1b2aa#diff-87fdb6d51b2e87bfa2c194fd3e8792a626657904ae658445c0739d27abece3f3   do not seem to be messing up with `make`. What is more important, I do not see `make` as depndence in any  criu tests, nor in generic packages in this PR, nor  *before* this PR.  So very likely I have overlooked some logic

```
cat   external/*criu*/test.properties | grep make
#nothing
```

However it is really obvious that criu tests are using make:
```
external/criu-functional/test.sh:make $1
external/criu-portable-checkpoint/test.sh:make $1
external/criu-ubi-portable-checkpoint/test.sh:make $1
```

Do you know wha(why it) was pulling it before? Does it now still work with ubuntu?

I had added the missing depndence. if you can try, it wouodl be awesome. TYVM!

**Comment by LongyuZhang** [comment]
The failed test is `criu-ubi-portable-checkpoint_test`. Rerun grinder to compare, results are shown in attached figure (left is master, right is this PR). It shows two problems here:
1. Originally the package for this test is `git wget perl`, and `make` installed automatically as a dependent. The `perl` got lost after this PR change.
2. This PR change added `unzip tar curl` which is not needed for this test. Considering each criu test runs on JDK_verisions&platforms&machine_os combos 90+ times for each build, I think the newly added `generic_packages` in this PR will not make our external tests more efficient. Take this case as an example, If one package in `generic_packages` is not need by a single CRIU test, it will take longer test time. @ShelleyLambert @llxia any suggestions? Thanks.
Screenshot 2024-09-18 at 23 42 55

**Comment by judovana** [comment]
Thanx a lot for heads up!

> The failed test is `criu-ubi-portable-checkpoint_test`. Rerun grinder to compare, results are shown in attached figure (left is master, right is this PR). It shows two problems here:
> 
> 1. Originally the package for this test is `git wget perl`, and `make` installed automatically as a dependent package. Though we can add `make` in the list, the `perl` got lost after this PR change for some reason.


Both perl and make are now installed - https://github.com/adoptium/aqa-tests/pull/5553/commits/cecf32367af67f04556ea093ee4f0485d7913b54#diff-87fdb6d51b2e87bfa2c194fd3e8792a626657904ae658445c0739d27abece3f3 - so it is fixed and runs, ok?
> 
>2. This PR change added `unzip tar curl` which is not needed for this test. Considering each criu test runs on JDK_verisions&platforms&machine_os combos 90+ times for each build, I think the newly added `generic_packages` in this PR will not make our external tests more efficient. Take this case as an example, If one package in `generic_packages` is not need by a single CRIU test, it will take longer test time and increase test failure risk with extra packages. @ShelleyLambert @llxia any suggestions? Thanks.
> 
> 

I'm in absolutel same  situation. I'm running many tests in combinations and on many versions, s I know what you mean. And I agree that soemtimes very small change one place when may lead to huge impact in such multiplied run.

I quite disagree that installation of tar or curl may lower stability of tests.

However I have absolutly no strong prefference. If you, @sophia-guo  or well anybody in the team have issues with that, I will happily remove the https://github.com/adoptium/aqa-tests/pull/5553/commits/5f8891585f459882e8f8a9c7eea3c43771f1b2aa#diff-3e2f7b162bce7417355d4fe0f18e6366b69e6abe5e73356f863274d3a4f9c04b and will copypaste those to tests  where they originallyc come from.

I think it is quite a good idea to have some shared known enviroenment with some shared known  deps, but I'm also well aware that such feuture may  bloat. 

Thanx again!

**Comment by sophia-guo** [comment]
I'm not sure how long it takes for the utils installation( each util may take different time). It may be worth comparing with the whole test process time  to see if it costs. Considering tests are usually running parallel in building farm we might not need to take account in our combos. Speaking of efficiency, we might also need to think about the maintainability, flexibility and scalability.

Regarding stability, adding unnecessary installations could affect test due to network problem or server being down, which make test false negative.

I would prefer at least split this PR to two different ones as I mentioned before. Open an separate issue to discuss the generic_packages one.

**Comment by judovana** [comment]
Sure! Fair enough! I will keep the logic in, but will move the `git wget unzip tar curl` back to original owners.

**Comment by judovana** [comment]
ok, the shared deps dropped. I hope I had not missed any. If so, that it was not intentional. Will take second eye after some timeout. Looking to that commit (https://github.com/adoptium/aqa-tests/pull/5553/commits/4c6217e210b5f97e13cdae37cdf20e036a05aa2b) I really dont like that :)

**Comment by judovana** [comment]
> ok, the shared deps dropped. I hope I had not missed any. If so, that it was not intentional. Will take second eye after some timeout. Looking to that commit ([4c6217e](https://github.com/adoptium/aqa-tests/commit/4c6217e210b5f97e13cdae37cdf20e036a05aa2b)) I really dont like that :)

ok, make double check and found miss in wycheproof and missed scala. Now it looks correct. Added few grinders:
https://ci.adoptium.net/view/Test_grinder/job/Grinder/10966/console
https://ci.adoptium.net/view/Test_grinder/job/Grinder/10967/console
https://ci.adoptium.net/view/Test_grinder/job/Grinder/10968/console
https://ci.adoptium.net/view/Test_grinder/job/Grinder/10969/console

All looks good

File: external/functional-test/test.properties [hunk 2]
```diff
@@ -2,4 +2,6 @@ github_url="https://github.com/adoptium/aqa-tests.git"
 test_results="testResults"
 gradle_version="5.1"
 environment_variable="MODE=java CC=gcc-7 CXX=g++-7"
-ubuntu_packages="ant-contrib build-essential git asciidoc"
\ No newline at end of file
+generic_packages="git"
+ubuntu_packages="ant-contrib build-essential asciidoc"
+fedora_packages="ant-contrib asciidoc"
```
sophia-guo, line 7: I'm confused what is the point of change property of ubuntu_packages to generic_package and ubuntu_packages? And what is the reason that some tests add fedora_packages and some not? I had thought if we were going to split the issue into supporting global image and discussion of enhancement of generic_package all test.properties would leave as it. Did I miss something?

LongyuZhang: In addition, the change from `ubi_packages` to `(fedora|ubi|rhel|centos)_packages` didn't work, see https://github.com/adoptium/aqa-tests/pull/5553#issuecomment-2359934873 which is already using SHA `[cecf323](https://github.com/adoptium/aqa-tests/pull/5553/commits/cecf32367af67f04556ea093ee4f0485d7913b54)` that contains `perl make` but `make` was not picked up.
Tested again, no packages picked up with current commit ` [exec] STEP 4/24: RUN dnf install -y    	&& dnf clean all` (hyc_grinder 43553). 
Do we currently have a case using `rhel`, `centos`, or `fedora` as the container OS? AFAIK, `ubi` is the container OS for `rhel`, so our customer only uses `ubi` as their supported container OS in their application. If not, I think we can keep only `ubi` for now.

judovana: Fedora do not have `build-essential` so the install would fail.
On contrary `git` is of same name everywhere, so it to not have sense to type it twice. Does it make more sense?

judovana: Per request I had removed usage of completely shared gnereic deps (https://github.com/adoptium/aqa-tests/pull/5553/commits/82b1c78fd18125db12037da1d4f55270cc62d717), which were declared once, and used everywhere (in all external tests). So now also the tar, wget and few more, which are copypasted in most of the test.properties, remains copypasted

judovana: > In addition, the change from `ubi_packages` to `(fedora|ubi|rhel|centos)_packages` didn't work, see [#5553 (comment)](https://github.com/adoptium/aqa-tests/pull/5553#issuecomment-2359934873) which is already using SHA `[cecf323](https://github.com/adoptium/aqa-tests/pull/5553/commits/cecf32367af67f04556ea093ee4f0485d7913b54)` that contains `perl make` but `make` was not picked up. Tested again, no packages picked up with current commit ` [exec] STEP 4/24: RUN dnf install -y && dnf clean all` (hyc_grinder 43553). Do we currently have a case using `rhel`, `centos`, or `fedora` as the container OS? AFAIK, `ubi` is the container OS for `rhel`, so our customer only uses `ubi` as their supported container OS in their application. If not, I think we can keep only `ubi` for now.

Yes, I'm trying to run them all.  I will verygladly  elaborate on  the ubi not picked by you. Can you please share `commandline` how to hit it?

judovana: > > In addition, the change from `ubi_packages` to `(fedora|ubi|rhel|centos)_packages` didn't work, see [#5553 (comment)](https://github.com/adoptium/aqa-tests/pull/5553#issuecomment-2359934873) which is already using SHA `[cecf323](https://github.com/adoptium/aqa-tests/pull/5553/commits/cecf32367af67f04556ea093ee4f0485d7913b54)` that contains `perl make` but `make` was not picked up. Tested again, no packages picked up with current commit ` [exec] STEP 4/24: RUN dnf install -y && dnf clean all` (hyc_grinder 43553). Do we currently have a case using `rhel`, `centos`, or `fedora` as the container OS? AFAIK, `ubi` is the container OS for `rhel`, so our customer only uses `ubi` as their supported container OS in their application. If not, I think we can keep only `ubi` for now.
> 
> 

[LongyuZhang](https://github.com/LongyuZhang) Yes, I'm trying to run them all. I will verygladly elaborate on the ubi not picked by you. Can you please share `commandline` how to hit it?

judovana: #[LongyuZhang](https://github.com/LongyuZhang)  I cannot acces syour logs, sorry:( I tried with crystall ball, and on "my" ubi images, and it worked fine. So we have to have different setup. Please provide mas much as you can. I will really galdly fix it. This PR is of most uttermost importance for me.

judovana: @LongyuZhang  ^^ (sorry, I messed up need-info character)

external/dockerfile_functions.sh
```diff
@@ -148,8 +162,6 @@ print_ubuntu_pkg() {
 # Select the ubuntu OS packages
 print_ubi_pkg() {
     local file=$1
-    local packages=$2
-
     echo -e "RUN dnf install -y ${packages} \\" \
```
LongyuZhang, line 165: The variable `packages` was removed, but still used here. This might be the reason why `RUN dnf install -y    	&& dnf clean all` cannot catch any packages.
If some core functions in this PR are needed in urgent, as Sophia suggested, we can extract this PR to only address the core functions, then work on non-core functions, e.g. `generice_packages` in other PRs, combining them all in this PR makes it easy to break current build. Thanks.

judovana:
The pacakges were moved to global scope. As you could have noticed, all - except your usecase - is passing.
Those two changes can not be separated - In the moment I'm want to support generic containers - thus at least fedora, rhel, centos and ubuntu, I need to specify packages for each of them, becasue the names and sets of packages are different.  Hand in hand with it,  I do not want to duplicate declarations of most common packages.

Hardcoded ubis are most terrible to adjust to the generic mages, because they were `if-else-hardcoded` to original code without much thinking.  @LongyuZhang  please provide reproducers  to your case. I would really love to fix it. It seems it is the only  failing one.

judovana:
When I'm in `maste` the test you are pointing out dies:
```
export BUILD_LIST=external/criu-ubi-portable-checkpoint ; export EXTRA_DOCKER_ARGS="-v $TEST_JDK_HOME:/opt/java/openjdk" ;  export TEST_JDK_HOME=/usr/lib/jvm/java-11-openjdk-11.0.24.0.8-2.portable.jdk.el.x86_64/
make compile
...
    [exec] DOCKERIMAGE_TAG nightly has been recognized.
     [exec] No credential available for container registry, will proceed without login...
     [exec] sudo podman pull ${env.DOCKER_REGISTRY_URL}/ubi8-with-criu/linux_x86-64-ubi8-criu:latest
     [exec] Error: parsing reference "${env.DOCKER_REGISTRY_URL}/ubi8-with-criu/linux_x86-64-ubi8-criu:latest": invalid reference format
```

When it is in `this pr` then it dies absolutely the same.

judovana:
From that I'm guessing that the issue will be in `function getMatchingPackagesKeys() {` and/or `function getImageOs() {` , because I'm reading OS here and it seems the os is `linux_x86-64-ubi8-criu`. The packages declaration is `(fedora|ubi|rhel|centos)_packages="git wget perl make"`  and if anything would go ok, the `.*(fedora|ubi|rhel|centos).*` matches  `linux_x86-64-ubi8-criu` so something went wrong in this particular usecase, where hardocded UBIs are present. Msot likely the EXTERNAL_AQA_IMAGE, which is workaround for that `if-ubi-if-critu`  unhappy blocks, was not set in time. But withotu reproducer od full verbose log, I can just guess


Actually the tes which you are unhappy with, have all packages installable on all tried OSes. I can mititgate now the issue by redeclaring it, but it is jsut moving the problem to `criu-portable-checkpoint` or to  improbable  future when criu will be tested on system with different package names

judovana:
the
```
 sh build_image.sh  criu-functional 17 openj9 ubi jdk full linux_x86-64 linux_x86-64-ubi8-criu  0 

```

correctly, although I dont know if it is at least close to reperoducer

LongyuZhang:
Re: https://github.com/adoptium/aqa-tests/pull/5553#issuecomment-2380434107
With the new update, the test passed (hyc_grinder 43691).

Since the CRIU tests have passed, I will leave this PR for @sophia-guo @smlambert  to check on their side then. Thanks.

judovana:
Thanx! I'm happy reading from crystal ball was successful!-) I 'm still very curious how to reproduce it locally, because I'm afraid hiding it behind "generic_packages" the issu is hidden. As the criu tests are heavily bounded to single container, the real issue maynever popup again, but still:(

@LongyuZhang  if you please can share, how can i run that particular tast, I t would be highly appreciated. TYVM!!

LongyuZhang:
The reason we set this test differently is because our internal machines don't have access to packages needed by CRIU, so the infra team pre-built CRIU into the container. If you have access to packages to build CRIU in ubi, you can follow the [CRIU](https://criu.org/Main_Page) instruction to build the ubi base image for this test. Thanks.

judovana:
@smlambert  @sophia-guo  hello! Anything more is needed?

smlambert (changes requested):
I have started reviewing this PR, but have stopped for now and will put a hold on it while I look at the current functionality versus what we used to have. I believe we had this functionality of using other images/OSes 'previously' but must have 'regressed'.  (related: https://github.com/adoptium/aqa-tests/pull/3363).

judovana:
> I have started reviewing this PR, but have stopped for now and will put a hold on it while I look at the current functionality versus what we used to have. I believe we had this functionality of using other images/OSes 'previously' but must have 'regressed'. (related: #3363).

Right you are, there was such a support, but it needed IFs for every distribution. I had made it as general as possible. There is only one real `if` left, which distinguish apt and dnf systems. I would like to get rid also of it no longer run (another property in test.properties I guess). But this PR is already too big, so not (yet)added.
In addition,   that support expected to pass each  part of container URI as it separated aprts, which - considering number of those parts - was  very uneasy to read, use or enhance.

external/common.properties
```diff
@@ -0,0 +1,6 @@
+# those packages are included in all tests and all distributions
```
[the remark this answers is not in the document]
judovana: Sure. But before I do, please check: https://github.com/adoptium/aqa-tests/pull/5553/commits/5f8891585f459882e8f8a9c7eea3c43771f1b2aa#diff-3e2f7b162bce7417355d4fe0f18e6366b69e6abe5e73356f863274d3a4f9c04bR2  Those `generic_packages="git wget unzip tar curl"`  are used in 99% of tests (see the rest of https://github.com/adoptium/aqa-tests/pull/5553/commits/5f8891585f459882e8f8a9c7eea3c43771f1b2aa and search for -ubuntu_packages or -ubi_packages).  I think such base really should be added. (note, this is not disagreeing with removal, just asking for brainstroming. I *will* remove it solemnly on this comment).

external/build_all.sh
```diff
@@ -16,6 +16,16 @@ set -o pipefail
 
 source $(dirname "$0")/common_functions.sh
 
+if [ -n "${EXTERNAL_AQA_IMAGE}" ] ; then\
+  echo "EXTERNAL_AQA_IMAGE is decalred as $EXTERNAL_AQA_IMAGE; it would break this script. Unset it."
```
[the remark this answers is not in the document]
judovana: Please see https://github.com/adoptium/aqa-tests/issues/5657 ;  I think that script shoud lbe dropped. It was broken since February. If such loop would be indeed needed, then my humble opinion is, that is should iterate through exact set of URIs, instead composing them by many IFs later in undelying scripts.

external/build_all.sh
```diff
@@ -16,6 +16,16 @@ set -o pipefail
...
+  exit 1
+fi
+
+if [ -z "${version}" ] ; then\
+  echo "'version' (of jdk, eg 17) is mandatory variable for this script to run"
+  exit 1
+fi
+
 # Cleanup any old containers and images
```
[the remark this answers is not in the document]
judovana: Please see https://github.com/adoptium/aqa-tests/issues/5657 ; I think that script should lbe dropped. It was broken since February,  when the version was dropped from loops but its usage was left on place. Its slightly earlier in this PR, without reply.

external/build_all.sh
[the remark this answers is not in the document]
smlambert: I will verify that no one else uses the script, then ask that build_all.sh be removed as 'no use case'.

external/external.sh
```diff
@@ -19,6 +19,33 @@ set -e
 
 source $(dirname "$0")/provider.sh
 
+if [ -z "${EXTRA_DOCKER_ARGS}" ] ; then
+  echo \
+"   # Warning Warning  Warning Warning Warning Warning Warning Warning Warning Warning Warning Warning #
```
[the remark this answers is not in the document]
judovana: Well but is well visible in logs O:). Sorry, especially with the markup, its indeed terrible. Will fix.

judovana:
decalred -> declared fixed. Any luxk with  [build_all.sh](https://github.com/adoptium/aqa-tests/pull/5553/files/0839a2771b2d148e659cf6586aae41a77ea87f7d#diff-2d631d3a282f5afb5fa112fa47f08911d787ad3cc299a7bf8cac1e57a35a4859) ?

external/build_all.sh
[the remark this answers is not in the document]
judovana: Typo fixed. Any progress on  removing of build-all?

external/build_all.sh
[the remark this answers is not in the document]
judovana: I hope that file will be removed. If not, then authors of commit  which made the build-all.sh should elaborate on correct usage.

external/external.sh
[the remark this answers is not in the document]
judovana: See `openj9/runtime/compiler/x/codegen/X86Recompilation.hpp:// WARNING WARNING WARNING WARNING WARNING WARNING WARNING WARNING WARNING WARNING`  but sure. Fixed

judovana:
fixed waht seemed to be fixable...
@smlambert  Any news on "the hold" ? It keeps to be missing feature...
@sophia-guo  You are ok with it as it is, or any more hints?

Thanx a lot both!

external/common.properties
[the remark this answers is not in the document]
judovana: @smlambert  Even if practically unused, I would liek to keep this file. Is is extremly helpful when new distribution is tried. It simply allows you to declare its specific dependencies on one place, and then you can easily pinpoint it. And if this file remains unused for future, then they can be tuned up for individual suites alter.

external/build_all.sh
```diff
@@ -16,6 +16,16 @@ set -o pipefail
 
 source $(dirname "$0")/common_functions.sh
 
+if [ -n "${EXTERNAL_AQA_IMAGE}" ] ; then\
+  echo "EXTERNAL_AQA_IMAGE is declared as $EXTERNAL_AQA_IMAGE; it would break this script. Unset it."
+  exit 1
+fi
+
+if [ -z "${version}" ] ; then\
+  echo "'version' (of jdk, eg 17) is mandatory variable for this script to run"
```
karianna, line 25:
```suggestion
  echo "'version' (of JDK, e.g., 17) is mandatory variable for this script to run"
```

external/common_functions.sh
```diff
@@ -161,11 +169,34 @@ function set_base_docker_registry_dir() {
...
+    local PROP_KEY="${1}"
+    local FILE=${2:-""}
+    if [ -z "${FILE}" ] ; then
+      FILE="${PROPERTY_FILE}"
+    fi
+    local PROP_VALUE=`cat "${FILE}" | grep -v "^#" | grep "^${PROP_KEY}=" | cut -d'=' -f 2-`
     echo  `sed -e 's/^"//' -e 's/"$//' <<<"$PROP_VALUE"`
 }
 
+# Getting matchning keys for given OS
```
karianna, line 181:
```suggestion
# Getting matching keys for given OS
```

external/dockerfile_functions.sh
```diff
@@ -212,6 +224,9 @@ print_ant_install() {
     local file=$1
     local ant_version=$2
     local os=$3
+    if isExternalImageEnabled ; then
+        os=$(getImageOs)
```
karianna, line 228: on line 512 this `os` variable is in Uppercase - which one should it be?

external/external.sh
```diff
@@ -19,6 +19,33 @@ set -e
...
+  echo \
+"   # ================= WARNING ================= WARNING ================= WARNING ================= #
+   # EXTRA_DOCKER_ARGS are not set. You will be testing java which is already in container and not TEST_JDK_HOME
+   # TEST_JDK_HOME is set to $TEST_JDK_HOME but will not be used. See test_base_functions.sh for order of search
+   # You should set your's TEST_JDK_HOME to mount to /opt/java/openjdk, eg:                          #
+   # export EXTRA_DOCKER_ARGS=\"-v \$TEST_JDK_HOME:/opt/java/openjdk\"                               #
+   # ================= WARNING ================= WARNING ================= WARNING ================= #"
+else
+  echo \
+"   # Info Info  Info Info Info Info Info Info Info Info Info Info Info Info Info Info Info Info Info #
```
karianna, line 32: Will this whitespacing produce a consistent output?

external/jacoco/test.properties
```diff
@@ -3,5 +3,5 @@ test_command="mvn clean verify"
 test_results="testResults"
 tag_version="v0.8.7"
 environment_variable="MODE=java"
-ubuntu_packages="git wget tar"
 maven_version="3.8.5"
```
karianna, line 6: Can we move to 3.9.9?

external/provider.sh
```diff
@@ -60,3 +66,115 @@ function getProviderTile() {
...
+###################################################################
+##              part *2*  which image to use                     ##
+## It recognizes one complex environment variable:               ##
+## EXTERNAL_AQA_IMAGE=optional_url/image:optional_version        ##
+##   defaults to:                                                ##
+## docker.io/library/eclipse-temurin:${JDK_VERSION}-jdk"         ##
+##   for most of the calls. Defaults to                          ##
+## docker.io/library/ibm-semeru-runtimes:open-${JDK_VERSION}-jdk ##
+##   for openj9                                                  ##
+##for prepare under "default" circumstances the domain is omitted##
```
karianna, line 78: These lines need a whitespace before the first word

external/jacoco/test.properties
judovana: No way in this PR:( But otherwise I would love to.

external/external.sh
judovana: I had aligned itto that "====" format Shelly suggested. They provide as consistent output as possible. Do you have any more exact concern with them? Like being to long or simiarly?

external/dockerfile_functions.sh
judovana: They have nothing in common.  The OS was introduced  4 years ago, os was introduced 2 years ago. They both are local. I handle the same additional logic to both of them, as both share the value of current os. I can refactor to be both lowercase (As are other local variables). But I do not strong preffernece, nor do I want to add more complexity to this PR. Still Sharp eye you have. TY!

Both probably share that theirs authors are originally form Asia origin :)

judovana:
Anybody alive around?

sophia-guo:
@judovana is it possible to run a grinder for the final check? All former grinders are obsolete.

sophia-guo (approved):

judovana:
> https://ci.adoptium.net/view/Test_grinder/job/Grinder/11615/console

Is green... Locally I tried also with EXTERNAL_AQA_IMAGE=fedora:41 and EXTERNAL_AQA_IMAGE=centos:stream9 passed.

As part as https://github.com/adoptium/aqa-tests/issues/5575 I will be doing extensive 3D matrix of external_test/os/jdk  

Do yuou think EXTERNAL_AQA_IMAGE woould be worthy to be used in grinder?

external/provider.sh
```diff
@@ -60,3 +66,115 @@ function getProviderTile() {
...
+  local JDK_VERSION="${1:-0}"
+  local jreSdk="${2:-jdk}"
+  if [ -z "${EXTERNAL_AQA_IMAGE}" ]; then
+    echo "docker.io/library/eclipse-temurin:${JDK_VERSION}-$jreSdk"
+  else
+    echo "${EXTERNAL_AQA_IMAGE}"
+  fi
+}
+
+function getFullTOpenJ9Image() {
```
smlambert, line 106: getFullTOpenJ9Image -> getFullOpenJ9Image (no T)

external/common_functions.sh
```diff
@@ -204,14 +238,28 @@ function set_test_info() {
     maven_version=$(getProperty "maven_version")
     environment_variable=$(getProperty "environment_variable")
     localPropertyFile=$(getProperty "localPropertyFile")
-    ubuntu_packages=$(getProperty "ubuntu_packages")
-    ubi_packages=$(getProperty "ubi_packages")
+    local local_generic_packages=$(getProperty "generic_packages")
+
+    packages="$global_generic_packages $local_generic_packages"
+    local gobalMatchingKeys=$(getMatchingPackagesKeys "${GLOBAL_PROPERTY_FILE}")
```
smlambert, line 244: gobalMatchingKeys -> globalMatchingKeys

smlambert:
> Do yuou think EXTERNAL_AQA_IMAGE woould be worthy to be used in grinder?

There is a strong desire to avoid bloating the already large set of Jenkins job parameters, which is why we have previously used EXTRA_DOCKER_ARGS to pass in alternate images.  

I am presuming with some modification we could still utilize the EXTRA_DOCKER_ARGS parameter to either check for --image or support other build args via parse_docker_args, and set EXTERNAL_AQA_IMAGE from that.

judovana:
Fair enough. I agree with bloating being undesired. Lets settle it down then for now, and if it works, then +1 for some `--iamge` like approach. TY!

judovana:
https://ci.adoptium.net/view/Test_grinder/job/Grinder/11636/console

external/external.sh
```diff
@@ -19,6 +19,33 @@ set -e
...
+   # TEST_JDK_HOME of $TEST_JDK_HOME is used in EXTRA_DOCKER_ARGS                                #
+   # =================== Info =================== Info =================== Info =================== #"
+  else
+    echo \
+"   # ================= WARNING ================= WARNING ================= WARNING ================= #
+   # TEST_JDK_HOME of $TEST_JDK_HOME is NOT used in EXTRA_DOCKER_ARGS                                #
+   # ================= WARNING ================= WARNING ================= WARNING ================= #"
+  fi
+fi
+
```
smlambert, line 47: For the record, not a great fan of this output.  My original review said to change the previous incarnation to look more professional.  Not going to block the PR on it, but think that this is excessively noisy and seems like its better suited for a debugging session than a production run.

smlambert (approved):
In order to extract ourselves from having to go deeper than the already 80+ comments in this PR, will merge as it is now (which should close off last of 'alignment' efforts, and direct any other activities and changes through Adoptium quarterly planning).

karianna (approved):

PR #5571, opened 2024-09-10
Fix pingPerf Test for JDK22

LongyuZhang:
Update dockerfile replace for JDK22


LongyuZhang:
Test Link:
- JDK22: hyc_grinder 43064
- JDK17: hyc_grinder 43067

karianna (approved):

external/criu/pingPerf.sh
```diff
@@ -50,7 +50,7 @@ getSemeruDockerfile() {
                 findCommandAndReplace '\-H \"\${CRIU_AUTH_HEADER}\"' '--user \"\${DOCKER_REGISTRY_CREDENTIALS_USR}:\${DOCKER_REGISTRY_CREDENTIALS_PSW}\"' $semeruDockerfile ";"
                 findCommandAndReplace 'RUN --mount.*' 'ARG DOCKER_REGISTRY_CREDENTIALS_USR \n ARG DOCKER_REGISTRY_CREDENTIALS_PSW \n RUN set -eux; \\' $semeruDockerfile
                 # in 21-ea, /opt/java/openjdk/legal/java.base/LICENSE does not exist. No need to replace
-                if [[ $jdkVersion != "21" ]]; then
+                if [[ $jdkVersion < 21 ]]; then
```
keithc-ca, line 53: Shouldn't this use `-lt` instead?
```
               if [[ $jdkVersion -lt 21 ]]; then
```
I didn't expect that `[[` would even see `<` (rather than connect stdin to a file named "21"), but according to https://www.man7.org/linux/man-pages/man1/bash.1.html,

> When used with [[, the < and > operators sort lexicographically using the current locale.

so `[[ 8 < 21 ]]` would yield false which is unwanted.

LongyuZhang: Updated in PR https://github.com/adoptium/aqa-tests/pull/5578. Thanks.

Issue #5600, opened 2024-09-13
Provide a mechanism for the jtrunner to inject custom jti parameters


andrew-m-leonard:
Currently the jck jtrunner get the required jti value from https://github.com/adoptium/aqa-tests/tree/master/jck/jtrunner/config
these values are not overridable via any custom means say a custom private value is required for say "ftpURL" (eg.https://github.com/adoptium/aqa-tests/compare/master...revert-5588-jckftp_pwd?expand=1), then it is not currently possible.

Can we provide a mechanism for the jck jtrunner to take a custom local jti additional/override option set?

PR #5601, opened 2024-09-13
Revert "redact jckftp url"

andrew-m-leonard:
Reverts adoptium/aqa-tests#5588

Although not ideal this password is to a secure firewalled machine, currently without this ftpURL set correctly we are unable to easily run Alpine Linux java_net/URL test that use ftpURL, as there is no current way to inject custom ftpURL via the jck jtrunner.
Until this feature is available, there is no harm in having this as it was.

Raised a new issue to request such a feature: https://github.com/adoptium/aqa-tests/issues/5600


steelhead31 (approved):
Looks ok.

sxa (approved):

karianna (approved):

PR #5607, opened 2024-09-13
Enhance diff.sh to include file difference

### LINKED ISSUE #5586: Enhance jck/diff/diff.sh to provide more details about the difference between two different jck
Currently diff.sh only checks if the test subfolders (depth 2) are same or not when comparing two jcks, which helps us know if jck test targets needs to be updated or not. 

A complete comparison  of the test subfolders will also help us know if new tests are added or former tests are removed. Correspondingly makefile or other related files can be updated easily.

sophia-guo:
Close #5586

### DIFF
--- a/jck/diff/diff.sh
+++ b/jck/diff/diff.sh
@@ -79,7 +79,7 @@ compareGroups()
 
 compareNewTestsInSubGroups() {
 	subGroups=("vm.verifier.instructions" "lang.CLSS" "lang.EXPR" "lang.CONV" "lang.INTF" "lang.LMBD" "lang.DASG" "lang.NAME" "lang.TYPE" "lang.ANNOT" "lang.STMT")
-	cat > "$WORKDIR/logs/NewTestsInSubGroups.txt"
+	touch "$WORKDIR/logs/NewTestsInSubGroups.txt"
 	for group in "${subGroups[@]}"; do
 		grep "${group}" "${WORKDIR}/${VERSION2}/JCK-compiler-${VERSION_VALUE2}/NewTests.txt" | sort | uniq >> "$WORKDIR/logs/NewTestsInSubGroups.txt"
 	done
@@ -87,7 +87,7 @@ compareNewTestsInSubGroups() {
 	if [ ! -s "$WORKDIR/logs/NewTestsInSubGroups.txt" ]; then
 		echo " No Subdir Group changes."
 	else
-		cat > "$WORKDIR/logs/DiffsInSubGroups.txt"
+		touch "$WORKDIR/logs/DiffsInSubGroups.txt"
 		while IFS= read -r line; do
 			subdir=$(echo "$line" | tr '.' '/' | sed 's/[[:blank:]]//g')
 			component="compiler"
--- a/jck/diff/diff.sh
+++ b/jck/diff/diff.sh
@@ -1,104 +1,147 @@
 #!/bin/bash 
 
-setup()
-{
+setup() {
 	VERSION1=$(echo "$REPO1" | sed 's/[^0-9]*//g')
 	VERSION2=$(echo "$REPO2" | sed 's/[^0-9]*//g')
-
-	if [[ $VERSION1 == $VERSION2 ]] ; then 
+	
+	if [[ "$VERSION1" == "$VERSION2" ]] ; then 
 		echo "Please provide two different repositories to compare"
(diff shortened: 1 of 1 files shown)


sophia-guo:
TC job: JCK_Check_For_New_Test_Group/31/

jck/diff/diff.sh
```diff
@@ -8,18 +8,21 @@ setup()
...
+		VERSION1="$VERSION2"
+		VERSION2="$temp"
+		temp="$REPO1"
+		REPO1="$REPO2"
+		REPO2="$temp"
+	fi
+
+	VERSION_VALUE1="$VERSION1"
+	VERSION_VALUE2="$VERSION2"
 	if [[ "$VERSION1" -eq 8 ]] ; then
```
andrew-m-leonard, line 24: I think we should use 8c,  as the input param, rather than coding in this script, eg 11 has a suffix as well (11a)

sophia-guo:
TC job: JCK_Check_For_New_Test_Group/32/

jck/diff/diff.sh
```diff
@@ -1,104 +1,147 @@
...
-		return 1
-	fi 
+		cat "$WORKDIR/logs/$dirName-groups-diff.txt"
+	fi
+	return $rc
+}
+
+compareNewTestsInSubGroups() {
+	subGroups=("vm.verifier.instructions" "lang.CLSS" "lang.EXPR" "lang.CONV" "lang.INTF" "lang.LMBD" "lang.DASG" "lang.NAME" "lang.TYPE" "lang.ANNOT" "lang.STMT")
+	cat > "$WORKDIR/logs/NewTestsInSubGroups.txt"
```
andrew-m-leonard, line 82: This line doesn't look right? Do you mean to echo "some header" > "$WORKDIR/logs/NewTestsInSubGroups.txt"

sophia-guo: No, it's basically to create the file NewTestsInSubGroups.txt. As I tested I found sometimes `diff "${WORKDIR}/${VERSION1}/JCK-${component}-${VERSION_VALUE1}/tests/${subdir}" "${WORKDIR}/${VERSION2}/JCK-${component}-${VERSION_VALUE2}/tests/${subdir}" >> "$WORKDIR/logs/DiffsInSubGroups.txt"` ">>" won't create the file as expected if the file doesn't exist.

andrew-m-leonard: So yeah that makes sense, but it's just the syntax seem odd, if I try say:
```
cat > file.txt
```
it hangs!!

sophia-guo: Oops! All I'm thinking is echo, but don't know how it come out with `cat`. Update with touch. Thanks @andrew-m-leonard !

smlambert (approved):

sophia-guo:
@andrew-m-leonard

jck/diff/diff.sh
```diff
@@ -1,104 +1,147 @@
...
 		rmdir "$WORKDIR/logs"
 		mkdir -p "$WORKDIR/logs"
-	fi 
+	fi
 }
 
 clone()	
 {
-	cd $WORKDIR
+	cd "$WORKDIR" || exit1
```
andrew-m-leonard, line 33: space missing
```
exit 1
```

jck/diff/diff.sh
```diff
@@ -1,104 +1,147 @@
...
-		return 1
-	fi 
+		cat "$WORKDIR/logs/$dirName-groups-diff.txt"
+	fi
+	return $rc
+}
+
+compareNewTestsInSubGroups() {
+	subGroups=("vm.verifier.instructions" "lang.CLSS" "lang.EXPR" "lang.CONV" "lang.INTF" "lang.LMBD" "lang.DASG" "lang.NAME" "lang.TYPE" "lang.ANNOT" "lang.STMT")
+	touch "$WORKDIR/logs/NewTestsInSubGroups.txt"
```
andrew-m-leonard, line 82: Can we guarantee this file doesn't exist from a previous run?

sophia-guo: setup() does the dir  $WORKDIR/logs/ clean. Though it's still safer to do the cleanup here. Update. @andrew-m-leonard .

karianna (approved):

andrew-m-leonard (approved):

Issue #5616, opened 2024-09-17
Windows openjdk tests fail with "application does not have desktop access" due to JDK-8336862


andrew-m-leonard:
Exclude openjdk tests that fail due to change in headless Windows detection, see https://bugs.openjdk.org/browse/JDK-8336862

ref: https://github.com/adoptium/aqa-tests/issues/5598#issuecomment-2352950695

Failing tests:
TEST: java/beans/PropertyEditor/6380849/TestPropertyEditor.java
TEST: java/beans/PropertyEditor/TestColorClass.java
TEST: java/beans/PropertyEditor/TestColorClassJava.java
TEST: java/beans/PropertyEditor/TestColorClassNull.java
TEST: java/beans/PropertyEditor/TestColorClassValue.java
TEST: java/beans/PropertyEditor/TestFontClass.java
TEST: java/beans/PropertyEditor/TestFontClassJava.java
TEST: java/beans/PropertyEditor/TestFontClassNull.java
TEST: java/beans/PropertyEditor/TestFontClassValue.java
TEST: java/beans/XMLEncoder/java_awt_ScrollPane.java
TEST: javax/imageio/plugins/jpeg/JPEGsNotAcceleratedTest.java

smlambert:
These were excldued in master, but the release branch was already created, so they would have needed to be cherrypicked to v1.0.4-release branch to avoid seeing the same failures in the October release.

sophia-guo:
Close it see https://github.com/adoptium/aqa-tests/issues/5616#issuecomment-2417563885

PR #5617, opened 2024-09-17
Exclude windows beans due to headless detection JDK-8336862

andrew-m-leonard:
Exclude Windows beans tests that fail with headless detection issue "application does not have desktop access"
ref: https://bugs.openjdk.org/browse/JDK-8336862

### DIFF
--- a/openjdk/excludes/ProblemList_openjdk23.txt
+++ b/openjdk/excludes/ProblemList_openjdk23.txt
@@ -18,16 +18,16 @@
 
 # jdk_beans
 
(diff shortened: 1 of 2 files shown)


smlambert (commented):
Please replace https://github.com/adoptium/aqa-tests/pull/5616 with https://bugs.openjdk.org/browse/JDK-8336862 as the referenced issue in each of the added lines.

In the case where there is an upstream issue, we prefer to use it for the exclusion, in that way, we do not have to leave a duplicate issue (aqa-tests/5616) open in our repo.

smlambert (approved):

sophia-guo (approved):

Issue #5621, opened 2024-09-18
Dynamic parallel via TEST_TIME issue


llxia:
When using dynamic parallel via TEST_TIME, there is a chance that we do not get the actual test execution time. This could result in default test execution being used (2 secs per test). In the following example, extended.jck is expected to be complete in 01m56s  on one machine.

```
00:05:04.866  TEST DURATION
00:05:04.866  ====================================================================================
00:05:04.866  Total number of tests searched: 94
00:05:04.866  Number of test durations found: 94
00:05:04.866  Top slowest tests: 
00:05:04.866  	00m02s jck-compiler-lang-CLSS-Group9_0
00:05:04.866  	00m02s jck-compiler-lang-ANNOT_0
00:05:04.866  	00m02s jck-runtime-lang-EXEC_0
00:05:04.866  ====================================================================================
00:05:04.866  
00:05:04.866  Test target is split into 1 lists.
00:05:04.866  Reducing estimated test running time from 01m56s to 01m56s.
00:05:04.866  
00:05:04.866  -------------------------------------testList_0-------------------------------------
00:05:04.866  Number of tests: 94
00:05:04.866  Estimated running time: 01m56s
```

[link](https://hyc-runtimes-jenkins.swg-devops.com/job/Test_openjdk17_j9_extended.jck_x86-64_linux/233/consoleFull)

This may be due to network issues. While we are investigating this, we should set the default minimal NUM_MACHINES for certain long-running test builds when dynamic parallel via TEST_TIME is used.

sophia-guo:
Closed by #6132  and #5626

PR #5626, opened 2024-09-18
Set default NUM_MACHINES for dynamic parallel via TEST_TIME

llxia:
related: https://github.com/adoptium/aqa-tests/issues/5621


llxia:
[link](https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/43297/)
```
00:02:31.872  NUM_LIST value 1 may not be calculated based on the actual execution time.
[Pipeline] echo
00:02:31.874  Regenerate parallel list with NUM_MACHINES=6 for TARGET extended.jck.
```

LongyuZhang (approved):

PR #5628, opened 2024-09-18
Add VectorGatherMaskFoldingTest to excludes for jdk23 s390x

jiekang:
Test material needs an update from JDK-8333248. Exclude until that is backported to 23.

Related: https://github.com/adoptium/aqa-tests/issues/5620

Issue #5620, opened 2024-09-17
VectorGatherMaskFoldingTest fails for s390x and riscv jdk 23


jiekang:
Logs:
```
[2024-09-16T17:39:11.476Z] Caused by: java.lang.RuntimeException: assertNotEquals: expected [1] to not equal [1]
[2024-09-16T17:39:11.476Z] 	at jdk.test.lib.Asserts.fail(Asserts.java:691)
[2024-09-16T17:39:11.476Z] 	at jdk.test.lib.Asserts.assertNotEquals(Asserts.java:451)
[2024-09-16T17:39:11.476Z] 	at jdk.test.lib.Asserts.assertNotEquals(Asserts.java:435)
[2024-09-16T17:39:11.476Z] 	at compiler.vectorapi.VectorGatherMaskFoldingTest.testTwoLongVectorLoadGatherMaskedNotEqualMask(VectorGatherMaskFoldingTest.java:227)
[2024-09-16T17:39:11.476Z] 	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
[2024-09-16T17:39:11.476Z] 	... 7 more
```

From:
https://ci.adoptium.net/job/Grinder/10882/consoleText

jiekang:
The test is newly introduced in 23+ via:
https://bugs.openjdk.org/browse/JDK-8325520

The test is known to fail when maximum vector bits is 64:
https://bugs.openjdk.org/browse/JDK-8333248

Though the comment specifically mentions aarch64 and risc-v, it may be that it applies to s390x as well. The test has been fixed but the code was not backported to 23 yet.

jiekang:
Maybe we can check what the vector bits setting is for s390x and then consider excluding this test.

jiekang:
Or apply the patch to 23 and run the test. Not sure what will take the least amount of work/effort though...

jiekang:
Could also try other vendor releases for s390x to see if the test failed

sophia-guo:
I guess only openj9 support s390x

Haroon-Khel:
I applied the changes in https://github.com/adoptium/jdk/commit/9b0a5c5cd056262fab99525e9260762b617152a3 and the test passes on s390x

jiekang:
I would vote to ignore this test result for the release then.

smlambert:
[...]

PR #5632, opened 2024-09-19
Add riscv exclude for jdk23 VectorGatherMaskFoldingTest

jiekang:
The test also fails on riscv:

Test material needs an update from JDK-8333248. Exclude until that is backported to 23.

Related: #5620 #5628


jiekang:
And this should also get cherry-picked to the v1.0.4-release branch

smlambert (approved):

sophia-guo (approved):

PR #5674, opened 2024-10-08
Exclude failing tests, won't fix 21.0.5 & 17.0.13

sophia-guo:
Exclude failing tests, won't fix 21.0.5 & 17.0.13


smlambert (commented):
Do we not also want to exclude them in aqa-tests master, or do they pass there?

sophia-guo:
Test failed now in aqa-tests. The reason is even though aqa-test master will grab openjdk tests the dev branch by default ( the fix is in dev and 21.0.6, 17.0.14, https://bugs.openjdk.org/browse/JDK-8338748) , the latest or nightly jdk itself are still tagged with 21.0.5 and 17.0.13. Once release is over the new tag would be 21.0.6 and 17.0.14, tests should pass with 21.0.6 and 17.0.14.

smlambert (approved):

PR #5686, opened 2024-10-15
Limit JDK23+ openj9 xmac testing on mac11+

llxia:

PR #5688, opened 2024-10-15
exclude JDK23+ and jdknext openj9 xmac testing on mac10

llxia:
related: https://github.com/adoptium/aqa-tests/pull/5686#discussion_r1801681881

### DIFF
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -303,8 +303,11 @@ timestamps{
             }
 
             // JDK23+ isn't supported on machines prior to Mac 11 https://github.com/eclipse-openj9/openj9/issues/19694
-            if (params.JDK_IMPL == "openj9" && PLATFORM == "x86-64_mac" && params.JDK_VERSION && params.JDK_VERSION.toInteger() >= 23) {
-                LABEL += "&&!sw.os.mac.10"
+            // assume if JDK_VERSION is set to next, JDK_VERSION is 23+
+            if (params.JDK_IMPL == "openj9" && PLATFORM == "x86-64_mac" && params.JDK_VERSION) {
+                if (!params.JDK_VERSION.isInteger() || params.JDK_VERSION.toInteger() >= 23) {
+                    LABEL += "&&!sw.os.mac.10"
+                }
             }
 
             if (params.DOCKER_REQUIRED) {


buildenv/jenkins/openjdk_tests
```diff
@@ -303,8 +303,11 @@ timestamps{
             }
 
             // JDK23+ isn't supported on machines prior to Mac 11 https://github.com/eclipse-openj9/openj9/issues/19694
-            if (params.JDK_IMPL == "openj9" && PLATFORM == "x86-64_mac" && params.JDK_VERSION && params.JDK_VERSION.toInteger() >= 23) {
-                LABEL += "&&!sw.os.mac.10"
+            // assume if JDK_VERSION is set to next, JDK_VERSION is 23+
+            if (params.JDK_IMPL == "openj9" && PLATFORM == "x86-64_mac" && params.JDK_VERSION) {
+                if (params.JDK_VERSION.equalsIgnoreCase("next") || params.JDK_VERSION.toInteger() >= 23) {
```
keithc-ca, line 308: I think this should be using `isInteger()` rather than make a special case for "next".
```
                if (!params.JDK_VERSION.isInteger() || params.JDK_VERSION.toInteger() >= 23) {
```

pshipton:
The grinder states "With AUTO_DETECT=on, JDK_VERSION value is required only if SDK_RESOURCE=nightly/releases. Otherwise, please leave it empty." Setting it to next with AUTO_DETECT off won't work.

buildenv/jenkins/openjdk_tests
llxia: updated

llxia:
Description `With AUTO_DETECT=on, JDK_VERSION value is required only if SDK_RESOURCE=nightly/releases. Otherwise, please leave it empty.` in Grinder is outdated. I will update it.

The test pipeline cannot consume next, so it will try to get Java version and set it at a later stage. AUTO_DETECT has to be set when using JDK_VERSION=next.

pshipton:
What does "JDK_VERSION=next" mean? Does it have special meaning, or is it just ignored and we could set any non-integer value?

keithc-ca (approved):

llxia:
`JDK_VERSION=next` means no JDK_VERSION is set in the test pipeline. The test pipeline will try to figure out JDK_VERSION value at a later stage (TKG level) if AUTO_DETECT is on. 

`JDK_VERSION=next` was initially introduced as the upstream pipeline cannot set the correct JDK_VERSION in the test pipeline.

pshipton (approved):

LongyuZhang (approved):

PR #5714, opened 2024-10-22
update AdoptOpenJDK references to Adoptium

gdams:
This will require some thorough testing but we should make this change


llxia:
In addition to ensuring all upstream pipelines (i.e., Adoptium, Openj9, Semeru, etc) pass the correct parameters, this change requires regenerating or updating all test jobs and Grinders across all sites. Coordination between all sites is necessary.

Move this PR into draft until we have a plan with all necessary changes.

cc @smlambert for awareness

smlambert:
If we are going to proceed with this change, it will need to be communicated clearly to everyone downstream who is using AQAvit and be impacted, as they will all need to update their setups.  I also would prefer to call it AQAVIT_REPO/BRANCH instead of ADOPTIUM_REPO/BRANCH.  There needs to be an update of documentation in other several other repositories that will be required (and possibly tagging a bunch of presentations as outdated).

PR #5740, opened 2024-11-11
Finish/fix SKIP_JCK_GIT_UPDATE variable in JCK

judovana:
It seems the SKIP_JCK_GIT_UPDATE was not finished/was single usage targeted/or was broken in later commits.

Before this PR, the JCK_GIT_REPO was mandatory to exists even if the SKIP_JCK_GIT_UPDATE was on place. In addition, the JCK_ROOT (which was supposed to be not updated) was deleted if the checks failed.

This PR is extending the usage of SKIP_JCK_GIT_UPDATE so the JCK_GIT_REPO is nto used at all and JCK_ROOT do not need to pass the git checs ( so it do not need to be repo art all)

Small side note - to fake git repo, which contains unpacked compatibility jars, takes some 30 minutes and used pretty quite a lot of disk space. This PR is making this creation of fake repository redundant. 

https://github.com/adoptium/aqa-tests/issues/5745

### DIFF
--- a/jck/build.xml
+++ b/jck/build.xml
@@ -119,41 +119,47 @@
 		<if>
 			<available file="${JCK_ROOT_USED}" type="dir" />
 			<then>
-				<!--Obtain local and remote SHA to figure out if local materials are up-to-date--> 
-				<exec executable="git" dir="${JCK_ROOT_USED}" failonerror="false" outputproperty="localSHA" resultproperty="return.code">
-					<arg line="rev-parse"/>
-					<arg line="HEAD"/>
-				</exec>
 				<if>
-					<not>
-						<equals arg1="${return.code}" arg2="0"/>
-					</not>
+					<isset property="env.SKIP_JCK_GIT_UPDATE" />
 					<then>
-						<echo message="Error: Git command failed with return code ${return.code}. Delete the directory and continue." />
-						<delete dir="${JCK_ROOT_USED}" failonerror="true" />
+						<echo message="env.SKIP_JCK_GIT_UPDATE is set. Skip Running git checks at ${JCK_ROOT_USED}. Continue." />
 					</then>
 					<else>
-						<exec executable="git" dir="${JCK_ROOT_USED}" failonerror="true" outputproperty="remoteSHA">
-							<arg line="ls-remote"/>
-							<arg line="${JCK_GIT_REPO_USED}"/>
-							<arg line="${jck_branch}"/>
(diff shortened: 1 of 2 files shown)


judovana:
Well if it is not used in  AzDo, I would like to use it. Because I have tck test-jars stored in binary, and distributed without repository. So I can prepare JCK_ROOT directly, and stop transforming it to repository which is quite expensive operation.

jck/build.xml
```diff
@@ -119,41 +119,47 @@
...
-				<!--Obtain local and remote SHA to figure out if local materials are up-to-date--> 
-				<exec executable="git" dir="${JCK_ROOT_USED}" failonerror="false" outputproperty="localSHA" resultproperty="return.code">
-					<arg line="rev-parse"/>
-					<arg line="HEAD"/>
-				</exec>
 				<if>
-					<not>
-						<equals arg1="${return.code}" arg2="0"/>
-					</not>
+					<isset property="env.SKIP_JCK_GIT_UPDATE" />
```
sophia-guo, line 123: Since SKIP_JCK_GIT_UPDATE is a boolean, the check will always return true as long as it is set. Maybe check to assess the specific value of env.SKIP_JCK_GIT_UPDATE instead of checking if it's set. <equal arg1="{env.SKIP_JCK_GIT_UPDATE}" arg2="true">

sophia-guo:
I opened an issue in temurin-compliance. More discussion will be moved there

jck/build.xml
judovana: The original code was just checking the "is set" . I have absolutely no objections to change it  to explicit value. TY!

judovana:
ty! Will follow

judovana:
@karianna   Sophia (@sophia-guo )is in favour to change the `SKIP_JCK_GIT_UPDATE  isSet` to `SKIP_JCK_GIT_UPDATE==true` that will break AzDo. Do you have any more info or anybody  to ping?

smlambert:
We will assume that the AzDo use case is very similar to RH QE use case where 1) they are not using our Jenkinsfile code and 2) the TCK materials are housed somewhere of their choosing and they will manage the update of them outside of this framework.  

Tagging @d3r3kk for awareness

judovana:
Tahnx all, changed to check for explicit "true".

sophia-guo:
Temurin-compliance grinder/2632,4634,4635 work as expected.

sophia-guo (approved):

smlambert (approved):

judovana:
tyvm!

Issue #4374, opened 2023-02-28
Add support for testing JDK binaries packaged in tar.xz format 


smlambert:
We fetch and install JDK binaries that are packaged in several forms (.tar.gz, .zip), we should additionally support .tar.xz files, so we can ensure to support all vendors listed in the marketplace.  Currently the Red Hat builds listed in the Adoptium marketplace use that format (example: [jdk17 RH binary](https://developers.redhat.com/content-gateway/file/pub/openjdk/adoptium/January_2023/java-17-openjdk-17.0.6.0.10-2.portable.jdk.el.x86_64.tar.xz)).

See a Grinder run that tries to fetch and test it: https://ci.adoptium.net/view/Test_grinder/job/Grinder/6747/

`tar xf java-17-openjdk-17.0.6.0.10-2.portable.jdk.el.x86_64.tar.xz` and requires xz-utils to be present on test machines (`sudo apt-get install xz-utils` if needs installation)

judovana:
//me watching  issue

TY!

PR #5762, opened 2024-11-19
Add support for decompression of tar.xz in get.sh

sxa:
Fixes https://github.com/adoptium/aqa-tests/issues/4374

A couple of points:
- Do we have a requirement for this for the debugimage etc.? I've covered it in this PR regardless
- We do not guarantee that `xz` is on any given machine - maybe we should add it ... At the moment it'll just fail to extract with a message along the lines of`xz: not found` or similar.

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -379,10 +379,15 @@ getBinaryOpenjdk()
 					unzip -q $file_name -d $extract_dir
 				else
 					# some debug-image tar has parent folder ... strip it
-					if tar --version 2>&1 | grep GNU 2>&1; then
-						gzip -cd $file_name | tar xof - -C $extract_dir --strip 1
+					if [[ $file_name == *xz ]]; then
+						DECOMPRESS_TOOL=xz
 					else
-						mkdir dir.$$ && cd dir.$$ && gzip -cd ../$file_name | tar xof - && cd * && tar cf - . | (cd ../../$extract_dir && tar xpf -) && cd ../.. && rm -rf dir.$$
+						DECOMPRESS_TOOL=gzip
+					fi
+					if tar --version 2>&1 | grep GNU 2>&1; then
+						$DECOMPRESS_TOOL -cd $file_name | tar xof - -C $extract_dir --strip 1
+					else
+						mkdir dir.$$ && cd dir.$$ && $DECOMPRESS_TOOL -cd ../$file_name | tar xof - && cd * && tar cf - . | (cd ../../$extract_dir && tar xpf -) && cd ../.. && rm -rf dir.$$
 					fi
 				fi
 			else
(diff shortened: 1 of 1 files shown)


smlambert:
The initial reason for the creation of https://github.com/adoptium/aqa-tests/issues/4374 to support xz was to support testing RHBO since it is packaged in that way (differently from the many other distributions that we support the testing of...).  

In that sense, I would not say it is a 'hard requirement' for us.  It is certainly a nice to have for those wishing to test artifacts that are compressed in that way.  I am not aware of a requirement for debugimage at the moment, but certainly fine to include it.

get.sh
```diff
@@ -367,6 +367,12 @@ getBinaryOpenjdk()
 	for file_name in "${jdk_file_array[@]}"
 	do
 		if [[ ! "$file_name" =~ "sbom" ]]; then
+			if [[ $file_name == *xz ]]; then
```
llxia, line 370: Can we make the check more specific? Instead of checking `xz`, check
`if [[ $file_name == *.tar.xz ]]; then`

sxa: The primary reason for not doing that is that .txz is also used by some people (same as `.tgz` is often used as a shorted `.tar.gz`), so this would catch that situation too .

llxia:
Please provide a Grinder link.

sxa:
> Please provide a Grinder link.

For what specifically?

smlambert:
Minimally 2 Grinders would be good, 1 to verify common use cases are unaffected, 1 to verify this works to decompress a RHBO from the marketplace would be helpful.

smlambert:
and apologies, we've just merged Andrew's PR, so you'll need to resolve the conflicts in the get.sh file

sxa:
> Minimally 2 Grinders would be good, 1 to verify common use cases are unaffected, 1 to verify this works to decompress a RHBO from the marketplace would be helpful.

- https://ci.adoptium.net/job/Grinder/11582/testReport/ is one with a marketplace tar.xz
- https://ci.adoptium.net/job/Grinder/11549/testReport/ is a "normal" temurin release
- https://ci.adoptium.net/job/Grinder/11602/ is a new one I've just kicked off to ensure it hasn't broken the Windows code path.

I figured I'd fire off an "upstream" one too to make sure those artifacts all work: https://ci.adoptium.net/job/Grinder/11608/

Hopefully that's adequate :-)

sxa:
> and apologies, we've just merged Andrew's PR, so you'll need to resolve the conflicts in the get.sh file

Ah yes ... that's unfortunate

sxa:
Re-runs of the four above after mergeing in Andrew's change:
- https://ci.adoptium.net/job/Grinder/11609/
- https://ci.adoptium.net/job/Grinder/11611/
- https://ci.adoptium.net/job/Grinder/11612/
- https://ci.adoptium.net/job/Grinder/11614/

smlambert (approved):
thanks @sxa

llxia (approved):
LGTM

PR #5766, opened 2024-11-20
Exclude failing UTC test on MacOS on Eclipse only

Haroon-Khel:
An extension of https://github.com/adoptium/aqa-tests/pull/5447/files

ref https://github.com/adoptium/infrastructure/issues/3361


Haroon-Khel:
Could someone with write access please merge. Thanks

PR #5770, opened 2024-11-21
Extra doc for EXTRA_OPTIONS

sxa:
Adding an example of what you might use this for. With the replace parameter being `JVM_OPTIONS` it's not entirely clear that `EXTRA_OPTIONS` (without `JVM` in the name) can be used for extra VM options as opposed to just parameters for the test harness - this should remove any ambiguity from me or others in the future.

Ref https://github.com/adoptium/infrastructure/issues/3065#issuecomment-2489009147

### DIFF
--- a/buildenv/jenkins/testJobTemplate
+++ b/buildenv/jenkins/testJobTemplate
@@ -361,8 +361,8 @@ ARCH_OS_LIST.each { ARCH_OS ->
                                                                        individual test in playlist.xml: floatSanityTests|MiniMix_5m|jdk_math|jdk_beans_0|jdk_custom<br/>
                                                                        a list of tests: testList TESTLIST=jit_jitt,jit_recognizedMethod_1,testSCCMLTests2_1<br/>
                                                                        Note: please do not use level, group or type (i.e., functional, sanity, etc) in the TESTLIST<br/>''')
-							stringParam('CUSTOM_TARGET', "", '''
-								Only used when the custom target is specified in TARGET , e.g. jdk_custom, langtools_custom, system_custom, etc., CUSTOM_TARGET=path to the test class to execute<br>
+							stringParam('CUSTOM_TARGET', "",
+								'''Only used when the custom target is specified in TARGET , e.g. jdk_custom, langtools_custom, system_custom, etc., CUSTOM_TARGET=path to the test class to execute (typically ending in .java or .sh)<br>
 								Example 1:<br>
 								TARGET=system_custom<br>
 								CUSTOM_TARGET=-test=MathLoadTest -test-args="workload=math,timeLimit=5m"<br>
@@ -386,7 +386,7 @@ ARCH_OS_LIST.each { ARCH_OS ->
 							}
 							choiceParam('TEST_FLAG', ['', 'JITAAS', 'AOT', 'FIPS', 'FIPS140_2', 'FIPS140_3_OpenJCEPlusFIPS', 'FIPS140_3_OpenJCEPlusFIPS.FIPS140-3'], "Optional. Only set for feature testing. (i.e.,  JITAAS, AOT, FIPS140_2, etc)")
 							stringParam('EXTRA_OPTIONS', "", "Use this to append options to the test command e.g. to add extra JVM flags")
-							stringParam('JVM_OPTIONS', "", "Use this to replace the test original command line options")
+							stringParam('JVM_OPTIONS', "", "Use this to replace the original options to the test command e.g. to override JVM flags")
 							stringParam('APPLICATION_OPTIONS', "", "Use this to append options to the test application")
 							stringParam('BUILD_IDENTIFIER', "", "build identifier")
 							stringParam('ITERATIONS',"1", '''Optional. Number of times to repeat execution of make target. <br/>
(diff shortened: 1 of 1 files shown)


buildenv/jenkins/testJobTemplate
```diff
@@ -385,7 +385,7 @@ ARCH_OS_LIST.each { ARCH_OS ->
 							    sectionHeaderStyle(sectionHeaderHelpTextStyleCss)
 							}
 							choiceParam('TEST_FLAG', ['', 'JITAAS', 'AOT', 'FIPS', 'FIPS140_2', 'FIPS140_3_OpenJCEPlusFIPS', 'FIPS140_3_OpenJCEPlusFIPS.FIPS140-3'], "Optional. Only set for feature testing. (i.e.,  JITAAS, AOT, FIPS140_2, etc)")
-							stringParam('EXTRA_OPTIONS', "", "Use this to append options to the test command")
+							stringParam('EXTRA_OPTIONS', "", "Use this to append options to the test command e.g. to add extra JVM flags")
 							stringParam('JVM_OPTIONS', "", "Use this to replace the test original command line options")
```
smlambert, line 389: While you are here, consider improving L389, in keeping with your suggestion for EXTRA_OPTIONS 
"Replaces the test's original JVM options" to make it clear versus options that would be passed into some other application called in the test command (which for awareness is set via APPLICATION_OPTIONS).

smlambert (approved):
Added a suggestion to also update the parameter below the one you changed, approving in any case.

karianna (approved):

buildenv/jenkins/testJobTemplate
sxa: Done - LMK what you think of the wording. I've also done a little adjustment to CUSTOM_TARGET based on a discussion about `.sh` files in slack this week.

Also I note that the full description for CUSTOM_TARGET isn't showing up so I've moved the location of `'''` to make it consistent with the the `TARGET` parameter and this is the obvious difference. At the moment only the first line is visible for some reason:

image

(Also noting that I'm not sure if there's an easy way to test this - maybe regen a test job somehow using this fork? But a revert will be easy if it's a problem)

buildenv/jenkins/testJobTemplate
```diff
@@ -361,8 +361,8 @@ ARCH_OS_LIST.each { ARCH_OS ->
                                                                        individual test in playlist.xml: floatSanityTests|MiniMix_5m|jdk_math|jdk_beans_0|jdk_custom<br/>
                                                                        a list of tests: testList TESTLIST=jit_jitt,jit_recognizedMethod_1,testSCCMLTests2_1<br/>
                                                                        Note: please do not use level, group or type (i.e., functional, sanity, etc) in the TESTLIST<br/>''')
-							stringParam('CUSTOM_TARGET', "", '''
-								Only used when the custom target is specified in TARGET , e.g. jdk_custom, langtools_custom, system_custom, etc., CUSTOM_TARGET=path to the test class to execute<br>
+							stringParam('CUSTOM_TARGET', "",
+								'''Only used when the custom target is specified in TARGET , e.g. jdk_custom, langtools_custom, system_custom, etc., CUSTOM_TARGET=path to the test class to execute (typically ending in .java or .sh)<br>
```
smlambert, line 365: This is statement is not incorrect, but doesn't cover several other cases, so may mislead people into thinking CUSTOM_TARGET can only be used for a single test class (when it can also be used to point to a directory or one can also pass a list of test classes separated by spaces).

buildenv/jenkins/testJobTemplate
```diff
@@ -385,8 +385,8 @@ ARCH_OS_LIST.each { ARCH_OS ->
 							    sectionHeaderStyle(sectionHeaderHelpTextStyleCss)
 							}
 							choiceParam('TEST_FLAG', ['', 'JITAAS', 'AOT', 'FIPS', 'FIPS140_2', 'FIPS140_3_OpenJCEPlusFIPS', 'FIPS140_3_OpenJCEPlusFIPS.FIPS140-3'], "Optional. Only set for feature testing. (i.e.,  JITAAS, AOT, FIPS140_2, etc)")
-							stringParam('EXTRA_OPTIONS', "", "Use this to append options to the test command")
-							stringParam('JVM_OPTIONS', "", "Use this to replace the test original command line options")
+							stringParam('EXTRA_OPTIONS', "", "Use this to append options to the test command e.g. to add extra JVM flags")
+							stringParam('JVM_OPTIONS', "", "Use this to replace the original options to the test command e.g. to override JVM flags")
 							stringParam('APPLICATION_OPTIONS', "", "Use this to append options to the test application")
```
smlambert, line 390: Improved, thanks!

buildenv/jenkins/testJobTemplate
smlambert: Noting that more exhaustive documentation for these parameters is captured here: https://github.com/adoptium/aqa-tests/blob/master/doc/pages/JenkinFeatures.md

sxa: Bear in mind I haven't changed the statement on `CUSTOM_TARGET` in this PR - just moving the quotes in the hope it will expose on the UI what's already there, which does include the - very useful - space-separated example, but it would definitely be great to have some sort of reference to the directory usage too somehow for the casual user! (Either via a link to another doc, or including another example in the list that's already in the parameter description)

> Noting that more exhaustive documentation for these parameters is captured here: https://github.com/adoptium/aqa-tests/blob/master/doc/pages/JenkinFeatures.md

I'm having a look at that page but CUSTOM_TARGET is only referenced in a table that only says `Required when TARGET=jdk_custom` and I can't see a reference to "directory" or "folder" in that doc. If there is a good list of those things it would be great to reference it in the description.

sxa: Maybe this should be a separate issue/PR now :-)

smlambert: Yes, let's get the EXTRA_OPTIONS and JVM_OPTIONS changes in

PR #5771, opened 2024-11-22
Unexclude LargeGatheringWrite everywhere except AIX

sxa:
Re-enabling after the investigation in https://github.com/adoptium/infrastructure/issues/3065#issuecomment-2493619252

A couple of points of note:
- I have not removed the excludes for OpenJ9 across all platforms which is still in place as I have not tested those builds. (FYI @AdamBrousseau)
- A related upstream bug which was listed in this comment is closed, but it has not resolved the specific failure we're seeing on AIX. Should we set the reference to something else?

In draft for now so opinions on the above can be voiced while some final testing is occurring on the newer JDK23/24 builds


sxa:
Taking out of draft based on the jdk23/24 tests failing on AIX but not on Windows. Versions of 21 have been explicitly tested on further platforms.

karianna (approved):

sxa:
Closes https://github.com/adoptium/infrastructure/issues/3065

smlambert (approved):

PR #5777, opened 2024-11-27
[v1.0.5-release] Partial update of properties files for Jan 2025 CPU

smlambert:
Update the AQAvit and OpenJDK repo tags, updates for OpenJ9 tags will come in a separate PR.


llxia (approved):
Thanks @smlambert

PR #5778, opened 2024-11-27
(v1.0.5) Update testenv.properties for openj9 related repos

llxia:


pshipton (approved):

pshipton:
Should I merge it? It's still in draft.

pshipton:
Lan said elsewhere, "I am not sure about OPENJCEPLUS_GIT_BRANCH values". They are the correct values for the next release and match what we did in 1.0.4, so I'll go ahead and merge. If there is a problem we can always fix it after.

Issue #5575, opened 2024-09-10
add logic for  jdk-version specific suites


judovana:
https://github.com/adoptium/aqa-tests/pull/5553#issuecomment-2340815163 have reminded me old issue, where some external tests, eg jacoco, works fine with 8-17 but do not work with jdk21

Unluckily, current logic to clone and checkout is in https://github.com/adoptium/aqa-tests/blob/master/external/jacoco/test.properties 

So I see several  options with many pros and cons:
 - narrow the checkout in test.sh ( I do not like this)
 - have duplicated testfolder (so jacoco target would be limited for jdk21 onwards, and will have updated tag, and new jacoco17 wold be limited to 17 and older)
 - create flexible checkout logic in properties file similar to what I did for os-based  properties (I'm not sure I like this)
   - tag_version -> tag(7|8)_version  and  tag(21|22)_version ....
   - this is less expendable to other non-external  test

This will be quite a precedent for other - at least external - tests

smlambert:
Thanks for raising this issue @judovana.  Before any option is chosen, it will be good to create a table of external apps, their versions and which ones work with which Java version.  That will be useful to visualize and help inform the best way forward.

judovana:
> Thanks for raising this issue @judovana. Before any option is chosen, it will be good to create a table of external apps, their versions and which ones work with which Java version. That will be useful to visualize and help inform the best way forward.

thats terrible task.... I would much rather go one by one:(

PR #5784, opened 2024-12-03
Fixed indentation in camel/test.properties

judovana:
The indentation was so wrong, that the file could not be parsed


judovana:
Found when running https://github.com/adoptium/aqa-tests/issues/5575#issuecomment-2341256974

karianna (approved):

sophia-guo:
The parsing error was caused by the updated strict parsing format from `PROP_VALUE=`cat $PROPERTY_FILE | grep "$PROP_KEY" | cut -d'=' -f 2-`` to  local [PROP_VALUE=`cat "${FILE}" | grep -v "^#" | grep "^${PROP_KEY}=" | cut -d'=' -f 2-`](https://github.com/adoptium/aqa-tests/pull/5553/files#diff-e1cbec3d1700bf35e3daef146d4667564133f0a528b700594195c7a0184e7ff5R177), which requires the no space before the each property line. So probably we need also update some documents.

sophia-guo (approved):

PR #5683, opened 2024-10-10
docker: switch to a bespoke test container

gdams:
Also moved `LIB_DIR` and `SYSTEM_LIB_DIR` to be inside the current workspace for docker builds due to permissions errors

PR #5800, opened 2024-12-05
Enhance the environment variable check

sophia-guo:
Check environment variable as set only if it's not empty.

Related https://github.com/adoptium/aqa-tests/pull/5683#issuecomment-2504457271

### DIFF
--- a/external/build.xml
+++ b/external/build.xml
@@ -25,8 +25,10 @@
 	<property environment="env" />
 	<property name="top" location="../" />
 
+	<set-property name="DOCKERIMAGE_TAG_ISSET" if-property-isset="env.DOCKERIMAGE_TAG"/>
+	<set-property name="EXTRA_DOCKER_ARGS_ISSET" if-property-isset="env.EXTRA_DOCKER_ARGS"/>
 	<if>
-		<isset property="env.DOCKERIMAGE_TAG"/>
+		<isset property="DOCKERIMAGE_TAG_ISSET"/>
 		<then>
 			<property name="dockerImageTag" value="${env.DOCKERIMAGE_TAG}"/>
 		</then>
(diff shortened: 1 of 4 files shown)


sophia-guo:
Using [macros "set-property"](https://github.com/adoptium/TKG/blob/master/scripts/build_test.xml#L78C3-L78C30) checking both the presence of the property and whether it is non-empty to handle the case where the env.variable might be set but empty.

smlambert (approved):

llxia:
Could you please test the change in Grinder?

sophia-guo:
Grinder shows the change won't affect current behaviour.
https://ci.adoptium.net/view/Test_grinder/job/Grinder/11989/
https://ci.adoptium.net/view/Test_grinder/job/Grinder/11992/

This change mainly handle the environment variable is set but empty, which seems can only trigger in this environment https://github.com/adoptium/aqa-tests/pull/5683#issuecomment-2502027532.  Set environment variable as empty in jenkinsfilebase  https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/JenkinsfileBase#L83 won't trigger this issue. Fix and causes is similar to https://github.com/adoptium/TKG/pull/641/files. 

Given the high cost of setting up this test environment ( trigger the dynamic agent, might also need to use George's PR) and current behaviour won't be changed I'm skipping the test the environment variable is set but empty. If any issues arise late we can address them at that time.

llxia:
I tested openJcePlusTests in 2 scenarios (Grinder_FIPS/2896, Grinder_FIPS/2897). Unfortunately, I cannot test all the cases. But I think this PR is safe to merge.

llxia (approved):
LGTM

PR #5799, opened 2024-12-05
Exclude TestCipherMode for OpenJ9 jdk17, 21

pshipton:
Issue https://github.com/eclipse-openj9/openj9/issues/19962

PR #5805, opened 2024-12-09
Revert "Exclude TestCipherMode for OpenJ9 jdk17, 21"

pshipton:
Reverts adoptium/aqa-tests#5799

The problem persists with the test excluded, the test wasn't the cause of the problem.


llxia (approved):
LGTM

LongyuZhang (approved):

PR #5817, opened 2024-12-11
JDK 24+ don't enable security manager and disable failing tests

theresa-m:
In [JEP 486](https://openjdk.org/jeps/486) for Java 24 any attempt to enable a security manager (such as -Djava.security.manager=allow) will result in an error. This change removes the use of java.security.manager=allow from JDK 24+.

The system test framework also has uses the security manager for load tests that needs to be removed. In the meantime this change disables tests that are failing because of it. See https://github.com/adoptium/STF/issues/142

### DIFF
--- a/system/system.mk
+++ b/system/system.mk
@@ -47,14 +47,15 @@ else
   OPENJ9_PRAM=""
 endif
 
+# In JDK 24+ any attempts to enable the security manager will result in an error.
 # In JDK18+, java.security.manager == null behaves as -Djava.security.manager=disallow.
 # In JDK17-, java.security.manager == null behaves as -Djava.security.manager=allow.
 # In case of system tests, the base infra (STF) which is used to launch tests utilizes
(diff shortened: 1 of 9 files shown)


theresa-m:
The remaining failures are due to use of the SecurityManager library in the system test framework. I opened https://github.com/adoptium/STF/issues/142 and will temporarily disable the failing tests.

theresa-m:
This is ready for review fyi @llxia

llxia (approved):
LGTM

system/system.mk
```diff
@@ -47,14 +47,15 @@ else
   OPENJ9_PRAM=""
 endif
 
+# In JDK 24+ any attempts to enable the security manager will result in an error
```
pshipton, line 50: This sentence is missing a period.

pshipton (changes requested):

system/system.mk
theresa-m: Fixed.

pshipton (approved):

smlambert (approved):
👍

PR #5819, opened 2024-12-12
Move SHA check function to TKG

### LINKED ISSUE #5818: Track all test materials or frameworks sha info
Part of #5759 

Currently some repos ( TKG, openj9, aqa-tests and/or vendors shas)  are checked by get.sh (https://github.com/search?q=repo%3Aadoptium%2Faqa-tests%20checkRepoSHA&type=code, which calls getSHA.sh in TKG), as those repos are visible when the get.sh is called.

Some repos are checked per test in build.xml (https://github.com/search?q=repo%3Aadoptium%2Faqa-tests+checkGitRepoSha&type=code, which calls [macrodef](https://github.com/adoptium/TKG/blob/7aa981fdbadc507fe0bdede063ff38dcc27958f3/scripts/build_test.xml#L54) in TKG, and then getSHA.sh in TKG)

Some repos sha are not checked and hence are missing in tap file. We could update each test build.xml as https://github.com/search?q=repo%3Aadoptium%2Faqa-tests+checkGitRepoSha&type=code. Though the solution feels more complicated when new tests material added.  The collecting process is scattered across different places and different repo with each test needing to make the call. This approach is cumbersome and prone to being overlooked.

Suggestion to  centralize the management of this process with the TKG framework can make it simpler and more convenient. Similar to testResultSummary, we can add a shaInfo target. This way collecting shas is invisible to aqa-tests hence no tests need to handle this itself.

sophia-guo:
Close #5818

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -659,8 +659,13 @@ getFunctionalTestMaterial()
 	else
 		mv openj9/test/functional functional
 	fi
-
-	rm -rf openj9
+   	
+	cd openj9
+	mkdir -p ../git-backup
+	mv .git ../git-backup
+	rm -rf *
+	mv ../git-backup .git
+	cd $TESTDIR
 }
 
 getVendorTestMaterial() {
@@ -745,7 +750,12 @@ getVendorTestMaterial() {
 		fi
 
 		# clean up
-		rm -rf $dest
+		cd $dest
+		mv .git ../git-backup
+		rm -rf *
+		mv ../git-backup .git
+		cd $TESTDIR
+
 	done
 }
 
@@ -807,11 +817,8 @@ testJavaVersion()
 
 checkRepoSHA()
 {
-	sha_file="$TESTDIR/TKG/SHAs.txt"
-	echo "$TESTDIR/TKG/scripts/getSHAs.sh --repo_dir $1 --output_file $sha_file"
-	$TESTDIR/TKG/scripts/getSHAs.sh --test_root_dir $1 --shas_file $sha_file
-
 	testenv_file="$TESTDIR/testenv/testenv.properties"
+
 	echo "$TESTDIR/TKG/scripts/getTestenvProperties.sh --repo_dir $1 --output_file $testenv_file --repo_name $2"
 	$TESTDIR/TKG/scripts/getTestenvProperties.sh --repo_dir $1 --output_file $testenv_file --repo_name $2
 }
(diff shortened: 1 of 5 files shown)


sophia-guo:
Depends https://github.com/adoptium/TKG/pull/651

sophia-guo:
functional.dev https://ci.adoptium.net/view/Test_grinder/job/Grinder/12068/
sanity.perf https://ci.adoptium.net/view/Test_grinder/job/Grinder/12069/

smlambert (approved):
LGTM

get.sh
```diff
@@ -660,8 +659,13 @@ getFunctionalTestMaterial()
...
 		mv openj9/test/functional functional
 	fi
-
-	rm -rf openj9
+   	
+	cd openj9
+	mkdir -p ../git-backup
+	mv .git ../git-backup
+	rm -rf *
+	mv ../git-backup .git
```
llxia, line 667: I do not think we should move `.git` file around. Can we move it into the subfolder? In this case, it should be `functional`.

sophia-guo: `.git` is not moved around. `.git` keeps in openj9/.git and all others under openj9 are removed. Same for all available vendors repo as vendor*/.git. 
`aqa-tests/functional` may also have other vendor's test material. For example eclispe adoptium https://github.com/adoptium/temurin-build/tree/master/test/functional/ so it's not good to move .git to `aqa-tests/functional`

llxia: To delete everything but .git, I would prefer to use git cmd https://stackoverflow.com/questions/22339837/linux-command-to-delete-all-files-except-git-folder

```
git rm -rf .
git clean -fxd
``` 

also, `rm -rf *` is dangerous to use.

sophia-guo: Updated.

get.sh
```diff
@@ -746,7 +748,11 @@ getVendorTestMaterial() {
 		fi
 
 		# clean up
-		rm -rf $dest
+		cd $dest
+		git rm -rf .
```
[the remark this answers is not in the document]
llxia: yes, for both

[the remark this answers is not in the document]
sophia-guo: https://ci.adoptium.net/view/Test_grinder/job/Grinder/12423/

llxia (approved):
LGTM

PR #5843, opened 2025-01-09
[v1.0.5-release] Update dryrun tags for Jan 2025 release

Haroon-Khel:
Theres no jdk-23.0.2 tag in https://github.com/adoptium/jdk23u/tags so we can use jdk-23.0.1 in the dryrun


smlambert (approved):
Temporarily modifying this file to trigger dryruns.

Issue #5852, opened 2025-01-14
January 2025 Dry-run Triage - JDK17


andrew-m-leonard:

andrew-m-leonard:
aarch64 linux:

https://ci.adoptium.net/job/Test_openjdk17_hs_sanity.openjdk_aarch64_linux/460/console
Re-run successful: https://ci.adoptium.net/job/Test_openjdk17_hs_sanity.openjdk_aarch64_linux_rerun/15/

https://ci.adoptium.net/job/Test_openjdk17_hs_extended.openjdk_aarch64_linux_rerun/32/console
```
Failed test cases: 
03:23:43  			TEST: sun/security/ssl/X509TrustManagerImpl/distrust/Symantec.java
```
Re-run: https://ci.adoptium.net/job/Grinder/12260/

andrew-m-leonard:
aarch64 mac

jdk_beans_1 :https://trss.adoptium.net/output/test?id=67819be3f66194006d37ca5a
[java/beans/PropertyEditor/6380849/TestPropertyEditor.java.TestPropertyEditor](https://ci.adoptium.net/job/Test_openjdk17_hs_extended.openjdk_aarch64_mac_rerun/18/testReport/junit/java_beans_PropertyEditor_6380849_TestPropertyEditor/java/TestPropertyEditor/)
[java/beans/PropertyEditor/TestColorClass.java.TestColorClass](https://ci.adoptium.net/job/Test_openjdk17_hs_extended.openjdk_aarch64_mac_rerun/18/testReport/junit/java_beans_PropertyEditor_TestColorClass/java/TestColorClass/)
[java/beans/PropertyEditor/TestColorClassJava.java.TestColorClassJava](https://ci.adoptium.net/job/Test_openjdk17_hs_extended.openjdk_aarch64_mac_rerun/18/testReport/junit/java_beans_PropertyEditor_TestColorClassJava/java/TestColorClassJava/)
[java/beans/PropertyEditor/TestColorClassNull.java.TestColorClassNull](https://ci.adoptium.net/job/Test_openjdk17_hs_extended.openjdk_aarch64_mac_rerun/18/testReport/junit/java_beans_PropertyEditor_TestColorClassNull/java/TestColorClassNull/)
[...]

PR #5863, opened 2025-01-15
Correct PropertyEditor exclusion mac+windows excludes under one entry

andrew-m-leonard:
Some corrections for existing excludes in jdk17 and jdk21 openjdk excludes for:
- PropertyEditor * on mac and windows had two entries, and should have been under the one entry for bug JDK-8173082
- java_awt_ScrollPane entry should have included windows-all
- jdk17 JPEGsNotAcceleratedTest should be windows-all

### DIFF
--- a/openjdk/excludes/ProblemList_openjdk17.txt
+++ b/openjdk/excludes/ProblemList_openjdk17.txt
@@ -18,19 +18,19 @@
 
 # jdk_beans
 
-java/beans/PropertyEditor/6380849/TestPropertyEditor.java https://bugs.openjdk.java.net/browse/JDK-8173082 macosx-all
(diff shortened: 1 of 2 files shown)


smlambert (approved):

sophia-guo (approved):

sophia-guo:
If there are more entries only the last one takes effect, which explains why there are still failure in jdk17 mac https://github.com/adoptium/aqa-tests/issues/5852#issuecomment-2593103250. It's better to check this duplicate exclusion in https://github.com/adoptium/aqa-tests/blob/master/scripts/disabled_tests/exclude_parser.py so it can be detected when added initially.

Issue #5869, opened 2025-01-16
jdk17 win32 test runtime/cds/appcds/applications/JavacBench.java#dynamic fails due to no client CDS


andrew-m-leonard:
Newly added jdk-17 test case: https://bugs.openjdk.org/browse/JDK-8341240
jdk17 win32 test runtime/cds/appcds/applications/JavacBench.java#dynamic fails due to no client CDS

https://ci.adoptium.net/job/Test_openjdk17_hs_extended.openjdk_x86-32_windows_rerun/36/consoleFull
```
15:01:23  STDERR:
15:01:23   stdout: [[0.011s][warning][logging] No tag set matches selection: cds+resolve. Did you mean any of the following? cds* cds+heap jni+resolve cds+mirror cds+unshareable
15:01:23  [0.023s][info   ][cds    ] trying to map c:\jenkins\workspace\Test_openjdk17_hs_extended.openjdk_x86-32_windows_rerun\jdkbinary\j2sdk-image\bin\client\classes.jsa
15:01:23  [0.023s][info   ][cds    ] UseSharedSpaces: Specified shared archive not found (c:\jenkins\workspace\Test_openjdk17_hs_extended.openjdk_x86-32_windows_rerun\jdkbinary\j2sdk-image\bin\client\classes.jsa).
15:01:23  [0.024s][info   ][cds    ] UseSharedSpaces: Unable to map shared spaces
15:01:23  Error occurred during initialization of VM
15:01:23  DynamicDumpSharedSpaces is unsupported when base CDS archive is not loaded
15:01:23  ];
15:01:23   stderr: []
15:01:23   exitValue = 1
15:01:23  
15:01:23  java.lang.RuntimeException: Expected to get exit value of [0]
15:01:23  
15:01:23  	at jdk.test.lib.process.OutputAnalyzer.shouldHaveExitValue(OutputAnalyzer.java:489)
15:01:23  	at jdk.test.lib.cds.CDSAppTester.executeAndCheck(CDSAppTester.java:137)
15:01:23  	at jdk.test.lib.cds.CDSAppTester.dumpDynamicArchive(CDSAppTester.java:188)
15:01:23  	at jdk.test.lib.cds.CDSAppTester.runDynamicWorkflow(CDSAppTester.java:237)
15:01:23  	at jdk.test.lib.cds.CDSAppTester.run(CDSAppTester.java:221)
15:01:23  	at JavacBench.main(JavacBench.java:55)
```

[...]

PR #5870, opened 2025-01-16
New testcase JavacBench.java#dynamic fails on win32

andrew-m-leonard:
Exclude for https://github.com/adoptium/aqa-tests/issues/5869

jdk17 win32 test runtime/cds/appcds/applications/JavacBench.java#dynamic fails due to no client CDS


andrew-m-leonard:
> Is the absence of a client CDS a bug or intentional?

win32 we build client & server
09:14:56  * JVM variants:   client server

sophia-guo:
```
15:01:23  [0.024s][info   ][cds    ] UseSharedSpaces: Unable to map shared spaces
15:01:23  Error occurred during initialization of VM
15:01:23  DynamicDumpSharedSpaces is unsupported when base CDS archive is not loaded
15:01:23  ];
```

Can't initialization of VM I would say it's testcase issue which should not block the release. As the testcase comes from a recent backport for jdk-21+ where there is no Win32 I would say this exclusion would be a permanent.

sophia-guo:
@andrew-m-leonard did we have same issue with arm32?

smlambert (approved):

PR #5878, opened 2025-01-17
Exclude failing CDS tests on jdk17u mac-x64 as Temurin does not build a CDS

andrew-m-leonard:
We don't build a CDS on Mac x64 (ref: https://github.com/adoptium/adoptium-support/issues/937#issuecomment-2312015504)
There are some tests not excluded on jdk17u for mac x64.
Note: jdk21+ excludes all CDS tests when no-CDS ref: https://github.com/openjdk/jdk21u/blob/e45287d1ad9dcadf8a23d3271f1b675b8dade0ac/test/hotspot/jtreg/TEST.groups#L56


steelhead31 (approved):
Looks good.

andrew-m-leonard:
> I'm ok with this in principle, but is the commented out line purely so you can have a second bug URL associated with it (i.e. can we not comma separate for example?)

Correct, multiple bug URL's is not supported and in fact I think causes the exclude line to be invalid

smlambert (approved):

smlambert:
> Correct, multiple bug URL's is not supported and in fact I think causes the exclude line to be invalid

Yes, upstream tooling only supports 1 entry per testcase, so if there are multiple entries, only the last one is honoured (earlier ones ignored), so we've decided to take this approach to understand which issues relate to which platforms.

PR #5876, opened 2025-01-16
Enable rerun job with langtools_custom target

sophia-guo:
Fix #5873

PR #5885, opened 2025-01-20
[v1.0.5-release] Enable rerun job with langtools_custom target (#5876)

sophia-guo:
Cherry-pick #5876


smlambert (approved):
thanks @sophia-guo !

Issue #5929, opened 2025-01-30
Kerberos Auth login frequently fails over UDP


andrew-m-leonard:
When using the default Kerberos UDP comms for login it frequently fails:
```
Using builtin default etypes for default_tkt_enctypes
default etypes for default_tkt_enctypes: 18 17 20 19.
>>> KrbAsReq creating message
>>> KrbKdcReq send: kdc=jckservices.adoptium.net UDP:88, timeout=30000, number of retries =3, #bytes=171
>>> KDCCommunication: kdc=jckservices.adoptium.net UDP:88, timeout=30000,Attempt =1, #bytes=171
SocketTimeOutException with attempt: 1
>>> KDCCommunication: kdc=jckservices.adoptium.net UDP:88, timeout=30000,Attempt =2, #bytes=171
SocketTimeOutException with attempt: 2
>>> KDCCommunication: kdc=jckservices.adoptium.net UDP:88, timeout=30000,Attempt =3, #bytes=171
using this Java property = java.security.auth.login.config
testURL = file:/home/jenkins/jck_root/JCK11-unzipped/JCK-runtime-11a/
The machine jckservices.adoptium.net does exist.
>>> JVM will use Java Login Config file: /home/jenkins/jck_root/JCK11-unzipped/JCK-runtime-11a/lib/jck.auth.login.config
Security Manager is null
Previous Property java.security.auth.login.config is null
Current Property java.security.auth.login.config is /home/jenkins/jck_root/JCK11-unzipped/JCK-runtime-11a/lib/jck.auth.login.config
Port #60834 was allocated as a free port.  This test will use this port.
Security Manager is null
<GSSCallbackHandler> callbacks.length = 1
<GSSCallbackHandler> Inside GSSCallbackHandler, callbacks is NameCallback.
Kerberos username [aleonardf4y]:  user2/jckservices.adoptium.net@ADOPTIUM_NET
<GSSCallbackHandler> callbacks.length = 1
<GSSCallbackHandler> Inside GSSCallbackHandler, callbacks is PasswordCallback.
Kerberos XXXX for user2/jckservices.adoptium.net@ADOPTIUM_NET: XX
Status return from the server thread: Failed. Authentication took too long, time expired.
```

Symptoms indicate likely intermittent problems with Firewall configurations.
Changing to use TCP rather than UDP, always succeeds.

PR #5930, opened 2025-01-30
Change Kerberos test runner default to use TCP rather than UDP

andrew-m-leonard:
Fixes https://github.com/adoptium/aqa-tests/issues/5929

Update krb5.conf default test config to use TCP to avoid login issues.


sxa (approved):
Interesting ... We have seen this on AIX and mac (Linux less so, although we haven't been running the tests so frequently).

PR #5937, opened 2025-02-01
Forward port DeflateIn_InflateOut.java exclude for OpenJ9

pshipton:
It's already excluded jdk8 - 17, and a failure is seen on jdk24.

Issue https://github.com/eclipse-openj9/openj9/issues/14948


pshipton:
Actually, may be a different problem.

PR #5969, opened 2025-02-21
Setting upstream job status to the worst status of downstream jobs

### LINKED ISSUE #5968: Improve status handling of job defined by aqaTestPipeline.groovy
The existing strategy for determining the parent job status is based on whether child jobs succeed. If any child job doesn't succeed the parent job is marked 'Failure' https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/aqaTestPipeline.groovy#L316-L318, even if the worst child jobs status is  'unstable'. 

This approach does not effectively represent the overall state of the pipeline, as it doesn't distinguish between different failure levels of child jobs. Instead, a better strategy would be to set the parent jobs status based on the worst status among child jobs. The benefits is to provide a more accurate summary of pipeline execution, which is  really helpful when child jobs are test jobs and we need to know if jobs is unstable ( tests finished) or failed , aborted ( tests might not be finished at all).

sophia-guo:
Close #5968

### DIFF
--- a/buildenv/jenkins/aqaTestPipeline.groovy
+++ b/buildenv/jenkins/aqaTestPipeline.groovy
@@ -39,7 +39,6 @@ if (params.BUILD_TYPE == "nightly") {
 }
 
 JOBS = [:]
-fail = false
 
 timestamps {
     JDK_VERSIONS.each { JDK_VERSION ->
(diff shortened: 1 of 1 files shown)


sophia-guo:
Build status can be only set to worse than it value, no need to worry the success child job will override the unstable or failed or aborted status.

karianna (approved):

sophia-guo:
temurin-compliance/job/AQA_Test_Pipeline/ : build 7849, 7850,7851 - shows the job status is the worse of the child job status.

smlambert (approved):

PR #6014, opened 2025-03-07
(v1.0.6-release) Update openj9 related repos and branches for 0.51

llxia:


llxia:
FYI @jasonkatonica

pshipton:
@llxia The 0.51 release is for jdk8, 11, 17, 21, 24.
The 0.50 release is for jdk 24.0 but we haven't created 0.50 branches yet and are currently using 0.51 branches. 
Not sure if v1.0.6 will be used for the April release? We are about to start M1 builds for 0.51, I think we want all the versions in the testenv.

llxia:
I did not find 0.50 (only 0.51) in openj9 or extension repo, so I thought we would use 0.51 for both. Since this v1.0.6 branch is for JDK24, I will change it to 0.50 for openj9 and extension repo.

> Not sure if v1.0.6 will be used for the April release? 

Since the openj9 and extension repo branch (specially for JDK24) will be different for the April release, I think we may need to create a separate AQA branch for the April release.

llxia:
PR is updated.

llxia:
Updated the file for 0.51.

pshipton (approved):

smlambert:
FYI @pshipton 
We will not use v1.0.6-release branch for April.  We have decided for simplicity to create a new branch for each release, including the March/Sept feature releases (where we previously thought we may get away without new branch).  This makes it clearer and easier to manage going forward.  So, that being said, we will soon branch for April release as well (with v1.0.7-release as the name).

smlambert:
thanks @llxia and @pshipton for the updates and merge

pshipton:
Thanks @smlambert for the update. In theory Semeru should have already started their Milestone 1 builds, it could start any time, and should have a testenv for running the tests.

PR #6010, opened 2025-03-06
Run OpenJCEPlus tests without FIPS properties

jasonkatonica:
The mvn based OpenJCEPlus tests are not intended to be run in FIPS mode this update runs these tests whenever FIPS flags are not present.

It should be noted that these tests are still expected to run and test from a functional perspective both FIPS and non FIPS related tests.

Signed-off-by: Jason Katonica <katonica@us.ibm.com>

PR #6016, opened 2025-03-10
(v1.0.6) Run OpenJCEPlus tests without FIPS properties

LongyuZhang:
Cherry-pick of https://github.com/adoptium/aqa-tests/pull/6010
The mvn based OpenJCEPlus tests are not intended to be run in FIPS mode this update runs these tests whenever FIPS flags are not present.

It should be noted that these tests are still expected to run and test from a functional perspective both FIPS and non FIPS related tests.


llxia (approved):
LGTM

Issue #6020, opened 2025-03-10
jdk_tools testcases fail in JDK 24


smlambert:
Many tools/jpackage testcases fail in JDK 24 with

```
[01:56:40.124] Output:
    Error: This JDK does not contain packaged modules and cannot be used to create another image with the jdk.jlink module
java.lang.IllegalArgumentException: This JDK does not contain packaged modules and cannot be used to create another image with the jdk.jlink module
	at jdk.jlink/jdk.tools.jlink.internal.JlinkTask.createImageProvider(JlinkTask.java:653)
	at jdk.jlink/jdk.tools.jlink.internal.JlinkTask.createImage(JlinkTask.java:489)
	at jdk.jlink/jdk.tools.jlink.internal.JlinkTask.run(JlinkTask.java:285)
	at jdk.jlink/jdk.tools.jlink.internal.Main.run(Main.java:50)
	at jdk.jlink/jdk.tools.jlink.internal.Main$JlinkToolProvider.run(Main.java:67)
	at jdk.jpackage/jdk.jpackage.internal.JLinkBundlerHelper.runJLink(JLinkBundlerHelper.java:191)
	at jdk.jpackage/jdk.jpackage.internal.JLinkBundlerHelper.execute(JLinkBundlerHelper.java:81)
	at jdk.jpackage/jdk.jpackage.internal.AppImageBundler.createAppBundle(AppImageBundler.java:181)
	at jdk.jpackage/jdk.jpackage.internal.AppImageBundler.execute(AppImageBundler.java:93)
	at jdk.jpackage/jdk.jpackage.internal.LinuxPackageBundler.execute(LinuxPackageBundler.java:137)
	at jdk.jpackage/jdk.jpackage.internal.Arguments.generateBundle(Arguments.java:707)
	at jdk.jpackage/jdk.jpackage.internal.Arguments.processArguments(Arguments.java:554)
	at jdk.jpackage/jdk.jpackage.main.Main.execute(Main.java:92)
	at jdk.jpackage/jdk.jpackage.main.Main.main(Main.java:53)
```

Possibly related to JEP 493 changes and the upstream tests not being updated to match.

These tests will need to be temporarily excluded until they can be fixed upstream.

jerboaa:
@smlambert Yes, those are JEP 493 related. They all have fixes upstream but not all ended up in the JDK 24 GA release:

[...]

PR #5973, opened 2025-02-24
Jmj/jan 2025 problemlist

jmjaffe37:
Updated msft problemlists for the jan 2025 psu, as mentioned in: https://github.com/adoptium/aqa-tests/issues/4714

I worked with @dhanalla to compile this problemlist :)

PR #6021, opened 2025-03-10
Exclude tests related with https://bugs.openjdk.org/browse/JDK-8347124

sophia-guo:
See #6020


sophia-guo:
Wrong exclude format was merged by https://github.com/adoptium/aqa-tests/pull/5973, which didn't trigger the Disable Tests Linter check. Need to update the check.

smlambert (approved):

andrew-m-leonard (approved):

PR #6038, opened 2025-03-12
Correct testcase IDs

sophia-guo:
Related #6020


sophia-guo:
Close this one as the ` [JDK-8345185](https://bugs.openjdk.org/browse/JDK-8345185): Majority of jpackage tests. To be fixed in 24.0.2 (July).` . Can just problemlist in March release.

PR #6125, opened 2025-04-01
Using jtreg.jar as declared in TKG

judovana:
depends on https://github.com/adoptium/TKG/pull/688
Is using once declared macro


get.sh
```diff
@@ -24,8 +24,8 @@ CLONE_OPENJ9="true"
 OPENJ9_REPO="https://github.com/eclipse-openj9/openj9.git"
 OPENJ9_SHA=""
 OPENJ9_BRANCH=""
-TKG_REPO="https://github.com/adoptium/TKG.git"
-TKG_BRANCH="master"
+TKG_REPO="https://github.com/judovana/TKG.git"
+TKG_BRANCH="setupAbleJtreg"
```
judovana, line 28: This is of course temporary

judovana:
> It is unclear to me what the purpose of this PR is. We should not reference any personal repo in the main repo. If this is for personal testing, please use Grinder instead.

Of course. "This is of course temporary". But the PR do not have sense without  https://github.com/adoptium/TKG/pull/688  so I hardcoded it. This one can go in only after https://github.com/adoptium/TKG/pull/688 and ocne https://github.com/adoptium/TKG/pull/688 goes in, I will remove the hardcoded repo/branch.  The change to see is the second file

smlambert:
You do not need to create a PR in aqa-tests to test changes that are done in TKG, there is a field in our test pipeline jobs (including Grinder) that can be set TKG_OWNER_BRANCH=judovana:setupAbleJtreg.  

Additionally, if you are to create an aqa-tests branch, you do not need to create a draft or a pr to refer to it, you can push it and decline to create a PR, and still be able to refer to it.

On the topic of https://github.com/adoptium/TKG/pull/688, there does not seem to be any issue that describes why that is needed.  I presume its a feature request to support the use of custom jtreg.jar (which no one uses except RH QE as far as I am aware).  In our previous discussions on the topic, we do not want to use random untagged/unchecked test dependencies, but I can add a note to the TKG PR to repeat that previously discussed topic.

PR #6136, opened 2025-04-02
JDK24/25 Consolidate duplicate tests

babsingh:
The excludes for the below issues have been updated to close duplicate issues:
- https://github.com/eclipse-openj9/openj9/issues/21406
- https://github.com/eclipse-openj9/openj9/issues/21416

PR #6139, opened 2025-04-03
Revert "JDK24/25 Consolidate duplicate tests"

babsingh:
Reverts adoptium/aqa-tests#6136

Additional follow-up issues were observed in the tests, so the decision was made to track them individually instead of consolidating them. As a result, I'm reverting my earlier PR.


babsingh:
fyi @pshipton @JasonFengJ9

JasonFengJ9 (approved):

llxia (approved):

Issue #6051, opened 2025-03-17
March 2025 Release Triage - JDK24 release triage


sophia-guo:
https://ci.adoptium.net/job/build-scripts/job/release-openjdk24-pipeline/3/

[Grid view](https://trss.adoptium.net/resultSummary?parentId=67d9c130105697006e0e5081)

---

✔️ results in these Tables means the activity has successfully completed.
⏳ results means that we are actively working on closing off the runs needed for this version, platform, binaryType.
⛔ means there is no build planned for that version/platform combination.
⏸️ means activity not yet started.

| Platform | Status | Results | Actions |
| --- | --- | --- | --- |
| **x86-64_mac** | ✔️  | [comment](https://github.com/adoptium/aqa-tests/issues/6051#issuecomment-2734811014)  | Refer https://github.com/adoptium/aqa-tests/issues/6072 |
| **x86-64_linux** | ✔️  |  [comment](https://github.com/adoptium/aqa-tests/issues/6051#issuecomment-2734896330) | Refer https://github.com/adoptium/aqa-tests/issues/6072 |
| **x86-64_windows** | ✔️  | [comment](https://github.com/adoptium/aqa-tests/issues/6051#issuecomment-2734949811)|  |
| **aarch64_linux** |  ✔️ | [comment](https://github.com/adoptium/aqa-tests/issues/6051#issuecomment-2734961971) |  |
| **aarch64_mac** | ✔️ |  [comment](https://github.com/adoptium/aqa-tests/issues/6051#issuecomment-2734980336) |  |
| x86-64_alpine-linux | ✔️  | [comment](https://github.com/adoptium/aqa-tests/issues/6051#issuecomment-2735104697) | Refer https://github.com/adoptium/aqa-tests/issues/6072 |
| ppc64le_linux |✔️ | [comment](https://github.com/adoptium/aqa-tests/issues/6051#issuecomment-2735124683) |
| s390x_linux | ✔️ | [comment](https://github.com/adoptium/aqa-tests/issues/6051#issuecomment-2735131131) |  |
| ppc64_aix | ✔️  |  [comment](https://github.com/adoptium/aqa-tests/issues/6051#issuecomment-2735152381)|  |
| aarch64_alpine-linux |  ✔️ | [comment](https://github.com/adoptium/aqa-tests/issues/6051#issuecomment-2735138890) |  |
[...]

PR #6151, opened 2025-04-04
Disable jdk_tools for hotspot jdk24

sophia-guo:
Related with https://bugs.openjdk.org/browse/JDK-8345185

### DIFF
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -1815,6 +1815,11 @@
 		<testCaseName>jdk_tools</testCaseName>
 		<disables>
 			<disable>
+				<comment>https://bugs.openjdk.org/browse/JDK-8345185</comment>
+				<impl>hotspot</impl>
+				<version>24</version>
+			</disable>
+			<disable>
 				<comment>https://github.com/eclipse-openj9/openj9/issues/10757</comment>
 				<impl>openj9</impl>
 			</disable>


sophia-guo:
Altogether ~290 tests, each run different platforms have different testcases failures ~30. https://bugs.openjdk.org/browse/JDK-8345185 says it's a general issue and affects 16, 17, 21, 23, 24. Night tests show 21 is not affected. https://bugs.openjdk.org/browse/JDK-8345185 label show only 24 affected. 

Disable for jdk24 only.

smlambert (approved):

fly1188 (approved):

karianna (approved):

jerboaa:
@sophia-guo Excluding `:jdk_tools` test group seems to broad for https://bugs.openjdk.org/browse/JDK-8345185 I'd suggest to exclude `:jdk_jpackage` from that group instead (if that's possible).

sophia-guo:
@jerboaa unfortunately exclude :jdk_jpackage  is not supported in aqa openjdk level.

smlambert:
> @jerboaa unfortunately exclude :jdk_jpackage is not supported in aqa openjdk level.

@sophia-guo ... we could consider breaking it down into 3 smaller targets, as per https://github.com/adoptium/jdk/blob/master/test/jdk/TEST.groups#L320C1-L322C18
    :core_tools \
    :svc_tools \
    :jdk_jpackage

jerboaa:
> > @jerboaa unfortunately exclude :jdk_jpackage is not supported in aqa openjdk level.
> 
> @sophia-guo ... we could consider breaking it down into 3 smaller targets, as per https://github.com/adoptium/jdk/blob/master/test/jdk/TEST.groups#L320C1-L322C18 :core_tools :svc_tools :jdk_jpackage

+1 That would be useful.

sophia-guo:
Yes, breaking into 3 smaller targets works. However it means regroup for all versions due to this temporary specific jdk24 issue ( suppose be good for July release).  

As jdk_tool is a moderate size target, @ShelleyLambert how about we keep  jdk_tool target as is and only do the regroup in April release branch? The PR can be done when release branch is ready?

smlambert:
> As jdk_tool is a moderate size target, @ShelleyLambert how about we keep jdk_tool target as is and only do the regroup in April release branch? The PR can be done when release branch is ready.

Yes, that seems like a reasonable approach I will create the branch tonight.  Also was it really that many individual testcases to exclude in the problemlist?  I recalled seeing only a handful (around 10).

sophia-guo:
yes, see https://github.com/adoptium/aqa-tests/issues/6051#issuecomment-2787411045. This is  for jdk24 only. If I have time I will quickly go through Jan release.

sophia-guo:
https://github.com/adoptium/aqa-tests/pull/6164

PR #6179, opened 2025-04-11
exclude JDK8 ValidateISO4217 test

sxa:
If this needs to have an issue in here and not just a reference to the OpenJ9 one let me know.

This has been seen on the Temurin builds too and is potentially a date-related issue as per [this comment](https://github.com/eclipse-openj9/openj9/issues/21539#issuecomment-2769448076) . Excluded for OpenJ9 in https://github.com/adoptium/aqa-tests/pull/6126

We'll also need to cherry-pick this across to the release branch when approved.

### DIFF
--- a/openjdk/excludes/ProblemList_openjdk8.txt
+++ b/openjdk/excludes/ProblemList_openjdk8.txt
@@ -361,6 +361,7 @@ com/sun/jdi/PrivateTransportTest.sh https://github.com/adoptium/aqa-tests/issues
 # jdk_util
 
 java/util/Arrays/TimSortStackSize2.java https://github.com/adoptium/aqa-tests/issues/1051 linux-s390x
+java/util/Currency/ValidateISO4217.java https://github.com/eclipse-openj9/openj9/issues/21539 generic-all
 java/util/Random/RandomTest.java https://github.com/adoptium/aqa-tests/issues/1051 linux-s390x
 java/util/ResourceBundle/Control/Bug6530694.java https://github.com/adoptium/aqa-tests/issues/137 macosx-all
 java/util/Spliterator/SpliteratorTraversingAndSplittingTest.java https://github.com/adoptium/aqa-tests/issues/1051 linux-s390x


sxa:
Upstream issue: https://github.com/openjdk/jdk8u-dev/pull/646

gnu-andrew:
It's definitely time related. The culprit is:
~~~
CW=ANG;2025-04-01-04-00-00;XCG
~~~
and similar for `SX`. Unfortunately, [the code in 8u which handles future currencies](https://github.com/openjdk/jdk8u/commit/439e910420ab9e1ca7c4645bb20634bc7cdbbe57) doesn't seem to have realised that this is no longer a future currency and so you get the `IllegalArgumentException`.  We have a quick fix of just dumping the historical data which resolves the test. The long-term fix is to make the logic actually work as it should (https://bugs.openjdk.org/browse/JDK-8353579)

andrew-m-leonard (approved):

openjdk/excludes/ProblemList_openjdk8.txt
```diff
@@ -361,6 +361,7 @@ com/sun/jdi/PrivateTransportTest.sh https://github.com/adoptium/aqa-tests/issues
 # jdk_util
 
 java/util/Arrays/TimSortStackSize2.java https://github.com/adoptium/aqa-tests/issues/1051 linux-s390x
+java/util/Currency/ValidateISO4217.java https://github.com/eclipse-openj9/openj9/issues/21539 generic-all
```
sendaoYan, line 364: I think it's better problemlist this testcase in [ProblemList_openjdk8-openj9.txt](https://github.com/adoptium/aqa-tests/blob/master/openjdk/excludes/ProblemList_openjdk8-openj9.txt), becase this test run passed by openjdk8, it only failed runed by openj9?

sendaoYan (changes requested):

openjdk/excludes/ProblemList_openjdk8.txt
sxa: No - as per the description `This has been seen on the Temurin builds too` (which are not J9)
Alhuogh since the upstream fix has gone in we probably don't need to exclude this any more. I'll re-check it.

PR #6223, opened 2025-04-25
Update arm32 testenv properties for release

steelhead31:
Update arm32 testenv properties for release


steelhead31:
Incorrectly branched, recreating

PR #6228, opened 2025-04-25
Porting JDK24 excludes to JDK25

JasonFengJ9:
Porting JDK24 excludes to JDK25 

Also adjusted entries in alphabetical order.

### DIFF
--- a/openjdk/excludes/ProblemList_openjdk25-openj9.txt
+++ b/openjdk/excludes/ProblemList_openjdk25-openj9.txt
@@ -22,16 +22,16 @@
 
 # jdk_beans
 
-java/beans/PropertyEditor/6380849/TestPropertyEditor.java https://github.com/eclipse-openj9/openj9/issues/20531 macosx-all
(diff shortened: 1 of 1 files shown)


pshipton:
Pls split this into two PRs, one to sort alphabetically, and another to add the missing excludes (or vice versa).

JasonFengJ9:
> Pls split this into two PRs, one to sort alphabetically, and another to add the missing excludes (or vice versa).

This is the `ProblemList_openjdk24-openj9.txt` sorted alphabetically.
Since there was no JDK25 specific exclude, i.e., the excludes only for JDK25, some excludes were added into JDK24 only which should be applied to JDK25 as well if we start running JDK25 tests regularly. It doesn't seem necessary to copy `ProblemList_openjdk24-openj9.txt` into `ProblemList_openjdk25-openj9.txt` first, then sort it in another PR.
The reason not sorting `ProblemList_openjdk24-openj9.txt` is that JDK24 is not LTS, and will be outdated in a few months.

pshipton (approved):

PR #6234, opened 2025-04-29
Exclude JDK 24/25 tests on selected platforms

JasonFengJ9:
Exclude JDK 24/25 tests on selected platforms


tajila:
I would remove excludes only from xlinux for now, then we can try amac after.

JasonFengJ9:
> I would remove excludes only from xlinux for now, then we can try amac after.

Excluded `linux-aarch64,linux-ppc64le,linux-s390x,macosx-all,windows-x64`.

pshipton (approved):

llxia (approved):

PR #6242, opened 2025-05-02
Exclude java/nio/channels/AsyncCloseAndInterrupt.java aix-all,linux-all

JasonFengJ9:
Exclude `java/nio/channels/AsyncCloseAndInterrupt.java` `aix-all,linux-all`

Match Openjdk and other OpenJ9 version excludes.

Related to
*  https://github.com/eclipse-openj9/openj9/issues/21777

### DIFF
--- a/openjdk/excludes/ProblemList_openjdk11-openj9.txt
+++ b/openjdk/excludes/ProblemList_openjdk11-openj9.txt
@@ -219,7 +219,8 @@ java/io/FileOutputStream/UnreferencedFOSClosesFd.java https://github.com/eclipse
 java/nio/Buffer/DirectBufferAllocTest.java https://github.com/eclipse-openj9/openj9/issues/4473 generic-all
 java/nio/Buffer/LimitDirectMemory.java	https://github.com/eclipse-openj9/openj9/issues/8063	generic-all
 java/nio/Buffer/LimitDirectMemoryNegativeTest.java	https://github.com/eclipse-openj9/openj9/issues/8065	generic-all
-java/nio/channels/AsyncCloseAndInterrupt.java https://github.com/adoptium/aqa-tests/issues/6092 z/OS-s390x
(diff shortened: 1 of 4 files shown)


pshipton (approved):

pshipton:
What about 21?

pshipton:
Why is it being excluded on linux, does it fail on linux too?

JasonFengJ9:
`java/nio/channels/AsyncCloseAndInterrupt.java` was excluded for `aix-all,linux-all` in all other `ProblemList_openjdk_xxx.txt` files. This is to match OpenJDK and other OpenJ9 versions.

pshipton:
But it it's not failing for Semeru on Linux, why would we exclude it there? If it is failing on Linux, the openj9 issue(s) should mention this. We don't need to duplicate what adoptium is doing. 

I see excludes for dead version and for jdk24+, but not for jdk21. Does it not fail there as well?

pshipton:
For the excludes jdk24+, they are excluded via https://github.com/adoptium/infrastructure/issues/1173 but this issue is closed saying the tests run successfully and the problematic machines are gone.

To be clear, pls exclude it on platforms where it doesn't work on Semeru infrastructure, but do it consistently and pls document it properly.

JasonFengJ9:
> But if it's not failing for Semeru on Linux, why would we exclude it there? If it is failing on Linux, the openj9 issue(s) should mention this. We don't need to duplicate what adoptium is doing.

So farm Semeru failure was only observed on JDK11 AIX, this PR added it to `ProblemList_openjdk11-openj9.txt`, updated `ProblemList_openjdk17-openj9.txt`, also removed `java/nio/channels/AsyncCloseAndInterrupt.java` exclude from JDK24+ since JDK21 doesn't have it. It can be added back if JDK24+ fails this test.

@pshipton this is ready for another look.

llxia (approved):

pshipton (approved):

Issue #6251, opened 2025-05-09
get.sh doesn't handle extra `*-jmods.tar.gz` archive correctly


jerboaa:
Currently smoke tests for JDK 25 fail when calling `get.sh` with something like this:

```
16:47:49  + ./get.sh -s /home/jenkins/workspace/build-scripts/jobs/jdk/jdk-linux-x64-temurin_SmokeTests/aqa-tests/.. -p x86-64_linux -r upstream -j 25 -i hotspot --clone_openj9 true --openj9_repo https://github.com/eclipse-openj9/openj9.git --openj9_branch master --tkg_repo https://github.com/adoptium/TKG.git --tkg_branch master --vendor_repos https://github.com/adoptium/temurin-build --vendor_branches master --vendor_dirs /test/functional
16:47:49  TESTDIR: /home/jenkins/workspace/build-scripts/jobs/jdk/jdk-linux-x64-temurin_SmokeTests/aqa-tests
16:47:49  get jdk binary...
16:47:49  --sdkdir is set to upstream. Therefore, skip download jdk binary
16:47:49  Uncompressing file: OpenJDK-jdk-sources_25_22-ea.tar.gz ...
16:47:53  List files in jdkbinary folder...
16:47:53  total 780060
16:47:53  -rw-r--r-- 1 jenkins jenkins 272564173 May  9 14:36 OpenJDK-debugimage_x64_linux_hotspot_25_22-ea.tar.gz
16:47:53  -rw-r--r-- 1 jenkins jenkins 119672045 May  9 13:50 OpenJDK-jdk-sources_25_22-ea.tar.gz
16:47:53  -rw-r--r-- 1 jenkins jenkins 138769445 May  9 14:36 OpenJDK-jdk_x64_linux_hotspot_25_22-ea.tar.gz
16:47:53  -rw-r--r-- 1 jenkins jenkins  84771867 May  9 14:36 OpenJDK-jmods_x64_linux_hotspot_25_22-ea.tar.gz
16:47:53  -rw-r--r-- 1 jenkins jenkins  59828932 May  9 14:34 OpenJDK-jre_x64_linux_hotspot_25_22-ea.tar.gz
16:47:53  -rw-r--r-- 1 jenkins jenkins    168160 May  9 14:42 OpenJDK-sbom_x64_linux_hotspot_25_22-ea.json
[...]

PR #6253, opened 2025-05-12
Skip JMOD files when extracting the JDK

jerboaa:
Closes: #6251

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -399,7 +399,8 @@ getBinaryOpenjdk()
 
 	for file_name in "${jdk_file_array[@]}"
 	do
-		if [[ ! "$file_name" =~ "sbom" ]]; then
+		# Skip sbom and jmods archives when extracting
+		if [[ ! "$file_name" =~ "sbom" ]] && [[ ! "$file_name" =~ "jmods" ]]; then
 			if [[ $file_name == *xz ]]; then
 				DECOMPRESS_TOOL=xz
 			else


jerboaa:
Test build with this is here:
https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk/job/jdk-linux-x64-temurin_SmokeTests/252/

Fails earlier here (without this patch):
https://ci.adoptium.net/job/build-scripts/job/jobs/job/jdk/job/jdk-linux-x64-temurin_SmokeTests/251

gdams (approved):

andrew-m-leonard (approved):
looks good

sophia-guo:
Would it be better to un-tar those jmod tar files as separate directory? similar to static-libs?
[elif [[ "$jar_dir_name" =~ "static-libs" ]]; then
						moveDirectorySafely $jar_dir_name ../static-libs](https://github.com/adoptium/aqa-tests/blob/master/get.sh#L467C6-L469)

Just in case they are needed for some tests either existing one or might be added later? @jerboaa

jerboaa:
> Would it be better to un-tar those jmod tar files as separate directory? similar to static-libs? [elif [[ "$jar_dir_name" =~ "static-libs" ]]; then moveDirectorySafely $jar_dir_name ../static-libs](https://github.com/adoptium/aqa-tests/blob/master/get.sh#L467C6-L469)

I wondered why that was being done, TBH. It seemed strange without precisely knowing that it's actually needed. Is this used in any test?

> Just in case they are needed for some tests either existing one or might be added later? @jerboaa

I'd suggest to only do this once the first use-case arises. By default extracting them just consumes disk space and unless a cross-link test is being added, it won't be needed (thanks to JEP 493).

PR #6259, opened 2025-05-13
Support vmfarm SDK mapping

llxia:
resolves: automation/issues/411

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -213,8 +213,39 @@ getBinaryOpenjdk()
 
 	if [ "$SDK_RESOURCE" == "nightly" ] && [ "$CUSTOMIZED_SDK_URL" != "" ]; then
 		if [[ ! "${CUSTOMIZED_SDK_URL}" =~ /$ ]]; then
-    			CUSTOMIZED_SDK_URL="${CUSTOMIZED_SDK_URL}/"
+			CUSTOMIZED_SDK_URL="${CUSTOMIZED_SDK_URL}/"
 		fi
+
+		if [[ $CUSTOMIZED_SDK_URL = *"vmfarm"* ]]; then
+			subdir=""
+			if [ $PLATFORM == "ppc32_aix" ]; then
+				subdir="ap3280"
+			elif [ $PLATFORM == "ppc64_aix" ]; then
+				subdir="ap6480"
+			elif [ $PLATFORM == "ppc64le_linux" ]; then
+				subdir="xl6480"
+			elif [ $PLATFORM == "s390x_linux" ]; then
+				subdir="xz6480"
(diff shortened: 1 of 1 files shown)


llxia:
- openj9 URL: [link](https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/50286/)
- Base URL contains `artifactory`: [link](https://hyc-runtimes-jenkins.swg-devops.com/job/Test_openjdk8_ibm_sanity.functional_s390x_linux/8/)
- Base URL contains `ui/native`: [link](https://hyc-runtimes-jenkins.swg-devops.com/job/Test_openjdk8_ibm_sanity.functional_s390x_linux/7/)
- [All platforms](https://hyc-runtimes-jenkins.swg-devops.com/job/IBM_Java8_J9_Integration/9/)

LongyuZhang (approved):
LGTM

get.sh
```diff
@@ -213,8 +213,39 @@ getBinaryOpenjdk()
...
+				subdir="wi3280"
+			elif [ $PLATFORM == "x86-64_windows" ]; then
+				subdir="wa6480"
+			else
+				echo "There is no $PLATFORM to Artifactory vmfarm SDK mapping. Please check PLATFORM value or update the map."
+				exit 1
+			fi
+			CUSTOMIZED_SDK_URL="${CUSTOMIZED_SDK_URL}${subdir}/"
+		fi
+		CUSTOMIZED_SDK_URL=(${CUSTOMIZED_SDK_URL//\/ui\/native\//\/artifactory\/})
```
pshipton, line 247: What does `//\/ui\/native\//` mean?

llxia: In the internal Artifactory, accessing the URL that contains `artifactory` in a browser automatically redirects to `ui/native`. As a result, users may provide two types of URLs: one containing `artifactory` and another with `ui/native`. This auto-redirect behavior does not occur in the external Artifactory.

In the pipeline, we query the base SDK Artifactory URL to retrieve a list of folders, determine the latest one, and use it accordingly. To support both URL formats consistently, the code replaces `ui/native` with `artifactory`, ensuring a unified output for processing.

I think `ui/native` is a feature in Artifactory to support native browser UI.

pshipton: Ok. I was asking more about the syntax, I'm not familiar with `//\/ui\/native\//\/artifactory\/`. We don't need two more slashes at the end?

llxia: I do not think so. 
```
${string//substring/replacement}
Replace all matches of $substring with $replacement.
```
https://tldp.org/LDP/abs/html/string-manipulation.html

pshipton (approved):

PR #3774, opened 2022-06-17
Use retry function to clone openjdk test material

renfeiw:
Signed-off-by: renfeiw <renfeiw@ca.ibm.com>

PR #6296, opened 2025-05-29
(v1.0.8) Hardcode Semeru M1 openjdk SHAs to avoid test failures

pshipton:


llxia (approved):
LGTM

llxia:
We may also want to check if `JDKxxx_BRANCH` will work for sha if we remove `getFileWithRetry` in the future. Related: https://github.com/adoptium/aqa-tests/pull/3774/files

PR #6307, opened 2025-06-04
Update j9 image name

LongyuZhang:
Update j9 image name

Issue: ibm_git/runtimes/backlog/issues/1542

### DIFF
--- a/external/provider.sh
+++ b/external/provider.sh
@@ -107,7 +107,8 @@ function getFullOpenJ9Image() {
   local JDK_VERSION="${1:-0}"
   local jreSdk="${2:-jdk}"
   if [ -z "${EXTERNAL_AQA_IMAGE}" ]; then
-    echo "docker.io/library/ibm-semeru-runtimes:open-${JDK_VERSION}-$jreSdk"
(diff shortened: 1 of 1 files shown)


external/provider.sh
```diff
@@ -107,7 +107,7 @@ function getFullOpenJ9Image() {
   local JDK_VERSION="${1:-0}"
   local jreSdk="${2:-jdk}"
   if [ -z "${EXTERNAL_AQA_IMAGE}" ]; then
-    echo "docker.io/library/ibm-semeru-runtimes:open-${JDK_VERSION}-$jreSdk"
+    echo "icr.io/appcafe/ibm-semeru-runtimes:open-${JDK_VERSION}-$jreSdk-jammy"
```
llxia: Is it possible to use a variable for this echo? If the actual image changes in the future, this echo message will be misleading.

ixlongz: Updated to use a variable. This `echo` actually output the commands to dockerfile directly. 
New test link: hyc_grinder 51224

llxia: Just for the record, talked with @LongyuZhang , this is not echo the message to the console output. This is to insert the command line into the dockerfile.

llxia (approved):

Issue #6269, opened 2025-05-21
Remove MachineInfo target in system tests


llxia:
MachineInfo was initially created to ensure we have test in each system test folder. For example: https://github.com/adoptium/aqa-tests/blob/13d358513f48d93f680ecf779281ead82063259f/system/otherLoadTest/playlist.xml#L8

Currently, we run MachineInfo in TKG. And as we changed the way of running the parallelization (dynamic smart parallel instead of folder-based parallel), I believe we do not need it anymore. Below is an example where multiple MachineInfo are put into one testList and executed multiple times in one child build.

![Image](https://github.com/user-attachments/assets/2c8065f7-b7c6-456c-9b01-c5d5c1563fea)

```
00:06:30.728  microArchOutput: 
00:06:34.698  ****************************** MACHINE INFO ******************************
00:06:34.698  File path : /home/jenkins/workspace/Test_openjdk11_j9_sanity.system_aarch64_linux_testList_3/aqa-tests/TKG/output_17477844866060/MachineInfo_0
00:06:34.698  Free Physical Memory Size : 5227606016
00:06:34.698  Free space (bytes) : 78290288640
00:06:34.698  Total Physical Memory Size : 8311341056
00:06:34.698  Total space (bytes) : 115771248640
00:06:34.698  Usable space (bytes) : 78290288640
00:06:34.698  antVersion : Apache Ant(TM) version 1.10.5 compiled on July 10 2018
00:06:34.698  bashVersion : 
00:06:34.698  	GNU bash, version 4.4.23(1)-release (aarch64-suse-linux-gnu)
00:06:34.698  	Copyright (C) 2016 Free Software Foundation, Inc.
00:06:34.698  	License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
00:06:34.698  	
00:06:34.698  	This is free software; you are free to change and redistribute it.
00:06:34.698  	There is NO WARRANTY, to the extent permitted by law.
00:06:34.698  cpuCores : 8
...
```

PR #6332, opened 2025-06-13
Remove machine info target from system playlists

MattyWeee123:
Removed machine info target from system playlists since it's already run in TKG level. 

Fixes: https://github.com/adoptium/aqa-tests/issues/6269


llxia:
@MattyWeee123 could you change the PR title from `Issue6269` to `Remove machine info target from system playlists`? Thanks

LongyuZhang (approved):
LGTM

Issue #4858, opened 2023-11-06
Trial the latest Dacapo release candidate v23.10-RC5


smlambert:
Create a branch to trial run https://github.com/dacapobench/dacapobench/releases/tag/v23.10-RC5-chopin ahead of GA (with the plan to move to it, if all looks well).

sxa:
Commenting so I get notifications - interested as I've seen flakiness on dacapo-xalan on arm32 and possibly other platforms :-)

smlambert:
https://download.dacapobench.org/chopin/dacapo-23.11-chopin.zip is 6.34 Gb, as it contains 2.4 Mb dacapo.jar file and a copious amount of test data used by the benchmarks.  

Will look at what is actually needed, as we do not want to be downloading everything on every run.

smlambert:
Now available: https://github.com/dacapobench/dacapobench/releases/tag/v23.11-MR1-RC1-chopin

llxia:
There is a [dacapo-23.11-MR2-chopin-minimal.zip](https://download.dacapobench.org/chopin/dacapo-23.11-MR2-chopin-minimal.zip), but it is still 1.3G. (see https://github.com/dacapobench/dacapobench/issues/345#issuecomment-2809300463)

llxia:
FYI @MattyWeee123

MattyWeee123:
Hi, I'd like to work on this!

MattyWeee123:
Using dacapo-23.11-MR2-chopin-minimal and Grinder parameters TARGET = perf, BUILD_LIST = perf/dacapo, Grinder is unstable for the following tests: 

x86-64_windows, jdk 8 (13291): 
dacapo-eclipse_0 => Java version '1.8.0_462-beta' is incompatable with eclipse: Eclipse requires Java 11 or newer.
dacapo-h2_0 => Java version '1.8.0_462-beta' is incompatable with h2: H2 requires Java 11 or newer.
dacapo-jython_0 => java.lang.ClassNotFoundException: com.kenai.jffi.Type, more info [here](https://ci.adoptium.net/job/Grinder/13291/console)
dacapo-luindex_0 => Java version '1.8.0_462-beta' is incompatable with luindex: Luindex requires Java 11 or newer.
dacapo-pmd_0 => Digest validation failed for pmd-report.txt, expecting 0x9ac4266f6d867118bf6365f3c935b97526705c2c found 0xf7fb29acd687098cb9646d419c36d553144b414b

[...]

PR #6338, opened 2025-06-14
Exclude failed dacapo tests 

MattyWeee123:
Exclude decapo tests that failed. Using 23.11 minimal (Disabled pmd, fop, jython, h2, eclipse, luindex)

PR #6341, opened 2025-06-15
Trial the latest Dacapo v23.11-chopin

MattyWeee123:
Fixes: #4858

### DIFF
--- a/perf/dacapo/build.xml
+++ b/perf/dacapo/build.xml
@@ -28,8 +28,10 @@
 
 	<target name="getDacapoSuite" depends="init">
 		<var name="FILE_NAME" value="dacapo.jar"/>
+    <var name="DIR_NAME" value="dacapo"/>
 		<var name="ZIP_FILE_NAME" value="dacapo-23.11-MR2-chopin.zip"/>
 		<var name="JAR_FILE_NAME" value="dacapo-23.11-MR2-chopin.jar"/>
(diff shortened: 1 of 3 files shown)


perf/dacapo/build.xml
```diff
@@ -21,32 +21,50 @@
...
+        <property name="DACAPO_DIR_NAME" value="dacapo-23.11-MR2-chopin-minimal"/>
+
+        <!-- value of fname in dacapo hash from getDependencies.pl -->
+        <property name="DEPENDENCY_FNAME" value="dacapo.zip"/>
+
+        <property name="LIB" value="dacapo"/>
+	
+        <target name="getDependentLibs">
+        <exec executable="perl" failonerror="true">
+                <arg line="${TEST_ROOT}/TKG/scripts/getDependencies.pl -path ${LIB_DIR} -task default -dependencyList ${LIB} -curlOpts -L"/>
```
llxia, line 33: We should not use getDependencies.pl. Please just update the original code `<var name="curl_options" value="-Lks -C - https://sourceforge.net/projects/dacapobench/files/latest/download -o dacapo.jar"/>` to download the dacapo-23.11-MR2-chopin-minimal jar.

llxia: Also, please see https://github.com/adoptium/aqa-tests/pull/6338#issuecomment-2980384450

Issue #6351, opened 2025-06-20
Auto generate 2nd-level job in Perf Pipeline


llxia:
Currently, the 2nd-level job needs to be generated manually. We need to automatically generate the 2nd-level jobs.

https://github.com/adoptium/aqa-tests/blob/c2cb30403aceac91d61b1840daf93f32093c42fc/buildenv/jenkins/perfPipeline_root.groovy#L31

reference: https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/testJobTemplate
how to use trigger it: https://github.com/adoptium/aqa-tests/blob/c2cb30403aceac91d61b1840daf93f32093c42fc/buildenv/jenkins/JenkinsfileBase#L427

previous work: https://github.com/adoptium/aqa-tests/pull/5695/files

llxia:
Given the fact that https://github.com/adoptium/aqa-tests/pull/5695/files was created 6 months ago, it may be easier to write a new DSL job based on https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/testJobTemplate. Our goal is to run the DSL in L1 and generate L2. Ideally, we should not loop PLATFORMS and trigger the L2 jobs regardless (as follows).
https://github.com/adoptium/aqa-tests/blob/479eb0e9ec4d501cc65a23d33d03f6e1b09b874a/buildenv/jenkins/perfPipeline_root.groovy#L7

It should use logic in L2, where the PLATFORM should match with perfConfigJson value. And L2 does not need to process PLATFORM and perfConfigJson anymore. L2 only gets one PLATFORM.

https://github.com/adoptium/aqa-tests/blob/479eb0e9ec4d501cc65a23d33d03f6e1b09b874a/buildenv/jenkins/perfPipeline.groovy#L91

In the DSL job, we should
- Add required parameters - all parameters in https://ci.adoptium.net/job/Perf_openjdk21_hs_sanity.perf_x86-64_linux_dacapo/build?delay=0sec except  PLATFORMS. PLATFORMS should be PLATFORM (no s). We expect a single platform.
- Remove unnessary parameters
- update SCRIPT_PATH to point to perfPipeline.groovy in aqa-test repo https://github.com/adoptium/aqa-tests/blob/479eb0e9ec4d501cc65a23d33d03f6e1b09b874a/buildenv/jenkins/testJobTemplate#L491C20-L491C31

PR #5695, opened 2024-10-16
 Add new template to enable auto generate parent level job in perf test

sophiaxu0424:
This change is to create a new scm script to for auto generate job usage

Fixes: https://github.ibm.com/runtimes/automation/issues/69

PR #6360, opened 2025-06-23
If the job is not runnable, use createJob.

MattyWeee123:
If the job has not been created, use a local
createJob that is a modified version of the one
found in JenkinsfileBase to automatically create
the job.

Fixes: #6351

[Grinder](https://ci.adoptium.net/view/Test_grinder/job/Perf_Pipeline/59/)

### DIFF
--- a/buildenv/jenkins/parentJobTemplate
+++ b/buildenv/jenkins/parentJobTemplate
@@ -15,7 +15,6 @@
 /**
  * A template that defines a parent perf job.
  */
-
 def getBaseJDKImpl(JDK_IMPL) {
 	if (JDK_IMPL != "openj9" && JDK_IMPL != "ibm" ) {
 		return "hotspot"
@@ -33,7 +32,6 @@ if (!binding.hasVariable('VENDOR_TEST_REPOS')) VENDOR_TEST_REPOS = ""
 if (!binding.hasVariable('VENDOR_TEST_BRANCHES')) VENDOR_TEST_BRANCHES = ""
 if (!binding.hasVariable('VENDOR_TEST_DIRS')) VENDOR_TEST_DIRS = "perf"
 if (!binding.hasVariable('VENDOR_TEST_SHAS')) VENDOR_TEST_SHAS = ""
-
 
 if (binding.hasVariable('ARCH_OS_LIST')) {
 	ARCH_OS_LIST = ARCH_OS_LIST.split(',')*.trim()
--- a/buildenv/jenkins/parentJobTemplate
+++ b/buildenv/jenkins/parentJobTemplate
@@ -1,0 +1,163 @@
+#!groovy
+
+/*
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+      https://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+/**
+ * A template that defines a parent perf job.
+ */
+
+def getBaseJDKImpl(JDK_IMPL) {
+	if (JDK_IMPL != "openj9" && JDK_IMPL != "ibm" ) {
+		return "hotspot"
+	}
+	return JDK_IMPL
+}
+
+if (!binding.hasVariable('JDK_IMPL')) JDK_IMPL = "openj9"
+if (!binding.hasVariable('TIME_LIMIT')) TIME_LIMIT = "10"
+if (!binding.hasVariable('KEEP_REPORTDIR')) KEEP_REPORTDIR = true
+if (!binding.hasVariable('SDK_RESOURCE')) SDK_RESOURCE = "upstream"
+if (!binding.hasVariable('JDK_VERSION')) JDK_VERSION = ""
+if (!binding.hasVariable('CUSTOMIZED_SDK_URL_CREDENTIAL_ID')) CUSTOMIZED_SDK_URL_CREDENTIAL_ID = ""
+if (!binding.hasVariable('VENDOR_TEST_REPOS')) VENDOR_TEST_REPOS = ""
+if (!binding.hasVariable('VENDOR_TEST_BRANCHES')) VENDOR_TEST_BRANCHES = ""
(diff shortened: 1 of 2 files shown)


buildenv/jenkins/perfPipeline_root.groovy
```diff
@@ -29,10 +30,43 @@ PLATFORMS.each { PLATFORM ->
...
+		jobParams.put('DAYS_TO_KEEP', DAYS_TO_KEEP)
+	}
+
+	if (params.BUILDS_TO_KEEP) {
+		jobParams.put('BUILDS_TO_KEEP', BUILDS_TO_KEEP)
+	}
+
+	def templatePath = 'aqa-tests/buildenv/jenkins/testJobTemplate'
+	if (!fileExists(templatePath)) {
+		sh "curl -Os https://raw.githubusercontent.com/adoptium/aqa-tests/master/buildenv/jenkins/testJobTemplate"
```
llxia: TestJobTemplate creates the child test job (L3). We need to create a new template for the Perf parent job (L2) only.

buildenv/jenkins/parentJobTemplate
```diff
@@ -0,0 +1,163 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+/**
+ * A template that defines a parent perf job.
+ */
+
```
karianna:
```suggestion
```

buildenv/jenkins/parentJobTemplate
```diff
@@ -0,0 +1,163 @@
...
+if (!binding.hasVariable('KEEP_REPORTDIR')) KEEP_REPORTDIR = true
+if (!binding.hasVariable('SDK_RESOURCE')) SDK_RESOURCE = "upstream"
+if (!binding.hasVariable('JDK_VERSION')) JDK_VERSION = ""
+if (!binding.hasVariable('CUSTOMIZED_SDK_URL_CREDENTIAL_ID')) CUSTOMIZED_SDK_URL_CREDENTIAL_ID = ""
+if (!binding.hasVariable('VENDOR_TEST_REPOS')) VENDOR_TEST_REPOS = ""
+if (!binding.hasVariable('VENDOR_TEST_BRANCHES')) VENDOR_TEST_BRANCHES = ""
+if (!binding.hasVariable('VENDOR_TEST_DIRS')) VENDOR_TEST_DIRS = "perf"
+if (!binding.hasVariable('VENDOR_TEST_SHAS')) VENDOR_TEST_SHAS = ""
+
+
```
karianna:
```suggestion
```

buildenv/jenkins/parentJobTemplate
```diff
@@ -0,0 +1,163 @@
...
+if (binding.hasVariable('ARCH_OS_LIST')) {
```
karianna, line 36: How does this if condition occur?

buildenv/jenkins/parentJobTemplate
```diff
@@ -0,0 +1,163 @@
...
+
+		properties {
+			copyArtifactPermission {
+				projectNames('*')
+			}
+		}
+
+		definition {
+			parameters {
+				stringParam('ADOPTOPENJDK_REPO', ADOPTOPENJDK_REPO, "AdoptOpenJDK git repo. Please use ssh for zos.")
```
karianna, line 93: Should we be replacing this naming with Adoptium?

karianna: Q for @llxia As I'm not sure how that propogates.

buildenv/jenkins/parentJobTemplate
```diff
@@ -0,0 +1,163 @@
...
+				stringParam('ADOPTOPENJDK_BRANCH', ADOPTOPENJDK_BRANCH, "AdoptOpenJDK branch")
+				stringParam('PERFORMANCE_REPO', PERFORMANCE_REPO, "PERF git repo. For PERF test only")
+				stringParam('PERFORMANCE_BRANCH', PERFORMANCE_BRANCH, "PERF branch. For PERF test only")
+
+				stringParam('PLATFORM', "${ARCH_OS}", '''Platform to be tested:
+                    <br/>arm_linux
```
karianna, line 99: Pretty sure we've dropped some of these platforms (arm32, win32), unless this is IBM J9 specific support?

karianna:
@MattyWeee123 Looks like Eclipse CLA needs signing as well.  LMK if you need a link

MattyWeee123:
> @MattyWeee123 Looks like Eclipse CLA needs signing as well. LMK if you need a link

Should be good now

buildenv/jenkins/parentJobTemplate
MattyWeee123: I think it should be the 'ARCH_OS_LIST' that I passed as additionalParameter in the jobDsl in perfPipeline_root

karianna: OK, so it's optionally passed in, and if not you effectively say: "Do all archs?"

MattyWeee123: For context, I borrowed this from [https://github.com/adoptium/aqa-tests/pull/5695/files](https://github.com/adoptium/aqa-tests/pull/5695/files). 

I was unsure of this purpose, but since the name doesn't change and the dsl plugin should be[ "updating previously generated jobs and views"](https://www.jenkins.io/doc/pipeline/steps/job-dsl/?utm_source=chatgpt.com), it seem like it's overwriting the same job until it has x86-64_windows?

Would it be preferable to have a single default platform or make ARCH_OS_LIST mandatory to pass in?

karianna: I'll defer to @llxia on that one, I suspect having a default if nothing gets passed in is good, but I'm not sure that default should be all.

llxia: https://github.com/adoptium/aqa-tests/pull/5695/files is just an example from a previous attempt that was never merged. It contains outdated logic and needs to be updated and cleaned up. We can't use it as-is.

That being said, `ARCH_OS_LIST` is correct. The job that triggers it should set `ARCH_OS_LIST` (the value should  be `PLATFORM`).

buildenv/jenkins/parentJobTemplate
llxia: Currently, `ADOPTOPENJDK_REPO` is used in AQA repo for all other jobs. We cannot use Adoptium because all downstream jobs are using ADOPTOPENJDK.

Renaming to Adoptium should be a separate task and handled with caution and coordination with all sites.

PR #6368, opened 2025-06-26
Updated 2nd-level job generation.

MattyWeee123:
1. According to L2 config UI, constructed perfL2JobTemplate
and updated SCRIPT_PATH to point to perfPipeline.groovy.

2. Assuming there is only one PLATFORM provided by params.PLATFORMS.trim()
in perfPipeline_root. This PLATFORM is used to generate the L2 jobs.

3. Since "if (params.PLATFORM == platform)'' must hold to trigger the L3
jobs, directly grab the machine and params.PLATFORM pair with a null check
rather than iterating through the entire PLAT_MACHINE_MAP.

Fixes: #6351

Signed-off-by: Matthew Wei <mwei2@andrew.cmu.edu>


buildenv/jenkins/perfL2JobTemplate
```diff
@@ -0,0 +1,145 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+/**
+ * A template that defines a L2 perf job.
+ */
+
```
karianna, line 20:
```suggestion
```

buildenv/jenkins/perfL2JobTemplate
```diff
@@ -0,0 +1,145 @@
...
+def getBaseJDKImpl(JDK_IMPL) {
```
karianna, line 21: is JDK_IMPL always not null? Wouldn't users just pass in "hotspot" if that's what's required?

buildenv/jenkins/perfL2JobTemplate
```diff
@@ -0,0 +1,145 @@
...
+def ADOPTOPENJDK_BRANCH = "master"
+
+// https://issues.jenkins.io/browse/JENKINS-42971 is resolved, but it seems to be a partial fix for Branch Specifier parameter.
+// Repository URL still has problem with lightweight checkout
+// Set Repository URL and Branch Specifier as build parameters by default
+// If LIGHT_WEIGHT_CHECKOUT is set to true, set Repository URL to its explicit value and keep Branch Specifier as parameter
+def ADOPTOPENJDK_REPO_BUILD_PARAM = '${ADOPTOPENJDK_REPO}'
+def ADOPTOPENJDK_BRANCH_BUILD_PARAM = '${ADOPTOPENJDK_BRANCH}'
+def SCRIPT_PATH = "aqa-tests/buildenv/jenkins/perfPipeline.groovy"
+
```
karianna, line 69:
```suggestion
```

buildenv/jenkins/perfL2JobTemplate
```diff
@@ -0,0 +1,145 @@
...
+                        copyArtifactPermission {
+                                projectNames('*')
+                        }
+                }
+                definition {
+                        parameters {
+                                stringParam('ADOPTOPENJDK_REPO', ADOPTOPENJDK_REPO, "AdoptOpenJDK git repo. Please use ssh for zos.")
+                                stringParam('ADOPTOPENJDK_BRANCH', ADOPTOPENJDK_BRANCH, "AdoptOpenJDK branch")
+                                stringParam('PLATFORM', PLATFORM, '''Platform to be tested:
+                                <br/>arm_linux
```
karianna, line 85: Not sure we support *all* of those platforms

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -66,52 +65,48 @@ node("ci.role.test&&hw.arch.x86&&sw.os.linux") {
...
-                                // baseline
-                                if (RUN_BASELINE) {
-                                    baselineParams << string(name: "BASELINE_NUM", value: "BASELINE_NUM_" + i.toString())
-                                    def baseRun = triggerJob("${item.BENCHMARK}", "${platform}", baselineParams, "baseline")
-                                    aggregateLogs(baseRun)
-                                } else {
-                                    echo "Skipping baseline run since RUN_BASELINE is set to false"
-                                }
-                            }
+                                for (int i = 0; i < PERF_ITERATIONS; i++) {
```
karianna, line 92: I think whitespace is out here slightly? Ignore me if I'm wrong :-)

buildenv/jenkins/perfL2JobTemplate
MattyWeee123: I think it might always be null here since I don't pass it in as additional param to jobDsl? Then the JDK_IMPL from childParams overrides it afterwards in the build call. They had the same pattern in testJobTemplate and the createJob from JenkinsfileBase, not sure why though.

buildenv/jenkins/perfL2JobTemplate
MattyWeee123: I got this from copying what is shown on the Perf Pipeline config UI: 
image

buildenv/jenkins/perfPipeline.groovy
MattyWeee123: I smoothed it out so everything in the block as 8 spaces per indent. Is there a strict convention for how many spaces in an indent? I see some places want 8 but some places want 4? I have switched to 8 per tab.

buildenv/jenkins/perfL2JobTemplate
llxia: This is just a description. As stated, `Please refer to PLATFORM_MAP in https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/openjdk_tests`. We can remove `arm_linux`, but it is still supported in code.

Issue #6329, opened 2025-06-12
Set current build result based on Perf metrics


llxia:
We can get the child files via https://github.com/adoptium/aqa-tests/issues/6328.

Calculate based on test and baseline runs:
- mean
- max
- min
- median
- std
- CI (optional)

Compare the test and baseline mean value: 
- Score > 98% - No Regression => do nothing
- Score in 91% - 98% - Possible Regression => set `currentBuild.result=UNSTABLE`
- Score < 90% - Regression => set `currentBuild.result=UNSTABLE`

related: https://github.com/adoptium/aqa-tests/issues/6268

MarkLi28:
can i work on this?

MattyWeee123:
Hi, I'd like to join @MarkLi28 on this issue

Issue #6330, opened 2025-06-12
Add exit early option in Perf build


llxia:
In Perf run, add the EXIT_EARLY flag.

If EXIT_EARLY is set to true, [loop](https://github.com/adoptium/aqa-tests/blob/61ba1997d607163ba0d008832dbcd9b5670ead89/buildenv/jenkins/perfPipeline.groovy#L74) only runs 80% of PERF_ITERATIONS ( `PERF_ITERATIONS * 0.8` and round to nearest integer). If Score > 98%, then break.

related: https://github.com/adoptium/aqa-tests/issues/6329 and https://github.com/adoptium/aqa-tests/issues/6268

MarkLi28:
Can i work on this

Issue #6328, opened 2025-06-12
Aggregate Perf log files in parent job


llxia:
https://github.com/adoptium/aqa-tests/issues/6327 will generate perf log file - `<env.JOB_NAME>_<env.BUILD_NUMBER>.log`

Inside [loop](https://github.com/adoptium/aqa-tests/blob/6a60719fdf58880a5e77ac14d09a92efeaaa370a/buildenv/jenkins/perfPipeline.groovy#L85), aggregate Perf log files in parent job. See https://github.com/adoptium/aqa-tests/blob/61ba1997d607163ba0d008832dbcd9b5670ead89/buildenv/jenkins/JenkinsfileBase#L1545 as reference.

related: https://github.com/adoptium/aqa-tests/issues/6268

MattyWeee123:
Hi, may I work on this?

Issue #6327, opened 2025-06-12
Get dacapo-h2 metric and store in a file


llxia:
Parse the Jenkins job console and find the dacapo-h2 metric. In the following example, we should find `4689` .

https://ci.adoptium.net/job/Perf_openjdk21_hs_sanity.perf_x86-64_linux_dacapo_test/9/console

```
00:02:30.514  ===== DaCapo 9.12-MR1 h2 PASSED in 4689 msec =====
```
TRSS regex can be used as an example: https://github.com/adoptium/aqa-test-tools/blob/f9e78de8e8d71654a727f226c29dcf615b7c5bee/TestResultSummaryService/parsers/BenchmarkMetric.js#L155

Getting job content: https://github.com/adoptium/aqa-tests/blob/61ba1997d607163ba0d008832dbcd9b5670ead89/buildenv/jenkins/openjdk_tests#L643

The output file should be in JSON format. For example,
```
[
    {
        "Benchmark": "dacapo-h2",
        "metrics": 
        {
          "dacapo-h2" : [    ====> if there is no metric name, use Benchmark name
            {"value": "4689"},
            {"higherbetter": true},
            {"units": "msec"}
        ]
      }
    }
]
```

Note: A benchmark could have multiple metrics. For example, in liberty_startup, we measure both `Footprint` and `Startup Time`.
```
[
    {
        "Benchmark": "liberty_startup",
        "metrics": 
        {
          "Footprint" : [
            {"value": 1234},
            {"higherbetter": false},
            {"units": "kb"}
        ],
          "Startup Time" : [
            {"value": 5678},
            {"higherbetter": false},
            {"units": "ms"}
        ]
      }
    }
]
```

The file name should be `<env.JOB_NAME>_<env.BUILD_NUMBER>.log`

@smlambert @LongyuZhang if I missed anything, please feel free to update.

MattyWeee123:
Could I have some clarification on this issue, since I believe dacapo h2 should be temporarily disabled due to #6331?

PR #6369, opened 2025-06-26
Set current build result based on Perf metrics.

MattyWeee123:
Modified aggregateLogs to read JSON logs and save
accumulate the runtimes as doubles in lists.
The stats functions returns a map of performance
statistics based on those runtimes. At the end of
each iteration, score is calculated to determine
possible regression or #6330 EXIT_EARLY. 

Note: Implementation not very robust and contains
a few hardcoded assumptions, waiting for issues to extend #6327 and #6328 first.

Fixes: #6329

### DIFF
--- a/buildenv/jenkins/perfPipeline.groovy
+++ b/buildenv/jenkins/perfPipeline.groovy
@@ -168,7 +168,7 @@ def generateChildJobViaAutoGen(newJobName) {
 
 def aggregateLogs(run, runtimes) {
         def json 
-        node(env.SETUP_LABEL) {
+        node(ci.role.test&&hw.arch.x86&&sw.os.linux) {
                 def buildId  = run.getRawBuild().getNumber()
                 def name = run.getProjectName()
                 def result = run.getCurrentResult()


MattyWeee123:
@llxia would you like to to archive the performance statistics as well? Currently I just echo them to the console.

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -99,16 +102,29 @@ timestamps {
...
                                     def baseRun = triggerJob("${item.BENCHMARK}", "${platform}", baselineParams, "baseline")
-                                    aggregateLogs(baseRun)
+                                    aggregateLogs(baseRun, baselineRuntimes)
                                 } else {
                                     echo "Skipping baseline run since RUN_BASELINE is set to false"
                                 }
+
+                                testStats = stats(testRuntimes)
+                                baselineStats = stats(baselineRuntimes)
+                                def score = (testStats.mean/baselineStats.mean) * 100
```
llxia, line 121: json can contain multiple benchmarks. We cannot assume we only have one (i.e., `dacapo-h2`). I am ok to deliver this as the initial version and add the enhancement in the future PR.

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -99,16 +102,29 @@ timestamps {
...
+
+                                echo "testRuntimes: ${testRuntimes}" 
+                                echo "baselineRuntimes: ${baselineRuntimes}"
+                                echo "score: ${score} %"
+
+                                if (score <= 98) {
```
llxia: We should only set the build result at the end of `PERF_ITERATIONS` or `EXIT_EARLY` (https://github.com/adoptium/aqa-tests/issues/6330). For example:

```
if (i == PERF_ITERATIONS || ( EXIT_EARLY && i >= PERF_ITERATIONS * 0.8) ) {
   if (score <= 98) {
    ....
   } else {
       echo "Perf iteration completed. EXIT_EARLY: ${EXIT_EARLY}. PERF_ITERATIONS: ${PERF_ITERATIONS}. Actual iterations: ${i}."
       break
   }
}
```

llxia: I missed `params`. It should be params.EXIT_EARLY.
Please test the PR after the change. Thanks.

llxia: ah, I noticed line 11. That works.

llxia (approved):
LGTM

LongyuZhang (approved):
LGTM

PR #6372, opened 2025-06-26
streamline dynamic machine label creation and cleanup process

mahdipub:
Uses create/free EBC pipeline calls which lets manage EBC from central job.

Signed-off-by: mahdi@ibm.com


mahdipub:
Sample run:
https://hyc-runtimes-jenkins.swg-devops.com/job/EBC/job/Test_openjdk11_j9_sanity.functional_x86-64_linux_mahdi/20/

buildenv/jenkins/openjdk_tests
```diff
@@ -588,13 +582,10 @@ def runTest() {
...
-        if (params.CLOUD_PROVIDER == "EBC" && ebcDemandId) {
+        if (params.CLOUD_PROVIDER == "EBC") {
             try {
-                println("Triggering EBC_Delete_Node job with demandId: ${ebcDemandId}")
-                def parameters = [
-                    string(name: 'demandId', value: ebcDemandId)
-                ]
-                build job: 'EBC_Delete_Node', parameters: parameters, propagate: false
+                println("Triggering EBC_Delete_Node for label: ${LABEL}")
+                freeEBCnodes(LABEL)
```
llxia: This may not be correct. Having CLOUD_PROVIDER = EBC does not mean we get an EBC node.

buildenv/jenkins/openjdk_tests
```diff
@@ -732,29 +723,28 @@ def generateJobViaAutoGen(testJobName) {
...
+        string(name: 'nodeType', value: 'testmachine'),
+        string(name: 'NUM_MACHINES', value: "1"), 
+        string(name: 'TIME_LIMIT', value: expectedNodeLifespan.toString()), 
+        string(name: 'EBC_ENV', value: 'prod'),
+        booleanParam(name: 'WAIT', value: wait),
+    ]
+}
+
+def freeEBCnodes(dynamicMachineLabel) {
+    build job: 'EBC/freeEBCnodes',
```
llxia: Can we standardize the naming style for the two job names (`EBC_Create_Node` and `freeEBCnodes`)? Let's choose either snake_case or CamelCase, but not mix both.

buildenv/jenkins/openjdk_tests
```diff
@@ -732,29 +723,28 @@ def generateJobViaAutoGen(testJobName) {
...
-                        'ebc_remotePlaybook_gitBranch=openj9',
-                        'ebc_remotePlaybook_file=ansible/playbooks/ibm_unix_playbook/ibm_jenkins_unix_test.yml',
-                        "ebc_ansible_extra_vars='is_ebc=true'",
-                        'ebc_remotePlaybook_jenkinsUserSshPubKey=unused']
-                    ]
-    echo "EBC Args: ${ebcArgs}"
-    def ebcResult = ebcJenkinsNodeDemand(ebcArgs);
-    echo "EBC Result: ${ebcResult}"
+def getEBCNode(dynamicMachineLabel) {
+    def int expectedNodeLifespan = params.TIME_LIMIT ? Integer.parseInt(params.TIME_LIMIT) : 1
```
llxia: def int is redundant. Either use def or a strong type like int, not both together.

buildenv/jenkins/openjdk_tests
```diff
@@ -386,28 +385,23 @@ timestamps{
...
-                                ebcDemandId = UUID.randomUUID().toString()
-                                def dynamicMachineLabel = 'ci.agent.dynamic.' + ebcDemandId
                                 try {
                                     timeout(time: 120, unit: 'MINUTES') {
                                         // ToDo: pass in platform once we support more platforms. For now, we only support xlinux.
-                                        getEBCNode(ebcDemandId, dynamicMachineLabel)
+                                        def dynamicMachineLabel = 'ci.agent.dynamic.' + UUID.randomUUID().toString()
+                                        getEBCNode(dynamicMachineLabel)
                                         node(dynamicMachineLabel) {
                                             echo "EBC node ${dynamicMachineLabel} is ready"
```
AdamBrousseau, line 394: Do we need to check for the node to be ready here? It appears that is what param WAIT is doing on the Create job.

buildenv/jenkins/openjdk_tests
mahdipub: Corrected

buildenv/jenkins/openjdk_tests
mahdipub: Corrected

buildenv/jenkins/openjdk_tests
mahdipub: Corrected

buildenv/jenkins/openjdk_tests
mahdipub: I would say it doesn't hurt if we keep it for when we decide to set `wait:false`.

AdamBrousseau (approved):

buildenv/jenkins/openjdk_tests
```diff
@@ -588,13 +582,10 @@ def runTest() {
                 }
             }
         }
-        if (params.CLOUD_PROVIDER == "EBC" && ebcDemandId) {
+        if (params.CLOUD_PROVIDER == "EBC" && LABEL.startsWith("ci.agent.dynamic.")) {
```
llxia: We could have other ways of creating `ci.agent.dynamic.` node. This may not be true. The original logic is correct, so why do we have to change it?

mahdipub: In EBC create/free architecture we decided to remove the requirements to handle demandID from request endpoint. So, we do not need to create/keep the demandID anymore and it happens indide create node and free mode pipelines behind the scene. 
Also in this change, the condition is combined with `params.CLOUD_PROVIDER == "EBC"` which seems to be right to the point.

llxia: From what I can see, `dynamicMachineLabel` is set by this pipeline code and passed into both EBC_Create_Node and EBC_Complete.

mahdipub: Yes. That is not demandID. I deliberately added a random string to it to prevent others take created nodes. But we can replace it with anything else to make it uniq to each pipeline run. For example job name+number or timestamp.

llxia: > we decided to remove the requirements to handle demandID from request endpoint

Could you help me understand the reason behind this change? If demandID is not set, EBC_Create_Node can generate it. But in the case where demandID is set, why do we remove the support for it? In the original code, the demandID could be set in EBC directly. By removing it, not only does it require additional logic in EBC_Complete to retrieve the demandID based on the unique LABEL, but it also introduces the overhead of generating two UUIDs - one for the LABEL and another for the demandID.

IMO, demandID is optional, and we should support both cases.

buildenv/jenkins/openjdk_tests
```diff
@@ -386,28 +385,22 @@ timestamps{
...
                                     echo 'Exception: ' + e.toString()
-                                    echo "Cannot launch EBC node ${dynamicMachineLabel}..."
-                                }
-                                if (ebcNode) {
-                                    LABEL = dynamicMachineLabel
-                                    echo "Use EBC node ${LABEL}..."
-                                } else {
+                                    echo "Cannot launch EBC node ..."
                                     echo "Use static node ${LABEL}..."
+                                    params.CLOUD_PROVIDER = ''
```
AdamBrousseau: I'm not sure if overriding a param like this is a good idea or if it will work? Has it been tested or can we set the variable without the `param.` prefix?

buildenv/jenkins/openjdk_tests
mahdipub: Thanks for your observation. Based on that I had a discussion with @AdamBrousseau and I reformed the code with higher level of abstraction. So now AQA does not need to take care of demanID nor dynamicLabel. The create node will provide it and I am [testing it](https://hyc-runtimes-jenkins.swg-devops.com/view/EBC/job/EBC/job/Test_openjdk11_j9_sanity.functional_x86-64_linux_mahdi/39/console) now.
The reason for removing demandID handling from user point of view is, adding a layer of abstraction of what caller do not need to care about. It cover the EBC complexity in pipelines and the user only needs to call the pipeline with node specification. That will help any other use cases later of follow this lead which means centralized work and avoid implementation of same logic for demandID every where.

buildenv/jenkins/openjdk_tests
```diff
@@ -732,29 +722,26 @@ def generateJobViaAutoGen(testJobName) {
...
+    def result = build job: 'EBC/EBC_Create_Node',
+    propagate: false,
+    wait: true,
+    parameters: [
+        string(name: 'platform', value: params.PLATFORM),
+        string(name: 'nodeType', value: 'testmachine'),
+        string(name: 'TIME_LIMIT', value: expectedNodeLifespan.toString()), 
+        booleanParam(name: 'WAIT', value: wait),
+    ]
+    return result.description.replaceAll("group_label: ", "").trim()
```
AdamBrousseau: It seems this relies on the job description to be of a certain format. What happens if/when that changes?
If you get it out of the environment variable, I think it will be safer long term.

buildenv/jenkins/openjdk_tests
mahdipub: That param is widely used. If I want to use a variable initiated by the param, I have to change them all. I did it first, but thought this way we do not reform all code. I can simulate a test and update here.

buildenv/jenkins/openjdk_tests
mahdipub: [This build](https://hyc-runtimes-jenkins.swg-devops.com/view/EBC/job/EBC/job/Test_openjdk11_j9_sanity.functional_x86-64_linux_mahdi/39/) uses env

buildenv/jenkins/openjdk_tests
mahdipub: That also changed now and [is testing](https://hyc-runtimes-jenkins.swg-devops.com/view/EBC/job/EBC/job/Test_openjdk11_j9_sanity.functional_x86-64_linux_mahdi/39/)

buildenv/jenkins/openjdk_tests
```diff
@@ -732,29 +724,26 @@ def generateJobViaAutoGen(testJobName) {
...
-                        "ebc_jenkins_agent_label=${dynamicMachineLabel}",
-                        'ebc_remotePlaybook_gitBranch=openj9',
-                        'ebc_remotePlaybook_file=ansible/playbooks/ibm_unix_playbook/ibm_jenkins_unix_test.yml',
-                        "ebc_ansible_extra_vars='is_ebc=true'",
-                        'ebc_remotePlaybook_jenkinsUserSshPubKey=unused']
-                    ]
-    echo "EBC Args: ${ebcArgs}"
-    def ebcResult = ebcJenkinsNodeDemand(ebcArgs);
-    echo "EBC Result: ${ebcResult}"
+def getEBCnode() {
```
llxia: Can we use getEBCNode() or getEbcNode()? Same for freeEBCnode()

buildenv/jenkins/openjdk_tests
```diff
@@ -588,13 +583,10 @@ def runTest() {
...
         }
-        if (params.CLOUD_PROVIDER == "EBC" && ebcDemandId) {
+        if (cloud_provider == "EBC") {
             try {
-                println("Triggering EBC_Delete_Node job with demandId: ${ebcDemandId}")
-                def parameters = [
-                    string(name: 'demandId', value: ebcDemandId)
-                ]
-                build job: 'EBC_Delete_Node', parameters: parameters, propagate: false
+                println("Triggering EBC_Delete_Node for label: ${LABEL}")
```
llxia: try and catch should be within freeEBCnode(), or there is no need to have freeEBCnode() function. freeEBCnode() only has one line.

buildenv/jenkins/openjdk_tests
```diff
@@ -126,7 +126,8 @@ def PLATFORMS = params.PLATFORM.trim().split("\\s*,\\s*");
 def JDK_VERSIONS = params.JDK_VERSION.trim().split("\\s*,\\s*");
 def JDK_IMPLS = params.JDK_IMPL.trim().split("\\s*,\\s*");
 
-@Field String ebcDemandId = ""
+def cloud_provider = params.CLOUD_PROVIDER ?: ""
```
llxia: Please use CLOUD_PROVIDER instead of cloud_provider.

buildenv/jenkins/openjdk_tests
mahdipub: That adjusted. 
[Here](https://hyc-runtimes-jenkins.swg-devops.com/view/EBC/job/EBC/job/Test_openjdk11_j9_sanity.functional_x86-64_linux_mahdi/40/console) is the test pipeline

buildenv/jenkins/openjdk_tests
mahdipub: That adjusted. 
[Here](https://hyc-runtimes-jenkins.swg-devops.com/view/EBC/job/EBC/job/Test_openjdk11_j9_sanity.functional_x86-64_linux_mahdi/40/console) is the test pipeline

buildenv/jenkins/openjdk_tests
mahdipub: That adjusted. 
[Here](https://hyc-runtimes-jenkins.swg-devops.com/view/EBC/job/EBC/job/Test_openjdk11_j9_sanity.functional_x86-64_linux_mahdi/40/console) is the test pipeline

llxia (approved):
LGTM

buildenv/jenkins/openjdk_tests
```diff
@@ -732,29 +718,32 @@ def generateJobViaAutoGen(testJobName) {
...
+    ]
+    return result.buildVariables.group_label
+}
+
+def freeEBCNode() {
+        try {
+            println("Triggering EBC_Delete_Node for label: ${LABEL}")
+            build job: 'EBC/EBC_Complete',
+                propagate: false,
+                wait: false,
```
AdamBrousseau: I might prefer we `wait` here. If we don't, there is no build number displayed in the test job. It just makes it slightly harder to track if there's a problem. The job only takes about 10 seconds to run as it simply makes a curl to the EBC API.
image

LongyuZhang (approved):
LGTM

Issue #6355, opened 2025-06-23
Generic logic to get get and store benchmark metric


llxia:
Extend based on https://github.com/adoptium/aqa-tests/pull/6339/, make checkDacapoH2Metric() and archiveDacapoH2Metric() generic.

On second thought, it may be better to have a separate script (i.e., Python - `benchmarkMetric.py`) to handle this.

- In `benchmarkMetric.py`, it contains a generic function `checkMetric()`. `checkMetric()` takes job console output and reads `benchmarkMetric.json` and runs each regex, and stores into one JSON output.

`benchmarkMetric.json` example:
```
[
    {
        "Benchmark": "dacapo-eclipse",
        "metrics": 
        {
          "dacapo-eclipse" : [
            {"regex": "DaCapo.*eclipse\\sPASSED\\sin\\s(\\d*\\.?\\d*)\\smsec"},
            {"higherbetter": true},
            {"units": "msec"}
        ]
      }
    },
{
        "Benchmark": "dacapo-h2",
        "metrics": 
        {
          "dacapo-h2" : [
            {"regex": "DaCapo.*h2\\sPASSED\\sin\\s(\\d*\\.?\\d*)\\smsec"},
            {"higherbetter": true},
            {"units": "msec"}
        ]
      }
    }
]
```

Note: Backslashes must be double-escaped (`\\`) in JSON strings.

Output: JSON file as specified in https://github.com/adoptium/aqa-tests/issues/6327. For example:

```
[
    {
        "Benchmark": "dacapo-eclipse",
        "metrics": 
        {
          "dacapo-eclipse" : [ 
            {"value": "123456"},
            {"higherbetter": true},
            {"units": "msec"}
        ]
      }
    },
{
        "Benchmark": "dacapo-h2",
        "metrics": 
        {
          "dacapo-h2" : [
            {"value": "4689"},
            {"higherbetter": true},
            {"units": "msec"}
        ]
      }
    }
]
```

related: https://github.com/adoptium/aqa-tests/issues/6327

top level issue: https://github.com/adoptium/aqa-tests/issues/6268

@smlambert @LongyuZhang if I missed anything, please feel free to update.

llxia:
For getting console output, try  `myjob.getRawBuild().getLog()` 
 
https://stackoverflow.com/questions/31521030/output-the-console-text-of-a-jenkins-job-that-my-job-is-running
https://stackoverflow.com/questions/42510489/console-output-in-pipelinejenkins

[...]

PR #6393, opened 2025-07-09
Generic Logic to get and store benchmark metrics

MattyWeee123:
metricConfig2JSON.py takes BenchmarkMetric.js and
manually converts it to JSON, since I didn't want
to mess with dependencies on the machine.

benchmarkMetric.py takes the console log, and the
JSON version of BenchmarkMetric and does a brute
force search for every possible regex, creating a list of dictionaries to describe each captured metric.

The aggregate JSON file is archived via checkMetric in JenkinsfileBase.

**Notes for additional issues:** 

- I kept metricConfig2JSON.py as a separate script,  if we want to run it once and cache it somewhere like how dependencies are handled in TKG. Wasn't sure where/how is preferable to do this. 

- benchmarkMetric.py is not optimized; Should probably retrieve exactly what regexes are needed by looking at env.BUILD_LIST and env.TARGET. Had trouble finding what part of the code handles that interaction with the playlist.xml files, and wasn't sure how to handle nuances among different test logs (didn't use outer regex or funcName). 

Fixes: #6355
[Successfully Archived Metrics on Grinder Job](https://ci.adoptium.net/job/Grinder/13535/)

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -820,9 +820,9 @@ def runTest( ) {
 				sh "genhtml --quiet codeCoverageInfoFinal.info --output-directory code_coverage_report"
 			}
 		}
-                node ('ci.role.test&&hw.arch.x86&&sw.os.linux') {
-                        generateBenchmarkMetric()
-                }
+
+                generateBenchmarkMetric()
+                
 		def resultSum = [:]
 		def matcher = manager.getLogMatcher(".*(TOTAL: \\d+)\\s*(EXECUTED: \\d+)\\s*(PASSED: \\d+)\\s*(FAILED: \\d+)\\s*(DISABLED: \\d+)?\\s*(SKIPPED: \\d+)\\s*\$")
 		if (matcher?.matches()) {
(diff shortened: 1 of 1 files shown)


buildenv/jenkins/JenkinsfileBase
```diff
@@ -1678,32 +1676,12 @@ def forceCleanWS() {
...
-                println("DaCapo-h2 metric found and archived")
-        } else {
-                println("no DaCapo-h2 metric found")
-        }
-}
-
-def archiveDacapoH2Metric(runtime) {
+def checkMetric() {
+        writeFile file : 'console.txt', text: currentBuild.rawBuild.getLog() 
+        sh "git clone --depth 1 https://github.com/adoptium/aqa-test-tools.git aqa-test-tools"
```
llxia: Instead of git clone the entire repo, let's just get the file

```
curl -Os  https://raw.githubusercontent.com/adoptium/aqa-test-tools/refs/heads/master/TestResultSummaryService/parsers/BenchmarkMetric.js
```

buildenv/jenkins/JenkinsfileBase
```diff
@@ -1678,32 +1676,12 @@ def forceCleanWS() {
...
-                def runtime = matcher.group(1) 
-                archiveDacapoH2Metric(runtime)
-                println("DaCapo-h2 metric found and archived")
-        } else {
-                println("no DaCapo-h2 metric found")
-        }
-}
-
-def archiveDacapoH2Metric(runtime) {
+def checkMetric() {
```
karianna: What does checkMetric really do?  Not sure if this is the right name for it...

buildenv/jenkins/JenkinsfileBase
```diff
@@ -821,9 +821,7 @@ def runTest( ) {
 			}
 		}
 
-                if (env.BUILD_LIST.startsWith('perf')) {
-                    checkDacapoH2Metric()
-                }
+                checkMetric()
```
llxia: Instead of running checkMetric() at L3. I think it should be triggered at [Perf L2](https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/perfPipeline.groovy).

This code requires a machine with Python installed. Let's leverage `SETUP_LABEL`
```
node (params.SETUP_LABEL) {
 ...
}
```

buildenv/jenkins/JenkinsfileBase
MattyWeee123: It creates and archives a JSON file containing data about the relevant benchmark metrics (throughput, latency, etc).

karianna: So it's more like generateBenchmarkMetrics?

MattyWeee123: yeah I'd say generateBenchmarkMetrics is more representative; I've updated the name

buildenv/jenkins/JenkinsfileBase
MattyWeee123: @llxia I don't think params.SETUP_LABEL has python 
image

buildenv/jenkins/JenkinsfileBase
MattyWeee123: fixed

buildenv/jenkins/JenkinsfileBase
MattyWeee123: @llxia I am thinking of delegating all python processing to L2, and all L3 needs to do is archive the relevant chunk of console log so L2 can use it. Is this what you mean? If so I will close this PR and combine with my work on #6398

llxia: I was thinking L3 does not need to take any action. L2 will process its child job (L3) log and archive JSON (for example: https://github.com/adoptium/aqa-tests/blob/49b3ffb29329a5d91330eced9a45ed7fc5d0a7c9/buildenv/jenkins/perfPipeline.groovy#L56). Since L3 code is common for all testing, changing this code will affect other testing (i.e., non-Perf pipeline testing).

karianna:
@MattyWeee123 was this closed in favour of another one?

MattyWeee123:
@karianna yes, working on #6418 now

Issue #6379, opened 2025-07-03
Add GENERATE_JOBS in Perf Pipeline


llxia:
Add the GENERATE_JOBS parameter in the Perf Pipeline. If GENERATE_JOBS is set to false (default), only generate the job if the job is not runnable. If GENERATE_JOBS is set to true, generate the job regardless.

https://github.com/adoptium/aqa-tests/blob/32fa2ac97e1b20cc09e1de61e8dcab3fd0b2c00c/buildenv/jenkins/JenkinsfileBase#L1444-L1447

MattyWeee123:
Hi, may I work on this?

PR #6397, opened 2025-07-09
Add GENERATE_JOBS in Perf Pipeline.

MattyWeee123:
Added GENERATE_JOBS in perfL2JobTemplate.

Fixes: #6379


karianna (approved):

llxia (changes requested):
Both https://github.com/adoptium/aqa-tests/blob/5610ac04f6f43810b0231ad7aa2b3f6058250c8e/buildenv/jenkins/perfPipeline_root.groovy#L78 and https://github.com/adoptium/aqa-tests/blob/5610ac04f6f43810b0231ad7aa2b3f6058250c8e/buildenv/jenkins/perfPipeline.groovy#L97 need to be updated if the GENERATE_JOBS is set to true.

PR #6443, opened 2025-07-21
Generalized regression monitoring for PerfPipeline.

### LINKED ISSUE #6418: L3 is non-unique to Perf Pipeline, move all logic related to generating benchmark metrics and checking for regression to L2
1. Remove logic related to checking metrics in L3 (checkDacapoH2Metric and helpers) 
2. Bring the generateBenchmarkMetric from #6355 to L2, calling getlog() on the RunWrapper object of the child instead 
3. Adjust logic from #6398 as needed

MattyWeee123:
metricConfig2JSON.py parses and rearranges BenchmarkMetrics.js into JSON. initBenchmarkMetrics.py generates templates for a groovy aggregate metric map and a single test map for archiving individual test/baseline child jobs. benchmarkMetrics.py updates a single test map with the metric value for a run.

testList is used to generalize regression monitoring and EXIT_EARLY for child jobs that run multiple tests (removing from testList will have both parent and child ignore the removed test).

Using specific rhel9 machine to run the python scripts and getPythonDependencies uses curl to retrieve the relevant python scripts.

BenchmarkMetric.js needs to be updated (lusearch-fix should be lusearch, tomcat is not there so the script thinks it should grab all dacapo metrics for tomcat which is not true, ...)

Fixes: #6418, #6355


buildenv/jenkins/perfPipeline.groovy
```diff
@@ -38,49 +36,69 @@ params.each { param ->
     }
 }
 
-node ("ci.role.test&&hw.arch.x86&&sw.os.linux") {
+env.L2_Machine = "test-rhibmcloud-rhel9-x64-1" //grab and execute python scripts here
```
llxia: This has to be a parameter that the user can set. It cannot be a hardcoded machine name. Different Jenkins servers could have different machines.

```

if (params.SETUP_LABEL) {
    SETUP_LABEL = params.SETUP_LABEL
} else {
    if (params.PROCESS_METRICS && params.EXIT_EARLY) {
       SETUP_LABEL = "ci.role.test&&hw.arch.x86&&sw.os.linux&&sw.os.rhel.9"
   } else {
       SETUP_LABEL = "ci.role.test&&hw.arch.x86&&sw.os.linux"
   }
}
```

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -38,49 +36,69 @@ params.each { param ->
...
+
+node (env.L2_Machine) {
```
llxia: Not an issue that is introduced by this PR. I just realized that we use 2 machines for one perf run.
- one is env.L2_Machine for process logs using python
- one is the actual perf machine

During the perf run, the L2_Machine is in idle, but cannot be used by other jobs. L2_Machine is only used before and after the actual perf run. Num of L2_Machine can be large, and the actual perf run can be very long. This needs to be improved in the future.

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -38,49 +36,69 @@ params.each { param ->
...
+
+                        def owner = params.ADOPTOPENJDK_REPO.tokenize('/')[2]
+                        def runBase = "runBase.json"
+                        def aggrBase = "aggrBase.json"
+                        getPythonDependencies(owner, params.ADOPTOPENJDK_BRANCH) 
+                        sh "curl -Os  https://raw.githubusercontent.com/adoptium/aqa-test-tools/refs/heads/master/TestResultSummaryService/parsers/BenchmarkMetric.js"
+                        sh "python3 metricConfig2JSON.py --metricConfig_js BenchmarkMetric.js"
+                        sh "python3 initBenchmarkMetrics.py --metricConfig_json metricConfig.json --testNames ${params.TARGET.split("=")[1]} --runBase ${runBase} --aggrBase ${aggrBase}"
+                        
+                        def testList = params.TARGET.split("=")[1].tokenize(",")
```
llxia: For now, I think process metrics and exit early should be optional flags. It means that we should wrap this code in if block.

```
if (params.PROCESS_METRICS && params.EXIT_EARLY) {
...
}
```

same for aggregateLogs and https://github.com/adoptium/aqa-tests/pull/6443/files#diff-ddb2cd746534db345c723274958566e7386caa85dd20558756e020817d23b5feR90-R101

buildenv/jenkins/perfPipeline.groovy
MattyWeee123: fixed, added comment that these params are not currently supported in UI or the auto gen job template I believe

buildenv/jenkins/perfPipeline.groovy
MattyWeee123: I wrapped relevant code with only params.PROCESS_METRICS and have params.EXIT_EARLY only for the break out of loop logic. Otherwise I'm not sure why we should we have both these parameters instead of one, since they both need to be true for the logic to hold.

MattyWeee123:
@llxia 

[Run with PROCESS_METRICS and EXIT_EARLY](https://ci.adoptium.net/view/Test_grinder/job/Perf_Pipeline/164/console)

[Run with default settings](https://ci.adoptium.net/view/Test_grinder/job/Perf_Pipeline/165/)

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -1,14 +1,19 @@
...
-def testStats = [:]
-def baselineStats = [:]
-def testRuntimes = []
-def baselineRuntimes = []
+
 def testParams = []
 def baselineParams = []
-int PERF_ITERATIONS = params.PERF_ITERATIONS ? params.PERF_ITERATIONS.toInteger() : 4
 boolean RUN_BASELINE = (params.RUN_BASELINE != null) ? params.RUN_BASELINE.toBoolean() : true
+int PERF_ITERATIONS = params.PERF_ITERATIONS.toInteger()
```
llxia, line 6: We no longer get PERF_ITERATIONS from the job parameter. It only gets set in `PERFCONFIG_JSON`.  `params.PERF_ITERATIONS.toInteger()` will cause job failure as params.PERF_ITERATIONS will be null.

buildenv/jenkins/perfPipeline_root.groovy
```diff
@@ -55,15 +55,14 @@ params.each { param ->
...
-                def BUILD_LIST = item.BUILD_LIST
-                def PLATMACHINE_MAP = item.PLAT_MACHINE_MAP
                 def baseParams = childParams.collect()
                 baseParams << string(name: "BENCHMARK", value: item.BENCHMARK)
                 baseParams << string(name: "TARGET", value: item.TARGET)
                 baseParams << string(name: "BUILD_LIST", value: item.BUILD_LIST)
                 baseParams << string(name: "PERF_ITERATIONS", value: item.PERF_ITERATIONS ? item.PERF_ITERATIONS.toString() : "4")
+                baseParams << string(name: "SETUP_LABEL", value: item.SETUP_LABEL) 
+                baseParams << string(name: "PROCESS_METRICS", value: item.PROCESS_METRICS ? item.PROCESS_METRICS.toString() : "false")
+                baseParams << string(name: "EXIT_EARLY", value: item.EXIT_EARLY ? item.EXIT_EARLY.toString() : "false")
```
llxia: These values should be set from `params`, not `item`. Because we set [all the params](https://github.com/adoptium/aqa-tests/pull/6443/files#diff-8e964073d0c915c9da98279d25ec128b88b1fa5f1e559ddad3a5c56b89ea201aR43-R55), I believe L63 - L65 can be removed.

With the current code, `test-rhibmcloud-rhel9-x64-1` is used regardless SETUP_LABEL value because it does not get set. https://ci.adoptium.net/job/Perf_openjdk21_hs_sanity.perf_x86-64_linux_dacapo/139/console

karianna:
@MattyWeee123 Will need a rebase (or that merge conflict fixed)

buildenv/jenkins/perfPipeline_root.groovy
MattyWeee123: @llxia I parse it via PerfConfigJSON in the root; will we be running L2 jobs independently as well?

llxia: SETUP_LABEL, PROCESS_METRICS, and EXIT_EARLY are not values in PerfConfigJSON. They are parameters that are set directly in the Jenkins job. With the current logic, SETUP_LABEL is not set, and `test-rhibmcloud-rhel9-x64-1` is used.

MattyWeee123: I see; Would you have to add those on your end? I'm not sure how to add new Jenkins job parameters to run Grinders to test the code.

llxia (dismissed):
Thanks @MattyWeee123

buildenv/jenkins/perfPipeline_root.groovy
llxia: yes, I added them in Perf_Pipeline
image

buildenv/jenkins/benchmarkMetric.py
```diff
@@ -0,0 +1,27 @@
+import argparse, pathlib, json, re
+
+def main():
```
karianna, line 3: Could do with a comment explaining what the script does

buildenv/jenkins/benchmarkMetric.py
```diff
@@ -0,0 +1,27 @@
...
+        p = argparse.ArgumentParser()
+        p.add_argument("--console", required=True)
+        p.add_argument("--benchmarkMetricsTemplate_json", required=True)
```
karianna, line 6: inconsistent name scheme?

buildenv/jenkins/benchmarkMetric.py
```diff
@@ -0,0 +1,27 @@
...
+        p.add_argument("--fname", required=True)
+        p.add_argument("--testNames", required=True)
+        args = p.parse_args()
+
+        console = pathlib.Path(args.console).read_text(encoding="utf-8")
+
+        benchmarkMetricsTemplate_json = pathlib.Path(args.benchmarkMetricsTemplate_json).read_text(encoding="utf-8")
```
karianna, line 13: Do we need more error handling around the Filo I/O or just let the script fail?  Will it leave artifacts behind if it does fail?

buildenv/jenkins/initBenchmarkMetrics.py
```diff
@@ -0,0 +1,41 @@
...
+                variant = test_info[1]
+                if (metricMap.get(variant) != None): 
+                        benchmarkMetrics.update({test : {variant : metricMap[variant]}})
+                        return 
+
+        benchmarkMetrics.update({test : metricMap})
+
+def main():
+        p = argparse.ArgumentParser()
+        p.add_argument("--metricConfig_json", required=True)
```
karianna, line 20: Naming convention?

buildenv/jenkins/initBenchmarkMetrics.py
```diff
@@ -0,0 +1,41 @@
+import argparse, pathlib, json
```
karianna, line 1: Comments for the functions please :-)

buildenv/jenkins/initBenchmarkMetrics.py
```diff
@@ -0,0 +1,41 @@
...
+
+def initBenchmarkMetrics(metricConfig, test, benchmarkMetrics): 
+        test_info = test.split("-")
+        benchmarkMap = metricConfig[test_info[0]]
+        metricMap = benchmarkMap["metrics"]
+        if len(test_info) > 1:
```
karianna: Is only one test not valid?

buildenv/jenkins/metricConfig2JSON.py
```diff
@@ -0,0 +1,38 @@
+import argparse, pathlib, json, re
```
karianna, line 1: Comments for the functions please

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -1,14 +1,21 @@
...
+boolean PROCESS_METRICS = (params.PROCESS_METRICS != null) ? params.PROCESS_METRICS.toBoolean() : false 
+boolean EXIT_EARLY = (params.EXIT_EARLY != null) ? params.EXIT_EARLY.toBoolean() : false 
 
-def EXIT_EARLY = (params.EXIT_EARLY) ? true : false 
+if (params.SETUP_LABEL) {
+    SETUP_LABEL = params.SETUP_LABEL
+} else {
+    if (PROCESS_METRICS && EXIT_EARLY) {
+       SETUP_LABEL = "test-rhibmcloud-rhel9-x64-1"
```
karianna: That seems very tightly coupled / brittle to be pinned to one machine only

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -38,48 +45,78 @@ params.each { param ->
...
+                        ["BUILD_LIST", "PLATFORM", "LABEL"].each { key ->
                                 [testParams, baselineParams].each { list ->
                                         list << string(name: key, value: params."${key}")
                                 }
                         }
+
+                        if (PROCESS_METRICS) {
+                                def owner = params.ADOPTOPENJDK_REPO.tokenize('/')[2]
+                                getPythonDependencies(owner, params.ADOPTOPENJDK_BRANCH) 
+                                sh "curl -Os  https://raw.githubusercontent.com/adoptium/aqa-test-tools/refs/heads/master/TestResultSummaryService/parsers/BenchmarkMetric.js"
```
karianna, line 68:
```suggestion
                                sh "curl -Os https://raw.githubusercontent.com/adoptium/aqa-test-tools/refs/heads/master/TestResultSummaryService/parsers/BenchmarkMetric.js"
```

karianna: Also do you always want the latest file and not something that was versioned?

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -38,48 +45,78 @@ params.each { param ->
...
+                                sh "python3 metricConfig2JSON.py --metricConfig_js BenchmarkMetric.js"
```
karianna, line 69: Should Python be invoking this in a virtual env?

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -38,48 +45,78 @@ params.each { param ->
...
+                        }
                         
                         echo "starting to trigger build..."
                         lock(resource: params.LABEL) {
                                 for (int i = 0; i < PERF_ITERATIONS; i++) {
+                                        //clone to avoid mutation
+                                        def thisTestParams = testParams.collect()
+                                        def thisBaselineParams = baselineParams.collect()
+                                        if (PROCESS_METRICS) {     
+                                                //set the target, testlist should change if some metrics regress while others do not
```
karianna:
```suggestion
                                                // Set the target, testlist should change if some metrics regress while others do not
```

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -38,48 +45,78 @@ params.each { param ->
...
-                                                        if (score <= 98) {
-                                                                currentBuild.result = 'UNSTABLE'
-                                                                echo "Possible regression, set build result to UNSTABLE."
+
+                                        if (PROCESS_METRICS) {
+                                                aggregateLogs(testRun, testNames, testList, runBase, metrics, "test")
+                                                aggregateLogs(baseRun, testNames, testList, runBase, metrics, "baseline")
+                                                writeJSON file: "metrics.json", json: metrics, pretty: 4
+                                                archiveArtifacts artifacts: "metrics.json" 
+                                                if (i == PERF_ITERATIONS-1 || (EXIT_EARLY && i >= PERF_ITERATIONS * 0.8)) {
```
karianna, line 114: A comment here would help to explain the predicate

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -120,49 +157,81 @@ def generateChildJobViaAutoGen(newJobName) {
...
+
+        testList.each { test ->
+                aggregateMetrics[test].each { metric -> 
+                        def value = runMetrics[test][metric.key]["value"]
+                        if (value != null) metric.value[testType]["values"] << value
+                }
+        }
+}
+
+def checkRegressions(aggregateMetrics, testList) {
```
karianna, line 191: So a regression is > 2%?

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -120,49 +157,81 @@ def generateChildJobViaAutoGen(newJobName) {
...
-        def variance = nums.collect{(it-mean)**2}.sum()/n
+        def variance = values.collect{(it-mean)**2}.sum()/n
         def stdev = Math.sqrt(variance as double)
         [mean: mean, max: sorted[-1], min: sorted[0], median: median, std: stdev]
 }
 
+def getPythonDependencies (owner, branch) {
+        def pythonScripts = ["benchmarkMetric.py", "initBenchmarkMetrics.py", "metricConfig2JSON.py"]
+        pythonScripts.each { pythonScript ->
+                sh "curl -Os https://raw.githubusercontent.com/${owner}/aqa-tests/refs/heads/${branch}/buildenv/jenkins/${pythonScript}"
```
karianna, line 239: Do we want to get versioned files?

buildenv/jenkins/perfPipeline_root.groovy
```diff
@@ -55,14 +55,11 @@ params.each { param ->
...
         perfConfigJson.each { item ->
-                def BENCHMARK = item.BENCHMARK 
-                def TARGET = item.TARGET
-                def BUILD_LIST = item.BUILD_LIST
-                def PLATMACHINE_MAP = item.PLAT_MACHINE_MAP
                 def baseParams = childParams.collect()
                 baseParams << string(name: "BENCHMARK", value: item.BENCHMARK)
                 baseParams << string(name: "TARGET", value: item.TARGET)
                 baseParams << string(name: "BUILD_LIST", value: item.BUILD_LIST)
+                baseParams << string(name: "PERF_ITERATIONS", value: item.PERF_ITERATIONS ? item.PERF_ITERATIONS.toString() : "4")
```
karianna: Magic number

karianna (changes requested):
A few nitpicks :-)

buildenv/jenkins/benchmarkMetric.py
MattyWeee123: @karianna I believe the script would terminate before the corresponding archive step occurs and the console would report the python error. Errors from read text mean the file can't be accessed, which I'm not sure how to bounce back from with error handling? Later iterations will probably have same issue since I read from same templates. Feels more like a fix code and rerun the script situation to me. @llxia thoughts?

buildenv/jenkins/initBenchmarkMetrics.py
MattyWeee123: @karianna It is, it would execute benchmarkMetrics.update({test:metricMap}) instead. The perf tests follow the convention "suite name-optional variant information", so if length is one suite name alone is enough to distinguish the unique benchmark information. This matches convention of benchmarkMetrics.js in aqa-test-tools for the perf tests.

test_info is a name of a test from the test list

buildenv/jenkins/perfPipeline.groovy
MattyWeee123: @llxia @karianna I wasn't sure what other machines had python3. The other rhel machine didn't work, so I stuck to rhel9.

buildenv/jenkins/perfPipeline.groovy
MattyWeee123: @llxia?

buildenv/jenkins/perfPipeline.groovy
MattyWeee123: @karianna yes

buildenv/jenkins/perfPipeline.groovy
MattyWeee123: @karianna I can determine owner and branch from jenkins UI, so changing versions should be same thing as moving around locally with git, pushing, and running again.

buildenv/jenkins/benchmarkMetric.py
MattyWeee123: @karianna naming convention is as follows: 
metricConfig_js (starter javascript file) -> metriConfig_json (converted to JSON format) -> benchmarkMetricsTemplate_json (optimized for metric processing)

buildenv/jenkins/initBenchmarkMetrics.py
MattyWeee123: explained naming convention in a previous comment

karianna (dismissed):

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -120,49 +160,82 @@ def generateChildJobViaAutoGen(newJobName) {
...
+        }
+
+        def runMetrics = readJSON file: fname 
+
+        testList.each { test ->
+                aggregateMetrics[test].each { metric -> 
+                        def value = runMetrics[test][metric.key]["value"]
+                        if (value != null) metric.value[testType]["values"] << value
+                }
+        }
```
smlambert:
```suggestion
        } }
```

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -120,49 +160,82 @@ def generateChildJobViaAutoGen(newJobName) {
...
+
+        def runMetrics = readJSON file: fname 
+
+        testList.each { test ->
+                aggregateMetrics[test].each { metric -> 
+                        def value = runMetrics[test][metric.key]["value"]
+                        if (value != null) metric.value[testType]["values"] << value
+                }
+        } }
```
smlambert, line 188:
```suggestion
        } 
```

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -1,14 +1,21 @@
...
-def testStats = [:]
-def baselineStats = [:]
-def testRuntimes = []
-def baselineRuntimes = []
+
 def testParams = []
 def baselineParams = []
-int PERF_ITERATIONS = params.PERF_ITERATIONS ? params.PERF_ITERATIONS.toInteger() : 4
 boolean RUN_BASELINE = (params.RUN_BASELINE != null) ? params.RUN_BASELINE.toBoolean() : true
+int PERF_ITERATIONS = params.PERF_ITERATIONS.toInteger()
```
smlambert:
```suggestion
int PERF_ITERATIONS = params.PERF_ITERATIONS ? params.PERF_ITERATIONS.toInteger() : 4
```

buildenv/jenkins/perfPipeline.groovy
```diff
@@ -120,49 +160,82 @@ def generateChildJobViaAutoGen(newJobName) {
...
+        }
+
+        def runMetrics = readJSON file: fname 
+
+        testList.each { test ->
+                aggregateMetrics[test].each { metric -> 
+                        def value = runMetrics[test][metric.key]["value"]
+                        if (value != null) metric.value[testType]["values"] << value
+                }
+        }
```
smlambert:
```suggestion
        } }
```

smlambert (changes requested):
Putting a block on this, as I have pushed some commits that I will need to remove/fix in the process of resolving merge conflicts.

PR #6452, opened 2025-07-24
Restrict dacapo-luindex test to Java 11+

phdiep2003:
The luindex benchmark requires Java 11 or newer to run.
Java 8 is incompatible due to API limitations.

This change replaces the <disables> block with a <versions> block
to explicitly declare version support in the playlist file. 

Fixes: #4858

### DIFF
--- a/perf/dacapo/playlist.xml
+++ b/perf/dacapo/playlist.xml
@@ -109,12 +109,9 @@
 	</test>
 	<test>
 		<testCaseName>dacapo-luindex</testCaseName>
-                <disables>
-                        <disable>
-                                <comment>https://github.com/adoptium/aqa-tests/issues/4858#issuecomment-2968494739</comment>
-                                <version>8</version>
-                        </disable>
-                </disables>
+		<versions>
+			<version>11+</version>
+		</versions>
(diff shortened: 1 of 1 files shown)


perf/dacapo/playlist.xml
```diff
@@ -109,12 +109,9 @@
...
 		<testCaseName>dacapo-luindex</testCaseName>
-                <disables>
-                        <disable>
-                                <comment>https://github.com/adoptium/aqa-tests/issues/4858#issuecomment-2968494739</comment>
-                                <version>8</version>
-                        </disable>
-                </disables>
+		<versions>
+			<version>11+</version>
+		</versions>
```
llxia: This is correct. But for readability and to align with others, can we move it after line 122 (`</groups>`)?

phat-henry-diep: > This is correct. But for readability and to align with others, can we move it after line 122 (`</groups>`)?

Sure. I'd move it now

llxia: We should move 
```
		<versions>
			<version>11+</version>
		</versions>
```

after line 122 `</groups>`.

phat-henry-diep: > We should move
> 
> ```
> 		<versions>
> 			<version>11+</version>
> 		</versions>
> ```
> 
> after line 122 `</groups>`.

Sorry, I have fixed that!

llxia (approved):

karianna (approved):

PR #6457, opened 2025-07-25
Fix Arctic dev.jck playback issues

andrew-m-leonard:
Fix various issues with running Arctic playback of jck test cases:
- OS detection was incorrect for Windows
- Arctic itself must be run using a specific latest Temurin jdk-21+, which needs to be installed on the target nodes:
  - Linux: /usr/bin/java
  - Mac: /Library/Java/JavaVirtualMachines/temurin-21.jdk/Contents/Home 
  - Windows: c:/openjdk/jdk-21
- If an Arctic glitch occurred and testcases were not run at all, the job was returning SUCCESS, ensure testcases are actually run... and overall success flag is set properly
- The Testcase launch command was incorrect for the various platforms
- The Testcase process detection logic was incorrect for Windows
-

### DIFF
--- a/jck/interactives/arctic_tests/runTests.sh
+++ b/jck/interactives/arctic_tests/runTests.sh
@@ -350,7 +350,7 @@ for i in "${active_versions[@]}"; do
 
             # Check testcase started successfully.
             ps -p $test_pid 2>/dev/null 1>&2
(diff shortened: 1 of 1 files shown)


sxa (approved):
Has been tested and LGTM

Issue #6451, opened 2025-07-24
Testcase exclusions from July 2025 Release


adamfarley:
**Summary**

This is a list of tests identified as non-blocking in the recent July 2025 release.

All tests listed here will be excluded in the Temurin exclude files.

To be excluded as part of this issue, a testcase must:
- Be a consistent failure, not an intermittent failure.
- Have run and failed in the July 2025 release
- Be identified as a non-blocker

**Sources:**
- JDK8u462-b08  - https://github.com/adoptium/aqa-tests/issues/6429
- JDK-11.0.28+6 - https://github.com/adoptium/aqa-tests/issues/6427
- JDK-17.0.16+8 - https://github.com/adoptium/aqa-tests/issues/6422
- JDK-21.0.8+9  - https://github.com/adoptium/aqa-tests/issues/6421
- JDK-24.0.2+12 - https://github.com/adoptium/aqa-tests/issues/6420

**JDK8**
sun/management/jmxremote/bootstrap/SSLConfigFilePermissionTest.sh https://github.com/adoptium/aqa-tests/issues/5917 solaris-x64
javax/net/ssl/ciphersuites/DisabledAlgorithms.java https://github.com/adoptium/aqa-tests/issues/5915 solaris-x64
sun/tools/native2ascii/Native2AsciiTests.sh https://github.com/adoptium/aqa-tests/issues/5893 alpine x64

**JDK11**
jdk/jshell/ExternalEditorTest.java https://github.com/adoptium/aqa-tests/issues/6458 alpine x

**JDK17**
jdk/jshell/ExternalEditorTest.java https://github.com/adoptium/aqa-tests/issues/6458 alpine x
runtime/ErrorHandling/MachCodeFramesInErrorFile.java https://bugs.openjdk.org/browse/JDK-8347908 linux-arm
compiler/loopopts/superword/TestEliminateAllocationWithCastP2XUse.java https://github.com/adoptium/aqa-tests/issues/6459 windows-x86

**JDK21**
runtime/memory/ReadFromNoaccessArea.java https://github.com/adoptium/aqa-tests/issues/5671 windows-aarch64
serviceability/jvmti/RedefineClasses/RedefineSharedClassJFR.java https://github.com/adoptium/aqa-tests/issues/5671 windows-aarch64
jdk/jshell/ExternalEditorTest.java https://github.com/adoptium/aqa-tests/issues/6458 x86-64_alpine-linux, aarch64_alpine-linux 
[...]

PR #6462, opened 2025-07-28
Exclude nonblockers from july 2025 release

adamfarley:
And add the ability to exclude tests on Alpine-only

Test runs:
- Non-Alpine: https://ci.adoptium.net/job/Grinder/13810/ - Passed, no extra exclude file.
- Alpine: https://ci.adoptium.net/job/Grinder/13809/ - Passed, exclude file added.
- Alpine with excluded file in target: https://ci.adoptium.net/job/Grinder/13813/ - Passed, excluded test was not run.

Resolves https://github.com/adoptium/aqa-tests/issues/6451

### DIFF
--- a/openjdk/openjdk.mk
+++ b/openjdk/openjdk.mk
@@ -246,17 +246,17 @@ else ifneq (,$(findstring OpenJCEPlus, $(TEST_FLAG)))
 endif
 
 # If we are on alpine, also use the exclude file specific to alpine.
-ALPINE_PROBLEM_LIST_FILE:=$(Q)$(TEST_ROOT)$(D)..$(D)jvmtest$(D)openjdk$(D)excludes$(D)alpine$(D)ProblemList_openjdk$(JDK_VERSION)_alpine.txt$(Q)
+ALPINE_PROBLEM_LIST_FILE:=$(TEST_ROOT)$(D)..$(D)jvmtest$(D)openjdk$(D)excludes$(D)alpine$(D)ProblemList_openjdk$(JDK_VERSION)_alpine.txt
 ifneq (,$(findstring alpine, $(SPEC)))
-	ifneq (,$(wildcard $(ALPINE_PROBLEM_LIST_FILE)))
+	ifneq (,$(wildcard $(realpath ALPINE_PROBLEM_LIST_FILE)))
 		ifeq (,$(FEATURE_PROBLEM_LIST_FILE))
-			FEATURE_PROBLEM_LIST_FILE:=-exclude:$(ALPINE_PROBLEM_LIST_FILE)
+			FEATURE_PROBLEM_LIST_FILE:=-exclude:$(Q)$(ALPINE_PROBLEM_LIST_FILE)$(Q)
 		else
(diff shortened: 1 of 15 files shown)


adamfarley:
Requesting reviews from @ShelleyLambert and @sophia-guo please.

It would be best to create an upstream fix. I see that as a potential long-term goal, but not something likely to happen in the next few weeks.

openjdk/openjdk.mk
```diff
@@ -245,6 +245,15 @@ else ifneq (,$(findstring OpenJCEPlus, $(TEST_FLAG)))
...
 
+# If we are on alpine, also use the exclude file specific to alpine.
+ifneq (,$(findstring alpine, $(SPEC)))
+	ifeq (,$(FEATURE_PROBLEM_LIST_FILE))
+		FEATURE_PROBLEM_LIST_FILE:=-exclude:$(Q)$(TEST_ROOT)$(D)openjdk$(D)excludes$(D)alpine$(D)ProblemList_openjdk$(JDK_VERSION)_alpine.txt$(Q)
+	else
+		FEATURE_PROBLEM_LIST_FILE+=-exclude:$(Q)$(TEST_ROOT)$(D)openjdk$(D)excludes$(D)alpine$(D)ProblemList_openjdk$(JDK_VERSION)_alpine.txt$(Q)
+	endif
+endif
+
```
sophia-guo, line 262: This needs one more check to deal with if ProblemList_openjdk$(JDK_VERSION)_alpine.txt doesn't exist. For example whenever the new version is available but the problemlist is not available yet or no need for that new version. 

[See example https://github.com/adoptium/aqa-tests/pull/6462/files#diff-c3d02f1e9a9f5dfed049a450b5d0fba4727b7b3d42f0a0771f2e41e408d6ef14L250](https://github.com/adoptium/aqa-tests/pull/6462/files#diff-c3d02f1e9a9f5dfed049a450b5d0fba4727b7b3d42f0a0771f2e41e408d6ef14L250-L254)

adamfarley: I considered that, but opted not to so that if the file doesn't exist (or has the wrong name) we fail fast when we try to use it.

What do you think?

sophia-guo: The file should not be necessary especially if upstream support that later so I would prefer not to fail the job.

adamfarley: Okie dokie. Will make that change. Is the rest ok?

adamfarley: Running a test with the requested change: https://ci.adoptium.net/job/Grinder/13882/
Didn't work due to the use of warning causing "recipe commences before first target".
Will fix tomorrow.

adamfarley: Fixed.

Test run with an incorrect ProblemList path (so the warning message can be tested):
https://ci.adoptium.net/job/Grinder/13887/

Test run with the correct ProblemList path: 
https://ci.adoptium.net/job/Grinder/13894

Both run and display the warning appropriately. 
@sophia-guo - What do you think?

sophia-guo (approved):

smlambert (approved):
thanks @adamfarley !

Issue #6447, opened 2025-07-23
Temporary exclusion of JDK21 tests on z/OS.


psoujany:
Below tests needs to be excluded temporarily on z/OS.
1. sun/security/ec/ed/TestEdDSA.java : refer openj9-openjdk-jdk21-zos/issues/212
2. java/nio/channels/FileChannel/directio/PreadDirect.java : refer openj9-openjdk-jdk21-zos/issues/174
3. java/nio/channels/FileChannel/directio/PwriteDirect.java : refer openj9-openjdk-jdk21-zos/issues/174
4. java/nio/channels/FileChannel/directio/ReadDirect.java : refer openj9-openjdk-jdk21-zos/issues/174
5. java/nio/channels/FileChannel/directio/WriteDirect.java : refer openj9-openjdk-jdk21-zos/issues/174

PR #6464, opened 2025-07-29
Exclude few tests on JDK17/21/25 on z/OS which are not supported

psoujany:
This PR is to exclude few tests mentioned in https://github.com/adoptium/aqa-tests/issues/6447 which are not supported on z/OS.


llxia (approved):

pshipton (approved):

karianna:
@psoujany - Can you address Peter's comment? Then we can merge :-)

psoujany:
@pshipton I've excluded 2 of the permanent excludes on JDK25 z/OS. Thank you.

PR #6469, opened 2025-07-31
Re-enable dacapo-tomcat test for JDK 11+

abdyus:
- Removed the disable block that was preventing dacapo-tomcat from running
- Added version constraint to restrict the test to JDK 11+ only

The test failures on JDK 8 confirm that restricting to JDK 11+ is the correct approach

### DIFF
--- a/perf/dacapo/playlist.xml
+++ b/perf/dacapo/playlist.xml
@@ -24,7 +24,7 @@
 		<command>$(JAVA_COMMAND) $(JVM_OPTIONS) -jar $(Q)$(TEST_RESROOT)$(D)dacapo.jar$(Q) eclipse; \
 		$(TEST_STATUS)</command>
 		<versions>
-			<version>8</version>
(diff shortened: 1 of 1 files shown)


karianna (approved):

perf/dacapo/playlist.xml
```diff
@@ -24,7 +24,7 @@
 		<command>$(JAVA_COMMAND) $(JVM_OPTIONS) -jar $(Q)$(TEST_RESROOT)$(D)dacapo.jar$(Q) eclipse; \
```
llxia, line 18: For dacapo-eclipse, can we remove `<disables>` block? We are going to run on JDK11+ only. Thanks

https://github.com/adoptium/aqa-tests/pull/6469/files#diff-46ae8839155061285da2bbeddeac88e0b4481fe6d049a8549c9b3ccdca360fd9R18-R23

PR #6470, opened 2025-07-31
JDK24+ fix aarch64_mac platform exclude

JasonFengJ9:
JDK24+ fix `aarch64_mac` platform exclude

Replace it with `macosx-aarch64`.

Related to
* https://github.com/eclipse-openj9/openj9/pull/22339


smlambert:
This was previously using aarch64_mac, but I think jtreg exclude format use an underscore, rather a dash, perhaps try aarch64-mac.

smlambert (approved):

Issue #6471, opened 2025-07-31
Consolidate intermittent test failure list issues


adamfarley:
**Summary**

This covers the work to consolidate many different intermittent test listing issues.

Note that this doesn't cover efforts to fix these issues. Just links to those efforts, or some indication of where the issue was first noted.

adamfarley:
**Issues that can be closed once this PR is merged**

- https://github.com/adoptium/aqa-tests/issues/1596

PR #6472, opened 2025-07-31
Creating lists of intermittently failing tests

adamfarley:
This change lists known unreliable tests in a consistent format.

Resolves https://github.com/adoptium/aqa-tests/issues/6471


adamfarley:
Requesting reviews from @ShelleyLambert and @sophia-guo please.

smlambert (changes requested):
Thank you for compiling this list @adamfarley.  Only openjdk testcase failures are handled by jtreg exclude Problemlist format. For the other test groups that are part of AQAvit, they get excluded via playlist exclusions.

I would have expected the perf and system targets to be listed via playlist files (which do not use a different file for each version as the problem lists do).  

I guess it depends on what we intend to do with these lists.  If we are adding a feature to exclude unreliable targets on the fly, or post-process failures against a list, we have to consider what format to use, and how to organize them within the directory (as different vendors may eventually have different lists).

adamfarley:
> Thank you for compiling this list @adamfarley. Only openjdk testcase failures are handled by jtreg exclude Problemlist format. For the other test groups that are part of AQAvit, they get excluded via playlist exclusions.

I noticed that. For Perf and Systemtest, I used the jtreg format for consistency, rather than explicit usability (though a consistent format does make any future parsing easier)

> I would have expected the perf and system targets to be listed via playlist files (which do not use a different file for each version as the problem lists do).

I considered that, but opted not to because I'm unaware of a mechanism for disabling targets via the playlist files *sometimes*. As far as I know, for a given vendor, platform, and version, a target is either enabled or disabled. An openjdk test can include an extra exclude file dynamically, so we can choose whether to exclude the test in some runs or not.

> I guess it depends on what we intend to do with these lists. If we are adding a feature to exclude unreliable targets on the fly, or post-process failures against a list, we have to consider what format to use, and how to organize them within the directory (as different vendors may eventually have different lists).

My first priority with this task was to provide a unified list of all intermittently failing tests.
- The lists should use a single format to allow parsing and easy-reading.
- The lists should be stored in the repository to allow use at runtime.

Specific uses were a secondary priority. My first thought was that they could be used as exclude files during releases (or during reruns, or reruns during a release), but I didn't want to sidetrack myself looking too much into applications. I figured that if we did pick an application for this data, the best thing I can do right now is to use a consistent format across the board, to make conversion/parsing easier later on.

I'll add a note to the General Retrospective to start discussion into some of our options here.

smlambert:
openjdk jtreg problemlists are the 'odd man out'... no other test group uses that format

I do not want to store non-openjdk test material in problemlist format as it might lead people to think that is the commonly used format, which it is not.  

Agree that we need to have a requirements gathering session and design discussion on this (in terms of formats, and how to structure and where to locate such files).  If this is merely a form of documentation, and not a plan for using for debug / temp exclude, then likely all could be in a doc folder.

adamfarley:
The perf and system tests are a goal for later, so I've split those files out into https://github.com/adoptium/aqa-tests/pull/6571

The work to exclude these tests at runtime (and the discussion around that) will be in a third PR so we can get [this PR](https://github.com/adoptium/aqa-tests/pull/6472) merged and ready for reference.

adamfarley:
@sophia-guo 

Heya. I see that the [exclude_parser.py](https://github.com/adoptium/aqa-tests/blob/c22139f36bafa9c8676de47b6ed3fdf65e5bf14c/scripts/disabled_tests/exclude_parser.py#L126) specifies that every uncommented line needs to be exactly 3 elements.

[In the jtreg FAQ](https://openjdk.org/jtreg/faq.html#how-do-i-list-the-tests-in-an-exclude-file-such-as-problemlist.txt), it allows for the 4+ elements to be comments.

What do you think about me changing [that code](https://github.com/adoptium/aqa-tests/blob/c22139f36bafa9c8676de47b6ed3fdf65e5bf14c/scripts/disabled_tests/exclude_parser.py#L126) to be a minimum of 3, rather than exactly 3?

sophia-guo:
@adamfarley yes, you can do that. Though personally I thinks description item is unnecessary as the second element of links( In adoptium we use links as bug information) should already have enough information include description.

smlambert (approved):

karianna:
@adamfarley will need a rebase also

PR #6487, opened 2025-08-07
Add unique group_label to build variables in getEBCNode function

mahdipub:
This avoid issue with merging similar run of create node pipeline into one and not creating a node per each test list

Signed-off-by: mahdi@ibm.com

### DIFF
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -739,9 +739,10 @@ def getEBCNode() {
     propagate: false,
     wait: true,
     parameters: [
-        string(name: 'platform', value: params.PLATFORM),
-        string(name: 'nodeType', value: 'testmachine'),
+        string(name: 'PLATFORM', value: params.PLATFORM),
+        string(name: 'NODE_TYPE', value: 'testmachine'),
         string(name: 'TIME_LIMIT', value: expectedNodeLifespan.toString()), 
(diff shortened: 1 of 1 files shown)


AdamBrousseau (approved):

buildenv/jenkins/openjdk_tests
```diff
@@ -742,6 +742,7 @@ def getEBCNode() {
         string(name: 'platform', value: params.PLATFORM),
         string(name: 'nodeType', value: 'testmachine'),
         string(name: 'TIME_LIMIT', value: expectedNodeLifespan.toString()), 
+        string(name: 'group_label', value: UID.randomUUID().toString()),
```
llxia: I think it is misleading to set UUID as `group_label`. Instead `group_label`, call it `JOB_UUID` as the parameter name. At the end of the day, this is a dummy parameter. 

btw, in EBC_Create_Node job, the parameters are mixed with upper case, camel case, and lower case. We should use one naming format.

mahdipub: Actually the function of that parameter is label in reality but we pass a random UUID here which can be replaced by any meaningful string instead. The `EBC Create Node` set that string on machine as label and with that test pipeline access created nodes. So it is not just a workaround as dummy param but instead by deciding the label  inside the AQA pipeline we avoid that mentioned issue.

For the param names, you are absolutely right. I will unify it.

LongyuZhang (approved):
LGTM

Issue #6378, opened 2025-07-03
BUILD_LIST test job parameter incorrectly set to openjdk


adamfarley:
**Summary**
Test jobs are being run with the wrong BUILD_LIST parameter, resulting in the following errors:

```
No tests found for target: \<Test target\>
```

and

```
java.lang.AssertionError: Build failed because cannot find NUM_LIST in parallelList.mk file
```

**Examples**
https://ci.adoptium.net/job/Test_openjdk21_hs_dev.system_x86-64_windows/41
https://ci.adoptium.net/job/Test_openjdk21_hs_sanity.functional_aarch64_linux/300

**Analysis**

These jobs are being run by build jobs, so I focused the investigation there.

It seems that the build jobs do not explicitly set the BUILD_LIST parameter for the test job ([evidence](https://github.com/adoptium/ci-jenkins-pipelines/blob/683fd106f242e96580464bed13671b3109e38e8c/pipelines/build/common/openjdk_build_pipeline.groovy#L498)), so I think the test job auto gen is incorrectly setting the BUILD_LIST somehow.

sophia-guo:
Seems the AUTO GENE job generate with the wrong BUILD_LIST https://ci.adoptium.net/job/Test_openjdk21_hs_sanity.functional_aarch64_linux/jobConfigHistory/showDiffFiles?timestamp1=2025-05-15_12-01-35&timestamp2=2025-07-01_06-00-21

sophia-guo:
https://ci.adoptium.net/job/Test_openjdk21_hs_sanity.functional_aarch64_linux/ was generated by with following params, which looks good `@string(name=GROUPS,value=functional)`

23:56:45  Use Test_Job_Auto_Gen to generate AQA test job with parameters: [@string(name=JDK_VERSIONS,value=21), @string(name=JDK_IMPL,value=hotspot), @string(name=TIME_LIMIT,value=25), @string(name=ARCH_OS_LIST,value=aarch64_linux), @booleanParam(name=LIGHT_WEIGHT_CHECKOUT,value=false), @string(name=LEVELS,value=sanity), @string(name=GROUPS,value=functional), @string(name=TEST_JOB_NAME,value=Test_openjdk21_hs_sanity.functional_aarch64_linux)]

Rege job is not available any more 00:00:22  Build [Test_Job_Auto_Gen #33711](https://ci.adoptium.net/job/Test_Job_Auto_Gen/33711/) completed: SUCCESS

[...]

PR #6510, opened 2025-08-15
Check BUILD_LIST default value is set correctly

adamfarley:
A check to ensure we never create a test job where the BUILD_LIST parameter has an invalid default value.

This should ensure we know https://github.com/adoptium/aqa-tests/issues/6378 is happening if the symptom crops up again.

### DIFF
--- a/buildenv/jenkins/testJobTemplate
+++ b/buildenv/jenkins/testJobTemplate
@@ -186,6 +186,10 @@ ARCH_OS_LIST.each { ARCH_OS ->
 				if (!expectedGroups.contains(GROUP)) {
 					assert false : "GROUP ${GROUP} is not in the the expected groups: ${expectedGroups}."
 				}
+				if ( BUILD_LIST && !BUILD_LIST.contains(GROUP) ) {
+					// Fail immediately, before we create a job that cannot build the tests it is expected to run.
+					assert false : "BUILD_LIST has been set (${BUILD_LIST}), but it does not contain the tests we expect the target job to run: ${GROUP}."
+				}
 				def ACTUAL_TEST_JOB_NAME = TEST_JOB_NAME ? TEST_JOB_NAME : "Test_openjdk${JDK_VERSION}_${JDK_IMPL_SN}_${LEVEL}.${GROUP}_${ARCH_OS}${SUFFIX}"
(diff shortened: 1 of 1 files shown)


adamfarley:
Please can @sophia-guo and @sxa review and merge when they can.

buildenv/jenkins/testJobTemplate
```diff
@@ -186,6 +186,10 @@ ARCH_OS_LIST.each { ARCH_OS ->
 				if (!expectedGroups.contains(GROUP)) {
 					assert false : "GROUP ${GROUP} is not in the the expected groups: ${expectedGroups}."
 				}
+				if ( BUILD_LIST && !BUILD_LIST.contains(GROUP) ) {
+					// Fail immediately, before we create a job that cannot build the tests it is expected to run.
+					assert false : "BUILD_LIST has been set (${BUILD_LIST}), but it does not contain the tests we expect the target job to run: ${GROUP}."
```
karianna:
```suggestion
					assert false : "BUILD_LIST has been set to (${BUILD_LIST}), but it does not contain the tests we expect the target job to run in the Group: ${GROUP}."
```

adamfarley: Good idea, thanks Martijn 🤓

llxia:
https://github.com/adoptium/aqa-tests/issues/6378 is not due job template. The original job parameters were generated correctly. The BUILD_LIST and TARGET were set wrong (mismatched) at runtime. 

https://ci.adoptium.net/job/Test_openjdk21_hs_dev.system_x86-64_windows/41/rebuild/parameterized

image

adamfarley:
> #6378 is not due job template. The original job parameters were generated correctly. The BUILD_LIST and TARGET were set wrong (mismatched) at runtime.

I'm not sure I understand what you mean by "generated correctly". They seem incorrect in the example you gave.

My understanding is that Test_Job_Auto_Gen, if run on specific node/s, runs with the mistaken impression that BUILD_LIST has been set to "openjdk", which (in the template) gets treated as an override to the default value (the name of the test group).

So the default BUILD_LIST value for the test job is now wrong, and any attempt to run that job without explicitly setting BUILD_LIST will now fail to run properly because we'll be building the wrong test material.

The template isn't the source of this problem, but it is a good opportunity for us to catch the issue, produce a useful error, and to prevent the test job being generated with an incorrect default BUILD_LIST.

PR #6515, opened 2025-08-20
Locale Display name update for Taiwanese

rishabh-ibm:
Applicable for JDK24+.
https://cldr.unicode.org/translation?languages-names-which-were-added-or-changed-in-english

As per the documentation linked above, for zh_TW the display name Taipei Time was changed to Taiwan Time. As this test case validates using the locale display name, we are updating the test case to use the updated timezone name.

Fixes : https://github.com/eclipse-openj9/openj9/issues/22454

### DIFF
--- a/functional/MBCS_Tests/language_tag/src/TWTagTest.java
+++ b/functional/MBCS_Tests/language_tag/src/TWTagTest.java
@@ -127,7 +127,7 @@ public class TWTagTest{
         String tag = "zh-TW-u-tz-twtpe"; //Defined in common/bcp47/teimezone.xml
         Locale l = Locale.forLanguageTag(tag);
         assertEquals("中文 (台灣，時區：台北時間)", l.getDisplayName(l));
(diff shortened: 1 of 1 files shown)


llxia:
Could you please run Grinder? Thanks

pshipton (approved):

Darshan-N3:
>On second thought, if the change is only in one file, checking the JDK version at runtime will be easier.

I’ve made the suggested change here: [Branch Link](https://github.com/adoptium/aqa-tests/compare/master...Darshan-N3:aqa-tests:Tw_TaiwanFix?expand=1). Could you please review it and let me know if any further modifications are needed? Also, kindly advise which release I should target for creating the PR. Thanks!

llxia:
Please test the PR with Grinder for JDK21 and JDK25. The PR should be created against `master`. Once it is merged, then cherry-pick it back into the `v1.0.10-release` branch.

Issue #6513, opened 2025-08-18
Add child build status/link at parent build level


llxia:
Displaying the child build status and link at the parent build level would improve usability, eliminating the need for users to search for the child job link in the console output.

For example: 

<img width="485" height="91" alt="Image" src="https://github.com/user-attachments/assets/b114bc3a-0bcb-4542-bfdf-238a0641455c" />

llxia:
PRs:
https://github.com/adoptium/aqa-tests/pull/6521
https://github.com/adoptium/aqa-tests/pull/6523

pshipton:
We could backport this to v1.0.9.

PR #6540, opened 2025-08-28
Sort child build status/link by invocation number

keithc-ca:
Addresses https://github.com/adoptium/aqa-tests/issues/6513#issuecomment-3221576539.


smlambert (approved):

buildenv/jenkins/JenkinsfileBase
```diff
@@ -1596,7 +1596,9 @@ def archiveChildJobTap(childJobs, originalStatus = 'SUCCESS') {
 		forceCleanWS()
 		try {
 			def buildPaths = ""
-			childJobs.each {
+			childJobs.stream()
+					.sorted { Comparator.comparing { it.value.getRawBuild().getNumber() } }
```
llxia: We should sort on the build name (getProjectName), not the build number.

keithc-ca: At first, I was going to respond that `getProjectName` would be the same for all those child jobs, but looking at https://openj9-jenkins.osuosl.org/job/Test_openjdk25_j9_sanity.openjdk_ppc64le_linux_Personal/2, for example, I
think `getProjectName` ends with `openjdk_ppc64le_linux_Personal_testList_2` and `getNumber` is `2`. I not sure it will make a difference to the sorting order, but will update this as you suggest.

Fixed in 1b548c4ab930a32c44efa7764bdf637cca4e92aa.

llxia: We set the value to `build name/build number`. The build number can be any number. It may not be the same as the build name suffix _testList_XXX. For example: 

image

It does not make sense to order based on `build number`.

keithc-ca: So the "build numbers" are 231, 238, 422, etc.?

keithc-ca: And in the grinder the 'name' fields are behind `getProjectName()`?
```
14:32:42      GeneratedJob{name='Grinder_testList_0'}
14:32:42      GeneratedJob{name='Grinder_testList_1'}
```

keithc-ca:
I'd like to see this through: I'm learning something new. ;-)

buildenv/jenkins/JenkinsfileBase
smlambert: What happens with an 11th child job (`_10`)?

keithc-ca: I expect the current proposal will put `_10` between `_1` and `_2`.

Do the child jobs have an integer property that could be queried? Where are the project names are constructed (that probably has what we want)? Alternatively I can construct a more involved comparator that would do the right thing when 10 or more child jobs are present.

llxia: I am not sure we can query API for the integer property. You may need to write a regex to get the integer.


> Where are the project names are constructed


https://github.com/adoptium/aqa-tests/blob/0905ea0d2a44744fe67c904a5313f77708a4a25d/buildenv/jenkins/JenkinsfileBase#L289
https://github.com/adoptium/aqa-tests/blob/0905ea0d2a44744fe67c904a5313f77708a4a25d/buildenv/jenkins/JenkinsfileBase#L291

keithc-ca: I have an update that I will push shortly that aims to have a fixed length suffix (use the same number of digits for `${i}` for all jobs) so we would have `_00`, ... `_10`, etc. when there are more than 10 jobs.

keithc-ca: See b46406b436fe637d89ed1d40cec7012588ed26c3.

keithc-ca:
Updated to avoid restricted Java API.

llxia:
[Grinder](https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/54357/console) - still have failure.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -281,14 +281,22 @@ def setupParallelEnv() {
 		echo "[PARALLEL: ${env.PARALLEL}] childJobNum is ${childJobNum}, creating jobs and running them in parallel..."
 		create_jobs = [:]
 		parallelTestParams =[:]
+
+		def format = "%d"
+		if (childJobNum > 10) {
+			def maxDigits = (int) Math.log10(childJobNum - 1) + 1
+			format = "%0${maxDigits}d"
+		}
```
keithc-ca: This part of the change requires a corresponding change in TKG's [TestDivider.java](https://github.com/adoptium/TKG/blob/master/src/org/testKitGen/TestDivider.java).

llxia: Could you please provide a Grinder?

keithc-ca: This part of the change will not be necessary. I intend to revert this and improve the sorting comparator.
I'll share a link to a grinder once I have that done.

keithc-ca: This [grinder](https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/54396) is testing what I plan to push to this branch.

keithc-ca: That grinder passed, showing the status links in sorted order; the updated change pushed here.

buildenv/jenkins/JenkinsfileBase
keithc-ca: Rather than change the suffixes, I improved the sorting comparator to handle more than 10 jobs.

keithc-ca:
@llxia Could you please have another look at this? Thanks.

buildenv/jenkins/JenkinsfileBase
```diff
@@ -1587,6 +1587,29 @@ def triggerRerunJob (String originalStatus, Map rerunTestJobParams = null) {
...
 
+// Sort the given list of jobs by project name.
+@NonCPS
+def sortByProjectName(jobs) {
+	jobs.sort { job1, job2 ->
+		def name1 = job1.getProjectName()
+		def name2 = job2.getProjectName()
+		def index1 = name1.lastIndexOf('_')
+		def index2 = name2.lastIndexOf('_')
+		def result = name1 <=> name2
```
llxia: Just a minor cleanup. Since this `def result = name1 <=> name2` gets overridden if both names have an underscore, should we put it after the if block?

```
...
		if ((index1 >= 0) && (index2 >= 0)) {
			def suffix1 = name1.substring(index1 + 1)
			def suffix2 = name2.substring(index2 + 1)

			return name1.substring(0, index1) <=> name2.substring(0, index2)
					?: suffix1.length() <=> suffix2.length()
					?: suffix1 <=> suffix2
		}
		return name1 <=> name2
}
```

keithc-ca: Updated to avoid computation of `name1 <=> name2` when not necessary in 86baaeeaed9de70a9428bf7cb4472cd843a6c727.

llxia:
[Grinder](https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/54462/)

Issue #6555, opened 2025-09-06
Sept 2025 - JDK 25 Dry-run triage


smlambert:
The dry-run occurred across several build pipelines:
All platforms: https://ci.adoptium.net/job/build-scripts/job/release-openjdk25-pipeline/5

aLinux: https://ci.adoptium.net/job/build-scripts/job/release-openjdk25-pipeline/7/
xMac: https://ci.adoptium.net/job/build-scripts/job/release-openjdk25-pipeline/8/
Aix: https://ci.adoptium.net/job/build-scripts/job/release-openjdk25-pipeline/9/

---

smlambert:
aarch64_mac ✔️

extended.openjdk - 1 testcase failing: java/net/CookieHandler/B6644726.java failure, appears to be hard-coded expired cookies in testcase, https://bugs.openjdk.org/browse/JDK-8365811 - happening across all platforms, non-blocking, should get excluded in v1.0.9-release branch

smlambert:
x86-64_linux

sanity.openjdk - 1 testcase failing, runtime/os/AvailableProcessors.java on test-docker-ubi9-x64-1, passes in rerun on test-docker-fedora42-x64-1

extended.openjdk - 2 testcases failing, java/net/CookieHandler/B6644726.java / [JDK-8365811](https://bugs.openjdk.org/browse/JDK-8365811) and runtime/os/AvailableProcessors.java failing on test-docker-ubuntu2404-x64-1

smlambert:
x86-64_alpine-linux

extended.openjdk - 2 testcase failures, non-blocking
- java/net/CookieHandler/B6644726.java
- java/nio/file/DirectoryStream/SecureDS.java - not supported on this platform, non-blocking, should be excluded, have to use an awkward workaround way of doing so

ACTION: exclude java/nio/file/DirectoryStream/SecureDS.java for alpine-linux platforms

smlambert:
aarch64_alpine-linux

sanity.openjdk - 2 testcase failures on test-docker-alpine320-armv8-4 pass in rerun on same node (intermittent)
- runtime/ErrorHandling/MachCodeFramesInErrorFile.java.MachCodeFramesInErrorFile
- runtime/ErrorHandling/StackWalkNativeToJava.java.StackWalkNativeToJava

extended.openjdk - 3 testcase failures
- java/nio/file/DirectoryStream/SecureDS.java (known, needs exclusion)
[...]

PR #6564, opened 2025-09-09
Exclude unsupported test on alpine-linux

sophia-guo:
java/nio/file/DirectoryStream/SecureDS.java

refer to https://github.com/adoptium/aqa-tests/issues/6555#issuecomment-3263769483


adamfarley (approved):
LGTM

PR #6568, opened 2025-09-09
Run AIX tests for jdk25+ with the 17.1 C++ Runtime

pshipton:
The label `sw.tool.c++runtime.17_1` is added only for jdk25+.

jdk25: https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/54595
`nodes with the 'ci.role.test&&hw.arch.ppc64&&sw.os.aix&&sw.tool.c++runtime.17_1&&ci.project.openj9' label`

jdk21: https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/54573/
`nodes with the 'ci.role.test&&hw.arch.ppc64&&sw.os.aix&&ci.project.openj9' label`

### DIFF
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -334,13 +334,18 @@ timestamps{
                 LABEL = params.LABEL
             } else {
                 LABEL = PLATFORM_MAP[params.PLATFORM]["LABEL"]
-                    if (params.BUILD_LIST.contains("perf")) {
-                        def perfLabel = LABEL.minus("ci.role.test&&").concat("&&ci.role.perf")
-                        if (areNodesWithLabelOnline(perfLabel)) {
+                if (params.BUILD_LIST.contains("perf")) {
+                    def perfLabel = LABEL.minus("ci.role.test&&").concat("&&ci.role.perf")
+                    if (areNodesWithLabelOnline(perfLabel)) {
                         LABEL = perfLabel
-                        }
                     }
                 }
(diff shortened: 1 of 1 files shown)


buildenv/jenkins/openjdk_tests
```diff
@@ -342,6 +342,12 @@ timestamps{
                     }
                 }
 
+            if (PLATFORM == "ppc64_aix" && env.JENKINS_URL.contains("hyc-runtimes")) {
```
pshipton: This isn't quite right as it still applies the LABEL addition even if the LABEL parameter in the grinder is set. Do you know how to check for that?

pshipton: I think I've got it.

buildenv/jenkins/openjdk_tests
```diff
@@ -342,6 +342,12 @@ timestamps{
                     }
                 }
 
+            if (PLATFORM == "ppc64_aix" && env.JENKINS_URL.contains("hyc-runtimes") && env.LABEL == "") {
```
pshipton: It seems to work but should it be `params.LABEL` instead?

karianna: I think the logic will override the LABEL assignment in the block above, is that the intent?

pshipton: I moved it into the block above.

pshipton: It doesn't override the LABEL, it appends to it.

karianna (approved):

PR #6579, opened 2025-09-11
Enable MiniMix 3hr system tests

annaibm:
- Enable MiniMix 3hr for Mode 501 and Mode 150-CS in extended.system

related:https://github.ibm.com/runtimes/backlog/issues/597

### DIFF
--- a/system/otherLoadTest/playlist.xml
+++ b/system/otherLoadTest/playlist.xml
@@ -181,6 +181,33 @@
 		</impls>
 	</test>
 	<test>
+		<testCaseName>MiniMix_extended_3h</testCaseName>
+		<disables>
+			<disable>
+				<comment>https://github.ibm.com/runtimes/backlog/issues/597</comment>
+				<variation>Mode150-CS</variation>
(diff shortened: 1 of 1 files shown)


system/otherLoadTest/playlist.xml
```diff
@@ -180,6 +180,26 @@
 			<impl>ibm</impl>
 		</impls>
 	</test>
+	<test>
+		<testCaseName>MiniMix_3h</testCaseName>
```
llxia: Let's rename MiniMix_3h to MiniMix_extended_3h for now. Otherwise, we will have 2 `MiniMix_3h_0` and `MiniMix_3h_1` tests. (one from special and the other one from extended). If they both failed, it would be difficult to distinguish between them.

llxia:
Note: Mode150-CS is restricted to xlinux only due to machine capacity concerns. We may enable it on other platforms later, depending on available resources.

JasonFengJ9 (approved):

PR #6612, opened 2025-09-19
Set parent job status to ABORTED if one of the parallel branches status is ABORTED

### LINKED ISSUE #6580: Jenkins test job status not reported correctly if sub jobs are aborted before tap files are available.
Test job status with rerun feature enabled might not report correctly if sub jobs are aborted before tap files are available. For example [ job ](https://ci.adoptium.net/job/Test_openjdk25_hs_sanity.openjdk_aarch64_windows/15/), which has three sub job and one rerun job. One sub job were aborted. The rerun job succeed. The aborted job will not be rerun automatically under currently strategy. So if aborted happened before its Tap files are available, the failure tests won't be considered by unstable rerun sub job. Hence the rerun job status might not correctly show the whole job status. 

Image

sophia-guo:
Fix #6580 

ABORTED jobs won't be triggered rerun automatically. It's possible that TAP file is not generated or archived, which can cause unstable rerun job doesn't have failed testcases of aborted job at all. This avoids parent job status is wrongly marked as SUCCESS due to the SUCCESS status of rerun job.

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -1584,6 +1584,9 @@ def triggerRerunJob (String originalStatus, Map rerunTestJobParams = null) {
 					def rerunResults = archiveChildJobTap(childJobs, originalStatus)
 					if (env.JDK_IMPL == "openj9" || env.JDK_IMPL == "ibm") {
 						currentBuild.result = originalStatus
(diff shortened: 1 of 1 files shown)


andrew-m-leonard (approved):

smlambert (approved):
👍

PR #6620, opened 2025-09-23
Ensure /tmp/.X11-unix is in place before starting weston

sxa:
`/tmp/` should be considered volatile and may be wiped at any point, for example on reboot.

Since `/tmp/.X11-unix` is required to be in place to avoid `weston` (used on EL10 to start an X11-compatibility session on Wayland systems) to start up otherwise a core dump occurs:
```
[15:20:26.798] failed to bind to /tmp/.X11-unix/X0: No such file or directory
Segmentation fault (core dumped)
```
This PR will ensure it is created with appropriate permissions to allow `weston` to start. 

Probably renders https://github.com/adoptium/infrastructure/pull/4069 irrelevant. Seen in https://github.com/adoptium/infrastructure/issues/3868#issuecomment-2884138589

Tested at https://ci.adoptium.net/job/Grinder/14604/ vs the failing version without this patch at https://ci.adoptium.net/job/Grinder/14601

FYI @aswinkr77

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -798,6 +798,7 @@ def runTest( ) {
 			if (env.SPEC.contains('linux') && !(LABEL.contains('ci.agent.dynamic') && CLOUD_PROVIDER == 'azure') && (BUILD_LIST != "external")) {
 				// Detect if Xvfb is on the machine, and if not use the wayland startup code (Initially for EL10)
 				if ( sh(script:"which Xvfb 2>&1", returnStatus:true) != 0 ) {
+					sh "mkdir -p -m 1777 /tmp/.X11-unix"
 					sh "weston --no-config --socket=wayland-vfb --backend=headless-backend.so --xwayland &"
 					env.DISPLAY = ":0"
 


aswinkr77:
ref: https://github.ibm.com/runtimes/automation/issues/632#issuecomment-135868444

cc: @llxia

buildenv/jenkins/JenkinsfileBase
```diff
@@ -798,6 +798,7 @@ def runTest( ) {
 			if (env.SPEC.contains('linux') && !(LABEL.contains('ci.agent.dynamic') && CLOUD_PROVIDER == 'azure') && (BUILD_LIST != "external")) {
 				// Detect if Xvfb is on the machine, and if not use the wayland startup code (Initially for EL10)
 				if ( sh(script:"which Xvfb 2>&1", returnStatus:true) != 0 ) {
+					sh "mkdir -p -m 1777 /tmp/.X11-unix"
```
adamfarley, line 801: Will that fail the pipeline if the folder already exists?
```suggestion
					sh "mkdir -p -m 1777 /tmp/.X11-unix || true"
```
Not a blocker, just curious. Seems fine otherwise.

adamfarley: Maybe this is better:

```suggestion
					if (!fileExists("/tmp/.X11-unix")) {
					    sh "mkdir -p -m 1777 /tmp/.X11-unix"
					}
```

adamfarley (approved):
Addendum

buildenv/jenkins/JenkinsfileBase
sophia-guo: I thinks if the folder exists nothing changes and no errors. Question is the permission will not change, either. So Separate commands will be better?

```
sh "mkdir -p  /tmp/.X11-unix && chmod 1777 /tmp/.X11-unix "
```

sophia-guo: If https://github.com/adoptium/infrastructure/pull/4069 is fixed then this should not be a problem. Leave to @sxa to decide. Approve for now.

sxa: > Will that fail the pipeline if the folder already exists?

The use of `-p` avoids a failure if it exists.

> I thinks if the folder exists nothing changes and no errors. 

If it already exists then I don't want to touch it. It's possible that it was created by another user and if that's the case executing a separate `chmod` would fail. That is a valid scenario to get into if someone gets access to a test machine to replicate a problem without using the `jenkins` ID (and regarding the next point below it will definitely occur if it's in the playbooks since the directory will be owned by `root` initially)

Ref the infrastructure PR, [as I mentioned in there](https://github.com/adoptium/infrastructure/pull/4069#pullrequestreview-3257888428) we should not rely in anything in the playbooks which create things in `/tmp` from being there permanently (It's `tmp` for "temporary" for a reason). I'm not against that playbook PR going in, but this is a much more reliable solution.

Issue #6611, opened 2025-09-19
Add add sleep and retry for cleanWs


llxia:
On Windows, cleanWs may not work as expected. We will try to add sleep and retry, just in case, some processes are still in the process of shutting down.

Related: runtimes/infrastructure/issues/11191

Agaba-derrick:
@llxia  am working on this Issue !

PR #6633, opened 2025-09-30
Add sleep and retry for cleanWs

Agaba-derrick:
This PR adds sleep and retry functionality to the forceCleanWS() function to handle Windows-specific issues where processes may still be shutting down when workspace cleanup is attempted.
Changes Made

Added retry mechanism with 3 attempts and 30-second sleep between retries
Preserved existing exception handling and fallback logic
Added logging for retry attempts
Follows the same retry pattern used in runTest() function

Fixes https://github.com/adoptium/aqa-tests/issues/6611


Agaba-derrick:
@llxia KIndly asking for a review

llxia (approved):

PR #6636, opened 2025-10-02
Allow user to set TRSS_BUILD_LIMIT

llxia:
related: runtimes/automation/issues/662


karianna:
I'm assuming the changing of the default timeout is deliberate

PR #6638, opened 2025-10-02
Remove archiving of cvsclient

### LINKED ISSUE #6637: Remove archiving cvsclient in getDependency script
Sincehttps://github.com/adoptium/aqa-systemtest/pull/502, we no longer fetch and use cvsclient, so we need to remove it from the getDependency script

https://github.com/smlambert/aqa-tests/blob/master/buildenv/jenkins/getDependency#L28

smlambert:
Fixes #6637 

cvsclient is no longer needed (was previously used to fetch mauve src code)


karianna (approved):

smlambert:
Need to also change the way we check for the mauve.jar file (may need to be a 2nd PR in aqa-systemtest)

Issue #5775, opened 2024-11-25
dev.functional suite on jdk8/mac/x64 is hitting 25 hour timeout


sxa:
This job seems to hit the limit consistently: https://ci.adoptium.net/job/Test_openjdk8_hs_dev.functional_x86-64_mac/

In the code below the `02:37` and `03:31` times are from a 25 hour gap, not a one-hour gap:
```
02:37:44  + [[ jdk == \j\d\k ]]
02:37:44  + [[ incompatiblefalse == \c\o\m\p\a\t\i\b\l\e\t\r\u\e ]]
02:37:44  + [[ incompatiblefalse == \i\n\c\o\m\p\a\t\i\b\l\e\f\a\l\s\e ]]
02:37:44  + run_swing_component_test_unset incompatible false
03:31:36  Cancelling nested steps due to timeout
03:31:36  Sending interrupt signal to process
03:31:51  make[3]: *** [dev.functional-testHeadlessComponents] Terminated: 15
03:31:51  make[4]: *** [TestHeadlessComponents_0] Terminated: 15
03:31:51  make[2]: *** [dev.functional-functional] Terminated: 15
03:31:51  /bin/sh: line 1:  5151 Terminated: 15          /Library/Developer/CommandLineTools/usr/bin/make -C .. -f autoGen.mk dev.functional
03:31:51  make[1]: *** [dev.functional-..] Error 143
03:31:51  make: *** [_dev.functional] Error 2
03:31:51  script returned exit code 2
```

![image](https://github.com/user-attachments/assets/480be3a3-406e-46f4-8b03-c0145366fdcf)

smlambert:
2 test targets are in the dev.functional top-level target:

TestHeadlessComponents:
Sept 2, fails on our x64 mac nodes with:
```
[2024-09-02T09:57:08.344Z] + declare -A resArray
[2024-09-02T09:57:08.344Z] /Users/jenkins/workspace/Test_openjdk8_hs_dev.functional_x86-64_mac/aqa-tests/TKG/../functional/testHeadlessComponents/TestHeadlessComponents/testHeadlessComponents.sh: line 191: declare: -A: invalid option
[2024-09-02T09:57:08.344Z] declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
[2024-09-02T09:57:08.344Z] -----------------------------------
[2024-09-02T09:57:08.344Z] TestHeadlessComponents_0_FAILED
```
most recently fails with:
```
[...]

PR #6648, opened 2025-10-09
Disabling TestHeadlessComponents on Mac due to consistent timeouts

adamfarley:
...that cause job failure and conceal other test failures.

Testing here: https://ci.adoptium.net/job/Grinder/14992

Related to (but does not close): #5775


adamfarley:
Thanks @smlambert and @karianna 😎

PR #6651, opened 2025-10-10
Fix for i18n for JDK22+ onwards

psawant19:
Month format is changed in Korean from M\uc6d4 to MMMM
in Chinese from 12 hr to 24 hr format -> ah:mm:ss to HH:mm:ss

Signed-off-by: Pratiksha Sawant <Pratiksha.Sawant@ibm.com>

### DIFF
--- a/functional/MBCS_Tests/i18n/src/DateFormatTest.java
+++ b/functional/MBCS_Tests/i18n/src/DateFormatTest.java
@@ -34,20 +34,14 @@ public class DateFormatTest {
 			version = version.substring(index1+1, index2);
 		}
 	}
-        long feature = JavaVersion.getFeature();
-        if (feature == 16L) {
-		try {
-			resource = ResourceBundle.getBundle("ResourceBundleTest_16", locale);
-		} catch (MissingResourceException e) {} // Do nothing
-	} else if (feature >= 22L) {
-		try {
-			resource = ResourceBundle.getBundle("ResourceBundleTest_22", locale);
-		} catch (MissingResourceException e) {} // Do nothing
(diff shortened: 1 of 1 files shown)


functional/MBCS_Tests/i18n/src/DateFormatTest.java
```diff
@@ -39,11 +39,15 @@ public DateFormatTest(Locale locale) {
...
+	} else if (feature >= 22L) {
+		try {
+			resource = ResourceBundle.getBundle("ResourceBundleTest_22", locale);
+		} catch (MissingResourceException e) {} // Do nothing
+	}else if (feature >= 19L) {
 		try {
 			resource = ResourceBundle.getBundle("ResourceBundleTest_19", locale);
 		} catch (MissingResourceException e) {} // Do nothing
-        }
+	}
```
llxia: Can we reduce try-catch duplication and put the logic into a function?

```
private ResourceBundle tryGetBundle(String baseName, Locale locale) {
	try {
		return ResourceBundle.getBundle(baseName, locale);
	} catch (MissingResourceException e) {
		// Do nothing
		return null;
	}
}

```
Then use:
```
if (feature >= 22L) {
	resource = tryGetBundle("ResourceBundleTest_22", locale);
} else if (feature >= 19L) {
	resource = tryGetBundle("ResourceBundleTest_19", locale);
} else if (feature == 16L) {
	resource = tryGetBundle("ResourceBundleTest_16", locale);
}

```

psawant19: Done. Fixes are made as per the comment.

Issue #6704, opened 2025-10-29
Propose to split jdk_util into smaller tests


llxia:
jdk_util execution time varies on different platforms.

In JDK25, it may take up to 3.5 hours on Windows.

```
ok 1 - jdk_util_1
  ---
    output:
      |
        Test results: passed: 974; excluded: 16; did not match keywords: 8; did not meet platform requirements: 4
    duration_ms: 12661968
```

On xlinux, it takes 40 mins.

```
ok 5 - jdk_util_1
  ---
    output:
      |
        Test results: passed: 973; excluded: 18; did not match keywords: 8; did not meet platform requirements: 3
    duration_ms: 1518169
  ...
```

To achieve a balanced workflow in parallel testing, I think we should split jdk_util into smaller tests.

According to https://github.com/ibmruntimes/openj9-openjdk-jdk25/blob/openj9/test/jdk/TEST.groups#L144C1-L148, let's split into 4 tests.

```
jdk_util_other
jdk_collections
jdk_concurrent
jdk_stream
```

llxia:
@Amrutha-Kanhirathingal could you please work on this issue? Thanks

llxia:
TODO:
- copy https://github.com/adoptium/aqa-tests/blob/8562c1a99947a87e242001304b9f47bd02764f69/openjdk/playlist.xml#L1678-L1709 4 times
- change `<testCaseName>jdk_util</testCaseName>` and `$(Q)$(JTREG_JDK_TEST_DIR):jdk_util$(Q); \` based on the 4 test above
- test all 4 tests on JDK8/11/17/25/26 xlinux and Windows respectively

Amrutha-Kanhirathingal:
|JDK|xlinux|Windows|
|---|---|---|
|JDK8|[jdk_util_other](https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/55957/console) & [jdk_collections,jdk_concurrent,jdk_stream](https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/55962/)||
|JDK11|[All 4 Tests](https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/55958/Console) :||
|JDK17|[All 4 Tests](https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/55959/console) :||
[...]

PR #6706, opened 2025-10-31
Split jdk_util into smaller tests

Amrutha-Kanhirathingal:
Broke down jdk_util to 4 tests jdk_util_other,jdk_collections,jdk_concurrent & jdk_stream

Closes: #6704

### DIFF
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -1675,7 +1675,7 @@
 		</features>
 	</test>
 	<test>
-		<testCaseName>jdk_util</testCaseName>
+		<testCaseName>jdk_util_other</testCaseName>
 		<variations>
 			<variation>$(TEST_VARIATION_DUMP) Mode150</variation>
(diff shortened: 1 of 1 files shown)


Amrutha-Kanhirathingal:
Grinder Links ;

https://github.com/adoptium/aqa-tests/issues/6704#issuecomment-3472732098

openjdk/playlist.xml
```diff
@@ -1674,7 +1674,7 @@
 			<feature>OpenJCEPlus:nonapplicable</feature>
 		</features>
 	</test>
-	<test>
+	<!-- <test>
```
karianna: Should we just delete that commented out block?

karianna (changes requested):

openjdk/playlist.xml
Amrutha-Kanhirathingal: Yes Absolutely,we can remove this block.

PR #6724, opened 2025-11-11
Exclude MauveSingleThrdLoad_J9, MauveSingleInvocLoad_J9 jdk17, 21, 26+

pshipton:
Issue https://github.com/eclipse-openj9/openj9/issues/22912

jdk26 grinder showing disabled tests
https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/4845/

### DIFF
--- a/system/mauveLoadTest/playlist.xml
+++ b/system/mauveLoadTest/playlist.xml
@@ -11,6 +11,14 @@
 				<impl>ibm</impl>
 				<version>8</version>
 			</disable>
+			<disable>
+				<comment>https://github.com/eclipse-openj9/openj9/issues/22912</comment>
+				<version>[17, 21]</version>
+			</disable>
+			<disable>
+				<comment>https://github.com/eclipse-openj9/openj9/issues/22912</comment>
+				<version>26+</version>
+			</disable>
 		</disables>
 		<variations>
 			<variation>Mode110</variation>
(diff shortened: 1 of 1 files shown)


system/mauveLoadTest/playlist.xml
```diff
@@ -11,6 +11,18 @@
 				<impl>ibm</impl>
 				<version>8</version>
 			</disable>
+			<disable>
+				<comment>https://github.com/eclipse-openj9/openj9/issues/22912</comment>
+				<version>17</version>
+			</disable>
+			<disable>
+				<comment>https://github.com/eclipse-openj9/openj9/issues/22912</comment>
+				<version>21</version>
```
JasonFengJ9: I think the two disables above can be combined into one `<version>[17,21]</version>`

pshipton: I tried something like that `[17, 21, 26]` and it didn't work for me.

llxia: we can only do `[17, 21]`. It means it applies to JDK17 to JDK26 inclusive. `[17, 21, 26]` is invalid.

pshipton: Updated it.

JasonFengJ9 (approved):

PR #6753, opened 2025-11-27
Fetching prebuilt mauve jar rather than building our own

adamfarley:
Because the mauve building process is currently broken.

Tested here: https://ci.adoptium.net/job/systemtest.getDependency/256/ - pass


sophia-guo:
Also wondered in this case is it sill necessary to separate systemtest.getDependency job from test.getDependency job considering job configuration is kindly identical? Having two separate jobs  can isolate failures but may increase resource usage and maintenance efforts.

PR #6754, opened 2025-11-27
Use LIBPATH Environment Variable to Run Serviceability JVMTI Tests

Darshan-N3:
Use LIBPATH environment variable option to run serviceability jvmti tests 

Signed-off-by: Darshan N [Darshan.N3@ibm.com](mailto:Darshan.N3@ibm.com)

### DIFF
--- a/openjdk/openjdk.mk
+++ b/openjdk/openjdk.mk
@@ -189,6 +189,9 @@ ifneq ($(JDK_VERSION),8)
 		# else if JDK_IMPL is openj9 or ibm
 		else ifneq ($(filter openj9 ibm, $(JDK_IMPL)),)
 			JVM_NATIVE_OPTIONS := -nativepath:"$(TESTIMAGE_PATH)$(D)openj9"
+			ifeq ($(OS),OS/390)
+			  ZOS_ONLY_NATIVE_LIBPATH := -e LIBPATH=$(TESTIMAGE_PATH)$(D)hotspot$(D)jtreg$(D)native
+			endif
 		endif
 	endif
 endif
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -606,7 +606,7 @@
 			<variation>Mode1000</variation>
 		</variations>
 		<command>$(JAVA_COMMAND) -Xmx512m $(APPLICATION_OPTIONS) -jar $(Q)$(TEST_RESROOT)$(D)jtreg$(D)lib$(D)jtreg.jar$(Q) \
-	$(JTREG_BASIC_OPTIONS) -nativepath:"$(TESTIMAGE_PATH)/hotspot/jtreg/native" -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS) $(VMOPTION_HEADLESS)$(Q) \
+	$(JTREG_BASIC_OPTIONS) -nativepath:"$(TESTIMAGE_PATH)/hotspot/jtreg/native" ${ZOS_ONLY_NATIVE_LIBPATH} -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS) $(VMOPTION_HEADLESS)$(Q) \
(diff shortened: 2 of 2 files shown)


karianna (approved):

pshipton:
I also notice serviceability_jdwp_j9 has the same native library setting, perhaps it needs the same change.

> On z/OS, simply specifying nativepath is not enough for these tests. The libraries are getting picked from the LIBPATH environment variable instead,

How does this work for any other playlist that uses native libraries?

openjdk/openjdk.mk
```diff
@@ -189,6 +189,10 @@ ifneq ($(JDK_VERSION),8)
 		# else if JDK_IMPL is openj9 or ibm
 		else ifneq ($(filter openj9 ibm, $(JDK_IMPL)),)
 			JVM_NATIVE_OPTIONS := -nativepath:"$(TESTIMAGE_PATH)$(D)openj9"
+			# z/OS dynamic loader ignores -nativepath and uses LIBPATH
+			ifeq ($(OS), zos)
+			export LIBPATH := $(TESTIMAGE_PATH)$(D)openj9:$(LIBPATH)
```
karianna:
```suggestion
			  export LIBPATH := $(TESTIMAGE_PATH)$(D)openj9:$(LIBPATH)
```

pshipton: I expected a variable to be defined on z/OS which sets the appropriate option using `-e` (like you had before), and then this new variable can be added to the test recipe in the playlist.

Darshan-N3: Thanks @pshipton. I’ve updated the changes as suggested, and `LIBPATH` is now being picked up correctly.

pshipton: Looks good.

openjdk/playlist.xml
```diff
@@ -606,7 +606,7 @@
 			<variation>Mode1000</variation>
 		</variations>
 		<command>$(JAVA_COMMAND) -Xmx512m $(APPLICATION_OPTIONS) -jar $(Q)$(TEST_RESROOT)$(D)jtreg$(D)lib$(D)jtreg.jar$(Q) \
-	$(JTREG_BASIC_OPTIONS) -nativepath:"$(TESTIMAGE_PATH)/hotspot/jtreg/native" -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS) $(VMOPTION_HEADLESS)$(Q) \
+	$(JTREG_BASIC_OPTIONS) -nativepath:"$(TESTIMAGE_PATH)/hotspot/jtreg/native" ${ZOS_NATIVE_LIBPATH} -vmoptions:$(Q)-Xmx512m $(JVM_OPTIONS) $(VMOPTION_HEADLESS)$(Q) \
```
karianna: Hmm, I know this variable will be "" when not ZOS but it feels 'odd' to have what looks like a ZOS flag being injected in the generic call here.  Not the hill I'm going to die on though.

pshipton: I believe it's what Shelley was suggesting in https://github.com/adoptium/aqa-tests/pull/6754#issuecomment-3690405829

smlambert: yes, perhaps a different name ZOS_ONLY_NATIVE_LIBPATH is clearer, but yes, this is the approach that makes most sense given our current options

smlambert (approved):
Thanks @Darshan-N3

smlambert (commented):
Hmm, one other way would be to just replace JVM_NATIVE_OPTIONS with `-e LIBPATH=$(TESTIMAGE_PATH)$(D)hotspot$(D)jtreg$(D)native` for OS/390, if the other -nativepath piece is ignored anyway

openjdk/playlist.xml
Darshan-N3: Thanks @smlambert. I’ve updated the changes as suggested.

PR #6775, opened 2025-12-05
Update genTestFolderList to dynamically detect test directories

### LINKED ISSUE #6761: Update genTestFolderList to search updated directories
As test material changes over time, we can improve the automation that checks if new test targets are needed. 

Given the setting of version number is already updated [here](https://github.com/adoptium/aqa-tests/blob/master/jck/jck.mk#L56), the setting can also be updated [here](https://github.com/adoptium/aqa-tests/blob/master/jck/jck_target_folder_check/jck_target_folder_check.sh#L57-L62).

Agaba-derrick:
## Summary

This PR updates `genTestFolderList` to dynamically discover JCK test directories instead of relying on hard-coded version-specific paths. This ensures new test targets are automatically detected as the JCK layout evolves.

## What Changed

- Removed version-based directory assumptions (e.g., `*-runtime-$ver`)
- Added dynamic directory scanning for:
  - `runtime`
  - `compiler`
  - `devtools`
- Improved resilience to future JCK repository structure changes

## Why

As test materials evolve over time, the previous static approach could miss new test folders. This update keeps target detection accurate and future-proof.

## Issue
Closes #6761

### DIFF
--- a/jck/jck_target_folder_check/jck_target_folder_check.sh
+++ b/jck/jck_target_folder_check/jck_target_folder_check.sh
@@ -55,7 +55,7 @@ genTargetList() {
 }
 
 genTestFolderList() {
-		if [ "$VERSION" -eq 8 ] ; then 
+	if [ "$VERSION" -eq 8 ] ; then 
 		ver="$VERSION"c
 	elif [ "$VERSION" -eq 11 ] ; then 
 		ver="$VERSION"a
@@ -71,7 +71,7 @@ genTestFolderList() {
 	cd $JCK_ROOT/*-compiler-$ver/tests
 	find . -maxdepth 2 -mindepth 2 -type d > $outputdir/compiler-dirs.txt
 
-	if [ "$ver" == "8c" ] || [ "$ver" == "11a" ] || [ "$ver" == "17a" ]; then 
+	if [ $ver == "8c" ]; then 
 		cd $JCK_ROOT/*-devtools-$ver/tests
 		find . -maxdepth 2 -mindepth 2 -type d > $outputdir/devtools-dirs.txt
 	fi


Agaba-derrick:
> hi @Agaba-derrick - this is not the change I expected for this one. I was hoping for the most minimal change to this file, essentially changing this [block](https://github.com/adoptium/aqa-tests/blob/master/jck/jck_target_folder_check/jck_target_folder_check.sh#L58-L62) to
> 
> ```
> 	if [ "$VERSION" -eq 8 ] ; then 
> 		ver="$VERSION"c
> 	elif [ "$VERSION" -eq 11 ] ; then 
> 		ver="$VERSION"a
> 	else 
> 		ver="$VERSION"
> 	fi 
> ```

thanks @smlambert  i will change this Asap !

smlambert:
@Agaba-derrick - see these [lines](https://github.com/adoptium/aqa-tests/blob/master/jck/jck.mk#L54-L58), you will need to deal with both 11 and 17 (my example only did 11)

jck/jck_target_folder_check/jck_target_folder_check.sh
```diff
@@ -55,8 +55,12 @@ genTargetList() {
 }
 
 genTestFolderList() {
-	if [ "$VERSION" -eq 8 ] ; then 
+		if [ "$VERSION" -eq 8 ] ; then
```
sophia-guo: A minor note: there is a unintended extra indentation, we can keep as before.

jck/jck_target_folder_check/jck_target_folder_check.sh
```diff
@@ -67,7 +71,7 @@ genTestFolderList() {
 	cd $JCK_ROOT/*-compiler-$ver/tests
 	find . -maxdepth 2 -mindepth 2 -type d > $outputdir/compiler-dirs.txt
 
-	if [ $ver == "8c" ]; then 
+	if [ "$ver" == "8c" ] || [ "$ver" == "11a" ] || [ "$ver" == "17a" ]; then
```
sophia-guo: There is no devtools for 11 and 17. So we can keep this as before.

PR #6890, opened 2026-02-10
Exclude AIX 7.1 machines from extended.openjdk tests on ppc64_aix

annaibm:
- Add LABEL exclusion for sw.os.aix.7_1 when running extended.openjdk tests
- Apply exclusion to both parent jobs and parallel child testList jobs

related: https://github.com/eclipse-openj9/openj9/issues/23340

### DIFF
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -344,6 +344,9 @@ timestamps{
                     if (!params.JDK_VERSION.isInteger() || params.JDK_VERSION.toInteger() >= 25) {
                         LABEL += "&&sw.tool.c++runtime.17_1"
                     }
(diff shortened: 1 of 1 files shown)


pshipton (approved):

smlambert (changes requested):
I am not sure this is the best place to make this change.  I'd like a chance to consider another approach.

smlambert:
If I understand correctly this issue is specific to OpenJ9/Semeru (related to OpenSSL), so we should not narrow the range of AIX machines for other JDK_IMPL cases which is what this PR currently does.

I would suggest moving the check inside https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/openjdk_tests#L343-L347

Issue #6867, opened 2026-02-01
Fix the "Parse disabled issues" workflow


smlambert:
.github/workflows/parse-issues.yml 

https://github.com/smlambert/aqa-tests/actions/runs/21557303904/job/62115840486

```
  DEBUG - Processing openjdk/excludes/alpine...
  ERROR - openjdk/excludes/alpine : 'openjdk/excludes/alpine' is not a file
  DEBUG - Processing openjdk/excludes/unreliables...
  ERROR - openjdk/excludes/unreliables : 'openjdk/excludes/unreliables' is not a file
  DEBUG - Processing openjdk/excludes/vendors...
  ERROR - openjdk/excludes/vendors : 'openjdk/excludes/vendors' is not a file
  DEBUG - Outputting to stdout
  DEBUG - Error found. Exiting with code 1
```

PR #6905, opened 2026-02-23
Finding all test exclusion files inside nested folders

adamfarley:
Currently, we try to parse a pair of folders as if they were files, which fails the parse-issues github workflow.

This change fixes that, and also allows us to parse the exclude files within the folders, allowing the workflow to proceed.

Many errors are reported, but the exclude files are parsed and the workflow eventually passes.

Tested here: https://github.com/adamfarley/aqa-tests/actions/runs/22314174344/job/64554212310

Resolves https://github.com/adoptium/aqa-tests/issues/6867

### DIFF
--- a/.github/workflows/parse-issues.yml
+++ b/.github/workflows/parse-issues.yml
@@ -26,7 +26,7 @@ jobs:
       - name: discover disabled tests
         run : |
           echo "::group::openjdk exclude files"
-          find ./openjdk/excludes -name '*ProblemList*.txt' | tee exclude_files.txt
+          find openjdk/excludes -name '*ProblemList*.txt' | tee exclude_files.txt
           echo "::endgroup::"
           echo "::group::playlist files"
           find . -name "playlist.xml" -not -path "scripts" | tee playlist_files.txt


.github/workflows/parse-issues.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: discover disabled tests
         run : |
           echo "::group::openjdk exclude files"
-          ls -1dq openjdk/excludes/* | tee exclude_files.txt
+          find ./openjdk -name *ProblemList*.txt | grep -o openjdk.* | tee exclude_files.txt
```
andrew-m-leonard: suspect the grep is not necessary, and also beware of globbing&cmd line expansion, since this is checking for excludes, shouldn't we just check under that directory?
ie.
```
find openjdk/excludes -name '*ProblemList*.txt' | tee exclude_files.txt
```

adamfarley: Changes made. Testing [here](https://github.com/adamfarley/aqa-tests/actions/runs/22346172953).

Will wait until the entire workflow finishes before re-requesting review.

andrew-m-leonard (approved):
lgtm

adamfarley:
Thanks @andrew-m-leonard and @smlambert 😎

Issue #6865, opened 2026-01-29
Use a more specific informational name for customUrl


smlambert:
Relating to this work/discussion, https://github.com/adoptium/aqa-tests/pull/6694#issuecomment-3816143091, we should improve the variable name 'customUrl' so that we better understand the Jenkinsfile code upon reading it.  We use a lot of URLs for a lot of different reasons, so we need to have better naming to make the code clearer and more understandable.

PR #6909, opened 2026-02-24
Improve the variable name 'customUrl' to 'testDependencyUrl'

Amrutha-Kanhirathingal:
Renamed customUrl to testDependencyUrl
Related :https://github.com/adoptium/aqa-tests/issues/6461
Closes :https://github.com/adoptium/aqa-tests/issues/6865

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -223,15 +223,15 @@ def setupParallelEnv() {
 				// check for each lib. If lib does not exist, donwload it.
 				// If lib exists, SHA will be checked. Re-download if SHA does not match.
 				timeout(time: 20, unit: 'MINUTES') {
-					def customUrl = getCustomUrl()
-					env.CUSTOM_URL = customUrl
+					def testDependencyUrl = getTestDependencyUrl()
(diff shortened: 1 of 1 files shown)


annaibm:
Also I think we need a companion TKG change as well, I see in TKG
https://github.com/adoptium/TKG/blob/master/scripts/getDependencies.pl#L51 which uses
```
customUrl = $ENV{'CUSTOM_URL'} if ...;
```
→ to be `$testDependencyUrl = $ENV{'TEST_DEPENDENCY_URL'} `
Please do verify other occurrences as well which might be failing when running through our JenkinsfileBase in release jobs..

Amrutha-Kanhirathingal:
> Also I think we need a companion TKG change as well, I see in TKG https://github.com/adoptium/TKG/blob/master/scripts/getDependencies.pl#L51 which uses
> 
> ```
> customUrl = $ENV{'CUSTOM_URL'} if ...;
> ```
> 
> → to be `$testDependencyUrl = $ENV{'TEST_DEPENDENCY_URL'} ` Please do verify other occurrences as well which might be failing when running through our JenkinsfileBase in release jobs..

I have already raised a PR for related changes in  TKG.Kindly have a look into this.
https://github.com/adoptium/TKG/pull/812

PR #6913, opened 2026-02-26
Update OpenJ9 Valhalla tests setup

theresa-m:
- modify hotspot_valhalla playlist to use the correct native libraries
- enable tests that are currently passing
- permanently disable tests that are not compatible with OpenJ9

personal build https://hyc-runtimes-jenkins.swg-devops.com/view/Valhalla%20Tests/job/Test_JDKnext_valhalla_openjdk_x86-64_linux/31/


openjdk/playlist.xml
```diff
@@ -57,7 +57,8 @@
 			<variation>-XX:ValueTypeFlatteningThreshold=99999 -XX:+EnableArrayFlattening</variation>
 		</variations>
 		<command>$(JAVA_COMMAND) -Xmx512m -jar $(Q)$(TEST_RESROOT)$(D)jtreg$(D)lib$(D)jtreg.jar$(Q) \
-			$(JTREG_BASIC_OPTIONS) $(JDK_NATIVE_OPTIONS) \
+			$(JTREG_BASIC_OPTIONS) \
+			-nativepath:"$(TESTIMAGE_PATH)/hotspot/jtreg/native" \
```
karianna: Why is this now hardcoded as such?

theresa-m: This test only runs with OpenJ9 builds so <s>JTREG_BASIC_OPTIONS</s> (edit: I meant JDK_NATIVE_OPTIONS) will be set to OpenJ9 natives which is incorrect for this test suite.

karianna: Would it be preferable to pass through the correct flag here though?

theresa-m: Sure. I added a variable for this and other playlist entries that were hardcoding this option.

theresa-m: Do you have other comments for this change @karianna ?

karianna (approved):

Issue #5161, opened 2024-03-20
Reduce duplication of test data in MBCS test suite


smlambert:
Currently the MBCS test suite pulls in test data from unicode.org public data, to be able to compare what the JDK produces, versus what is expected based on the test data.

For expediency, a copy of this test data is in the MBCS_Tests directory.  In fact, certain files are duplicate within subdirectories like https://github.com/adoptium/aqa-tests/blob/master/functional/MBCS_Tests/codepoint/data/UnicodeData-15.1.0.txt and https://github.com/adoptium/aqa-tests/blob/master/functional/MBCS_Tests/unicode/data/UnicodeData-15.1.0.txt.  

We can improve this in 2 ways:
- not keeping 2 copies of files, but referring to a single copy that is in a single location (possibly pulled in by the getDependencies job)
- pulling test data as part of getDependencies job (and possibly noting when new data shows up at the unicode site)

smlambert:
Items to move out of aqa-tests and pulled in as testDependency

- CLDC_11/ext/icu4j-61_1.jar
- CLDC_11/ext/icu4j-localespi-61_1.jar
- codepoint/data/*
- unicode/data/*


Items to consider `**/data/*` in other subdirs, though they do not appear to be duplicating content

sophia-guo:
@samyoon20 test dependency job https://ci.adoptium.net/view/Test_grinder/job/test.getDependency/, which is defined by aqa-tests/buildenv/jenkins/getDependency file and underlying calls getDependencies.pl under TKG TKG/scripts/getDependencies.pl. So first step would be working on TKG/scripts/getDependencies.pl to add those jars.

sophia-guo:
> Noting that in a recent update to the MBCS suite https://github.com/adoptium/aqa-tests/pull/5150, 9 of 13 files were test data, while the 4 src files changed only required 10 lines of code changed. Managing the test data differently would reduce maintenance costs of this suite.

Just to clarify 7 of 13 files were test data, which can be obtained from unicode website.  2 of files are identical resource files ( https://github.com/adoptium/aqa-tests/blob/master/functional/MBCS_Tests/codepoint/resources/UnicodeVers.properties and https://github.com/adoptium/aqa-tests/blob/master/functional/MBCS_Tests/unicode/resources/UnicodeVers.properties), which setup the mapping of unicode version with jdk version.

sophia-guo:
[...]

PR #6990, opened 2026-04-01
Set minimum NUM_MACHINES=6 for special.functional on AIX to prevent OOM

annaibm:
AIX machines (e.g., paix820) with limited RAM (8GB) run into OOM errors in resultsSummary when special.functional runs as a single unsplit job. Adding a minimum of 6 parallel lists ensures the job is always split, preventing Perl OOM failures during test result processing.

related: https://github.ibm.com/runtimes/automation/issues/921

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -269,9 +269,10 @@ def setupParallelEnv() {
 							PARALLEL_OPTIONS = PARALLEL_OPTIONS_TEMP + " NUM_MACHINES=${minParallelNum} TEST_TIME="
 							NUM_LIST = genParallelList(PARALLEL_OPTIONS)
 						}
-					} else if (TARGET == "sanity.functional" && PLATFORM.contains('windows')) {
+					} else if ((TARGET == "sanity.functional" && PLATFORM.contains('windows')) || (TARGET == "special.functional" && PLATFORM.contains('aix'))) {
 						if (NUM_LIST < 6) {
-							echo "Regenerate parallel list with NUM_MACHINES=6 for TARGET ${TARGET} due to long running tests (runtimes/automation/issues/654)."
+							def reason = TARGET == "sanity.functional" ? "due to long running tests (runtimes/automation/issues/654)" : "on AIX due to memory constraints (runtimes/automation/issues/921)"
+							echo "Regenerate parallel list with NUM_MACHINES=6 for TARGET ${TARGET} ${reason}."
 							PARALLEL_OPTIONS = PARALLEL_OPTIONS_TEMP + " NUM_MACHINES=6 TEST_TIME="
 							NUM_LIST = genParallelList(PARALLEL_OPTIONS)
 						}


annaibm:
@pshipton , yes this was tested via Grinder run https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/59332/ on `ppc64_aix` and the fix correctly split the job into 6 parallel lists. However `testList_5` still failed with OOM — this suggests NUM_MACHINES=6 may not be sufficient and we may need to increase it to 7 or 8. 
`testList_5` still hit OOM with NUM_MACHINES=6. Would you suggest increasing the minimum to 7 or 8, or is there a better approach you would recommend to handle this?

pshipton:
Perhaps we should be looking at the tests which produce so much output and see about reducing it.

pshipton:
Or am I misunderstanding why the OOM occurs?

annaibm:
my understanding is the OOM occurs in [resultsSum.pl](https://github.com/adoptium/TKG/blob/master/scripts/resultsSum.pl) which reads the entire `TestTargetResult` file and builds the TAP output string in memory before writing it to disk. With MBCS tests producing very verbose output, this in-memory string grows very large on AIX machines with limited RAM (8GB).
So as you suggest, reducing test output from MBCS tests would reduce the memory pressure. The NUM_MACHINES increase could help when initial test time from TRSS is not available. I will investigate which MBCS tests produce the most verbose output and see if it can be reduced.

smlambert:
> I will investigate which MBCS tests produce the most verbose output and see if it can be reduced.

This is a very good initiative.  There are several other ways to improve MBCS tests, and we should likely make a plan to address all of them.  Related: https://github.com/adoptium/aqa-tests/issues/5161

buildenv/jenkins/JenkinsfileBase
```diff
@@ -275,6 +275,12 @@ def setupParallelEnv() {
 							PARALLEL_OPTIONS = PARALLEL_OPTIONS_TEMP + " NUM_MACHINES=6 TEST_TIME="
 							NUM_LIST = genParallelList(PARALLEL_OPTIONS)
 						}
+					} else if (TARGET == "special.functional" && PLATFORM.contains('aix')) {
+						if (NUM_LIST < 6) {
+							echo "Regenerate parallel list with NUM_MACHINES=6 for TARGET ${TARGET} on AIX due to memory constraints (runtimes/automation/issues/921)."
+							PARALLEL_OPTIONS = PARALLEL_OPTIONS_TEMP + " NUM_MACHINES=6 TEST_TIME="
+							NUM_LIST = genParallelList(PARALLEL_OPTIONS)
+						}
```
sophia-guo, line 278: This logic is exactly same as the logic for
```
else if (TARGET == "sanity.functional" && PLATFORM.contains('windows')) {
						if (NUM_LIST < 6) {
							echo "Regenerate parallel list with NUM_MACHINES=6 for TARGET ${TARGET} due to long running tests (runtimes/automation/issues/654)."
							PARALLEL_OPTIONS = PARALLEL_OPTIONS_TEMP + " NUM_MACHINES=6 TEST_TIME="
							NUM_LIST = genParallelList(PARALLEL_OPTIONS)
						}
					}
```

It would be better to update the condition instead of duplicating the whole block.

annaibm: I have updated the PR to combine both conditions into a single block to avoid code duplication

karianna (approved):

pshipton:
Not sure we'll still need this change, I'm setting it to draft.

Issue #1853, opened 2020-06-26
Create a Disabled tests pipeline 


smlambert:
This test pipeline could check state of the issue that the test was disabled under and then run all disabled test targets whose issues are resolved to check if they really are in a state to reinclude/reenable.

Since there are various granularities on how we can and do disable tests, this work would look at several parts:

- [ ] disabled test targets in playlists
- [ ] disabled openjdk tests in openjdk subdirectory (in exclude*.txt files)
- [ ] disabled functional tests in openj9 repo (several types of exclusions, testNG, cmdlinetester, others?)

Likely do it as a pipeline that launched Grinders (similar to Grinder of Grinders, see https://github.com/AdoptOpenJDK/openjdk-tests/issues/1816).  

One job was already added, but is a WIP, see https://ci.adoptopenjdk.net/view/Test_grinder/job/Test_jdk8_j9_Disabled_Weekly/

derek-keeler:
At Microsoft we have a plan to do something along this line as well, and we would love to collaborate here.

Currently our idea is to:

For each platform/OS:

- Obtain the latest known good build (latest nightly) 
- Obtain the ProblemList.txt files (there will certainly be more than 1).
- Parse the ProblemList.txt files to enumerate the list of tests.
- Run each test 1 at a time against the last known good build and record the status.
- Generate a report that shows the status for each test run
- Collate week-over-week runs of these tests to show trends toward success (candidates for re-introduction).


...something like that.

d3r3kk:
At Microsoft we have a plan to do something along this line as well, and we would love to collaborate here.

Currently our idea is to:

For each platform/OS:

- Obtain the latest known good build (latest nightly) 
- Obtain the ProblemList.txt files (there will certainly be more than 1).
- Parse the ProblemList.txt files to enumerate the list of tests.
- Run each test 1 at a time against the last known good build and record the status.
- Generate a report that shows the status for each test run
- Collate week-over-week runs of these tests to show trends toward success (candidates for re-introduction).


...something like that.

d3r3kk:
cc: @junyzheng

smlambert:
Great!  since AQA invokes multiple underlying test frameworks there are several ways tests can be excluded.  

You've identified how one could approach the openjdk test group.  It will be good to break down the components of the solution into runnable units so we can use some of the same pieces for the different types of excludes.

Generically, we 
- create a list of tests and the issues they are excluded under and platform
[...]

PR #6993, opened 2026-04-02
Script for running test exclusion parser in sandbox

adamfarley:
A quickstart script that prepares a sandbox and runs the three core python scripts used for identifying disabled tests and determining whether their associated issue is closed.

Related to Step 4 in https://github.com/adoptium/aqa-tests/issues/1853

assisted-by: bob


karianna (approved):

karianna (approved):

karianna (approved):

scripts/disabled_tests/parse_disabled_tests.sh
```diff
@@ -0,0 +1,285 @@
...
+
+Must be run from the aqa-tests repository root directory.
+EOF
+    exit 0
+}
+
+# Parse command line arguments
+CLEAN_OUTPUT=false
+VERBOSE=false
+OUTPUT_DIR="./output"
```
andrew-m-leonard, line 79: Should we add ./output to the ".gitignore" to prevent people accidently merging the files?

scripts/disabled_tests/parse_disabled_tests.sh
```diff
@@ -0,0 +1,285 @@
...
+        requirementsPresent=false
+    fi
+done < scripts/disabled_tests/requirements.txt
+
+if [ "$requirementsPresent" = "true" ]; then
+    log_success "Python requirements already installed"
+else
+    log_info "One or more Python requirements not found"
+    log_info "Installing Python requirements..."
+    pip3 install -q -r scripts/disabled_tests/requirements.txt
```
andrew-m-leonard: not sure we should just go and install the required python modules, especially if they are coming from a variable? could be malware?

scripts/disabled_tests/parse_disabled_tests.sh
adamfarley: That's a good idea.

adamfarley: Done.

adamfarley:
> thanks @adamfarley - Remind me from our previous discussion, how this script is different than the workflow we already have https://github.com/adoptium/aqa-tests/blob/master/.github/workflows/parse-issues.yml and why it is needed or preferred.

Hi Shelley, and thank you for your patience.

This script is designed to allow people to easily use and develop the python scripts locally.

This ensures that:
- We do not proceed to the next python script without verifying the output json files produced by the previous script.
- We have full access to the intermediary files generated at each step, and can rerun individual scripts against individual jsons (allowing us to test fixes).
- New functionality can be tested without needing to commit changes containing debug code that the user needs to remember to remove prior to merging a PR.

scripts/disabled_tests/parse_disabled_tests.sh
adamfarley: That's a fair comment. I've replaced the code that installs the modules with instructions on how the user can do it.

.gitignore
```diff
@@ -4,6 +4,7 @@
 # misc auto generated folders
 bin
 !perf/liberty/scripts/bin
+/output
```
andrew-m-leonard, line 7: should that be just "output"

andrew-m-leonard (approved):

PR #7048, opened 2026-04-20
Exclude jdk26 problematic testcases

sophia-guo:

PR #7050, opened 2026-04-21
[v1.0.13-release] - Exclude jdk26 problematic testcases (#7048)

sophia-guo:
* Exclude jdk26 problematic testcases

* Update openjdk/excludes/ProblemList_openjdk26.txt

---------


adamfarley (approved):
lgtm

PR #7088, opened 2026-04-30
Simplify TAP verification to use new Semeru release structure

annaibm:
- Remove old structure fallback from tapVerification.groovy
- Add target filtering to only collect specified AQAvit targets

related: https://github.ibm.com/runtimes/automation/issues/988

### DIFF
--- a/buildenv/jenkins/tapVerification/tapVerification.groovy
+++ b/buildenv/jenkins/tapVerification/tapVerification.groovy
@@ -47,22 +47,43 @@ timestamps{
 
                     if (SDK_RESOURCE == "customized" ) {
                         if (params.TOP_LEVEL_SDK_URL) {
-                            // example: <jenkins_url>/job/build-scripts/job/openjdk17-pipeline-IBM/354/artifact/target/linux/s390x/openj9/AQAvitTaps/*zip*/AQAvitTaps.zip
+                            // Use release structure with TAP files at root
+                            // example: <jenkins_url>/job/AQA_Test_Pipeline_Release/473/artifact/*zip*/archive.zip
+                            download_url = params.TOP_LEVEL_SDK_URL + "artifact/*zip*/archive.zip"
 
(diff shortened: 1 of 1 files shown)


annaibm:
Test:
[Collect_AQA_TAP](https://hyc-runtimes-jenkins.swg-devops.com/job/Collect_AQA_TAP/lastBuild/)

Collected as below:
Screenshot 2026-04-30 at 4 53 10 PM

annaibm:
@ShelleyLambert @sophia-guo Kindly review!

smlambert:
@annaibm - fyi my github id is @smlambert, not ShelleyLambert which was somehow associated with an old IBM email address I no longer have

smlambert:
Also, I will like to understand how we are diverging... 
@sophia-guo - presume that this may be a change that will be in use as we move to decoupled story

annaibm:
Hi @smlambert, sorry about the wrong GitHub ID!
Regarding the divergence — `tapVerification.groovy `was using the old Semeru build pipeline structure where TAP files were nested under `target/${os}/${arch}/${variant}/AQAvitTaps/` With new process, this structure changed and TAP files are now at the root of the artifact archive (`artifact/*zip*/archive.zip`). This caused [Collect_AQA_TAP](https://hyc-runtimes-jenkins.swg-devops.com/job/Collect_AQA_TAP/) to break at some point in past and we had to collect TAP files manually to upload. This PR removes the old path and adds target filtering so only the specified AQAvit targets are collected rather than all TAP files. Its basically to make the  Semeru Aqavit Verification step quick and easy.

Tested it or used to complete this  [Semeru CE April 2026 AQAvit Verification](https://github.ibm.com/runtimes/semeru-releases/issues/531)

smlambert:
Thanks @annaibm - I understand what it is for, I just want to move towards a common solution again (and have less 'if j9' then do this...), from a quick glance it also looks like there are semeru specific comments in code used by non-semeru projects.  In any case we can discuss what are our common requirements and ensure we have a clean, generic solution.

smlambert (approved):

karianna (approved):

PR #7091, opened 2026-05-01
Add aggregated TEST TARGETS SUMMARY when ITERATIONS > 1

annaibm:
- Add the summary display when ITERATIONS > 1  to avoid incorrectly reporting all tests passed when earlier iterations had failures.

Related: #[978](https://github.ibm.com/runtimes/automation/issues/978)

### DIFF
--- a/buildenv/jenkins/JenkinsfileBase
+++ b/buildenv/jenkins/JenkinsfileBase
@@ -109,6 +109,15 @@ def setupEnv() {
 	env.TKG_ITERATIONS = params.TKG_ITERATIONS ? "${params.TKG_ITERATIONS}".toInteger() : 1
 	env.EXIT_FAILURE = params.EXIT_FAILURE ? params.EXIT_FAILURE : false
 	env.EXIT_SUCCESS = params.EXIT_SUCCESS ? params.EXIT_SUCCESS : false
+
+	// Initialize iteration result accumulators for multi-iteration test runs
+	env.ITER_COUNT = '0'
+	env.ITER_TOTAL = '0'
+	env.ITER_EXECUTED = '0'
+	env.ITER_PASSED = '0'
+	env.ITER_FAILED = '0'
+	env.ITER_DISABLED = '0'
+	env.ITER_SKIPPED = '0'
 	NUM_MACHINES = params.NUM_MACHINES ? params.NUM_MACHINES.toInteger() : 1
 	if (CLOUD_PROVIDER == 'azure') {
 		// Needs to be inside the workspace as the docker container won't have permissions to write to higher level directories
@@ -858,7 +867,23 @@ def runTest( ) {
 					makeTest("${RUNTEST_CMD}")
 				}
 			}
+
+			// Accumulate results after each iteration
+			def resultSum = [:]
+			def matcher = manager.getLogMatcher(".*(TOTAL: \\d+)\\s*(EXECUTED: \\d+)\\s*(PASSED: \\d+)\\s*(FAILED: \\d+)\\s*(DISABLED: \\d+)?\\s*(SKIPPED: \\d+)\\s*\$")
+			if (matcher?.matches()) {
+				for (int j = 1; j <= matcher.groupCount(); j++) {
+					if (matcher.group(j) != null) {
+						def matchVals = matcher.group(j).split(": ")
+						resultSum[matchVals[0]] = matchVals[1] as int
+					}
+				}
+				checkTestResults(resultSum)
+			}
 		}
+
+		// Write aggregated summary after all iterations complete
+		writeSummary()
 
 		if (params.CODE_COVERAGE) {
 			echo 'Generating Code Coverage Reports...'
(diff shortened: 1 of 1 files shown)


annaibm:
**Test Result:**

https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder_iteration_1/6924/
https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder_iteration_0/7127/ 
https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/59897/ 

- **Before** (misleading - ITERATIONS=10, failed iterations reported as PASSED):
https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder_iteration_1/6905
https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder_iteration_1/6905
Screenshot 2026-05-01 at 4 43 15 PM

- **After** - (fix - TEST TARGETS SUMMARY skipped when ITERATIONS > 1):
https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder_iteration_1/6924/
Screenshot 2026-05-01 at 4 42 36 PM

buildenv/jenkins/JenkinsfileBase
```diff
@@ -916,6 +916,13 @@ def checkTestResults(results) {
 		summary = manager.createSummary("accept.svg")
 	}
 
+	// Skip TEST TARGETS SUMMARY for Grinder iteration jobs (when parent has multiple iterations)
```
karianna: Would we not want to produce the summary for each iteration?

smlambert:
I do not think the right approach is to skip test targets summary.

The summary should ideally indicate how many iterations occured, report on percentage of pass/fail and generally reflect a summary of all iterations.  

But that is not the only thing that we should consider.  If I recall correctly we had different behaviours or modes for how we iterate.  By default the behaviour is to run through all iterations, in this case, typically to determine what percentage of failures versus passes out of the total number of iterations.  I believe there were two other modes, one where we want to iterate until a successful pass and then exit the iteration loop, and conversely a mode where we want to iterate until a failure and then exit the iteration loop.

smlambert:
Since the reporting issue is not publicly visible and there is a lot of discussion and explanation in that private issue, I will pull some of the key points into the public space here so everyone has a better understanding of the behaviour.  I had not looked at it until now, which means it was not possible to understand why this change was being proposed.

#### One TEST TARGET SUMMARY and TAP file per iteration
Each iteration writes its own TAP file and its own TEST TARGET SUMMARY. For iteration=10, then the console output will have 10 TEST TARGET SUMMARY. The TEST TARGET SUMMARY on jenkins job description part is not correct.

#### Test Job Status Reports Correctly
It is important to note that the overall job status is correctly shown as UNSTABLE when one or more of the iterations fail.

image

#### TKG correctly summarizes the overall iterations in the console
image

#### Chiclets report correctly
And the chiclets at the top of the Jenkins view reports correctly when the iterations are run in parallel

image

#### Jenkins parent job displays TEST TARGET SUMMARY of the last iteration

image

This is the incorrect part of the whole story.  When a human looks at the Jenkins job description page and sees a green TEST TARGET SUMMARY (as it only reports the last iteration, which if it happened to pass, will show as green), they miss the other cues like an unstable test job, yellow chiclets, and they don't proceed to looking at console output.

annaibm:
When ITERATIONS > 1 or TKG_ITERATIONS > 1, display an aggregated TEST TARGETS SUMMARY that sums results across all iterations, instead of only showing the last iteration's summary (which could misleadingly show all-green when earlier iterations failed).

Grinder results: 

- https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/60017/  => 5 iterations

Screenshot 2026-05-20 at 11 18 08 AM




 - https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/59992/ - > 3 Iteration builds
 
Screenshot 2026-05-20 at 11 20 01 AM


- https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder_iteration_0/7137/ -> 3 iterations

Screenshot 2026-05-20 at 11 18 26 AM

- https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/59984/ ✅  no iterations

Screenshot 2026-05-20 at 11 18 48 AM

- https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder_jck/4724/  sanity.jck using Dynamic playlists ✅

annaibm:
@smlambert @sophia-guo Kindly review the changes when you have some time.

sophia-guo:
To clarify - the issue is Jenkins child leaf jobs display  TESTS TARGET SUMMARY of the last iteration. If iteration >1 the  TESTS TARGET SUMMARY may be incorrect to only show the last iteration. Parent jobs don't display  TESTS TARGET SUMMARY at all.

1. With this change  TESTS TARGET SUMMARY still only applies to downstream leaf jobs. https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/59992/.

2. Downstream leaf jobs with iteration >0,  https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder_iteration_0/7137/ shows aggregated test targets correctly. We can get the idea about passing or failure percentage and this status information is consistent with the job status. However it won't give more information about which iteration fails if there is. Developers sill need to go to failed job and look at the console output and search the failed targets manually.   

I think this extra aggeration information can be added to parent jobs for parallel run.

sophia-guo (approved):

smlambert (approved):

PR #7116, opened 2026-05-15
Remove ant-contrib dependency

gdams:
## Summary

Eliminates the ant-contrib dependency by migrating all usage to native Ant 1.9.1+ equivalents. This removes an external dependency that was only used for a small subset of its features.

### Build files (108 files, net -253 lines)

- **Removed** `<taskdef resource="net/sf/antcontrib/antlib.xml"/>` from all 92 XML build files
- **Replaced** `<if>/<then>/<else>` with `xmlns:if="ant:if"` / `xmlns:unless="ant:unless"` namespace declarations and `if:set`/`unless:set` attributes on tasks
- **Replaced** `<propertyregex>` with `<loadresource>` + `<filterchain>` containing `<replaceregex>` (for regex replace/select) and `<striplinebreaks/>` (for z/OS newline removal)
- **Replaced** `<var>` (mutable properties) with `<property>` using Ant's first-set-wins immutability and `<condition>` with `value`/`else`
- **Flattened** deeply nested if/else chains into sequential `<condition>` blocks with first-set-wins `<property>` fallbacks

### Infrastructure cleanup

- Removed `ant-contrib` from `buildenv/docker/Dockerfile` and `external/system-test/dockerfile/Dockerfile`
- Removed `ant-contrib` from package lists in `test.properties` files (system-test, functional-test, criu-functional, criu-portable-checkpoint)
- Removed `ant_contrib_version` from `external/tck-ubi-test/test.properties`
- Removed `print_ant_contrib_install()` function from `external/dockerfile_functions.sh`
- Removed `ant_contrib_version` variable from `external/common_functions.sh`

### Documentation

- Updated `docs/pages/Prerequisites.md`: removed ant-contrib prerequisite and installation section
- Updated `jck/README.md`: removed ant-contrib.jar from prerequisites

## Migration patterns used

| ant-contrib task | Native Ant replacement |
|---|---|
| `<if>/<then>/<else>` wrapping target body | Target `if`/`unless` attributes |
| `<if>/<then>/<else>` for conditional property | `<condition property="X" value="Y" else="Z">` |
| `<if>/<then>/<else>` for conditional task execution | `if:set`/`unless:set` attributes on individual tasks |
| `<propertyregex>` (regex replace) | `<loadresource>` + `<tokenfilter>` + `<replaceregex>` |
| `<propertyregex>` (regex select) | `<loadresource>` + `<tokenfilter>` + `<replaceregex pattern="..." replace="\1">` |
[...]

PR #7124, opened 2026-05-20
Fix missing ant dependency in criu containers after ant-contrib removal

annaibm:
- PR#7116 removed ant-contrib from test.properties package lists. However, ant itself was being installed as a transitive dependency of ant-contrib. Removing ant-contrib also removed ant, causing container builds to fail with:
 ` /bin/sh: ant: not found`
- Add `--no-same-owner` to tar command in `pingPerf.sh` alongside `--no-same-permissions`
  to fully fix: "Cannot change mode to rwxr-xr-x: Operation not permitted" 
- Re-enable rhel.8 nodes in imageUploadMap now that -`-no-same-permissions` `--no-same-owner` resolves the ubi10 tar permission

related: https://github.com/adoptium/aqa-tests/pull/7116


annaibm:
**Grinder runs:**
- Before fix: criu-portable-checkpoint_test_0  ❌ 
https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder_criu_imageUpload/228/ ⚠️ 
```
10:15:46  if [ -z "$(tail -1 /aqa-tests/TKG/../TKG/output_compilation/compilation.log | grep 0)" ]; then perl scripts/moveDmp.pl --compileLogPath="/aqa-tests/TKG/../TKG/output_compilation/compilation.log" --testRoot="/aqa-tests/TKG/.."; false; else rm -f -r "/aqa-tests/TKG/../TKG/output_compilation"; fi
10:15:46  /bin/sh: 3: ant: not found
10:15:46  127
```

- After fix: `criu-portable-checkpoint_test_0` ✅ passed
https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder_criu_imageUpload/248/  ✅

smlambert (approved):

karianna (approved):

PR #7152, opened 2026-06-05
Skip download "tap.zip"

Amrutha-Kanhirathingal:
Exclude the tap.zip name from download because in test pipeline unrecognized zip is treated as SDK.

Related: https://github.ibm.com/runtimes/automation/issues/1069

### DIFF
--- a/get.sh
+++ b/get.sh
@@ -337,7 +337,7 @@ getBinaryOpenjdk()
 						if [ "$TEST_IMAGES_REQUIRED" == "true" ]; then
 							download_url+=" ${download_url_base}${n}"
 						fi
-					elif [[ $n != *"install"* && $n != *"unsigned"* ]]; then
+					elif [[ $n != *"install"* && $n != *"unsigned"* && $n != *"tap"* ]]; then
 						download_url+=" ${download_url_base}${n}"
 					fi
 				done


get.sh
```diff
@@ -337,7 +337,7 @@ getBinaryOpenjdk()
 						if [ "$TEST_IMAGES_REQUIRED" == "true" ]; then
 							download_url+=" ${download_url_base}${n}"
 						fi
-					elif [[ $n != *"install"* && $n != *"unsigned"* ]]; then
+					elif [[ $n != *"install"* && $n != *"unsigned"* && $n != *"tap"* ]]; then
```
annaibm, line 340: LGTM but may be could to tighten=>  `*"tap"*` to `*".tap.zip"*` for more clarity.

annaibm:
**Adding Context**:
The `tap.zip` file is a TAP metadata artifact published alongside the SDK in Artifactory. Since its filename contains `jdk`, the pipeline's unrecognized-zip fallback mistakenly treats it as an SDK archive, causing a conflict when the real SDK tarball is also downloaded.

get.sh
Amrutha-Kanhirathingal: As per the existing code we are using the file name  without "zip" extension
for example to skip downloading of "installmgr.zip" we  are using keyword "install"* similarly we are using keyword "unsigned" to skip downloading of "unsigned.tar.gz" were we have not added the extensions.so do we really have to add extension with "tap" for this case?

karianna:
> **Adding Context**: The `tap.zip` file is a TAP metadata artifact published alongside the SDK in Artifactory. Since its filename contains `jdk`, the pipeline's unrecognized-zip fallback mistakenly treats it as an SDK archive, causing a conflict when the real SDK tarball is also downloaded.

Worth adding that as an inline comment in the code

karianna (approved):

smlambert (approved):

sophia-guo:
@Amrutha-Kanhirathingal Please do a command cherry-pick

/cherry-pick v1.0.14-release

PR #7220, opened 2026-06-24
Correct job name for other vendor

sophia-guo:
Related with https://github.com/adoptium/ci-jenkins-pipelines/issues/1388

### DIFF
--- a/buildenv/jenkins/aqaTestPipeline.groovy
+++ b/buildenv/jenkins/aqaTestPipeline.groovy
@@ -295,15 +295,17 @@ def generateJobs(jobJdkVersion, jobTestFlag, jobPlatforms, jobTargets, globalBui
         if (os.contains("windows")) {
             filter = "*.zip"
         }
-        def short_name = "hs"
+        def short_name = params.VARIANT
         def jdk_impl = "hotspot"
         if (params.VARIANT == "openj9") {
             short_name = "j9"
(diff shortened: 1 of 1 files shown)


buildenv/jenkins/aqaTestPipeline.groovy
```diff
@@ -295,15 +295,17 @@ def generateJobs(jobJdkVersion, jobTestFlag, jobPlatforms, jobTargets, globalBui
         if (os.contains("windows")) {
             filter = "*.zip"
         }
-        def short_name = "hs"
+        def short_name = params.VARIANT
```
karianna, line 298: This logic comes across oddly, you're assigning to short_name but then rechecking the params.VARIANT value later.

sophia-guo: yes, it's due to some history naming convention for short_name ( used as part of test job name). Initially we only have 'temurin' and 'openj9' and short_name is defined as 'hs' and 'j9' correspondingly. Then there are more variants as  'dragonwell', 'correttto' , 'bisheng', etc. Using the variant itself works fine, no need to name a short_name for each new variant. However for consistence 'temurin' and 'openj9' will still use 'hs' and 'j9.

karianna (approved):

andrew-m-leonard (approved):

PR #7251, opened 2026-07-06
[v1.0.14-release] Updates for disabled test parser

adamfarley:
Includes the contents of several PRs, all centered around the disabled tests parser and PR-checker.

A list of PRs being cherry-picked can be found in the subject line of the associated commits.

Update: The checks have passed.


Haroon-Khel (approved):

smlambert:
Noting that we try to limit what gets cherrypicked to release branches to only what is needed for a release (in this case, updated problemlists IF they contained new exclusions needed for the release).  In future, I would not expect to see workflows being cherrypicked to a release branch, as we would not expect them to be run against that branch.

PR #7344, opened 2026-07-12
Clean up openjdk problemlist files

### LINKED ISSUE #6574: Update openjdk problemlist_*.txt to remove comment lines for recording different issues for same failure testcases.
Non-comment line format of Jtreg problemlist.txt is

 _test-name list-of-bug-numbers list-of-platforms description_

list-of-bug-numbers can be comma-separated lists with no embedded whitespace. So if we need to exclude one testcase on different platform for different reason we can use comma-separated links instead of adding extra duplicate line, which will also simply parsing the problemlist files to do some automation check jobs.

For example , the following case we can updated to one line instead of two lines

```
compiler/7184394/TestAESMain.java https://github.com/adoptium/aqa-tests/issues/1051,https://github.com/adoptium/infrastructure/issues/2893 linux-s390x,solaris-sparcv9
```

https://github.com/adoptium/aqa-tests/blob/master/openjdk/excludes/ProblemList_openjdk8.txt#L23C1-L25
```
# compiler/7184394/TestAESMain.java https://github.com/adoptium/infrastructure/issues/2893
compiler/7184394/TestAESMain.java https://github.com/adoptium/aqa-tests/issues/1051 linux-s390x,solaris-sparcv9
```

This issue to update all problemlist files to remove comment lines, which was added to record different issues for same failure testcases.

Akshat005Chaudhary:
This pull request removes the legacy duplicate and commented-out testcase exclude entries across the `ProblemList_*.txt` files.

### Why?
Previously, to record different issue tickets or URLs for the same test case, duplicate entries or commented-out entries were added to the problemlists. This ended up cluttering the excludes files and created parsing issues for automated verification scripts. Automated tools were having a hard time parsing the problems correctly due to these duplicate entries in simple terms right.

Since the `jtreg` format natively supports a comma-separated list of issue urls and bug numbers (with no spaces) in the bug-list field of an active entry, what this PR does is that it merges those references into a single active exclusion line per testcase.

### Changes Made
- Scanned all `ProblemList_*.txt` files in `openjdk/excludes/` (including vendor-specific lists).
- Merged the issue URLs and JDK bug IDs (e.g. `JDK-xxxxxxx`) from commented-out entries into their corresponding active exclusion entries as a comma-separated list.
- Removed the redundant commented-out duplicate lines.
- **Impact:** Cleaned up 37 files, resulting in a net reduction of ~590 redundant lines from the exclude lists.

Fixes: #6574

Signed-off-by: Akshat Chaudhary <chaudharyakshat555@gmail.com>

### DIFF
--- a/openjdk/excludes/ProblemList_openjdk11-openj9.txt
+++ b/openjdk/excludes/ProblemList_openjdk11-openj9.txt
@@ -172,9 +172,7 @@ javax/management/security/HashedPasswordFileTest.java https://github.com/adoptiu
 java/net/DatagramSocket/SetGetSendBufferSize.java https://github.com/adoptium/aqa-tests/issues/2245 macosx-all
 java/net/Inet4Address/PingThis.java	https://github.ibm.com/runtimes/backlog/issues/1595	aix-all
 java/net/InetAddress/CheckJNI.java https://github.ibm.com/runtimes/backlog/issues/684 macosx-all
-# java/net/Inet6Address/B6206527.java on macosx-all issue https://github.com/adoptium/aqa-tests/issues/1524
-# java/net/Inet6Address/B6206527.java on macosx-all issue https://github.com/adoptium/infrastructure/issues/1085
-java/net/Inet6Address/B6206527.java https://github.com/adoptium/infrastructure/issues/1105 linux-all,macosx-all
+java/net/Inet6Address/B6206527.java	https://github.com/adoptium/aqa-tests/issues/1524,https://github.com/adoptium/infrastructure/issues/1085,https://github.com/adoptium/infrastructure/issues/1105	linux-all,macosx-all
 java/net/MulticastSocket/Promiscuous.java https://github.com/adoptium/infrastructure/issues/699 linux-s390x,aix-all
 java/net/MulticastSocket/SetOutgoingIf.java  https://github.com/adoptium/aqa-tests/issues/2246 aix-all,linux-s390x
 java/net/NetworkInterface/UniqueMacAddressesTest.java https://github.ibm.com/runtimes/backlog/issues/719 macosx-all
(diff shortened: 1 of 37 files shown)


karianna:
@Akshat005Chaudhary Thanks! Please do sign the Eclipse CLA - https://www.eclipse.org/legal/eca/

Akshat005Chaudhary:
Hi @karianna, I have signed the agreement.

adamfarley:
I disagree with this change.

This change will make it more complicated to identify which platforms are covered by which issues. 

When an issue is resolved, we'll need to interrogate all of the other issues to ensure that none of them overlap with the platform linked to the resolved issue.

smlambert:
> This change will make it more complicated to identify which platforms are covered by which issues.

I was about to add the same note.  

The reason we decided to keep comments in the files is to keep context on which bug is associated with which platform.  While one could argue that "if" people kept the proper order when listing comma-separated issues and comma-separated platforms, one could intuit the association, there is no guarantee of that, and it doesn't cover a case where the number of issues doesn't match the number of platforms being excluded (2 issues, where one issue covers several platforms).

I am not particularly concerned about reducing the size of these text files especially if extra comments hold information.

Akshat005Chaudhary:
Yeah @smlambert I was doubtful on that too. Is there anything else that might be a more befitting fix. Or the issue itself is better left dismissed?

smlambert:
> Yeah @smlambert I was doubtful on that too. Is there anything else that might be a more befitting fix. Or the issue itself is better left dismissed?

@Akshat005Chaudhary - I would like to change the problem you are trying to fix.  Instead of removing comments, can you structure the comment as a correctly formatted exclude line.  I will update the wording in #6574 

Example: 
This: 
```
# java/net/Inet6Address/B6206527.java on macosx-all issue https://github.com/adoptium/infrastructure/issues/1085
java/net/Inet6Address/B6206527.java https://github.com/adoptium/infrastructure/issues/1105 linux-all,macosx-all
```
changes to this:
```
# java/net/Inet6Address/B6206527.java https://github.com/adoptium/infrastructure/issues/1085 macosx-all
java/net/Inet6Address/B6206527.java https://github.com/adoptium/infrastructure/issues/1105 linux-all,macosx-all
```

In that way, there is a chance that if 1105 happened to be fixed, the change to reinclude that testcase would be to delete the second line and uncomment the first.

smlambert:
Close in favour of https://github.com/adoptium/aqa-tests/pull/7350

PR #7370, opened 2026-07-15
Update getEBCNode to map PLATFORM to new EBC_Create_Node OS/ARCH params

mahdipub:
Signed-off-by: mahdi@ibm.com

### DIFF
--- a/buildenv/jenkins/openjdk_tests
+++ b/buildenv/jenkins/openjdk_tests
@@ -790,17 +790,74 @@ def generateJobViaAutoGen(testJobName) {
 
 def getEBCNode() {
     def expectedNodeLifespan = params.TIME_LIMIT ?: 1
-    def wait = true
-    def result = build job: 'EBC/EBC_Create_Node',
-    propagate: false,
-    wait: true,
-    parameters: [
-        string(name: 'PLATFORM', value: params.PLATFORM),
-        string(name: 'NODE_TYPE', value: 'testmachine'),
-        string(name: 'TIME_LIMIT', value: expectedNodeLifespan.toString()),
-        string(name: 'GROUP_LABEL', value: UUID.randomUUID().toString()),
-        booleanParam(name: 'WAIT', value: wait),
+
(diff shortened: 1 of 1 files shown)


annaibm (approved):
LGTM

annaibm:
@smlambert Could you please review this PR.

smlambert (dismissed):

buildenv/jenkins/openjdk_tests
```diff
@@ -790,17 +790,71 @@ def generateJobViaAutoGen(testJobName) {
...
+    def ebcOs   = osMap[rawOs]
+
+    if (!ebcArch || !ebcOs) {
+        error "PLATFORM '${platformStr}' is not supported by EBC (unsupported arch='${rawArch}' or os='${rawOs}')"
+    }
+
+    echo "EBC node request: PLATFORM=${platformStr} → OS=${ebcOs}, ARCH=${ebcArch}"
+
+    def result = build job: 'EBC/EBC_Create_Node',
+        propagate: false,
```
pshipton, line 845: According to AI, `propagate: false` suppresses failures silently. If the downstream build fails, `buildVariables` may be empty or missing `group_label`, causing a silent null return rather than a clear error.

mahdipub: The whole reason propagate: false was chosen here was that an EBC provisioning failure is treated as a recoverable event, not a pipeline failure which obviously is not a wanted behaviour. This was the aim of the `propagate: false` and fallback to static farm when EBC fails provisioning which this PR didn't change it. 
But to address your/AI true concern, I have added an `if blcok` to error when the build fails, then the [try..catch](https://github.com/mahdipub/aqa-tests/blob/ec87ff9e792ff1fb615e37c9e3c8a8c059b8baa5/buildenv/jenkins/openjdk_tests#L409) will handle and fail-back to static farm.

pshipton (approved):

smlambert (approved):

PR #7522, opened 2026-08-07
Exclude the jdk27+ jdk_doublevector128_j9 tests

pshipton:
Issue https://github.com/eclipse-openj9/openj9/issues/24360

I missed one in https://github.com/adoptium/aqa-tests/pull/7509

### DIFF
--- a/openjdk/playlist.xml
+++ b/openjdk/playlist.xml
@@ -3270,6 +3270,10 @@
 				<comment>https://github.com/eclipse-openj9/openj9/issues/20389</comment>
 				<platform>aarch64_linux</platform>
 			</disable>
+			<disable>
+				<comment>https://github.com/eclipse-openj9/openj9/issues/24360</comment>
+				<version>27+</version>				
+			</disable>
 		</disables>
 		<variations>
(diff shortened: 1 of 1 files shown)


openjdk/playlist.xml
```diff
@@ -3270,6 +3270,10 @@
 				<comment>https://github.com/eclipse-openj9/openj9/issues/20389</comment>
 				<platform>aarch64_linux</platform>
 			</disable>
+			<disable>
+				<comment>https://github.com/eclipse-openj9/openj9/issues/24360</comment>
+				<version>27+</version>
```
JasonFengJ9, line 3275: `jdk_doublevector128_j9` only applies to JDK 27+, this disable `version` is unnecessary. In addition, the disable for `aarch64_linux` can be removed now.

pshipton: These excludes are all temporary and will be removed as soon as we can fix vectorization for jdk27+ (next week or two?). I had added the version to all the excludes in https://github.com/adoptium/aqa-tests/pull/7509, none of them are necessary. Not sure it's worth fixing.

Similarly, I don't want to remove the exclude for aarch64_linux because I assume it will be necessary after the jdk27+ excludes are removed.

JasonFengJ9: Okay

JasonFengJ9 (approved):

sophia-guo (approved):

PR #7643, opened 2026-08-31
Enable Double128VectorTests on OpenJ9 AArch64 Linux

knn-k:
This commit enables Double128VectorTests on OpenJ9 AArch64 Linux back again.  The testcase was disabled by #7063 and #7106.


knn-k:
The relevant issue is not resolved yet.

PR #7650, opened 2026-09-04
Use xlc_r for AIX native builds

babsingh:
Use `xlc_r` instead of `xlc` to enable thread-safe `errno` handling on AIX.

### DIFF
--- a/jck/jtrunner/makefile
+++ b/jck/jtrunner/makefile
@@ -188,7 +188,7 @@ ifeq ($(OS),sunos)
 endif
 
 ifeq ($(OS),aix)
-	CC=/opt/IBM/xlC/16.1.0/bin/xlc
+	CC=/opt/IBM/xlC/16.1.0/bin/xlc_r
 	CFLAGS=-qstaticinline -q$(BitMode) -I$(UNIX_PATH) -I$(OSX_PATH) -I$(SRC_PATH)
 	LDFLAGS=-G -q$(BitMode)
 endif


babsingh:
fyi @pshipton @JasonFengJ9

JasonFengJ9 (approved):

babsingh:
Local testing passed after manually rebuilding the native libraries with `xlc_r`. The grinders have been failing due to infra issues, so I haven’t been able to validate the change through the full test framework yet.

I’m also not completely sure this is the right place to make the change, or whether it will work as expected in the full build. I opened the PR based on the current build logic so we can review the approach.

JasonFengJ9:
From the [test build](https://hyc-runtimes-jenkins.swg-devops.com/job/Test_openjdk27_j9_extended.jck_ppc64_aix_testList_7/5/consoleFull)
```
11:46:08       [exec] /home/jenkins/workspace/Test_openjdk27_j9_extended.jck_ppc64_aix_testList_7/aqa-tests/TKG/../jck/jtrunner/makefile:93: JAVA_HOME set to 
```
It is the file path specified in this PR though there is an internal version of [jck/jtrunner/makefile](https://github.ibm.com/runtimes/jck/blob/main/jck/jtrunner/makefile).

Yeah, we do need a grinder to verify the sanity of PR change.

pshipton (approved):

pshipton:
@annaibm pls take a look.

JasonFengJ9:
I think we need a verification grinder run before merging as per https://github.com/adoptium/aqa-tests/pull/7650#issuecomment-5545422452.

annaibm:
This change will also need to be mirrored internally — the JCK grinder runs use `jck/jtrunner/makefile` from https://github.ibm.com/runtimes/jck, which is a separate copy from this repo. Once that's updated, kick off a verification Grinder run for PLATFORM=ppc64_aix, TARGET=sanity.jck at https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder_jck/ to confirm the fix

PR #7694, opened 2026-09-17
Exclude arm32 tests

adamfarley:
A pair of arm32 tests have been seen to fail during previous releases, as detailed in the associated issues.

After discussion in the General Retrospective, it has been agreed that these tests should be excluded.


Haroon-Khel (approved):

steelhead31 (approved):

adamfarley:
Thanks @Haroon-Khel and @steelhead31 😎

PR #7697, opened 2026-09-18
Add OpenJCEPlusSemeruDefaults provider

JinhangZhang:
OpenJCEPlusSemeruDefaults is an additional security provider added in the master java.security file. We should add it in the openjceplus test java.security file to match the number of providers in the master one.


pshipton:
If OpenJCEPlusSemeruDefaults is the first security provider, why did you add it as the second?

jasonkatonica:
@JinhangZhang We also want to update all our other java.security test files which are used by the test flags. I believe that all the files under https://github.com/JinhangZhang/aqa-tests/tree/8dbb6a058be28db37b3bf24193a8c7fb52af5980/config/openjceplus/jdk will need a similar change as they are applied on different arch and java versions accordingly.

JinhangZhang:
> If OpenJCEPlusSemeruDefaults is the first security provider, why did you add it as the second?

@pshipton `OpenJCEPlusSemeruDefaults` is an additional provider in the provider list of the master java.security file. Adding `OpenJCEPlusSemeruDefaults` here is to match the number of providers same as the providers in the master java.security file. 

Also `OpenJCEPlusSemeruDefaults` and `OpenJCEPlus` are two separate providers, `OpenJCEPlus` test flag has openjceplus provider list first in the list, so OpenJCEPlusSemeruDefaults has to be in the second position in this test java.security file.

jasonkatonica (approved):
LGTM

pshipton (approved):

pshipton:
Pls open a PR again the v1.0.16-release branch as well.

JasonFengJ9 (approved):

pshipton:
@JinhangZhang is there any problem with your Eclipse ECA? The check is failing.

pshipton:
Obsoleted by https://github.com/adoptium/aqa-tests/pull/7710

